PR #62, opened 2020-09-16
Decoupling initialize providers

andreogle:
This is the first part of decoupling the lower parts of the node.

A lot of these functions were previously receiving the entire state just to access the current Ethereum provider so it could log that out. Now, they will instead return a tuple that looks like `[PendingLog[], Error | null, T]`. The calling function is then required to handled each of these elements otherwise ESLint complains.

I've also moved a lot of the Ethereum specific code (from the "triggers" directory) into the "ethereum" directory (I plan to rename this to "evm" later). This is in anticipation of eventually adding support for other blockchains.


bbenligiray (approved):
I can't say I was able to follow 100% but looks good

andreogle:
Thanks. There isn't really anything major in here. Mostly just rewriting and moving stuff anyway

Issue #54, opened 2020-09-09
Ignore API calls and withdrawals with insufficient balance


andreogle:
- API calls are currently set to `Errored` (wrong)
- Withdrawals do not have their balance validated

PR #63, opened 2020-09-16
Ignore requests with insufficient balance

andreogle:
Addresses: https://github.com/api3dao/airnode/issues/54

- Ignore API calls instead of setting them to be "errored" when the wallet does not have sufficient balance
- Do the same for withdrawals which are currently not getting validated

I branched off the `decoupling` branch so let's merge that first before this one


bbenligiray (approved):
:ship:

PR #65, opened 2020-09-20
Decoupling Part 2

andreogle:
Decoupling more modules & functions. There aren't really any significant logic changes here, mostly just simplifying and moving stuff around.

I also split the authorization fetching and application into 2 separate modules so that it's consistent with the way templates work


packages/node/src/core/ethereum/authorization/authorization-application.ts
```diff
@@ -0,0 +1,101 @@
...
+      return [[], null, apiCall];
+    }
+
+    const log = logger.pend(
+      'WARN',
+      `Client:${apiCall.requesterAddress} is not authorized to access Endpoint ID:${apiCall.endpointId} for Request ID:${apiCall.id}`
+    );
+    const updatedApiCall = {
+      ...apiCall,
+      status: RequestStatus.Errored,
```
andreogle, line 61: Is it correct that a request is set to "Errored" if the authorization returns false? This will result in an error transaction at the end. Can a malicious user drain someone else's wallet by spamming unauthorised requests?

bbenligiray (approved):
Looks good :+1:

packages/node/src/core/ethereum/authorization/authorization-application.ts
andreogle: Yeah but does that stop a malicious client from spamming unauthorized requests? Since they all get set to `Errored` they still result in `error()` transactions using the requester's wallet, right? Maybe I'm misunderstanding that part

bbenligiray: I think you're imagining a use case where the dapp looks like this
https://clc-group.github.io/workshop-honeycomb-react/
where the dapp owner pays for the gas but the user freely triggers however many requests they want. In reality, the client (i.e. the dapp) would be implemented in a way that either makes a set request that cannot be parametrized by the user or the user puts down an ETH deposit that covers the gas fee of the request + possibly more etc.

PR #66, opened 2020-09-23
Decoupling Part 3

andreogle:
This PR is mostly just shuffling modules and functions around. I've also decoupled more functions from the general app state. This makes it much easier to test & adjust them if needed. There aren't many notable logic changes.

This is what the `core` directory structure looks like now. I'm not sold on all of the folder names yet but those are easy enough to change.

```js
// Different ways that Airnode can connect to the outside world
adapters/
  // The only "adapter" at the moment that executes aggregated API calls
  http/

// Airnode configuration settings
config/

// The "main" coordinator modules and state definition
coordinator/
  // Modules that aggregate, execute and disaggregate individual API calls (don't really like this folder name) 
  http/

// EVM specific logic for most of the app. Soon to be renamed "evm"
ethereum/
  authorization/
  contracts/
  requests/
  templates/
  transactions/
  triggers/

// Provider workers and state definition
providers/

// Generic modules for requests that we can hopefully re-use between between different blockchains
requests/

// General utility functions
utils/

// "Worker" function that allow for "forking"
workers/ 
```

Let me know what you think of this structure. It should be easier to change it going forward


bbenligiray (approved):
I agree, `coordinator/http/` doesn't sound accurate. `calls` maybe?
Can you copy paste the directory structure to a readme somewhere? I think it's very helpful.

andreogle:
Yeah `calls` is nicer. Thanks

I was thinking of putting something like this directory structure somewhere in the technical documentation anyway

PR #98, opened 2020-10-09
Dynamic config

andreogle:
## Note to self

Update the API3 docs with the relevant changes

### Settings

This pull request adds the ability for users to pass dynamic configuration to the node. The only options that can be changed are (for now):
1. The logging format: either JSON or "plain". I'm not sure where we want this to be configured, so it's hardcoded to "plain" for now.
2. The minimum number of confirmations. This defaults to 6 for now but has no impact on the node (yet)
3. The "block history limit" - how many blocks the node should look back when fetching events. This default to 600.

2 & 3 can be configured per provider, not by chain.

```js
{
  "chains": [
    {
      "id": 1,
      "type": "evm",
      "providers": [
        {
          "name": "infura-mainnet",
          "url": "https://mainnet.infura.io/...",
          "minConfirmations": 6, // optional
          "blockHistoryLimit": 600 // optional
        }
      ],
      // the below key and array items are all optional
      "contracts": [{ "name": "Airnode", "address": "0x123456" }] 
    }
  ]
}
```

### Contracts

On top of this, `contracts` can now also be (optionally) dynamically configured. Any contracts not provided, will use the Airnode defaults for that chain ID. After the contracts have been "built", there is some basic validation that happens: are all of the contracts present and do they start with `0x` (for EVM contracts).

NOTE: EVM contract addresses for chain ID: 1 (mainnet) cannot be overwritten

These contracts & settings are stored in the coordinator's state and then copied to each provider's state so the states now look like:

```js
{
  contracts: { Airnode: '0x123456, ... },
  settings: { blockHistoryLimit: 600, logFormat: 'plain', ... },
  ...remaining_state
}
```
### Logging Overhaul

The logging module and formats have been updated. The "main" functions are now:

1. `logger.info('Something happened', { format: 'json', meta: { coordinatorId: '1234', providerName: 'infura-mainnet', ... } })`. There are also `logger.warn`, `logger.debug` and `logger.error` helpers.

2. `logger.pend('INFO', 'Something happened')`

3. `logger.logPending(pendingLogs, { format: 'json', meta: { ... } })`

### "Main" functions moved to handlers/ directory

The entry point functions have been grouped and moved to a new `handlers/` directory. The idea is that it's easier to find the different places where the code starts. 

### Aggregated API calls are now stored by index

Previously, the coordinator would store aggregated (unique) API calls in a flat array. This has now been changed to store these requests keyed by `requestId`. This was done in anticipation of dropping the `walletDataByIndex` in favour of a simpler, flat array of all requests.


bbenligiray (approved):
Gave up halfway through. I'm okay with the changes in the PR note though.

Issue #76, opened 2020-10-07
Request fulfillment events changed


bbenligiray:
Request fulfillment events used to be
https://github.com/api3dao/airnode/blob/fd1fdf209ea22aba8b2bf980d4980011e850c955/packages/protocol/contracts/interfaces/IAirnode.sol#L41-L45
https://github.com/api3dao/airnode/blob/fd1fdf209ea22aba8b2bf980d4980011e850c955/packages/protocol/contracts/interfaces/IAirnode.sol#L47-L51
https://github.com/api3dao/airnode/blob/fd1fdf209ea22aba8b2bf980d4980011e850c955/packages/protocol/contracts/interfaces/IAirnode.sol#L53-L57
https://github.com/api3dao/airnode/blob/fd1fdf209ea22aba8b2bf980d4980011e850c955/packages/protocol/contracts/interfaces/IAirnode.sol#L59-L62
Now they are
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/interfaces/IAirnode.sol#L44-L49
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/interfaces/IAirnode.sol#L51-L56
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/interfaces/IAirnode.sol#L58-L61

The ABI needs to be updated

Issue #73, opened 2020-10-07
Request creation events changed


bbenligiray:
Request events used to be
https://github.com/api3dao/airnode/blob/fd1fdf209ea22aba8b2bf980d4980011e850c955/packages/protocol/contracts/interfaces/IAirnode.sol#L9-L19
https://github.com/api3dao/airnode/blob/fd1fdf209ea22aba8b2bf980d4980011e850c955/packages/protocol/contracts/interfaces/IAirnode.sol#L21-L27
https://github.com/api3dao/airnode/blob/fd1fdf209ea22aba8b2bf980d4980011e850c955/packages/protocol/contracts/interfaces/IAirnode.sol#L29-L39
Now they are
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/interfaces/IAirnode.sol#L9-L20
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/interfaces/IAirnode.sol#L22-L29
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/interfaces/IAirnode.sol#L31-L42
Types representing API call requests need to be updated.

bbenligiray:
Changes are due to #74 and #77

PR #102, opened 2020-10-13
Update API call request and fulfillment event names

andreogle:
Addresses: https://github.com/api3dao/airnode/issues/76 and https://github.com/api3dao/airnode/issues/73

1. Updated the contract ABIs for `Airnode` and `Convenience` by building the `protocol` package with `npm run build`

2. Updated the event names and topics


bbenligiray (approved):
Topic names not matching the event names is a bit triggering tbh

andreogle:
My thinking was that `ApiCallRequestCreated` is probably more intuitive than `ClientRequestCreated` for someone opening the code for the first time, but I can see how the names not matching the contract names is even more confusing and might cause problems. I'll update the topic names to be the same

bbenligiray:
It feels like the event names will change before we tie the final bow but it's better to have them consistent now even if that's the case

Issue #79, opened 2020-10-07
No need to fetch requester data from Convenience


bbenligiray:
The node used to get requester data from Convenience:
https://github.com/api3dao/airnode/blob/566fc7b72e1b22486cdc6e87d1e86c27d93ecacd/packages/node/src/core/evm/requests/requester-data.ts
It got `requesterId`, `walletAddress`, `walletInd`, `walletBalance`, `minBalance`.

Now, all requests include `requesterInd` and `designatedWallet`, there is no separate `walletInd`, and the minimum balance check is done at an Authorizer.

All code related to fetching requester data should be removed.

PR #104, opened 2020-10-15
Drop requester data

andreogle:
Addresses: https://github.com/api3dao/airnode/issues/79

Drops all requester data codes and references. I've kept the fields for now but hardcoded the values until they get removed towards the end.

Also dropped the `BaseRequest` type. This was basically the request before the requester data has been applied


bbenligiray (approved):
Awesome

Issue #85, opened 2020-10-07
Authorization check arguments have changed


bbenligiray:
Authorization checks used to be for endpoint-client pairs:
https://github.com/api3dao/airnode/blob/fd1fdf209ea22aba8b2bf980d4980011e850c955/packages/protocol/contracts/interfaces/IConvenience.sol#L48-L54

Now they are request specific and have a bunch of other arguments (`providerId` is not an array!):
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/interfaces/IConvenience.sol#L43-L53

Also note that the non-batch version of this method is also in Convenience (will be used as fallback if the batch version reverts):
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/interfaces/IConvenience.sol#L31-L41

andreogle:
It looks like the client `clientAddresses` argument is being sourced from the event `requester` field. Maybe we should rename this argument to be consistent? 

My understand is that generally `requester` will refer to a user, while `client` refers to one or more contracts, right?

bbenligiray:
Good point, that's a leftover. I pushed a commit to `protocol-overhaul` to fix that c2b7556

That also means that the contracts on testnets need to be redeployed. I assume they are not needed for now, let me know when they are.

andreogle:
Yeah not using the testnet contracts yet so you're free to update them as needed

PR #107, opened 2020-10-16
Authorization argument updates

andreogle:
Addresses: https://github.com/api3dao/airnode/issues/85

Passing in the extra arguments to the `checkAuthorizationStatuses` function calls


bbenligiray (approved):
Nice, I was concerned that this was going to be a lot more difficult

Issue #75, opened 2020-10-07
Fulfillment and failure method arguments changed


bbenligiray:
Fulfillment and failure arguments used to be
https://github.com/api3dao/airnode/blob/fd1fdf209ea22aba8b2bf980d4980011e850c955/packages/protocol/contracts/interfaces/IAirnode.sol#L95-L105
https://github.com/api3dao/airnode/blob/fd1fdf209ea22aba8b2bf980d4980011e850c955/packages/protocol/contracts/interfaces/IAirnode.sol#L131-L132
Now, they are
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/interfaces/IAirnode.sol#L93-L105
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/interfaces/IAirnode.sol#L121-L127
(`fail()` still doesn't call back `fulfillAddress`/`fulfillFunctionId`, it just needs these arguments to check `designatedWallet`)

This is done for the fulfillment parameters to be checked on-chain. In the previous state, the provider was able to fulfill an existing request using `providerId`, `designatedWallet`, `fulfillAddress` and `fulfillFunctionId` that are different than the ones defined by the requester.

Calls to these methods need to be updated.

Issue #74, opened 2020-10-07
Requests are no longer able errorable


bbenligiray:
There used to be a method to error a call
https://github.com/api3dao/airnode/blob/fd1fdf209ea22aba8b2bf980d4980011e850c955/packages/protocol/contracts/interfaces/IAirnode.sol#L119-L129
Now, the `errorCode` is returned in `statusCode` of `fulfill`:
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/interfaces/IAirnode.sol#L93-L105

All code about erroring needs to be removed. The node should try to call `fulfill()` and if that reverts, it should call `fail()` directly.

andreogle:
Just to confirm, is `statusCode` supposed to be `0` or `null` if no error happened?

bbenligiray:
`0`

andreogle:
If the request errored and has no response data to return, what should be used for the `data` argument here? The value shouldn't matter since the client should check `statusCode` first, but I still need to call the `fulfill` function with something. 

Should I just use `0x0`? I'm sure if this is a valid value yet though

bbenligiray:
Yes, `0x0` (or rather `ethers.constants.HashZero`) is fine

PR #109, opened 2020-10-18
Drop error transaction calls and update arguments

andreogle:
Addresses: https://github.com/api3dao/airnode/issues/75 & https://github.com/api3dao/airnode/issues/74

1. Updates the arguments to `fulfill()` and `fail()`
2. Drops all references to `error()`
3. Updated some log messages with status code

As discussed, status code will be `0` for successful requests. For errored requests, the `responseValue` with be `ethers.constants.HashZero`


bbenligiray (approved):
:shipit:

Issue #81, opened 2020-10-07
Need to verify template ID-contents


bbenligiray:
Template ID is derived from its contents:
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/TemplateStore.sol#L70-L78

When the node receives a template, it should verify this to ensure that the Ethereum provider didn't tamper with any of the template fields.

bbenligiray:
Example:
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/test/UserFlow.js#L221-L238

PR #111, opened 2020-10-21
Verify templates

andreogle:
Addresses: https://github.com/api3dao/airnode/issues/81

1. Adds a `template-verification` module. Templates are verified before the contents are mapped into the API calls. If a template is invalid, I'm assume that something has been maliciously modified so the API call is set to `Ignored`. What do you think about this? Also, if an API call is not pending at template application time, then the template is not applied. This is existing functionality. 

2. I've adjusted the `template-application` module to work with a flat list of API calls. This is anticipation of dropping `walletDataByIndex` shortly. It made things significantly easier to integrate with the new verification module. I did have to add a temporary workaround where the API calls and withdrawals are regrouped again.


packages/node/src/core/evm/templates/template-verification.test.ts
```diff
@@ -0,0 +1,129 @@
+import { ethers } from 'ethers';
+import * as fixtures from 'test/fixtures';
+import * as verification from './template-verification';
+import { RequestErrorCode, RequestStatus } from 'src/types';
+
+describe('verify', () => {
+  const TEMPLATE_ID = '0x101c36202b92b358adda664515ea920f76b1edcf5f1285ab0c845f54638b45aa';
```
andreogle, line 7: For the verification tests, I used actual valid values, as this was necessary to actually validate templates. Most of the other unit tests in the app use invalid values like `apiCallId` or `templateId`. I'm still undecided which way is preferable.

Invalid values are much easier to read at a glance. Valid values are obviously closer to the real thing.

packages/node/src/core/evm/templates/template-verification.test.ts
```diff
@@ -0,0 +1,129 @@
...
+      encodedParameters: '0x1234',
+      endpointId: ethers.constants.HashZero,
+      fulfillAddress: ethers.constants.AddressZero,
+      fulfillFunctionId: '0x12341464',
+      providerId: ethers.constants.HashZero,
+      requesterIndex: '2',
+    };
+
+    Object.keys(invalidFields).forEach((field) => {
+      it(`is invalid if ${field} has been changed`, () => {
```
andreogle, line 110: Pretty neat thing: you can do dynamically named tests. This is preferable to just having everything in a single test and looping through values as it's easier to debug when one field is causing problems

packages/node/src/core/evm/templates/template-verification.test.ts
```diff
@@ -0,0 +1,129 @@
...
+      endpointId: '0x2f3a3adf6daf5a3bb00ab83aa82262a6a84b59b0a89222386135330a1819ab48',
+      fulfillAddress: '0x197F3826040dF832481f835652c290aC7c41f073',
+      fulfillFunctionId: '0xd3bd1464',
+      id: TEMPLATE_ID,
+      providerId: '0x9e5a89de5a7e780b9eb5a61425a3a656f0c891ac4c56c07037d257724af490c9',
+      requesterIndex: '1',
+    });
+
+    const invalidFields = {
+      designatedWallet: ethers.constants.AddressZero,
```
andreogle, line 100: Note to self. I think I would rather use "real" addresses instead of the zero addresses so I'll update this

packages/node/src/core/evm/templates/template-verification.ts
```diff
@@ -0,0 +1,89 @@
...
+): LogsData<ClientRequest<ApiCall>[]> {
+  const logsWithVerifiedApiCalls: LogsData<ClientRequest<ApiCall>>[] = apiCalls.map((apiCall) => {
+    if (!apiCall.templateId) {
+      return [[], apiCall];
+    }
+
+    // If the request is already errored, blocked or ignored, then we don't want to make further
+    // updates to either the status or error code
+    const ignoredStatuses = [RequestStatus.Blocked, RequestStatus.Errored, RequestStatus.Ignored];
+    if (ignoredStatuses.includes(apiCall.status)) {
```
andreogle, line 43: Note to self: this should ignore ALL non-pending statuses

packages/node/src/core/evm/templates/template-verification.test.ts
bbenligiray: I think only using valid values when you have to makes sense

packages/node/src/core/evm/templates/template-application.ts
```diff
@@ -46,82 +39,67 @@ function mergeRequestAndTemplate(
...
+      status: RequestStatus.Blocked,
+      errorCode: RequestErrorCode.TemplateNotFound,
+    };
+    return [[log], updatedApiCall];
+  }
+
+  // Attempt to decode the template parameters
+  const templateParameters = evm.cbor.safeDecode(template.encodedParameters);
+
+  // If the template contains invalid parameters, then we can't use execute the request
```
bbenligiray, line 75:
```suggestion
  // If the template contains invalid parameters, then we can't use it to execute the request
```

packages/node/src/core/evm/templates/template-verification.ts
bbenligiray: Why not check for if the request status is pending?

bbenligiray (approved):
1. Agreed with both, as talked about today.
2. Got it

packages/node/src/core/evm/templates/template-verification.ts
andreogle: I'm going to update this to this which is consistent with other modules:

```ts
if (apiCall !== RequestStatus.Pending) {
  ...
}
```

I want to be very careful with updating statuses as it could potentially have a critical impact if something goes wrong. All API calls start out as "pending" and then can only have their status (and error code) changed once after that. I might make a helper that updates status & error code that also checks that status is pending first.

Issue #82, opened 2020-10-07
Need to verify request ID-parameters


bbenligiray:
Request IDs are derived from request parameters. The derivation changes with type:

Regular:
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/Airnode.sol#L51-L55
Short:
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/Airnode.sol#L100-L104
Full:
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/Airnode.sol#L156-L161

When the node receives a request, it should verify this derivation to ensure that the Ethereum provider didn't tamper with any of the request parameters.

bbenligiray:
Example:
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/test/UserFlow.js#L212-L219
Would be slightly different for a full request

andreogle:
Just to be clear, should this validate withdrawals too?

bbenligiray:
No, the node already provides all the withdrawal request parameters while fulfilling it, so we can do the check on-chain
https://github.com/api3dao/airnode/blob/c17870b5884aa25464504deb94cf02e51ff45f05/packages/protocol/contracts/ProviderStore.sol#L130-L139
In theory, we can also do the check for API requests on-chain, but it would require the node to send all the request parameters back (`parameters` for example), which would create quite the gas cost overhead. So there is a somewhat arbitrary tradeoff decision that I made here.

Issue #80, opened 2020-10-07
Need to verify designatedWallet-requesterInd


bbenligiray:
All requests pass `designatedWallet` and `requesterInd`.

When the node receives a request, it should verify that `designatedWallet` is the address of `m/0/${requesterInd}` and ignore the request if this is not the case.

bbenligiray:
Example:
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/test/UserFlow.js#L240-L244

andreogle:
This will need to be done after templates have been applied, right? Short requests don't have `requesterInd` and `designatedWallet` until that point

bbenligiray:
Yes

PR #113, opened 2020-10-23
Verify request IDs and designated wallets

andreogle:
Addresses: https://github.com/api3dao/airnode/issues/80 and https://github.com/api3dao/airnode/issues/82

1. Adds a new "verification" module to the EVM folder. There are 2 new verification functions `verifyDesignatedWallets` and `verifyRequestIds`. Each of these accepts an array of API call requests and returns an array of API call requests.

2. Request IDs are verified immediately after they have been mapped from the raw event logs.

3. Designated wallets are verified later - after templates have been fetched and applied.

4. Added a `type` field to API call requests that is `short`, `regular` or `full`. This is set based on the topic for each log.

I've added some more workarounds for `walletDataByIndex` but they will be easy to remove later.


bbenligiray (approved):
Thick. Solid. Tight.

Issue #78, opened 2020-10-07
Requests don't need to be compared deeply


bbenligiray:
Request IDs are derived from request parameters, for example:
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/Airnode.sol#L51-L55
and these IDs are verified when the request is received. After this point, two requests with the same IDs have to be identical

Here, we only need to test for `requestId`:
https://github.com/api3dao/airnode/blob/fd1fdf209ea22aba8b2bf980d4980011e850c955/packages/node/src/core/requests/api-calls.ts#L7

bbenligiray:
We can skip this because we'll now be doing #82

PR #114, opened 2020-10-23
Aggregated requests

andreogle:
Addresses: https://github.com/api3dao/airnode/issues/78

1. It is no longer necessary to aggregate API calls into an array of requests where the ID is the same, but other parameters are different. Requests with the same ID are now considered to be the same. So, API calls are now stored simply by their ID (not in an array)


bbenligiray (approved):
Big improvement :ok_hand:

Issue #87, opened 2020-10-08
Withdrawal transactions should estimate gas


bbenligiray:
There are two withdrawals:
- Made upon a withdrawal request
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/ProviderStore.sol#L120
- Made after a provider record is created
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/ProviderStore.sol#L31

In both cases, we need the node to send `balance - txCost` to the destination. Here, we should estimate the gas cost to estimate the transaction cost
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/test/UserFlow.js#L131-L147
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/test/UserFlow.js#L297-L321

This (i.e. not hardcoding the gas cost) good for three reasons:
1. Future forks may change instruction gas costs
2. The destination may be a `payable` contract instead of a regular wallet, in which case the gas cost would be variable
3. Gas cost may vary with the chain

and withdrawals won't be common, so the additional Ethereum call is not significant.

bbenligiray:
My bad, withdrawals already estimate gas.

bbenligiray:
Reopened for referencing

PR #115, opened 2020-10-24
Update withdrawals

andreogle:
Addresses: https://github.com/api3dao/airnode/issues/87

1. Rename the `evm/transactions/` directory to `evm/fulfillments`. I'm not sure that I want to include other future transaction logic with the fulfillment logic so I've separated this. 

2. Fix the `fulfillWIthdrawal` arguments.

3. Pad the estimated withdrawal gas limit by 20 000.

4. Validate the withdrawal amount is positive. Ignore the request and log an error if the amount is negative. Maybe it should be `WARN` instead of `ERROR` though since the API provider doesn't have anything to action here so they don't really need to know about this happening. Or even `INFO`?

5. Clean up the entry point for submitting API call and withdrawals to always output a log entry, regardless of status


packages/node/src/core/evm/fulfillments/withdrawals.ts
```diff
@@ -0,0 +1,108 @@
...
+    );
+    return [[log], null, null];
+  }
+
+  const estimateTx = airnode.estimateGas.fulfillWithdrawal(
+    request.id,
+    request.providerId,
+    request.requesterIndex,
+    request.destinationAddress,
+    // We need to send some funds for the gas price calculation to be correct
```
bbenligiray, line 28: The receiving address may be a `payable` contract that does one thing when it receives 1 Wei and  another thing when it receives 1 ETH so this estimation is not guaranteed to be correct. I'm being pedantic though, this is a silly thing to be concerned about.
I guess the guaranteed way of dealing with is having the user provider the gas limit and gas price, but that's terrible UX.

bbenligiray (commented):
4. `INFO` sounds right to me.

There are a bunch of references to `walletIndex`. Shouldn't you switch to `requesterIndex`?

packages/node/src/core/evm/fulfillments/withdrawals.ts
andreogle: That's not a problem at all. I can just shuffle the code around so that the balance is fetched first then that is used as the value (before subtracting the transaction cost). I guess that would be more accurate than 1 Wei.

Good catch on the `walletIndex` references.

bbenligiray: If you don't subtract it it would probably revert. I think leaving it as is will be cleaner.

Issue #84, opened 2020-10-07
Node needs to check for provider data at the start and create a record if necessary


bbenligiray:
Instead of getting the block number while initalizing the providers, the node should call Convenience to get the block number + provider data:
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/interfaces/IConvenience.sol#L9-L16
Before calling this, the node computes `providerId` off-chain as `keccak256(address of path 'm')`. If this method returns `xpub` as `''`, it means that the record for this provider hasn't been created on this chain yet. Then, the node should call this:
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/contracts/interfaces/IProviderStore.sol#L36-L42
using the wallet with path `m`, where `admin is` from `config.json` and `xpub` is derived from the private key.

Note that this method is `payable`. This is because the provider will have to send some ETH to the wallet with path `m` for the node to be able to do this. While doing this, the node will send the remaining funds to `admin`.

bbenligiray:
See:
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/test/UserFlow.js#L120
Note that the master wallet is funded beforehand here:
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/test/UserFlow.js#L92

PR #117, opened 2020-10-27
Find or create provider with current block

andreogle:
Addresses: https://github.com/api3dao/airnode/issues/84

1. Updated the contract ABIs

2. Added a new `evm/providers.ts` module that contains several functions:
a. `findWithBlock` which calls `convenience.getProviderAndBlockNumber`
b. `create` which calls `airnode.createProvider` if xpub is returned as falsey (emptystring)
c. `findOrCreateProviderWithBlock` which does both

3. Added a new `computeProviderId` function that uses the mnemonic to determine that `providerId` - in the same way that `UserFlow.js` does. Is this necessary though or does the `nodeSettings.providerId` field not exist on the first run?


packages/node/src/core/providers/initialize.test.ts
```diff
@@ -47,13 +43,19 @@ import * as providers from './initialize';
...
+      xpub: '0xxpub1',
+    });
+    contract.getProviderAndBlockNumber.mockResolvedValueOnce({
+      admin: '0xadmin2',
+      blockNumber: ethers.BigNumber.from(987654),
+      xpub: '0xxpub2',
+    });
 
-    const getLogs = provider.getLogs as jest.Mock;
+    const getLogs = jest.spyOn(ethers.providers.JsonRpcProvider.prototype, 'getLogs');
```
andreogle, line 58: It's better to use `spyOn` rather than mocking the entire module, then unmocking everything again with `jest.requireActual()` except for certain modules. Not necessary to update the older code, it's just something I've noticed.

andreogle: I haven't tried to fix this for the `ethers.Contract` parts yet though. Just had problems with this line so that's why I fixed this specifically

packages/node/src/core/evm/providers.ts
```diff
@@ -0,0 +1,113 @@
...
+    // should always be a valid number
+    blockNumber: res.blockNumber.toNumber(),
+    providerExists: !!res.xpub && res.xpub !== '',
+    xpub: res.xpub,
+  };
+
+  const blockLog = logger.pend('INFO', `Current block: ${res.blockNumber}`);
+
+  if (!data.providerExists) {
+    const providerLog = logger.pend('INFO', 'EVM provider not found');
```
bbenligiray, line 57: The "provider" in the protocol is short for "API provider", though I never use it as such.  I think this "EVM provider" is referring to the same provider, so it's not accurate.

packages/node/src/core/evm/providers.ts
```diff
@@ -0,0 +1,113 @@
...
+    const logs = [fetchLog, blockLog, providerLog];
+    return [logs, data];
+  }
+
+  const addressLog = logger.pend('INFO', `Admin address: ${res.admin}`);
+  const xpubLog = logger.pend('INFO', `Admin extended public key: ${res.xpub}`);
```
bbenligiray, line 63: Admin extended public key -> Provider extended public key

There are two separate entities:
The node (or provider), which has a private key, its mnemonic, xpub, designatedWallets, etc.
The provider admin, which is an Ethereum account that is probably controlled by some employee of the API provider, which has weaker credentials compared to the node itself.

packages/node/src/core/evm/providers.test.ts
```diff
@@ -0,0 +1,272 @@
...
+    getProviderAndBlockNumberMock.mockResolvedValueOnce({
+      admin: '0xadminAddress',
+      blockNumber: ethers.BigNumber.from('12'),
+      xpub: '',
+    });
+    const [logs, res] = await providers.findWithBlock(options);
+    expect(logs).toEqual([
+      { level: 'INFO', message: 'Fetching current block and provider admin details...' },
+      { level: 'INFO', message: 'Current block: 12' },
+      { level: 'INFO', message: 'EVM provider not found' },
```
bbenligiray, line 63:
```suggestion
      { level: 'INFO', message: 'Provider not found' },
```

bbenligiray: See below

packages/node/src/core/evm/providers.test.ts
```diff
@@ -0,0 +1,272 @@
...
+    const [logs, res] = await providers.findOrCreateProviderWithBlock(options);
+    expect(logs).toEqual([
+      {
+        level: 'DEBUG',
+        message:
+          'Computed provider ID from mnemonic: 0x9e5a89de5a7e780b9eb5a61425a3a656f0c891ac4c56c07037d257724af490c9',
+      },
+      { level: 'INFO', message: 'Fetching current block and provider admin details...' },
+      { level: 'INFO', message: 'Current block: 12' },
+      { level: 'INFO', message: 'EVM provider not found' },
```
bbenligiray, line 223:
```suggestion
      { level: 'INFO', message: 'Provider not found' },
```

bbenligiray: See below

packages/node/src/core/evm/providers.test.ts
```diff
@@ -0,0 +1,272 @@
...
+      {
+        level: 'DEBUG',
+        message:
+          'Computed provider ID from mnemonic: 0x9e5a89de5a7e780b9eb5a61425a3a656f0c891ac4c56c07037d257724af490c9',
+      },
+      { level: 'INFO', message: 'Fetching current block and provider admin details...' },
+      { level: 'INFO', message: 'Current block: 12' },
+      { level: 'INFO', message: 'Admin address: 0xadminAddress' },
+      { level: 'INFO', message: 'Admin extended public key: 0xxpub' },
+      { level: 'DEBUG', message: 'Skipping provider creation as the EVM provider exists' },
```
bbenligiray, line 258:
```suggestion
      { level: 'DEBUG', message: 'Skipping provider creation as the provider exists' },
```

bbenligiray: See below

packages/node/src/core/evm/providers.ts
bbenligiray: I didn't have any problems with this implementing the protocol because I didn't deal with EVM providers much but obviously it's pretty confusing. I'm open to suggestions.

packages/node/src/core/evm/providers.ts
```diff
@@ -0,0 +1,113 @@
...
+      ...options,
+      adminAddress: providerBlockData.adminAddress,
+      xpub: providerBlockData.xpub,
+    };
+    const [createLogs, _providerId] = await create(createOptions);
+    const logs = [idLog, ...providerBlockLogs, ...createLogs];
+    return [logs, providerBlockData];
+  }
+
+  const existsLog = logger.pend('DEBUG', `Skipping provider creation as the EVM provider exists`);
```
bbenligiray, line 111:
```suggestion
  const existsLog = logger.pend('DEBUG', `Skipping provider creation as the provider exists`);
```

packages/node/src/core/evm/providers.ts
```diff
@@ -0,0 +1,113 @@
...
+    return [logs, null];
+  }
+
+  // If the extended public key was returned as an empty string, it means that the provider does
+  // not exist onchain yet
+  if (!providerBlockData.providerExists) {
+    const createOptions = {
+      ...options,
+      adminAddress: providerBlockData.adminAddress,
+      xpub: providerBlockData.xpub,
```
bbenligiray, line 104: Isn't `providerBlockData.xpub` an empty string here? This should be derived from the mnemonic.
https://github.com/api3dao/airnode/blob/95b1e663a7fff81c27a72dd3a7650636fea54f18/packages/protocol/test/Airnode.js#L175-L176

packages/node/src/core/evm/providers.ts
```diff
@@ -0,0 +1,113 @@
...
+    const logs = [idLog, ...providerBlockLogs];
+    return [logs, null];
+  }
+
+  // If the extended public key was returned as an empty string, it means that the provider does
+  // not exist onchain yet
+  if (!providerBlockData.providerExists) {
+    const createOptions = {
+      ...options,
+      adminAddress: providerBlockData.adminAddress,
```
bbenligiray, line 103: If the provider isn't created yet, this will return `0x0`. This should be read from `nodeSettings.chains.*.adminAddress`.

packages/node/src/core/evm/wallet.ts
```diff
@@ -20,6 +20,13 @@ function getPathFromIndex(index: number | string) {
   return `m/0/0/${index}`;
 }
 
+export function computeProviderId(provider: ethers.providers.JsonRpcProvider) {
+  const masterHdNode = ethers.utils.HDNode.fromMnemonic(security.masterKeyMnemonic);
+  const designatorHdNode = masterHdNode.derivePath('m');
```
bbenligiray, line 25: This line probably does nothing. You can remove it unless you wanted to be explicit.

bbenligiray (changes requested):
One thing to note: If the node creates the provider record with the wrong `xpub`, there is no easy way of fixing that. The user would have to run a script that extracts the mnemonic from SSM and re-create the provider record (and there is no such script as of now). I think this is okay because I don't really see it happening.

3. `nodeSettings.providerId` does not exist on the first run.

`adminAddress` needs to be sourced from `config.json` and `xpub` needs to be derived from the mnemonic.

packages/node/src/core/evm/providers.ts
andreogle: Thanks, you're right. I didn't think too much about these log messages. There are so many uses of the word "provider", it all gets quite confusing 😅

packages/node/src/core/evm/providers.ts
andreogle: Correct 👍🏻

packages/node/src/core/evm/providers.ts
andreogle: I don't have any better suggestions at the moment

packages/node/src/types.ts
```diff
@@ -285,8 +286,9 @@ export interface ChainProvider {
 }
 
 export interface ChainConfig {
-  readonly id: number;
+  readonly adminAddress: string;
```
bbenligiray, line 289: This is not needed if the provider record has already been created. Should we make this optional? We would be increasing the complexity of what a valid `config.json` is, but having the user enter a placeholder is also bad.

andreogle: Assuming we don't change the structure of `config.json`, I prefer to just leave it as required. I think having certain fields required and then later optional will just end up more confusing. 

Plus it means more checks/error catching in the node if the field exists before accessing it.

What do you think?

bbenligiray: The only problem is that the user will change the `adminAddress` in `config.json` thinking that this will do anything after the provider is created, but having this field optional doesn't solve this problem.

Renaming this here and in `config.json` should help though, something like `adminAddressForCreatingProviderRecord`.

packages/node/src/core/evm/providers.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  const addressLog = logger.pend('INFO', `Admin address:${res.admin}`);
+  const xpubLog = logger.pend('INFO', `Provider extended public key:${res.xpub}`);
+  const logs = [fetchLog, blockLog, addressLog, xpubLog];
+  return [logs, data];
+}
+
+export async function create(options: CreateOptions): Promise<LogsData<ethers.Transaction | null>> {
+  const airnode = new ethers.Contract(options.airnodeAddress, Airnode.ABI, options.provider);
+  const contractCall = () => airnode.createProvider(options.adminAddress, options.xpub);
```
bbenligiray, line 71: Oops, forgot about something. `createProvider` is `payable` for the wallet to be able to send back the remaining amount to `providerAdmin`.
https://github.com/api3dao/airnode/blob/protocol-overhaul/packages/protocol/test/UserFlow.js#L131-L147

Also you didn't specify the wallet/signer making this call so ethers will probably error. It should be the `m` wallet.

packages/node/src/types.ts
```diff
@@ -119,7 +119,7 @@ export interface WalletDataByIndex {
 }
 
 export interface ProviderSettings {
-  readonly adminAddress: string;
+  readonly adminAddressForCreatingProviderRecord?: string;
```
bbenligiray, line 122: Did we not decide to make this required?

bbenligiray (approved):

PR #119, opened 2020-10-29
General execution fixes

andreogle:
A few general fixes related to running the node

1. Renamed `requesterAddress` to `clientAddress` to be consistent with the contracts

2. Update the contract default address for development network 1337

3. Set the GasPricedFeed address to `ethers.constants.AddressZero`. The gas price calculation will also not try to fetch the gas price if the address is `AddressZero`. This was outputting unnecessary errors when running locally.


packages/node/src/core/evm/gas-prices.test.ts
```diff
@@ -22,18 +22,34 @@ import * as utils from './utils';
 import * as gasPrices from './gas-prices';
 
 describe('getGasPrice', () => {
-  const baseFetchOptions = {
+  const baseOptions = {
     address: '0x3071f278C740B3E3F76301Cf7CAFcdAEB0682565',
   };
+
+  it('only checks the node price if the contract address is AddressZero', async () => {
```
bbenligiray, line 29:
```suggestion
  it('checks only the node price if the contract address is AddressZero', async () => {
```

bbenligiray: Made me scratch my head a bit :)

bbenligiray (approved):

PR #121, opened 2020-10-30
Validator script

### LINKED ISSUE #118: Running the validator as a script
It would be nice to be able to run the validator from the command line, passing in a JSON file along with a template that isn't hardcoded

Something like:

```sh
npm run validate --file=config.json --template=templates/api.json
```

Drgy:
Usage explained in README:

> Validator can be run as a npm script, by providing specifications file that will be validated and template (validator specification structure) file: `npm run validate --specs="[specsFile]" --template="[templateFile]"`

> In case specifications file is provided first, the command can be simplified to: `npm run validate [specsFile] [templateFile]`

closes #118


andreogle:
@Drgy would you be able to merge master into this branch to resolve the conflicts please?

packages/validator/src/validate.ts
```diff
@@ -0,0 +1,49 @@
...
+    return validateSpecs(parsedSpecs, parsedTemplate, '', nonRedundant, {
+      specs: parsedSpecs,
+      nonRedundantParams: nonRedundant,
+    });
+  } catch (e) {
+    return { valid: false, messages: [{ level: 'error', message: `${e.name}: ${e.message}` }] };
+  }
+}
+
+if (process.env.npm_config_specs || process.env.npm_config_template) {
```
andreogle, line 45: Using env variables isn't mentioned in the README. Worth adding it there I think.

Also, I think let's drop the `npm_` prefix as it implies it's an NPM setting and uppercase the name for consistency.

packages/validator/src/validate.ts
```diff
@@ -0,0 +1,49 @@
...
+  console.log(validate(process.env.npm_config_specs, process.env.npm_config_template));
+} else {
+  console.log(validate(process.argv[2], process.argv[3]));
```
andreogle, line 48: Does process.argv work when providing arguments in the form of `--variable=value`. I thought this would mean they come through as just that, so `process.argv[2]` would be `--specs=config.json` when we want just `config.json`.

It might be worth using a very basic [yargs](https://github.com/yargs/yargs) setup. Burak is also using yargs the `aws-deployment` branch, although that usage is much more complex

packages/validator/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules/
```
andreogle, line 1: 👍🏻

andreogle (changes requested):
Sorry for the late review. I was really expecting a notification to let me know when to review (although I may have missed it too)

packages/validator/src/validate.ts
Drgy: Its not supposed to be used as environment variables, its how npm passes arguments when running script with arguments in `--variable="value"` format, the value of `variable` can be accessed as `process.env.npm_config_variable`. Thats the reason no environment variables are mentioned in readme and prefix must stay.

packages/validator/src/validate.ts
Drgy: If arguments are passed as `--variable="value"` first part of condition is executed, otherwise it will try to retrieve regular whitespace separated arguments and execute with them.
yargs seems like powerful tool, isnt it a bit overkill for this though? Imho this is a simple and elegant solution for this problem, allowing to specify arguments in 2 different ways without any robust 3rd party tools.

packages/validator/src/validate.ts
andreogle: Wasn't aware of that behaviour. Good to know, thanks

packages/validator/src/validate.ts
andreogle: Yeah I'm happy with this then. No need for yargs

andreogle (approved):

packages/validator/README.md
```diff
@@ -20,6 +20,9 @@ Where array `messages` contains message objects:
 ```
 If provided OIS is valid, parameter `valid` will be true, however parameter `messages` still might contain messages with `level` set to `warning`. If `valid` is `false`, there will be always one or more error messages. Similarly functions `isApiSpecsValid` and `isEndpointsValid` check only the API and endpoint specifications contained in the OIS.
 
+Validator can be run as a npm script, by providing specifications file that will be validated and template (validator specification structure) file: `npm run validate --specs="[specsFile]" --template="[templateFile]"`
```
andreogle, line 23: This wording sounds a bit weird to me so I've tried to reword it a bit. What do you think?

"
The validator can be run as an NPM script, by providing the paths to the JSON file that will be checked and the JSON file to use as the template (validator specification structure). You can run the validator as script using the following command:

```sh
npm run validate --specs="[path-to-specs-json]" --template="[path-to-template-json]"
```
"

Issue #120, opened 2020-10-30
Get secrets from environment variables


bbenligiray:
Currently, secrets are being read from `security.json`. Instead, they should be read from the environment variables named as:
`MASTER_KEY_MNEMONIC`
`${oisTitle}_${securityScheme.securitySchemeName}`

The node shouldn't require `security.json` to be built/run, and the deployer needs to be updated accordingly.

Issue #103, opened 2020-10-14
Provide config when starting the coordinator


andreogle:
This was briefly discussed internally today. The node having a hard build dependency on `config.json` (and `security.json` for now) existing is not ideal. The idea for now is to move this to the coordinator.

So, instead of starting the coordinator as:

```js
import { config } from '../core/src/config';

await coordinator.start(config.nodeSettings);
```

it would be called as something like:

```js
await coordinator.start({ config });
```
where `config` is being passed in by the deployer or the cloud provider (Lambda) itself on each invocation. This will probably need some experimentation to get right.

bbenligiray:
I'll wait for you to implement the local version and I'll match that at the deployer. I don't want to break the tests.

PR #127, opened 2020-11-08
Decouple config.json from the node package

andreogle:
### Changes
1. Remove the importing of `config.json` and `security.json` from the node. `config` is now expected to be provided when calling: `start()` and a `security` object is no longer required. `config` is an expected parameter for all cloud provider functions, not just the `startCoordinator` function, as they all need this context. After receiving the `config` initially, it is stored in the relevant state object. For the coordinator, this means in the `CoordinatorState`. For provider related workers, it means in the `ProviderState`. The config is "scrubbed" from the provider state before being returned to the coordinator. This is because we don't want to duplicate the config in the coordinator state (in case it's quite large" 

The node can now build without any external file dependencies because of this change.

2. The following environment variables are now required when running the node:

- `MASTER_KEY_MNEMONIC` - the mnemonic phrase of the master wallet

Security credentials in the form of:
- `<OIS title>_<security scheme name>`

3. All references to "serverless" have now been dropped from the `node` package. The node only exposes functions with the signatures necessary for the given cloud provider.

4. A new `__dev__` folder has been added with an `invoke.ts` script that can be run with `npm run invoke:aws`. This simulates a coordinator run using AWS configuration. It also loads environment variables from a `.env` file inside of the `__dev__` folder. A valid `config.json` file is also expected in this folder.  Examples of both `config.json` and `.env` are provided in the `__dev__` folder.

5. Renamed the `start()` function to `startCoordinator`, I'm anticipating usage looking like: 

```ts
import * as airnode from '@airnode/node'; 
airnode.AWS.startCoordinator({ config })`
```
which is a bit clearer than just `airnode.start();`

6. Lots of testing cleanup. All of the `jest.mock('../config')` stuff has been dropped as we no longer need to mock the `config.json` file.

7. A few small bug fixes here and there

### Addresses
- https://github.com/api3dao/airnode/issues/120
- https://github.com/api3dao/airnode/issues/103


packages/node/test/setup/set-env-vars.ts
```diff
@@ -0,0 +1 @@
+process.env.MASTER_KEY_MNEMONIC = 'achieve climb couple wait accident symbol spy blouse reduce foil echo label';
```
andreogle, line 1: This file gets run when tests start (so we don't need to set it everywhere)

packages/node/src/core/providers/state.ts
```diff
@@ -23,17 +24,19 @@ export function createEVMState(
     blockHistoryLimit: chainProvider.blockHistoryLimit || 600,
     chainId: chain.id,
     chainType: 'evm' as ChainType,
-    logFormat: settings.logFormat,
+    logFormat: config.nodeSettings.logFormat,
     minConfirmations: chainProvider.minConfirmations || 6,
     name: chainProvider.name,
-    providerId: settings.providerId,
+    // TODO: derive from mnemonic
```
bbenligiray, line 30: It would be good to derive `providerIdShort` with `providerId`. It feels like we may need to change the derivation rule (first 7 characters of `providerId`) in the future so doing the derivation once at the start will make it easier to modify.
I also need this `providerIdShort` to invoke the Lambda function.

packages/node/test/setup/set-env-vars.ts
bbenligiray: It was looking awkward :pray:

bbenligiray (approved):
The new fixtures look very clean. It all looks good to me and I will extract the cloud provider handlers to `deployer`

packages/node/src/core/providers/state.ts
bbenligiray: Or I guess `providerIdShort` would need to be in the coordinator state

andreogle: I don't think we need to keep `providerId` or `providerIdShort` in the state anymore. If we're deriving it from `MASTER_KEY_MNEMONIC`, then we can just derive it when needed. Something like:

```ts
export function getMasterWallet(provider?: ethers.providers.JsonRpcProvider) {
  const mnemonic = config.getMasterKeyMnemonic();
  const masterHdNode = ethers.utils.HDNode.fromMnemonic(mnemonic);
  return new ethers.Wallet(masterHdNode.privateKey, provider);
}

export function getProviderId(provider?: ethers.providers.JsonRpcProvider) {
  const masterWallet = getMasterWallet(provider);
  return ethers.utils.keccak256(ethers.utils.defaultAbiCoder.encode(['address'], [masterWallet.address]));
}

export function getProviderIdShort(provider?: ethers.providers.JsonRpcProvider) {
  return getProviderId(provider).substring(2, 9);
}
```
What do you think?

bbenligiray: We shouldn't need a provider to derive `providerId`, this also works
```js
const masterHdNode = ethers.utils.HDNode.fromMnemonic(mnemonic);
const providerId = ethers.utils.keccak256(ethers.utils.defaultAbiCoder.encode(['address'], [masterHdNode.address]));
```
but it still takes almost 80ms. Deriving from the private key directly
```js
const masterWallet = new ethers.Wallet(privateKey);
const providerId = ethers.utils.keccak256(ethers.utils.defaultAbiCoder.encode(['address'], [masterWallet.address]));
```
takes 55ms so it's not only about the mnemonic. It's the first line that uses 90%+ of that time, which makes me think that we should use a common `HDNode` for deriving different `designatedWallet`s (maybe something to think about for the future).

I guess we will run a profiler at some point to detect these inefficiencies but we can be preemptive about this one. `providerId` gets used a lot and there is one per node so no reason not to derive it only once (though deriving `providerIdShort` should be fine).

andreogle: Thanks. I'm removing the `provider` argument for now as it's unnecessary

Even 55ms is a long time to derive the `providerId` so it's probably better to cache this (and a common `HDNode`) in the state. I'll make this change now

Issue #123, opened 2020-11-03
Require contracts in nodeSettings.chains of config.json


bbenligiray:
`contracts` will be a required key in `nodeSettings.chains` of `config.json`. Its value will be in the form:
```js
{
  Airnode: "0x123...456",
  Convenience: "0xabc...def"
}
```

andreogle:
Do we want any validation to protect the user here? i.e. for chain ID 1, should it check the address against a hardcoded address? Or do we just take whatever is configured?

bbenligiray:
I think the node should take whatever is configured, but we can have the `validator` display a warning

PR #128, opened 2020-11-11
Protocol overhaul cleanup

andreogle:
### Changes

1. Remove the following fields from API calls and withdrawals as they are no longer needed:

```
- errorAddress
- errorFunctionId
- requesterId
- walletIndex
- walletAddress
- walletBalance
- walletMinimumBalance
```

2. Update `config.nodeSettings` structure and typings to be in line with https://github.com/api3dao/api3-docs/blob/master/airnode/2-7-config-json.md#nodesettings

3. Cache `providerIdShort` in the provider state. I'm not sure if this is really necessary, but I don't think it hurts to have it easily available.

4. `chains.x` objects now expect a `contracts` object to always be present. The 2 expected keys are `Airnode` and `Convenience`. There is no validation for this in the node.

5. Update the default `minConfirmations` for each EVM provider to be 0, although this still isn't used yet

### Linked issues
https://github.com/api3dao/airnode/issues/123


bbenligiray (commented):
`providerIdShort` is needed in the coordinator state, even before the providers are initialized because it's used to build the Lambda function name to initialize the provider.

PR #129, opened 2020-11-15
Fix EVM initialize provider flow 

andreogle:
### Changes

1. Fixes the flow of initializing an EVM provider to be inline with the latest flow chart. Specifically:
  a. Templates are fetched and applied immediately after fetching requests, not concurrently with anything else
  b. Transaction counts and authorization statuses are fetched at the same time

2. Transaction counts are now fetched by `requesterIndex` instead of simply address. They are stored this way to allow for easier/faster lookup later on.

3. Renamed `address` in some function arguments to be more descriptive


bbenligiray (approved):
:ok_hand: 

> Transaction counts are now fetched by `requesterIndex` instead of simply address.

This will have to be updated when a `requesterIndex` has multiple designated wallets (`m/0/requesterIndex`, `m/1/requesterIndex`, etc.). Not something we should care about now though.

andreogle:
Yeah that would likely be a much bigger change to the node

PR #130, opened 2020-11-16
Add local handler

andreogle:
### Changes

1. Adds a new `local-handler` module which is intended to call the node functions through regular promises. This is useful for development, testing and when the node functions get wrapped as a docker image. Local handlers can be used by setting: `nodeSettings.cloudProvider = 'local'`.

2. Change the dev script to `dev:invoke`. `invoke:aws` doesn't really make sense anymore. `config.json` and `.env` files are expected in the `__dev__` folder when using this command 

3. Standardize the responses from each handler (AWS & local) function into a `WorkerResponse` object which looks like:

```ts
interface WorkerResponse {
  ok: boolean;
  data?: any;
  errorLog?: PendingLog;
}
```
Not all of the functions have immediate access to the context needed to output an error log if one happens (coordinatorId, chainId etc). So I thought it better to simply return a pending log which the calling function can handle.

4. Add more tests around workers


bbenligiray (approved):
This PR is a bit abstract for me but feels right

PR #132, opened 2020-11-17
Config security validation

### LINKED ISSUE #99: Validation across config.json and security.json
There are two things to be validated between the `config.json` and `security.json` files:
- `id` fields of the two files must be the same
- All security schemes in `config.json` (`ois.*.security`) must exist in `security.json` (`apiCredentials.{oisTitle}`)

Drgy:
- Add validation for `config.json` and `security.json`
- Add validation from root into if/then conditions (`__rootThen`)
- Move reading of arguments from `validate.ts` into `validateCmd.ts`, so validation with custom templates can be run from script as well
- Update examples in readme and add them to tests

closes #99


packages/validator/README.md
```diff
@@ -18,7 +18,7 @@ Where array `messages` contains message objects:
     message: string
 }
 ```
-If provided OIS is valid, parameter `valid` will be true, however parameter `messages` still might contain messages with `level` set to `warning`. If `valid` is `false`, there will be always one or more error messages. Similarly functions `isApiSpecsValid` and `isEndpointsValid` check only the API and endpoint specifications contained in the OIS.
+If provided OIS is valid, parameter `valid` will be true, however parameter `messages` still might contain messages with `level` set to `warning`. If `valid` is `false`, there will be always one or more error messages. Similarly functions `isApiSpecsValid` and `isEndpointsValid` check only the API and endpoint specifications contained in the OIS, as well as `isConfigSecurityValid` validates `config.json` and `security.json`
 
 The validator can be run as an NPM script, by providing the paths to the JSON file that will be checked and the JSON file to use as the template (validator specification structure). You can run the validator as script using the following command: `npm run validate --specs="[specsFile]" --template="[templateFile]"`
```
bbenligiray, line 23: I think it makes more sense for the command to go `npm run validate <specs> <template>` because "validate according to `<template>`" is the action and `<specs>` is the argument

bbenligiray: A copy-pasteable command that uses an example `specs` file would be useful here. Because the first thing I do after reading this first paragraph is to go out and look for example files to try this out.

packages/validator/specs/api.json
```diff
@@ -12,7 +12,7 @@
     "__conditions": [
       {
         "__if": {
-          "__this": "(?<={)[^\\/{}]+(?=})"
+          "__this_name": "(?<={)[^\\/{}]+(?=})"
```
bbenligiray, line 15: I think it would be less confusing to name this file exactly the same as the field it refers to (so `apiSpecifications.json` instead of `api.json`)

packages/validator/specs/configSecurity.json
```diff
@@ -0,0 +1,94 @@
+{
```
bbenligiray, line 1: Shouldn't this directory be called `templates` and not `specs`?

bbenligiray (changes requested):
I just started the review but I'll have to ask for a change before continuing.
I added some example specs files under `exampleSpecs/`.
1. This seems to error when it shouldn't.
```
npm run validate exampleSpecs/ois.specs.json specs/ois.json
```
It looks like the OIS template isn't correct. Can you look into this?

2. I like the workaround of keeping the `config.json` and `security.json` templates in the same file. However, the user will be providing these as two separate files and putting them in the same JSON file-logic should be in this validator. We will extract the generalized JSON validation-logic from this package to its independent npm package soon (and import that into this) anyway, so don't worry about this disturbing the generalizability of the package.

Drgy:
1. Validator is written according to whats in [the docs](https://github.com/api3dao/api3-docs/blob/master/airnode/2-6-ois.md), could you please either update the docs or the examples you added? These are the differences I found and which the validator is correctly reporting:
- `apiSpecifications.security` in the example is an array, however in the docs it is object
- `apiSpecifications.info` is in the example and also is mentioned in the docs, but not under apiSpecifications

Drgy:
2. Would something like adding `npm run validate --config="[configFile]" --security="[securityFile]"` work?

packages/validator/README.md
Drgy: I don't quite follow "the command to go `npm run validate <specs> <template>`", currently you can launch the command exactly as you say, did you mean `npm run validate <template> <specs>`?

Example added in 943f6df

packages/validator/specs/api.json
Drgy: Renamed in 4b9caca

packages/validator/specs/configSecurity.json
Drgy: Yes, I'm having some issues with the naming, in the validator `template` is called `specsStruct` (validator specification structure), in one of the PRs Andre suggested to call it template instead, however at that point the validator was already pretty much finished and the specsStruct was used everywhere. So in some more recent commits template is used, but specsStruct is still used in most of places and im not sure if its worth it to go through all the code and rename it everywhere (code, JSDoc, readme etc...) or leave it as it is and rename it only in instances such as this one, what do you think?

packages/validator/README.md
bbenligiray: I was just saying that it would make more sense if the order of the arguments were the other way around

packages/validator/specs/configSecurity.json
bbenligiray: I personally do this kind of stuff all the time (a commit today for example: https://github.com/api3dao/api3-docs/commit/e55c526b9b4d8d00fdddcf63b04fb10c10e9e6c1) so I would say it's worth it, especially considering that this code will be used by a lot of people for a long time. It doesn't look that bad either, a search returned me 47 results.

bbenligiray:
1. Thanks, it was the example that was wrong. `security` can be an array in OAS, but not in OIS (https://github.com/api3dao/api3-docs/blob/master/airnode/ois.md#43-security).

2. Yeah that would do. I think keeping the `validate` script as a general purpose one and then having a second `validateConfigSecurity` with `config` and `security` parameters would work better.

I'll continue reviewing the rest.

packages/validator/specs/configSecurity.json
Drgy: Renamed in 526ceae

bbenligiray (changes requested):
This section is extremely obscure.
https://github.com/api3dao/airnode/blob/config-security-validation/packages/validator/README.md#conditions
Each rule needs a separate example (a minimal format example that could not have been validated without this rule and how you would use the rule to solve that)

bbenligiray (approved):
Thank you, the new README is great. I couldn't wrap my head around it before, now I only had to do a few double takes for `# Regular expression matches in if/then conditions` but it's not that bad considering the concept is complex in the first place.

API operations in the endpoints are not being matched to the ones in the API specifications (and endpoint parameters are not matched to API operation parameters). So one can add an endpoint that refers to an API operation that doesn't exist in the API specifications.
Similar thing with triggers in `config.json` and OIS endpoints, and security schemes in `security.json` and OIS security schemes.
I suppose we would need one giant template to support this, but a validator that doesn't actually validate everything can do more harm than good so this is not really optional for us.

There should be a `validateOIS` command that only takes the OIS file to be validated as the argument, reads its OIS version (`oisFormat`), then load the corresponding template for validation.
Similarly, `validateConfigSecurity` should read `nodeSettings.nodeVersion` and use the corresponding `configSecurity.json` template.
Proposed directory structure:
```
templates/
├── ois/
│   └── 1.0.0/
│       ├── apiSpecifications-1.0.0.json
│       ├── endpoints-1.0.0.json
│       └── ois-1.0.0.json
└── config/
    └── 0.1.0/
        └── configSecurity-0.1.0.json
```
So in most cases the user shouldn't be providing their own template. Update the README.md to reflect that, i.e., `# Usage` explains how to use `validateOIS` and `validateConfigSecurity`, and ends with "by the way you can use `validate` and provide your own template".

So in general I'm asking for 2 things:
1. A more end-to-end validation, specifically between `apiSpecifications` and `endpoints`, and also between `ois` and `config.json`/`security.json`. I'd be fine with doing this with a giant template file. (This would probably mean a node version would only work with a specific `oisFormat`. This is probably fine considering `oisFormat` will not change much.)
2. Prioritizing the scenario where the user just feeds in their file and doesn't care about the template. The validator reads the version from the user's file, finds the corresponding template and validates.

These can be done in two separate PRs. This one already got bloated with the `specsStruct`->`template` change and we're at a good place so you can merge it.

Drgy:
@bbenligiray can this PR be merged as [convertor PR](https://github.com/api3dao/airnode/pull/195) depends on it and for [your last comment](https://github.com/api3dao/airnode/pull/132#pullrequestreview-558912567) issues #196 and #197 are opened?

PR #133, opened 2020-11-17
Fix lambda function names

bbenligiray:
Fixes Lambda function name build order as `airnode-${providerIdShort}-${stage}-${functionName}`


andreogle (approved):
Really thought I had the correct ordering heh 👍🏻

PR #134, opened 2020-11-17
Fix provider admin config name

bbenligiray:
- Renames `adminAddressForCreatingProviderRecord` as `providerAdminForRecordCreation` to conform to the [`config.json` docs](https://github.com/api3dao/api3-docs/blob/master/airnode/2-7-config-json.md)

- Sets a `gasLimit` of `300,000` while estimating gas for provider record creation. `estimateGas` fails if `gasLimit * getGasPrice()` is larger than the master wallet balance. Sometimes `ethers` (or whatever is doing this in the background) picks absurdly large `gasLimit`s, which may result in the gas estimation failing even if the master wallet balance is decent like 0.1 ETH. I didn't encounter this here specifically, but still, setting a large `gasLimit` as such doesn't really have any downsides in this context.

### DIFF
--- a/packages/node/src/core/evm/providers.ts
+++ b/packages/node/src/core/evm/providers.ts
@@ -7,7 +7,7 @@ import * as wallet from './wallet';
 import { LogsData } from '../../types';
 
 interface BaseFetchOptions {
-  adminAddressForCreatingProviderRecord?: string;
+  providerAdminForRecordCreation?: string;
   airnodeAddress: string;
   convenienceAddress: string;
   masterHDNode: ethers.utils.HDNode;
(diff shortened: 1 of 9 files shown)


packages/node/src/core/evm/providers.ts
```diff
@@ -83,7 +80,10 @@ export async function create(options: CreateOptions): Promise<LogsData<ethers.Tr
 
   // Gas cost is 160,076
   const [estimateErr, estimatedGasCost] = await go(
-    airnode.estimateGas.createProvider(options.adminAddressForCreatingProviderRecord, options.xpub, { value: 1 })
+    airnode.estimateGas.createProvider(options.providerAdminForRecordCreation, options.xpub, {
+      value: 1,
+      gasLimit: 300_000,
```
andreogle, line 85: 👍🏻

andreogle (approved):
I definitely prefer the new variable name to the old one :shipit:

Issue #94, opened 2020-10-08
API call requests being validated after forking


bbenligiray:
Isn't it a bit wasteful to validate these after forking?

https://github.com/api3dao/airnode/blob/fd1fdf209ea22aba8b2bf980d4980011e850c955/packages/node/src/core/adapters/http/execution.ts#L13

PR #137, opened 2020-11-18
Validate aggregated API calls before spawning workers

andreogle:
### Changes

1. As the title suggests, this PR changes the coordinator to validate aggregated API calls first before spawning workers. It's wasteful to spawn a `callApi` worker only to have it return immediately because there was an existing error code.

2. Changed a few function signatures and moved the validation logic to `src/core/coordinator/calls/validation`

### Addresses

https://github.com/api3dao/airnode/issues/94


bbenligiray (approved):
Neat :+1:

Issue #93, opened 2020-10-08
Adapter casting numbers to Number


bbenligiray:
I remember discussing this but it bothered me again.

https://github.com/api3dao/airnode/blob/fd1fdf209ea22aba8b2bf980d4980011e850c955/packages/adapter/src/response-processor/caster.ts#L32

This would be a problem if the API returned a Wei value for example. Why not parse as a big number directly?

andreogle:
Yeah I don't think the existing behaviour is correct

PR #138, opened 2020-11-19
Cast to BigNumber instead of Number

andreogle:
### Changes

1. When converting to `int256` in the adapter, we now use `new BigNumber()` instead of `Number()`. This is done in case a large number is returned, outside of what JS can normally handle precisely.

2. Extracted some common functions when casting

### Addresses
https://github.com/api3dao/airnode/issues/93


packages/adapter/src/response-processor/caster.ts
```diff
@@ -75,13 +73,21 @@ export function castValue(value: unknown, type: ResponseType) {
   }
 }
 
-export function multiplyValue(value: number | string, times: number | string): string {
+export function multiplyValue(value: string | BigNumber, times: number | string): string {
```
bbenligiray, line 76: The `times` here is `encoded` in the request `parameters` and may or may not be a big number. For example, it is easy to think of a use case where it would be `10^18`. But I guess we depend on `cbor.js` to decode it, which can't do big numbers? Tomorrow I will try encoding a 256 bit number and decoding it with cbor-custom to see what it spits out.

bbenligiray: I decoded `{key: 1000000000000000000}` using `cbor-custom` and received a big number.  So we can also avoid using `number` for `_times` here. lmk what you think.

andreogle: Good point. I don't think we should be using regular JS numbers at all for anything number related. Either strings or big numbers

PR #139, opened 2020-11-19
Migrate to yarn

bbenligiray:
npm was causing some issues during deployer installation so we decided to migrate to yarn a while ago. I updated all references to npm to yarn and updated the lockfiles.


bbenligiray:
Except I didn't update the Github workflow. I don't know if it supports yarn, though I'd assume it does.

andreogle (approved):

PR #140, opened 2020-11-21
Aws deployment

bbenligiray:
This PR implements a deployer package that:
- For the first deployment, creates a mnemonic and stores it in AWS SSM using Terraform
- Builds and deploys the node using Serverless and the mnemonic stored at AWS SSM
The code is plain JS and has no tests :ok_hand: 

There is also a Docker container configuration at the root that allows the use to use the deployer commands. Both the [deployer package](https://github.com/api3dao/airnode/blob/aws-deployment/packages/deployer/README.md) and the [container](https://github.com/api3dao/airnode/blob/aws-deployment/Docker.md) have their docs.

I moved the Lambda handlers to the deployer package and made small adjustments to them by debugging. I only debugged until provider initialization, though I don't expect any further problems.

There also seems to be two small deployment scripts for the protocol contracts.


Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN cd /usr/local/bin \
+    && wget "https://releases.hashicorp.com/terraform/0.13.4/terraform_0.13.4_linux_amd64.zip"  -O terraform.zip \
+    && unzip terraform.zip \
+    && rm terraform.zip
+
+RUN yarn run bootstrap
+RUN yarn run build
+
+CMD cp out/config.json packages/deployer/src/config-data/config.json | true \
+    && cp out/security.json packages/deployer/src/config-data/security.json | true \
```
bbenligiray, line 20: `config.json` and `security.json` is copied from the current working directory to where the deployer expects them to be

Docker.md
```diff
@@ -0,0 +1,86 @@
...
+3. If you will be running [`deploy-first-time`](#deploy-first-time) or [`redeploy`](#redeploy), your `config.json` and `security.json` must be in the current working directory.
+
+4. Run the image with one of the following commands:
+
+### `deploy-first-time`
+```sh
+docker run -it --rm \
+  --env-file .env \
+  --env COMMAND=deploy-first-time \
+  -v $(pwd):/airnode/out \
```
bbenligiray, line 19: For the docker commands that require to read files from the current working directory (`config.json`, `security.json` or the receipt), we're creating a volume that maps the current working directory to `/airnode/out` in the container.

.github/workflows/master.yml
```diff
@@ -23,6 +23,10 @@ jobs:
     - name: Lint code
       run: npm run lint
 
+    - name: Prepare config
+      run: |
+        cp packages/deployer/src/config-data/config.json.example packages/deployer/src/config-data/config.json
```
bbenligiray, line 28: Even though the node package no longer needs `config.json` to build, the deployer still does.

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    && cp out/$RECEIPT_FILENAME packages/deployer/ | true \
```
bbenligiray, line 21: Tries to copy the receipt, doesn't raise an error if it fails because of the `| true`. Used the same trick for all copy operations.

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    && cd /airnode/packages/deployer \
+    && yarn run sls:config \
+    && yarn run terraform:init \
+    && yarn run command:$COMMAND \
+    && cd /airnode \
+    && cp packages/deployer/*.receipt.json out | true
```
bbenligiray, line 27: Copies the receipt to the current working directory (if a receipt exists)

package.json
```diff
@@ -26,19 +27,22 @@
     "test:node": "(cd packages/node && yarn run test)",
     "test:node:watch": "(cd packages/node && yarn run test:watch)"
   },
+  "dependencies": {
+    "@types/node": "^14.10.2",
+    "lerna": "^3.22.1",
+    "rimraf": "^3.0.2",
+    "typescript": "^4.0.2"
+  },
```
bbenligiray, line 35: I had to define all dependencies that I need to build the node inside the deployer container as non-development dependencies because I'm using `NODE_ENV`-`production`. This is probably not a good idea. Let me know what you think.

bbenligiray: This also happens in packages btw.

.gitignore
```diff
@@ -6,6 +6,7 @@
 
 # Log files
 *.log
+*.receipt.json
```
bbenligiray, line 9: Deployments and redeployments output a `*.receipt.json` file

packages/deployer/.env.example
```diff
@@ -0,0 +1,2 @@
+AWS_ACCESS_KEY_ID=
```
bbenligiray, line 1: This file normally needs to be in the directory the container is being run from. It's also here if you want to use the deployer locally/outside the container (which you should during development, rebuilding the container takes too long).

packages/deployer/README.md
```diff
@@ -0,0 +1,117 @@
...
+Keep the outputted receipt file, or send it to API3 if you are an API3 provider.
+
+1. Fund the master wallet address with the displayed amount to create your provider record.
+This will be done for you if you are an API3 provider.
+
+1. Use the `redeploy` command when you need to update your node.
+Note that your `providerIdShort` from your receipt file must be inserted into the `nodeSettings.providerIdShort` field of your `config.json` file.
+This will be done for you if you are an API3 provider.
+
+## Commands
```
bbenligiray, line 43: These are NPM scripts

packages/deployer/serverless.aws.yml
```diff
@@ -0,0 +1,39 @@
...
+plugins:
+  - serverless-plugin-typescript
+  - serverless-plugin-optimize
+
+provider:
+  name: aws
+  region: ${opt:region}
+  stage: ${opt:stage}
+  runtime: nodejs12.x
+  memorySize: 256
```
bbenligiray, line 13: 128 would probably be fine, but I felt generous

packages/deployer/serverless.aws.yml
```diff
@@ -0,0 +1,39 @@
+service:
+  name: airnode-${opt:providerIdShort}
+
+plugins:
+  - serverless-plugin-typescript
+  - serverless-plugin-optimize
```
bbenligiray, line 6: This both packs the whole thing down to 1/100 of the size and also replaces the monorepo symlinks with actual files.

packages/deployer/serverless.aws.yml
```diff
@@ -0,0 +1,39 @@
...
+  name: aws
+  region: ${opt:region}
+  stage: ${opt:stage}
+  runtime: nodejs12.x
+  memorySize: 256
+  environment: ${file(secrets.json)}
+  iamRoleStatements:
+    - Effect: 'Allow'
+      Action: 'lambda:InvokeFunction'
+      Resource: '*'
```
bbenligiray, line 18: All functions are allowed to invoke all functions of the user

packages/deployer/serverless.aws.yml
```diff
@@ -0,0 +1,39 @@
...
+
+package: 
+  individually: true
+  
+functions:
+  startCoordinator:
+    handler: ./src/handlers/aws/index.startCoordinator
+    timeout: 60
+    reservedConcurrency: 1
```
bbenligiray, line 27: Does `startCoordinator` wait `processProviderRequests` to return to exit?

packages/deployer/src/commands.ts
```diff
@@ -0,0 +1,132 @@
+import ora from 'ora';
```
bbenligiray, line 1: https://i.ytimg.com/vi/o5lj1Pgjruo/maxresdefault.jpg

packages/deployer/src/commands.ts
```diff
@@ -0,0 +1,132 @@
...
+  await applyTerraformWorkaround(configParams.region);
+
+  const mnemonic = generateMnemonic();
+  const providerId = deriveProviderId(mnemonic);
+  const providerIdShort = shortenProviderId(providerId);
+  if (await ssm.checkIfProviderIdShortExists(providerIdShort)) {
+    throw new Error('Randomly generated providerIdShort is already used at SSM. This should not have happened.');
+  }
+
+  ora().warn('Write down the 12 word-mnemonic below on a piece of paper and keep it in a safe place\n');
```
bbenligiray, line 35: What I don't like the most about all this is how tightly the business logic is coupled with the user interaction bits.

packages/deployer/src/config.ts
```diff
@@ -0,0 +1,78 @@
...
+    );
+    throw new Error('Attempted to deploy node configuration with the wrong node version');
+  }
+}
+
+export function generateServerlessSecretsFile(providerIdShort, apiCredentials) {
+  const secrets = {};
+  // The mnemonic will be fetched from AWS SSM
+  secrets['MASTER_KEY_MNEMONIC'] = `$\{ssm:/airnode/${providerIdShort}/masterKeyMnemonic~true\}`;
+  secrets['LOG_LEVEL'] = 'debug';
```
bbenligiray, line 67: Should be removed at some point

packages/deployer/src/handlers/aws/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export async function startCoordinator() {
+  await node.handlers.startCoordinator(config);
+  const response = { ok: true, data: { message: 'Coordinator completed' } };
+  return { statusCode: 200, body: encodeBody(response) };
+}
+
+export async function initializeProvider(event: any) {
+  // State and config are sent separately
+  const stateWithConfig = node.providerState.update(event.state, { config });
```
bbenligiray, line 18: Parameters were being read from a `parameters` field. That's not needed according to my debugging
https://github.com/api3dao/airnode/blob/f74e2a011aa56daed1c8f38c77c8fd90cea3ff93/packages/node/src/aws/handler.ts#L21

packages/deployer/src/infrastructure/serverless.ts
```diff
@@ -0,0 +1,34 @@
...
+    await removeServerless(providerIdShort, region, stage);
+    spinner.succeed('Removed the serverless functions');
+  } catch (e) {
+    spinner.fail('Failed removing the serverless functions.');
+    throw e;
+  }
+}
+
+async function deployServerless(providerIdShort, region, stage) {
+  await exec(`PROVIDER_ID_SHORT=${providerIdShort} REGION=${region} STAGE=${stage} yarn run sls:deploy`);
```
bbenligiray, line 29: This `exec()` business is extremely brittle, but that's not a huge issue when you're running the deployer in a container

packages/deployer/terraform/variables.tf.workaround
```diff
@@ -0,0 +1,24 @@
+variable "region" {
+  type    = string
+  default = "<UPDATE_REGION>"
```
bbenligiray, line 3: Terraform only works correctly when the region is hardcoded in this file. That's why the deployer copies this file and replaces `<UPDATE_REGION>` with the correct value each time it starts.

packages/node/package.json
```diff
@@ -1,6 +1,9 @@
 {
   "name": "@airnode/node",
   "version": "0.1.0",
+  "license": "MIT",
+  "main": "dist/index",
+  "types": "dist/index",
```
bbenligiray, line 6: I had difficulty importing node to deployer without doing this

packages/node/tsconfig.build.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "baseUrl": "./",
     "rootDir": "src",
-    "outDir": ".build"
+    "outDir": "./dist"
```
bbenligiray, line 6: Similarly, had to do this and the one below

packages/deployer/export-env.sh
```diff
@@ -0,0 +1,3 @@
+# !/bin/bash
```
bbenligiray, line 1: You need to run this script to export your AWS credentials before running the npm scripts locally

andreogle (commented):
Really excited to get this in 🚀  

Some general comments:

1. Is there a way that we could only have a single deploy command instead of: `deploy-first-time` and `redeploy`. For example, how about a process like:

a. The user uses a script we provide to generate a mnemonic. We force them to re-enter it by typing it back, although there wouldn't really be consequences if they exit instead.
b. They enter this mnemonic into a .env file or something similar
c. They run a single `deploy` command that loads the mnemonic and derives the `providerIdShort` field internally.
d. The deployer checks if functions for that providerIdShort, region and stage exist. If they do, overwrite. If they don't, deploy new functions.

This would mean one less field the user needs to worry about (`providerIdShort`). It also makes deployment a bit easier. I had a `providerIdShort` in my config.json by mistake and it took 5 - 10 minutes to realise this because the container takes a very long time to build.

One downside would be that the user needs to keep track of which mnemonic links to which `providerIdShort`, if they have multiple deployments. We would need a script/tool to input a mnemonic and get a `providerIdShort` out.

2. It takes a long time to get feedback about something missing or not being right. I spent 30 minutes figuring out that config.json and security.json need to be in `src/config-data` and that I shouldn't have `providerIdShort`.

3. The docs feel disjointed and a bit all over the place. Some things are missing like where to put config.json and security.json. We can work on this though as this evolves.

4. I don't think we've considered running this from Windows yet (or at least I haven't).

Let me know what you think

Dockerfile
bbenligiray: They are only needed for `deploy-first-time` and `redeploy` and the errors are raised in the scripts

package.json
bbenligiray: Okay then. I think this is not a good enough reason to use global packages.

packages/deployer/serverless.aws.yml
bbenligiray: Yeah it's possible, I just didn't look into it further.

packages/deployer/serverless.aws.yml
bbenligiray: Okay then this is enough to prevent concurrent iterations.

packages/deployer/src/commands.ts
bbenligiray: Not that I know of

packages/deployer/src/handlers/aws/index.ts
bbenligiray: Both of them right? 293bfd6

packages/deployer/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+1. Deploy the serverless functions at the target `region` using Serverless Framework
+
+The mnemonic deployment needs to be performed only when deploying for the first time on a region.
+For the following redeployments, mnemonic deployment must not be repeated.
+
+## Regular user flow
+
+1. Use the `deploy-first-time` for your first deployment.
+Note down the displayed mnemonic.
```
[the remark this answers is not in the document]
bbenligiray: They can deploy their own mnemonic using [`deploy-mnemonic`](https://github.com/api3dao/airnode/blob/aws-deployment/Docker.md#deploy-mnemonic), use the corresponding `providerIdShort` in their `config.json` and use `redeploy`. I didn't bother optimizing this flow because it's for pros only.

>ensuring they have it written down, by forcing them to enter it again

and probably allow to disable that with a flag. I think having them type in the mnemonic is a security issue with potential keyloggers. We should probably ask a single word at random, like "what was your Xth word?". Let me know what you think.

packages/deployer/README.md
```diff
@@ -0,0 +1,117 @@
...
+Keep the outputted receipt file, or send it to API3 if you are an API3 provider.
```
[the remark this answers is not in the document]
bbenligiray: This was more a note to ourselves. I hope the instructions we send to the API providers are nothing like this.

packages/protocol/deployments/rinkeby/solcInputs/0x9b0904b42a34b3f725927312286e3a592aaf3712ac6e91ee1bd77e406a72dd9c.json
```diff
@@ -0,0 +1,82 @@
+{
```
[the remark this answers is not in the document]
bbenligiray: The inputs the compiler used to create the bytecode. I think they are here for you to be able to post them on Etherscan for verification for example.

Docker.md
```diff
@@ -0,0 +1,86 @@
+# Docker instructions
+
+1. Build the Docker image
+```sh
+docker build . -t airnode-alpha
+```
+
+2. Ensure that your `.env` file looks like [`/packages/deployer/.env.example`](/packages/deployer/.env.example) and is the current working directory.
```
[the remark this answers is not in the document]
bbenligiray: Right, I shouldn't include the path.I think it's good now?

packages/deployer/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+*The recommended way to deploy Airnode is by using the deployment Docker image.
+This package simply implements the scripts used by that image and is not meant to be used directly by the end user.*
+
+*If you will be using the provided scripts locally, run `export-env.sh` to export your AWS credentials as environment variables first.*
+
+## Setup
+
+- Download the [Terraform v0.13.* binary](https://www.terraform.io/downloads.html) and move it to your `PATH`
+- Get an *Access Key ID* and *Secret Access Key* from AWS and insert into `.env` (refer to `.env.example`)
```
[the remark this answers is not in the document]
bbenligiray: Clarified

bbenligiray:
1. I think it's really really really important for the API provider to get the `config.json`, `security.json`, `.env` in a directory, run a single command there and have the node up.

There are two separate commands because there are two possible `config.json` configurations (with and without `providerIdShort`). Having the user state what they intend explicitly is supposed to catch related mistakes.

a. We can verify somehow if they wrote the mnemonic down. This may encourage them to copy paste it, which is probably worse than not writing it down.
b. I like the current setup because the mnemonic is never in a file that can be seen directly by the host machine's file system. It's not encrypted inside the container but at least it's something.
d. Feels very error-prone to me (and this is more or less how the deployer worked before). Requesting an exact thing and throwing when it's not met is safer and that safety is needed in this case imo. Btw you need the region+stage both for first and following deployments.

> I had a providerIdShort in my config.json by mistake and it took 5 - 10 minutes to realise this because the container takes a very long time to build.

I don't get what happened here. Once you built the container, it immediately throws if you use the wrong command with the `config.json`. You can just update `config.json` (or your command) and try again.

2. Again, I don't get it. Are you talking about the package or the container? A built container checks `providerIdShort` the first thing.

3. Yeah I'm not considering these to be docs really.

4. I think a built container should work seamlessly, assuming they got Docker working. That's one of the reasons I wanted to containerize the deployer.

packages/deployer/README.md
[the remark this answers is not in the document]
andreogle: > type in the mnemonic is a security issue with potential keyloggers

Good point. Much better to **not** have them enter the full mnemonic. I like having them enter a random Xth word 👍🏻

packages/deployer/src/handlers/aws/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  const { aggregatedApiCall, logOptions } = event;
+  const [logs, apiCallResponse] = await node.handlers.callApi(config, aggregatedApiCall);
+  node.logger.logPending(logs, logOptions);
+  const response = encodeBody({ ok: true, data: apiCallResponse });
+  return { statusCode: 200, body: JSON.stringify(response) };
+}
+
+export async function processProviderRequests(event: any) {
+  // State and config are sent separately
+  const stateWithConfig = { ...event.state, config };
```
andreogle, line 42: Github doesn't seem to want to let me respond to the original comment since it's now outdated, but yeah this looks good (in both places). The comment above can also go now since the config isn't "sent" anymore

packages/deployer/serverless.aws.yml
andreogle: Ok I think it's worth doing it before merging or logging an issue for this. It will be pretty easy to forget about later on. What do you think?

andreogle (commented):
tl;dr: I'm happy with these responses. It was probably me who messed things up while testing because I was hopping between branches and the docs weren't very clear.

1. 
> run a single command there and have the node up.
Definitely. Agreed

> two possible config.json configurations
I was just thinking it would be nicer to only have a single configuration

> the current setup because the mnemonic is never in a file that can be seen directly by the host machine's file system
Good point. Much better to never have this in the directory.

> it immediately throws if you use the wrong command with the config.json
Right. I think this was actually because my config.json was invalid somehow when it tried to build.

2. It may have been because of other issues, sorry. I'll need to test it again later

3. Ok great 

4. Yeah the docker related stuff should all work fine on Windows. I meant more the scripts that run bash commands (NPM scripts, `export-env.sh` etc). Just something we'll need to test later

packages/deployer/serverless.aws.yml
bbenligiray: Agreed https://github.com/api3dao/airnode/issues/142

packages/deployer/README.md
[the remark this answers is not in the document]
bbenligiray: Created an issue https://github.com/api3dao/airnode/issues/143

packages/deployer/src/handlers/aws/index.ts
bbenligiray: Removed the comment. I also removed the AWS handlers from the node package.

bbenligiray:
> I was just thinking it would be nicer to only have a single configuration

I think it's obvious that there should be one deploy command, `config.json` should never include `providerIdShort` and it should be provided as a CLI argument. But if we do that, API providers will forget to provide `providerIdShort` while redeploying, which will result in them deploying a new node with a new mnemonic and be extremely confused. In this setup, its "download the files we sent you and copy paste this generic command" both if they are deploying for the first time or redeploying.
It will have the user bang their heads to the wall a couple of times if they are not being provided with a `config.json` and have to do it themselves, but that's better than them shooting themselves in the foot.

> I meant more the scripts that run bash commands (NPM scripts, export-env.sh etc)

Yeah I wasn't thinking about the dev environment. I think Windows devs will have to run the deployer in the container for development as well.

bbenligiray:
Can you look into the failing test? I couldn't figure out how to have Jest not refer to the AWS handlers anymore.

packages/deployer/src/handlers/aws/index.ts
andreogle: Cool thanks. I was planning to do that after this was merged anyway 👍🏻

andreogle:
> It will have the user bang their heads to the wall a couple of times if they are not being provided with a config.json and have to do it themselves, but that's better than them shooting themselves in the foot.

Yeah fair enough. 

> Windows

For deployment, we should definitely provide first class support for Windows. i.e. Windows users don't need anything special to deploy. For development, I'm less concerned. I think most things should probably work as is barring a few possible issues (mainly in the deployer I would guess).

> Failing tests

Heh just needed to remove an old mock

bbenligiray:
>needed to remove an old mock

I know but couldn't do it nonetheless :facepalm: 

So what do we do with this one, merge? I won't deal with the 2 issues for now.

andreogle:
Yeah I'm happy to merge this and then we can implement improvements as separate PRs

PR #141, opened 2020-11-23
Adapter cleanup

andreogle:
### Changes

1. The adapter code wasn't great. The `processByX` functions were really confusing and code was all over the place with a bunch of pretty useless wrapper functions. I've tried to clean some of this up


packages/adapter/src/response-processing/index.ts
```diff
@@ -0,0 +1,43 @@
+export * from './casting';
+export * from './encoding';
+export * from './extraction';
+
+// export function isNumberType(type: ResponseType) {
```
bbenligiray, line 5: What's up with the commented out code?

bbenligiray (approved):

packages/adapter/src/response-processing/index.ts
andreogle: Woops I missed that. It was supposed to be removed

Issue #144, opened 2020-11-25
Blocked requests should be ignored after x blocks have passed


andreogle:
Currently blocked requests would block all subsequent requests until they are no longer fetched

Ignore instead of blocking after 20 blocks have passed from the current block

PR #146, opened 2020-11-25
Configure number of blocks before a blocked request becomes ignored

andreogle:
### Changes

1. Adds an optional `ignoreBlockedRequestsAfterBlocks` number to `chains.providers[n]` that can be configured to set how many blocks should have passed before a request goes from being "blocked" to "ignored". It defaults to 20 blocks if not set.

2. Store the current block and the `ignoreBlockedRequestsAfterBlocks` setting in the metadata of each request. It does mean some duplicated data, but it makes it very easy to access this information when needed. You don't need to fetch it from the state and pass it through to the lower level functions.

3. Added a fixture to configure metadata. It was a bit messy overwriting the nested object when building an API call in tests

### Addresses

https://github.com/api3dao/airnode/issues/144

I was a bit hasty in logging this issue, the `nonces.assign()` function already handles this logic in a central place.


bbenligiray (changes requested):
Similar to `providerAdminForRecordCreation`, I think `ignoreBlockedRequestsAfterBlocks` should be a field of `ChainConfig` rather than `ChainProvider`

andreogle:
Yep, you're right. I realised the same thing last night. 

This also actually includes `blockHistoryLimit` and `minConfirmations`. If they are set differently for different providers, it could result in nonces being different

bbenligiray (approved):
Good catch. I updated the docs to reflect the changes:
https://github.com/api3dao/api3-docs/commit/f2908f70c916d14edd921d94d503aa7859f7b775

PR #157, opened 2020-12-08
Add Airnode-ABI package

andreogle:
### Changes

1. Adds a new package `@airnode/airnode-abi` that exposes two functions for encoding and decoding according to the [specifications](https://github.com/api3dao/api3-docs/blob/master/airnode/airnode-abi-specifications.md)

**Encoding**
```ts
import { encode } from '@airnode/airnode-abi';

const parameters = 
  { type: 'byte32', name: 'from', value: 'ETH' },
  { type: 'uint256', name: 'amount', value: '100000' },
]; 
const encodedData = encode(parameters);

console.log(encodedData);
// '0x...'
```

**Decoding**
```ts
import { decode } from '@airnode/airnode-abi';

const encodedData = '0x...';
const decoded = decode(encodedData);

console.log(decoded);
// { from: 'ETH', amount: ethers.BigNumber.from('100000') }
```

NOTE: I have not yet integrated this with the node, will do that in a separate PR. Hoping to get the package merged first.


packages/airnode-abi/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+console.log(encoded);
+// '0x...'
+```
+
+### Decoding
+
+Decoding returns an object where the keys are the "names" and the values are the "values" from the initial encoding.
+
+It is important to note that `int256` and `uint256` types are returned as [ethers BigNumbers](https://docs.ethers.io/v5/api/utils/bignumber/). This is done to preserve precision. These values can be converted to strings if necessary using `.toString()`. See the [BigNumber](https://docs.ethers.io/v5/api/utils/bignumber/) docs for more information.
```
bbenligiray, line 51: In our use case with the node, we will send the numbers to the API as strings so we don't really need big numbers.
For other use cases, numbers as strings would be more universal, because for example the user may be using web3.js etc.

packages/airnode-abi/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+# Yarn
+yarn add @airnode/airnode-abi
+```
+
+## Usage
+
+### Encoding
+
+**NB:** The `types`, `names` and `values` array inputs must all have the same number of elements
```
bbenligiray, line 29: I thought we agreed on an array of objects instead of 3 arrays

packages/airnode-abi/README.md
```diff
@@ -0,0 +1,71 @@
...
+const encodedData = '0x...';
+const decoded = decode(encodedData);
+
+console.log(decoded);
+// { from: 'ETH', amount: ethers.BigNumber.from('100000') }
+```
+
+## Security
+
+API3 and Airnode take security very seriously. If you find a security related bug please let the team know privately.
```
bbenligiray, line 65: This line has always sounded off to me. Everyone claims that they take security very seriously so the first sentence is redundant. Although the second sentence is instructions on how you should be reporting security related bugs, it feels like a plea to the public to help find these bugs.

packages/airnode-abi/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Contributing
+
+Airnode welcomes all contributors, regardless of how big or small the change is!
```
bbenligiray, line 69: I think we should have these in the main README and not duplicate them

packages/airnode-abi/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+If you find a bug, please log an [issue](https://github.com/api3dao/api3-docs)
```
bbenligiray, line 71: Why in api3-docs :thinking:

packages/airnode-abi/src/decoding.ts
```diff
@@ -0,0 +1,74 @@
...
+
+type TransformationFunction = (value: any) => string;
+
+type TransformationReference = {
+  [key in ABIParameterType]: TransformationFunction | null;
+};
+
+// Certain types need to be parsed after ABI decoding happens
+const TRANSFORMATIONS: TransformationReference = {
+  bytes: ethers.utils.parseBytes32String,
```
bbenligiray, line 14: If ethers decodes a `bytes` type as `"0xabc123"`, we want that exact value and not have it parsed as a string. So this should also be null.

packages/airnode-abi/src/decoding.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      '0x31750000000000000000000000000000000000000000000000000000000000005465737455496e744e616d65000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000007d0';
+    const res = decoding.decode(data);
+    expect(res).toEqual({ TestUIntName: ethers.BigNumber.from('2000') });
+  });
+
+  it('decodes bytes types', () => {
+    const data =
+      '0x31420000000000000000000000000000000000000000000000000000000000005465737442797465734e616d650000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000020536f6d652062797465732076616c756500000000000000000000000000000000';
+    const res = decoding.decode(data);
+    expect(res).toEqual({ TestBytesName: 'Some bytes value' });
```
bbenligiray, line 30: A correct `bytes` example would be a hexadecimal string (and the number of hexadecimals has to be even)

bbenligiray (changes requested):
- I thought we agreed on `encode()` arguments being an array of objects rather than 3 arrays
- I think `decode()` should output numbers as strings
^ Let me know what you think about these
- `bytes` shouldn't be parsed as a string

packages/airnode-abi/README.md
andreogle: Yeah I see what you mean. Should I just remove it?

packages/airnode-abi/README.md
andreogle: Agreed

packages/airnode-abi/README.md
andreogle: Ok I misunderstood the conclusion. I'll adjust

packages/airnode-abi/README.md
bbenligiray: Yeah I think we should just have a main Contributing section in the main README and reword this there

packages/airnode-abi/README.md
andreogle: Yes great point. Makes things easier for me too

packages/airnode-abi/README.md
andreogle: Removed it

PR #159, opened 2020-12-10
Replace cbor-custom with airnode-abi

andreogle:
### Changes

1. See title. Replaces `cbor-custom` with `@airnode/airnode-abi` in the node


packages/node/src/core/evm/abi-encoding.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    );
+    expect(res).toEqual({
+      from: 'ETH',
+      amount: '1000',
+    });
+  });
+
+  it('returns null if parameters are empty', () => {
+    const res = safeDecode('0x');
+    expect(res).toEqual(null);
```
bbenligiray, line 16: Wouldn't `{}` be more accurate here?

packages/node/src/core/evm/abi-encoding.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  });
+
+  it('returns an empty object if parameters are falsey', () => {
+    const res = safeDecode('');
+    expect(res).toEqual({});
```
bbenligiray, line 21: and `null` here because `"0x"` is a valid empty encoding and `""` is not

bbenligiray: For context, if you declare a `bytes` variable in a contract and retrieve it before initializing it, it returns `"0x"`

bbenligiray (commented):
Some comments about how empty parameters are treated

packages/node/src/core/evm/abi-encoding.test.ts
andreogle: Yeah good point. I added some special cases for zero addresses to the airnode-abi package here: https://github.com/api3dao/airnode/pull/159/commits/db4adfe1257d39109d5f7ed0515474e17a54438b

I don't know if I need to cover both `AddressZero` and `HashZero` but it feels a bit safer

bbenligiray: I'm not sure about that. `AddressZero` and `HashZero` are not any different than `"this is not a valid ABI encoding"`.

andreogle: Ok so just handling `0x` is all we need then?

bbenligiray: I think so

andreogle: Decoding `AddressZero` or `HashZero` now throws an error

bbenligiray (approved):

Issue #156, opened 2020-12-08
The node accepts reserved parameters that weren't in the OIS


bbenligiray:
The node should only accept reserved parameters listed under `endpoints.*.reservedParameters`. At the moment, it accepts all reserved parameters. This doesn't break anything or cause a security problem, only unexpected behavior.

PR #160, opened 2020-12-10
Require response parameters to be whitelisted in the OIS

andreogle:
### Changes

1. See title. Response parameters must now be specified in the `endpoints.x.reservedParameters` - even if no default or fixed value is set.

### Addresses

https://github.com/api3dao/airnode/issues/156

### DIFF
--- a/packages/node/src/core/adapters/http/parameters.ts
+++ b/packages/node/src/core/adapters/http/parameters.ts
@@ -9,8 +9,10 @@ export function getResponseParameterValue(
   requestParameters: ApiCallParameters
 ): string | undefined {
   const reservedParameter = endpoint.reservedParameters.find((rp) => rp.name === name);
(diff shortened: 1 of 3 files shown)


packages/node/src/core/adapters/http/parameters.ts
```diff
@@ -30,9 +32,5 @@ export function getResponseParameters(endpoint: Endpoint, requestParameters: Api
   const _times = getResponseParameterValue('_times', endpoint, requestParameters);
   const _type = getResponseParameterValue('_type', endpoint, requestParameters);
 
-  return {
-    _type,
-    _path,
-    _times: _times ? Number(_times) : undefined,
```
andreogle, line 36: This shouldn't have been converted to a number

bbenligiray (approved):

PR #165, opened 2020-12-13
Rename requester index

bbenligiray:
I was only meaning to rename `requestInd`->`requestIndex` and `requestIndex`->`requestIndices` at the contracts and redeploy, but updating the contract ABIs and addresses manually started to get annoying so I exported them from the protocol package.
I'm pretty sure that I didn't miss any `requestInd`s, I did all the updates manually and double checked.


packages/node/src/evm/contracts/airnode.test.ts
```diff
@@ -1,11 +1,12 @@
...
+import { AirnodeAddresses } from '@airnode/protocol';
 
 describe('Airnode', () => {
   it('exposes the addresses for each network', () => {
     expect(Airnode.addresses).toEqual({
       1: '<TODO>',
-      3: '0x9AF16dE521f41B0e0E70A4f26F9E0C73D757Bd81',
-      4: '0xf8f2f4a8317947B6037186e5aC87D015bcD756C1',
+      3: AirnodeAddresses[3],
+      4: AirnodeAddresses[4],
```
bbenligiray, line 9: I don't know if importing the addresses here too makes sense

packages/protocol/index.js
```diff
@@ -0,0 +1,20 @@
+const AirnodeArtifact = require('./artifacts/contracts/Airnode.sol/Airnode.json');
+const ConvenienceArtifact = require('./artifacts/contracts/Convenience.sol/Convenience.json');
+
+const networkChainIds = { ropsten: 3, rinkeby: 4 };
```
bbenligiray, line 4: I initially exported all deployments present in the directory, but I guess hardcoding these is a better idea. This way it will error if the deployments can't be found.

packages/protocol/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@airnode/protocol",
   "version": "0.1.0",
   "license": "MIT",
+  "main": "index",
```
bbenligiray, line 5: I think it's fine to leave this at the package root (?)

andreogle: You mean as opposed to in `src/`? Yeah I think it's fine

andreogle (approved):
Looks good. Just left a few small comments

bbenligiray:
Normally I treat abbreviations as regular words re casing conventions but I don't have a strong preference so I capitalized ABIs in variable names. We can follow that going forwards.

PR #168, opened 2020-12-23
Hardcode artifact dependencies

bbenligiray:
- Unrolled protocol contract artifact dependencies and hardcoded them so that they are added to the package by `serverless-plugin-optimize`
- I copied the additions from https://github.com/api3dao/airnode/blob/operation-package/packages/protocol/index.js in advance. However I exported the artifacts more explicitly as `AirnodeArtifact` instead of `Airnode` instead of importing them with aliases later because the names `Airnode`/`Convenience` are already used (as in https://github.com/api3dao/airnode/blob/operation-package/packages/node/src/evm/contracts/airnode.ts#L2).

### DIFF
--- a/packages/protocol/index.js
+++ b/packages/protocol/index.js
@@ -1,16 +1,19 @@
 const AirnodeArtifact = require('./artifacts/contracts/Airnode.sol/Airnode.json');
 const ConvenienceArtifact = require('./artifacts/contracts/Convenience.sol/Convenience.json');
 
-const networkChainIds = { ropsten: 3, rinkeby: 4 };
-
 const AirnodeAddresses = {};
 const ConvenienceAddresses = {};
-for (const network in networkChainIds) {
-  const AirnodeDeployment = require(`./deployments/${network}/Airnode.json`);
-  AirnodeAddresses[networkChainIds[network]] = AirnodeDeployment.receipt.contractAddress;
-  const ConvenienceDeployment = require(`./deployments/${network}/Convenience.json`);
-  ConvenienceAddresses[networkChainIds[network]] = ConvenienceDeployment.receipt.contractAddress;
-}
+let AirnodeDeployment, ConvenienceDeployment;
+// Ropsten - 3
+AirnodeDeployment = require(`./deployments/ropsten/Airnode.json`);
+AirnodeAddresses[3] = AirnodeDeployment.receipt.contractAddress;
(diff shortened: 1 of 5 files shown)


packages/protocol/index.js
```diff
@@ -1,20 +1,29 @@
...
 const ConvenienceAddresses = {};
-for (const network in networkChainIds) {
-  const AirnodeDeployment = require(`./deployments/${network}/Airnode.json`);
-  AirnodeAddresses[networkChainIds[network]] = AirnodeDeployment.receipt.contractAddress;
-  const ConvenienceDeployment = require(`./deployments/${network}/Convenience.json`);
-  ConvenienceAddresses[networkChainIds[network]] = ConvenienceDeployment.receipt.contractAddress;
-}
+let AirnodeDeployment, ConvenienceDeployment;
+// Ropsten - 3
+AirnodeDeployment = require(`./deployments/ropsten/Airnode.json`);
```
andreogle, line 9: No need for backticks here anymore (and for the other `require`s)
```suggestion
AirnodeDeployment = require('./deployments/ropsten/Airnode.json');
```

packages/protocol/index.js
```diff
@@ -1,20 +1,29 @@
...
 
 const AirnodeAddresses = {};
 const ConvenienceAddresses = {};
-for (const network in networkChainIds) {
-  const AirnodeDeployment = require(`./deployments/${network}/Airnode.json`);
-  AirnodeAddresses[networkChainIds[network]] = AirnodeDeployment.receipt.contractAddress;
-  const ConvenienceDeployment = require(`./deployments/${network}/Convenience.json`);
-  ConvenienceAddresses[networkChainIds[network]] = ConvenienceDeployment.receipt.contractAddress;
-}
+let AirnodeDeployment, ConvenienceDeployment;
```
andreogle, line 7: Having these mutable variables that get overwritten feels like a weird pattern to me. It may even cause unexpected behaviour if requires aren't executed in order. Last I checked, using `import` syntax doesn't respect ordering.

I would rather have unique names for each require just to be safe. Something like:

```ts
const RopstenAirnode = require('./deployments/ropsten/Airnode.json');
const RopstenConvenience = require('./deployments/ropsten/Convenience.json');

const RinkebyAirnode = require('./deployments/rinkeby/Airnode.json');
const RinkebyConvenience = require('./deployments/rinkeby/Convenience.json');
```

andreogle (approved):

PR #169, opened 2020-12-23
Add @airnode/operation package

andreogle:
### Changes

Adds a new `@airnode/operation` package that allows for easily setting up an Airnode development environment.

The following commands are provided:

1. `yarn run dev:eth-node` - starts an Eth node on 127.0.0.1:8545. Uses Hardhat behind the scenes
2. `yarn run dev:eth-deploy` - deploys a full Airnode setup (API providers, templates, requesters etc) to the Eth node and funds the necessary accounts. Also saves the output in a "deployment" file in a (git ignored) `deployments/` folder. 
3. `yarn run dev:eth-requests` - loads the deployment file and makes requests.

The functions for these scripts have been extracted into smaller exported modules. The intention is to re-use them for E2E tests in the node package.

Configuration is done through the use of a JSON config file in `src/config/evm-dev-config.json`. You can find more details of this in the operation package README.

This package will need future work for things like supporting different client and authorizer contracts, but I figured it's better to get the initial version merged sooner rather than later


packages/node/src/evm/contracts/types.ts
```diff
@@ -1,10 +1,7 @@
-import { ChainID } from '../networks';
-
 // TODO: Can't find the ethers.js type for ABIs
 export type ABI = string[] | any;
 
 export interface Contract {
   ABI: ABI;
-  addresses: { [chainId in ChainID]: string };
```
bbenligiray, line 8: lol good point

packages/operation/src/evm/deploy/account-assignment.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export async function assignDesignatedWallets(state: State) {
+  const requestersById: { [id: string]: RequesterAccount } = {};
+  for (const configRequester of state.config.requesters) {
+    const requester = state.requestersById[configRequester.id];
+    const designatedProviderNames = Object.keys(configRequester.apiProviders);
+
+    const designatedWallets = designatedProviderNames.map((providerName) => {
+      const apiProvider = state.apiProvidersByName[providerName];
+      const wallet = deriveWalletFromPath(apiProvider.mnemonic, `m/0/0/${requester.requesterIndex}`, state.provider);
```
bbenligiray, line 55: An extra `/0` here

packages/operation/src/evm/deploy/funding.ts
```diff
@@ -0,0 +1,44 @@
...
+    });
+  }
+  return state;
+}
+
+export async function fundRequesterAccounts(state: State): Promise<State> {
+  for (const requesterId of Object.keys(state.requestersById)) {
+    const requester = state.requestersById[requesterId];
+
+    // Ensure that the API provider address has enough ETH to create the onchain provider
```
bbenligiray, line 21: Update

packages/operation/src/evm/deploy/funding.ts
```diff
@@ -0,0 +1,44 @@
...
+  return state;
+}
+
+export async function fundDesignatedWallets(state: State): Promise<State> {
+  for (const configRequester of state.config.requesters) {
+    const requester = state.requestersById[configRequester.id];
+    for (const designatedWallet of requester.designatedWallets) {
+      const ethAmount = configRequester.apiProviders[designatedWallet.apiProviderName].ethBalance;
+
+      // Ensure that the API provider address has enough ETH to create the onchain provider
```
bbenligiray, line 36: Update

packages/operation/src/evm/deploy/providers.ts
```diff
@@ -0,0 +1,45 @@
+import { ethers } from 'ethers';
+import { deriveProviderId } from '../utils';
+import { DeployState as State } from '../../types';
+
+export async function createProviders(state: State): Promise<State> {
+  for (const apiProviderName of Object.keys(state.apiProvidersByName)) {
+    const apiProvider = state.apiProvidersByName[apiProviderName];
+
+    // Ensure that the API provider address has enough ETH to create the onchain provider
+    await state.deployer.sendTransaction({ to: apiProvider.address, value: ethers.utils.parseEther('5') });
```
bbenligiray, line 10: There is also a `fundProviderAccounts()` function

packages/operation/src/evm/deploy/providers.ts
```diff
@@ -0,0 +1,45 @@
...
+  for (const providerName of Object.keys(state.apiProvidersByName)) {
+    const configApiProvider = state.config.apiProviders[providerName];
+    const apiProvider = state.apiProvidersByName[providerName];
+    const providerId = deriveProviderId(apiProvider.address);
+
+    for (const endpointName of Object.keys(configApiProvider.endpoints)) {
+      const configEndpoint = configApiProvider.endpoints[endpointName];
+
+      const { keccak256, defaultAbiCoder } = ethers.utils;
+      const endpointId = keccak256(defaultAbiCoder.encode(['string'], [endpointName]));
```
bbenligiray, line 31: It's `${oisTitle}/${endpointName}`, I assume you're only using the endpoint name for simplicity (which is okay)

packages/operation/src/evm/deploy/providers.ts
```diff
@@ -0,0 +1,45 @@
...
+
+      const authorizerAddresses = configEndpoint.authorizers.reduce((acc: string[], authorizerName: string) => {
+        const address = state.authorizersByName[authorizerName];
+        return [...acc, address];
+      }, []);
+
+      // Ethers can't estimate a gas limit here so just set it really high
```
bbenligiray, line 38: This is probably not related but ethers can't estimate a gas limit for `updateEndpointAuthorizers()` when the provider has not yet been created. Just wanted to point this out because you will encounter it in the future.

packages/operation/src/evm/requests/make-requests.ts
```diff
@@ -0,0 +1,96 @@
+import { ethers } from 'ethers';
+import { encode } from '@airnode/airnode-abi';
+import { deriveProviderId, getDesignatedWallet } from '../utils';
+import { FullRequest, RegularRequest, RequestsState as State, RequestType, ShortRequest } from '../../types';
+
+const CLIENT_ABI = [
```
bbenligiray, line 6: Should import from `@airnode/protocol` tbh

packages/operation/src/evm/requests/make-requests.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export async function makeShortRequest(state: State, request: ShortRequest) {
+  const { privateKey } = state.deployment.requesters[request.requesterId];
+  const signer = new ethers.Wallet(privateKey, state.provider);
+
+  const clientAddress = state.deployment.clients[request.client];
+  const client = new ethers.Contract(clientAddress, CLIENT_ABI, state.provider);
+  const encodedParameters = encode(request.parameters);
+
+  const templateAddress = state.deployment.apiProviders[request.apiProvider].templates[request.template];
```
bbenligiray, line 22: Why not call this `templateId`?

packages/operation/src/evm/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+  const designatorHdNode = masterHdNode.derivePath(path);
+  return new ethers.Wallet(designatorHdNode.privateKey, provider);
+}
+
+export function getDesignatedWallet(
+  mnemonic: string,
+  requesterIndex: string,
+  provider: ethers.providers.JsonRpcProvider
+) {
+  return deriveWalletFromPath(mnemonic, `m/0/0/${requesterIndex}`, provider);
```
bbenligiray, line 24: Extra `/0`

packages/operation/README.md
```diff
@@ -0,0 +1,257 @@
...
+  "requesters": [],
+  "requests": []
+}
+```
+
+### 1. apiProviders
+
+`apiProviders` must have a unique name as the key.
+
+`mnemonic` - must be a unique 12 or 24 list of dictionary words. You can generate a mnemonic [here](https://iancoleman.io/bip39/). **DO NOT SEND REAL FUNDS TO A WALLET LINKED TO A TEST MNEMONIC**
```
bbenligiray, line 95: You can add here that the master wallet will be derived from the mnemonic, and master wallet address will also be the provider admin. I was a bit confused about not being able to find the provider admin address anywhere.

bbenligiray (approved):
Some minor comments but looks good otherwise.

My big comment is that I exported some small functions to interact with the contracts
https://github.com/api3dao/airnode-admin/blob/main/src/index.ts
and enjoyed using them here
https://github.com/api3dao/airnode-starter/tree/main/scripts

So I think we should maintain a single SDK to interact with the contracts and you can also use it in this package. If we are going to do that, it would make sense to move it in the monorepo as `@airnode/admin`. I'm not in a rush to do that, I'm waiting for us to be able to claim the `@airnode` organization from npmjs.com, which should happen in 2 weeks.

packages/operation/src/evm/deploy/providers.ts
andreogle: This is a leftover from before I refactored the functions and isn't needed so I'll remove. Thanks

packages/operation/src/evm/deploy/providers.ts
andreogle: Nice catch - I forgot about this. I'd rather keep it the same just to avoid confusion.

packages/operation/src/evm/requests/make-requests.ts
andreogle: Yeah this is a mistake

packages/operation/src/evm/deploy/providers.ts
andreogle: Thanks - I wasn't sure why this was the case. Is it fine to just leave the gas limit really high here then?

bbenligiray: Yeah I think so.

PR #170, opened 2020-12-24
Fix designated wallet derivation convention

bbenligiray:
I think we miscommunicated about this. It's better to use `m/0/#` as the path because it cuts down on running time, we definitely won't need more than 2^31 requesters and we can still extend this by using `m/0/*/...`.
The designated wallet addresses updated in the tests are calculated with a separate script.


andreogle (approved):
Nice

PR #171, opened 2020-12-24
Update build scripts

bbenligiray:
- `yarn run build` no longer builds the deployer
- The deployer build script is renamed as `_build` for lerna to exclude it. I considered something like `build-separately` but it loses its meaning when you consider the package by itself and not as a part of the monorepo. I don't feel strongly about it though.
- `yarn run build-all` deploys all packages (and will fail if `config.json`&`security.json` are missing)
- `Dockerfile` uses `build` and not `build-all` because Serverless already builds the deployer package. The previous setup was redundant.


andreogle (approved):
🚀

PR #172, opened 2020-12-25
Lambda fixes

bbenligiray:
- Changed the `triggers` field from `requests` to [`request`](https://github.com/api3dao/api3-docs/blob/master/airnode/config-json.md#triggers)
- AWS Lambda InvocationResponse is
```ts
export interface InvocationResponse {
    StatusCode?: Integer;
    FunctionError?: String;
    LogResult?: String;
    Payload?: _Blob;
    ExecutedVersion?: Version;
  }
```
where `data` is in `Payload` under `body` in stringified form. So updated the worker to return the parsed `data` and updated the related tests.
- `ApiCall` was stringifying the response unlike the provider functions causing inconsistency, so removed that.

### DIFF
--- a/packages/node/src/workers/cloud-platforms/aws.ts
+++ b/packages/node/src/workers/cloud-platforms/aws.ts
@@ -18,7 +18,7 @@ export function spawn(params: WorkerParameters): Promise<WorkerResponse> {
       if (err) {
         reject(err);
       }
-      resolve(data as WorkerResponse);
+      resolve(JSON.parse(JSON.parse(data.Payload as string).body) as WorkerResponse);
     });
   });
 }
--- a/packages/node/test/fixtures/config/config.ts
+++ b/packages/node/test/fixtures/config/config.ts
@@ -6,7 +6,7 @@ export function buildConfig(config?: Partial<Config>): Config {
   return {
     id: 'my-config',
     triggers: {
-      requests: [{ endpointId: 'endpointId', endpointName: 'endpointName', oisTitle: 'oisTitle' }],
+      request: [{ endpointId: 'endpointId', endpointName: 'endpointName', oisTitle: 'oisTitle' }],
     },
     ois: [ois.buildOIS()],
     nodeSettings: settings.buildNodeSettings(),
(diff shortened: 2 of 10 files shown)


packages/node/src/workers/cloud-platforms/aws.ts
```diff
@@ -18,7 +18,7 @@ export function spawn(params: WorkerParameters): Promise<WorkerResponse> {
       if (err) {
         reject(err);
       }
-      resolve(data as WorkerResponse);
+      resolve(JSON.parse(JSON.parse(data.Payload as string).body) as WorkerResponse);
```
andreogle, line 21: The double parse is pretty weird 🤷🏻‍♂️

packages/node/test/fixtures/config/config.ts
```diff
@@ -6,7 +6,7 @@ export function buildConfig(config?: Partial<Config>): Config {
   return {
     id: 'my-config',
     triggers: {
-      requests: [{ endpointId: 'endpointId', endpointName: 'endpointName', oisTitle: 'oisTitle' }],
+      request: [{ endpointId: 'endpointId', endpointName: 'endpointName', oisTitle: 'oisTitle' }],
```
andreogle, line 9: You missed the example `config.json` files. I think these are only two:

```
packages/node/__dev__/config.json.example
packages/deployer/src/config-data/config.json.example

```

andreogle (approved):

packages/node/src/workers/cloud-platforms/aws.ts
bbenligiray: The problem is that `JSON.parse()` is not deep so
```js
const doublyStringifiedJson = JSON.stringify({name1: JSON.stringify({name2: 'value'})});
console.log(typeof JSON.parse(doublyStringifiedJson).name1);
```
prints `"string"`. Here's a related question:
https://stackoverflow.com/questions/31662849/parse-json-object-that-stringify-twice
We're not stringifying anything explicitly at the functions so I don't see any way around it.

packages/node/test/fixtures/config/config.ts
bbenligiray: Oops

PR #173, opened 2020-12-31
E2E tests

andreogle:
### Changes

This PR adds the ability to run end-to-end (E2E) tests using the node. Each E2E test goes through the full process of deploying Airnode, creating the relevant onchain data, making requests and invoking the node. These tests can be run in parallel as they are completely isolated and are especially helpful as they test that the node is working as a whole.

Along with these tests, a mock web API has been included in the `operation` package. This API uses Express.js and exposes a simple endpoint. This will grow over time with the complexity of E2E tests.

In order to run the E2E tests, a user first needs to start the web API and a Hardhat Ethereum node. Simple scripts have been provided to do this for them - the simplest being `yarn run dev:background` that starts both processes in the background using [PM2](https://pm2.keymetrics.io/). See the operation README for more information.


.github/workflows/build-test.yml
```diff
@@ -0,0 +1,206 @@
+name: Continuous Build
```
andreogle, line 1: Made some changes to the Github CI workflow so that tests run in parallel

packages/node/test/fixtures/config/node-settings.ts
```diff
@@ -14,9 +14,9 @@ export function buildNodeSettings(settings?: Partial<NodeSettings>): NodeSetting
           Airnode: '0x197F3826040dF832481f835652c290aC7c41f073',
           Convenience: '0x2393737d287c555d148012270Ce4567ABb1ee95C',
         },
-        id: 1337,
+        id: 31337,
```
andreogle, line 17: Hardhat default is 31337

packages/operation/src/evm/deploy/funding.ts
```diff
@@ -5,21 +5,23 @@ export async function fundProviderAccounts(state: State): Promise<State> {
   for (const apiProviderName of Object.keys(state.apiProvidersByName)) {
     const apiProvider = state.apiProvidersByName[apiProviderName];
     // Ensure that the API provider address has enough ETH to create the onchain provider
-    await state.deployer.sendTransaction({
+    const tx = await state.deployer.sendTransaction({
       to: apiProvider.address,
       value: ethers.utils.parseEther('5'),
     });
+    await tx.wait();
```
andreogle, line 12: These `wait` calls are not strictly necessary but better to just have them to avoid any weird behaviour I think

PR #174, opened 2021-01-01
Migrate admin

bbenligiray:
- Migrated the [`airnode-admin`](https://github.com/api3dao/airnode-admin) repo as a package in the monorepo
- Created additional `package.publish.json` files to be used to publish monorepo packages as standalone npm packages
- Created some ugly `publish.sh` scripts
- Deployed `@api3/airnode-admin`, `@api3/airnode-abi` and `@api3/airnode-protocol`. Used these packages as dependencies for the packages to be published instead of the placeholder npm packages I had published earlier. Also started using these in the [starter example](https://github.com/api3dao/airnode-starter)


packages/admin/publish.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+sed -i 's=@airnode/protocol=@api3/airnode-protocol=g' ./src/evm.ts
```
bbenligiray, line 3: `sed` stuff is absolutely disgusting. Let me know if you can think of an alternative way of doing this.

bbenligiray:
By the way I wonder what you didn't like with the admin interface, I liked using it quite a lot

packages/admin/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "main": "dist/index",
+  "scripts": {
+    "build": "yarn run clean && yarn run compile",
+    "clean": "rimraf -rf ./dist",
+    "compile": "tsc -p tsconfig.build.json"
+  },
+  "dependencies": {
+    "@airnode/protocol": "^0.1.0",
+    "@airnode/airnode-abi": "^0.1.0",
+    "ethers": "^5.0.24",
```
andreogle, line 15: I don't think we need `ethers` here since it's in the root package.json

packages/admin/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "yargs": "^16.2.0"
+  },
+  "devDependencies": {
+    "@types/node": "^14.14.11",
```
andreogle, line 19: This one is also in the root package.json

packages/admin/package.json
andreogle: Also applies to the package.json below

packages/admin/publish.sh
andreogle: Nothing comes to mind

packages/airnode-abi/package.publish.json
```diff
@@ -0,0 +1,18 @@
...
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "yarn run clean && yarn run compile",
+    "clean": "rimraf -rf ./dist",
+    "compile": "tsc -p tsconfig.build.json"
+  },
+  "dependencies": {
+    "lodash": "^4.17.20"
```
andreogle, line 16: Is it necessary to add lodash here? This is also available in the root package.json

andreogle (approved):

packages/airnode-abi/package.publish.json
bbenligiray: Yeah because the one in the root package.json is not included in the independently published package. That's essentially why a separate package.publish.json is needed

packages/admin/package.json
bbenligiray: That's true. I'm removing ethers from package.json but not package.publish.json

PR #175, opened 2021-01-03
Add withdrawals to the operation package

andreogle:
### Changes

1. Fixes the `WithdrawalRequested` and `WithdrawalFulfilled` event topics in the node package with the updated signature. This was overlooked previously during the protocol overhaul.

2. Add a new requests type (`withdrawal`) to the `operation` package


bbenligiray (approved):
Based

Issue #64, opened 2020-09-17
Defensiveness against templates with very large parameters fields


bbenligiray:
One can create a template with a very large `parameters` field, which would cause the `Convenience` method to revert due to the exceeding the gas limit, resulting in not being able to retrieve other 9 templates. I'll investigate further.

bbenligiray:
Same thing for endpoint authorizations

bbenligiray:
Even with fairly long parameters/authorizer lists, gas costs are acceptable:
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/test/Convenience.sol.js#L64
https://github.com/api3dao/airnode/blob/696ebad7207b02ffba9b85f2bb57430425c7cd32/packages/protocol/test/Convenience.sol.js#L114
so I don't expect this to be a problem under normal conditions.

A malicious requester may want to create templates with very long `parameters` fields and make a lot of requests that way, but creating the template costs x10 the gas of fetching it, so this attack is almost not feasible. The provider may use a malicious/mis-implemented authorizer that runs an infinite loop, in which case Convenience will revert. In either case, we would want to fallback on making individual calls for each of the 10 elements, but ignore requests whose templates/authorizer revert, rather than have them block.

andreogle:
> In either case, we would want to fallback on making individual calls for each of the 10 elements,

Yeah agreed

> but ignore requests whose templates/authorizers revert

Not sure I understand this part. How do I check this?

bbenligiray:
Correcting, I should have said error, not ignore.
So if a batch convenience call reverts, individual calls need to be made. If an individual call reverts, the corresponding request should be errored.

PR #184, opened 2021-01-26
Template fetching fallback

andreogle:
### Changes

1. If a call to `convenience.getTemplates` fails, the node will now fall back to fetching individual templates using `airnode.getTemplate()` (in parallel)
2. I've exported the new `fetchTemplate` function since it seems pretty useful as is
3. Added some explanation comments to the `constants.ts` file

### Addresses

https://github.com/api3dao/airnode/issues/64 although there are other areas where the node should fallback to individual calls. I'm going to add a checklist to the Github issue and leave it open

### DIFF
--- a/packages/node/src/constants.ts
+++ b/packages/node/src/constants.ts
@@ -1,12 +1,38 @@
-//API calls
+// The maximum TOTAL time an API call has before it is timed out (including retries).
 export const API_CALL_TOTAL_TIMEOUT = 29_000;
+
+// The maximum time an API call has before it is timed out (and retried).
 export const API_CALL_TIMEOUT = 20_000;
+
+// The number of past blocks to lookup when fetching Airnode events.
 export const BLOCK_COUNT_HISTORY_LIMIT = 300;
+
+// Certain events cause requests to be "blocked" (e.g. the template cannot be fetched)
+// In order to preserve nonce ordering, these blocked requests also cause later requests
+// to become blocked. Once this number of blocks has passed, these blocked requests will become
+// "ignored" and no longer block later requests.
 export const BLOCK_COUNT_IGNORE_LIMIT = 20;
+
+// The minimum number of block confirmations required.
 export const BLOCK_MIN_CONFIRMATIONS = 0;
+
+// The Convenience contract allows for returning multiple items in order to reduce calls
+// to the blockchain provider. This number is the maximum number of items that can get returned
+// in a single call.
 export const CONVENIENCE_BATCH_SIZE = 10;
-export const DEFAULT_RETRY_OPERATION_TIMEOUT = 5000;
+
+// The default amount of time before a "retryable" promise is timed out and retried
+export const DEFAULT_RETRY_OPERATION_TIMEOUT = 5_000;
+
+// The amount of time EVM provider calls are allowed
 export const EVM_PROVIDER_TIMEOUT = 10_000;
+
+// The default number of times a "retryable" promise is retried
+// TODO: this should be 1 to represent a single retry
 export const OPERATION_RETRIES = 2;
+
+// The maximum amount of time the "initialize provider" worker is allowed before being timed out
 export const WORKER_PROVIDER_INITIALIZATION_TIMEOUT = 20_000;
+
+// The maximum amount of time the "call API" worker is allowed before being timed out
 export const WORKER_CALL_API_TIMEOUT = 29_500;
--- a/packages/node/src/constants.ts
+++ b/packages/node/src/constants.ts
@@ -1,7 +1,7 @@
 //API calls
 export const API_CALL_TOTAL_TIMEOUT = 29_000;
 export const API_CALL_TIMEOUT = 20_000;
-export const BLOCK_COUNT_HISTORY_LIMIT = 600;
+export const BLOCK_COUNT_HISTORY_LIMIT = 300;
 export const BLOCK_COUNT_IGNORE_LIMIT = 20;
 export const BLOCK_MIN_CONFIRMATIONS = 0;
 export const CONVENIENCE_BATCH_SIZE = 10;
--- a/packages/node/src/constants.ts
+++ b/packages/node/src/constants.ts
@@ -32,7 +32,7 @@ export const EVM_PROVIDER_TIMEOUT = 10_000;
 export const OPERATION_RETRIES = 2;
 
 // The maximum amount of time the "initialize provider" worker is allowed before being timed out
-export const WORKER_PROVIDER_INITIALIZATION_TIMEOUT = 20_000;
+export const WORKER_PROVIDER_INITIALIZATION_TIMEOUT = 19_500;
 
 // The maximum amount of time the "call API" worker is allowed before being timed out
 export const WORKER_CALL_API_TIMEOUT = 29_500;
--- a/packages/node/src/evm/templates/template-fetching.test.ts
+++ b/packages/node/src/evm/templates/template-fetching.test.ts
@@ -1,3 +1,4 @@
+const getTemplateMock = jest.fn();
 const getTemplatesMock = jest.fn();
 jest.mock('ethers', () => {
   const original = jest.requireActual('ethers');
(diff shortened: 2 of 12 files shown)


packages/node/src/evm/templates/template-fetching.ts
```diff
@@ -64,11 +103,12 @@ export async function fetch(
   // Requests are made for up to 10 templates at a time
   const groupedTemplateIds = chunk(uniq(templateIds), CONVENIENCE_BATCH_SIZE);
 
-  // Create an instance of the contract that we can re-use
+  // Create an instance of earch contract that we can re-use
```
bbenligiray, line 106:
```suggestion
  // Create an instance of each contract that we can re-use
```

packages/node/src/evm/templates/template-fetching.test.ts
```diff
@@ -175,28 +173,182 @@ describe('fetch (templates)', () => {
...
-    const rawTemplates = {
-      designatedWallets: ['designatedWallet-0'],
-      endpointIds: ['endpointId-0'],
-      providerIds: ['providerId-0'],
-      fulfillAddresses: ['fulfillAddresses-0'],
-      fulfillFunctionIds: ['fulfillFunctionId-0'],
-      parameters: ['0x6874656d706c6174656576616c7565'],
-      requesterIndices: ['requesterIndex-0'],
+  it('fetches templates individually if the template group cannot be fetched', async () => {
+    getTemplatesMock.mockRejectedValueOnce(new Error('Server says no'));
```
bbenligiray, line 177: `OPERATION_RETRIES` is hardcoded in all these tests. Probably not worth it to fix right now but we can create an issue.

bbenligiray (approved):
Looks good, left a few comments. I think in the final version, a too large `parameters` will revert during template creation
https://github.com/api3dao/airnode/blob/3bed327c64b72a73dc6dc3a8a8416aa3be9c73be/packages/protocol/contracts/TemplateStore.sol#L86
so that it'll be less likely for `getTemplates` to fail.

I'm thinking of a `bytes` length limit of about `1024`. The main reason we would want to allow a large `parameters` is for off-chain computation specifications (i.e. `_jsin` and `_jsout`). So this limit will have to be chosen carefully (because it will be hardcoded to the protocol).

packages/node/src/constants.ts
```diff
@@ -1,12 +1,38 @@
...
+
+// The amount of time EVM provider calls are allowed
 export const EVM_PROVIDER_TIMEOUT = 10_000;
+
+// The default number of times a "retryable" promise is retried
+// TODO: this should be 1 to represent a single retry
 export const OPERATION_RETRIES = 2;
+
+// The maximum amount of time the "initialize provider" worker is allowed before being timed out
 export const WORKER_PROVIDER_INITIALIZATION_TIMEOUT = 20_000;
```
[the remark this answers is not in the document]
andreogle: Yeah I'll adjust this

packages/node/src/constants.ts
```diff
@@ -1,12 +1,38 @@
...
+export const DEFAULT_RETRY_OPERATION_TIMEOUT = 5_000;
+
+// The amount of time EVM provider calls are allowed
 export const EVM_PROVIDER_TIMEOUT = 10_000;
+
+// The default number of times a "retryable" promise is retried
+// TODO: this should be 1 to represent a single retry
 export const OPERATION_RETRIES = 2;
+
+// The maximum amount of time the "initialize provider" worker is allowed before being timed out
```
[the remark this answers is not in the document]
andreogle: Would that allow users to customise the timeouts? I don't think it's that big of a deal being hardcoded like this

packages/node/src/evm/templates/template-fetching.test.ts
andreogle: Not sure I understand?

bbenligiray: No, I'm saying changing the `2` here
https://github.com/api3dao/airnode/blob/7472b30861d442602ea6176ad7a6a628de6a0e5a/packages/node/src/constants.ts#L32
would result in the tests failing. I think the point of keeping the parameters in one file is that you would be able to change the value in that file and wouldn't have to do anything else.

packages/node/src/constants.ts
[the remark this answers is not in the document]
bbenligiray: What I mean is this line
https://github.com/api3dao/airnode/blob/7472b30861d442602ea6176ad7a6a628de6a0e5a/packages/node/src/constants.ts#L35
implies that you can update the value as `30_000` to give the node more time to initialize the provider. It really not be the case though.

bbenligiray:
My main complaint is that updating the values at `constants.ts` either breaks things or results in unexpected outcomes. This PR is not about `constants.ts` so you can merge it. It's just that the added comments brought it to my attention as an issue.

packages/node/src/constants.ts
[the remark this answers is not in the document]
andreogle: Well technically you could adjust these values if running locally, like in a Docker container or Node.js process.

I don't know if we should make the values environment variables. If it means we can have one source of truth for the values (for the node itself and the deployer) then maybe that makes sense

packages/node/src/evm/templates/template-fetching.test.ts
andreogle: It shouldn't break any tests as functionality would be the same. So these changes:

1. Change the `OPERATION_RETRIES` value to be `1`
2. Change the internal parts of the `retryOperation` function to work with values that are one less.

The end result would be:

```ts
// Good: Retry once on failure
retryOperation(1, () => myFunction());

// Bad: implies 2 retries
retryOperation(2, () => myFunction());
```

packages/node/src/constants.ts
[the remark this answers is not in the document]
bbenligiray: Good point about running it locally. I'll just create an issue that we can potentially work on at some point.

packages/node/src/evm/templates/template-fetching.test.ts
bbenligiray: I meant all the mocks in the tests assume `OPERATION_RETRIES` is 2

bbenligiray: Updated 2->3, 24 tests fail. I created the issue, we can deal with it later.

andreogle: Yeah most mocks only expect a single retry so that makes sense.

Going from 2 -> 1 wouldn't break anything (but we need to update the internals of the `retryOperation()` function too)

bbenligiray: Ah I see what you mean now. I wasn't referring to the TODO about changing 2 to 1 at all. What I'm saying is that say someone wants the node to retry one more time and they increment `OPERATION_RETRIES` at `constants.ts`. They would also have to update all the tests. Similarly if they didn't want any retries, they would have to decrement that variable and also update all the tests. Ideally the tests would import `OPERATION_RETRIES` and create the mocks with a loop.

andreogle: Yeah I figured we were talking about different things. With some of the tests we could make the following changes which would work regardless of how many times the mock is called

```js
mockReturnValueOnce() -> mockReturnValue()
mockResolvedValueOnce() -> mockResolvedValue()
mockRejectedValueOnce() -> mockRejectedValue()
```

I didn't do this initially because I preferred more fine grained control over what gets mocked, but it's not a big deal to switch. This wouldn't work for tests where the initial call fails and the subsequent call succeeds.

bbenligiray: I actually like the fine grained mocks. I think if we ever make this change
```js
getTemplatesMock.mockRejectedValueOnce(new Error('Server says no'));
getTemplatesMock.mockResolvedValueOnce(rawTemplates);
```
should be replaced with
```js
for (var i = 0; i < OPERATION_RETRIES - 1; i++) {
  getTemplatesMock.mockRejectedValueOnce(new Error('Server says no'));
}
getTemplatesMock.mockResolvedValueOnce(rawTemplates);
```

andreogle: Yeah something like that could work

Issue #143, opened 2020-11-24
Check if the user has noted down the mnemonic during deployment


bbenligiray:
Currently, we ask the user to note down the randomly generated mnemonic during the first deployment
https://github.com/api3dao/airnode/blob/421822c2799dda9d6e9a8e6dab79b939f7a1984b/packages/deployer/src/commands.ts#L35
However, the user may skip to do this. To make sure, we should clear the terminal afterwards and do a check such as "Enter the Nth word of you mnemonic" where N is a randomly generated integer between 1-12.
This check should be able to be disabled with a flag for debugging, etc.

Arrowana:
What behaviour do we expect if one of the word is wrong when verifying? Should it give the mnemonic sentence again, then disappear after pressing return only to go throught the verification process again?

bbenligiray:
Yes, that sounds right.
I think it should be the same mnemonic, because it's likely because they have actually noted it down but simply made a typo while entering a word.

bbenligiray:
Implemented in https://github.com/api3dao/airnode/pull/186

PR #186, opened 2021-01-31
Mnemonic verification process

Arrowana:
Implements #143 

Remaining work to do:

- Bypass verification while debugging, do i need to add an new --debug argument?

### DIFF
--- a/packages/deployer/src/commands.ts
+++ b/packages/deployer/src/commands.ts
@@ -18,61 +18,7 @@ import {
   shortenProviderId,
   deriveMasterWalletAddress,
 } from './util';
-import readline from 'readline';
-
-function shuffleArray(array) {
-  for (let i = array.length - 1; i > 0; i--) {
-    const j = Math.floor(Math.random() * (i + 1));
-    [array[i], array[j]] = [array[j], array[i]];
-  }
-  return array;
-}
-
-function ask(rl, question) {
-  return new Promise((resolve) => {
-    rl.question(question, (answer) => resolve(answer));
-  });
-}
-
-function clearLine() {
-  readline.moveCursor(process.stdout, 0, -1);
-  readline.clearLine(process.stdout, 1);
-}
-
-async function verifyMnemonic(mnemonic) {
-  const mnemonics = mnemonic.split(' ');
-  const shuffledIndexedMnemonics = shuffleArray(
-    mnemonics.map((element, index) => {
-      return { mnemonic: element, index: index + 1 };
-    })
-  );
-
-  const rl = readline.createInterface({
-    input: process.stdin,
-    output: process.stdout,
-  });
-
-  console.log(mnemonic);
-  await ask(rl, `press return to start mnemonic verification, the mnemonic will disappear!`);
-  clearLine();
-  clearLine();
-
-  let correct = false;
-  while (!correct) {
-    correct = true;
-    for (const indexedMnemonic of shuffledIndexedMnemonics) {
-      const word = await ask(rl, `Enter ${indexedMnemonic.index}-th word:`);
-      clearLine();
-      if (word != indexedMnemonic.mnemonic) {
-        correct = false;
-        await ask(rl, `Incorrect ${indexedMnemonic.index}-th word, press return to verification again`);
-        break;
-      }
-    }
-  }
-  console.log('Mnemonic verification passed');
-  rl.close();
-}
+import { verifyMnemonic } from './io';
 
 export async function deployFirstTime(configPath, securityPath, nodeVersion) {
   const configParams = await parseFiles(configPath, securityPath);
--- a/packages/deployer/src/commands.ts
+++ b/packages/deployer/src/commands.ts
@@ -19,6 +19,7 @@ import {
   waitForEnter,
   deriveMasterWalletAddress,
 } from './util';
+import { verifyMnemonic } from './io';
 
 export async function deployFirstTime(configPath, securityPath, nodeVersion) {
   const configParams = await parseFiles(configPath, securityPath);
@@ -33,8 +34,7 @@ export async function deployFirstTime(configPath, securityPath, nodeVersion) {
   }
 
   ora().warn('Write down the 12 word-mnemonic below on a piece of paper and keep it in a safe place\n');
-  console.log(mnemonic);
-  await waitForEnter();
+  await verifyMnemonic(mnemonic);
   await ssm.addMnemonic(mnemonic, providerIdShort);
 
   const masterWalletAddress = deriveMasterWalletAddress(mnemonic);
--- a/packages/deployer/src/commands.ts
+++ b/packages/deployer/src/commands.ts
@@ -16,9 +16,9 @@ import {
   generateMnemonic,
   deriveProviderId,
   shortenProviderId,
+  waitForEnter,
   deriveMasterWalletAddress,
 } from './util';
-import { verifyMnemonic } from './io';
 
 export async function deployFirstTime(configPath, securityPath, nodeVersion) {
   const configParams = await parseFiles(configPath, securityPath);
@@ -33,7 +33,8 @@ export async function deployFirstTime(configPath, securityPath, nodeVersion) {
   }
 
   ora().warn('Write down the 12 word-mnemonic below on a piece of paper and keep it in a safe place\n');
-  await verifyMnemonic(mnemonic);
+  console.log(mnemonic);
+  await waitForEnter();
   await ssm.addMnemonic(mnemonic, providerIdShort);
 
   const masterWalletAddress = deriveMasterWalletAddress(mnemonic);
--- a/packages/deployer/src/commands.ts
+++ b/packages/deployer/src/commands.ts
@@ -16,7 +16,6 @@ import {
   generateMnemonic,
   deriveProviderId,
   shortenProviderId,
-  waitForEnter,
   deriveMasterWalletAddress,
 } from './util';
 import { verifyMnemonic } from './io';
--- a/packages/deployer/src/commands.ts
+++ b/packages/deployer/src/commands.ts
@@ -16,9 +16,63 @@ import {
   generateMnemonic,
   deriveProviderId,
   shortenProviderId,
-  waitForEnter,
   deriveMasterWalletAddress,
 } from './util';
+import readline from 'readline';
+
+function shuffleArray(array) {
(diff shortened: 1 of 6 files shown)


packages/deployer/src/commands.ts
```diff
@@ -16,9 +16,63 @@ import {
   generateMnemonic,
   deriveProviderId,
   shortenProviderId,
-  waitForEnter,
   deriveMasterWalletAddress,
 } from './util';
+import readline from 'readline';
+
+function shuffleArray(array) {
```
bbenligiray, line 23:
```suggestion
// Adapted from https://stackoverflow.com/a/12646864/14558682
function shuffleArray(array) {
```

packages/deployer/src/commands.ts
```diff
@@ -16,9 +16,63 @@ import {
...
+  let correct = false;
+  while (!correct) {
+    correct = true;
+    for (const indexedMnemonic of shuffledIndexedMnemonics) {
+      const word = await ask(rl, `Enter ${indexedMnemonic.index}-th word:`);
+      clearLine();
+      if (word != indexedMnemonic.mnemonic) {
+        correct = false;
+        await ask(rl, `Incorrect ${indexedMnemonic.index}-th word, press return to verification again`);
+        break;
```
bbenligiray, line 69: When the first word is right and the second word is wrong, this takes the user back to the first word. I think it should have the user repeat the second word.

packages/deployer/src/commands.ts
```diff
@@ -16,9 +16,63 @@ import {
...
+      }
+    }
+  }
+  console.log('Mnemonic verification passed');
```
bbenligiray, line 73:
```suggestion
  ora().succeed('Mnemonic verified successfully');
```

packages/deployer/src/commands.ts
```diff
@@ -16,9 +16,63 @@ import {
...
+    mnemonics.map((element, index) => {
+      return { mnemonic: element, index: index + 1 };
+    })
+  );
+
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
+
```
bbenligiray, line 54:
```suggestion

  ora().info('When you press Enter, the mnemonic below will disappear and you will be asked to provide 3 of the words selected at random.\n');
```

packages/deployer/src/commands.ts
```diff
@@ -16,9 +16,63 @@ import {
...
+  console.log(mnemonic);
+  await ask(rl, `press return to start mnemonic verification, the mnemonic will disappear!`);
```
bbenligiray, line 56:
```suggestion
  await ask(rl, '');
```

packages/deployer/src/commands.ts
```diff
@@ -16,9 +16,63 @@ import {
...
+  clearLine();
+  clearLine();
```
bbenligiray, line 58:
```suggestion
```

packages/deployer/src/commands.ts
bbenligiray: The `ora` symbols just look nice

packages/deployer/src/commands.ts
```diff
@@ -16,9 +16,63 @@ import {
...
+
+  let correct = false;
+  while (!correct) {
+    correct = true;
+    for (const indexedMnemonic of shuffledIndexedMnemonics) {
+      const word = await ask(rl, `Enter ${indexedMnemonic.index}-th word:`);
```
bbenligiray, line 64:
```suggestion
      const word = await ask(rl, `Enter word #${indexedMnemonic.index}`);
```

packages/deployer/src/commands.ts
```diff
@@ -16,9 +16,63 @@ import {
...
+      clearLine();
+      if (word != indexedMnemonic.mnemonic) {
+        correct = false;
+        await ask(rl, `Incorrect ${indexedMnemonic.index}-th word, press return to verification again`);
```
bbenligiray, line 68:
```suggestion
        await ask(rl, `Enter word #${indexedMnemonic.index} again, or exit and start over`);
```

packages/deployer/src/commands.ts
```diff
@@ -16,9 +16,63 @@ import {
...
+  await ask(rl, `press return to start mnemonic verification, the mnemonic will disappear!`);
+  clearLine();
+  clearLine();
+
+  let correct = false;
+  while (!correct) {
+    correct = true;
+    for (const indexedMnemonic of shuffledIndexedMnemonics) {
+      const word = await ask(rl, `Enter ${indexedMnemonic.index}-th word:`);
+      clearLine();
```
bbenligiray, line 65:
```suggestion
```

packages/deployer/src/commands.ts
```diff
@@ -16,9 +16,63 @@ import {
...
+      if (word != indexedMnemonic.mnemonic) {
+        correct = false;
```
bbenligiray, line 67:
```suggestion
        correct = false;
        clearLine();
```

bbenligiray: Only clear a line if we have to. This also allows the user to know how many words they have entered and how many are left.

packages/deployer/src/commands.ts
bbenligiray: Causing things like "2-th"

bbenligiray (changes requested):
Hey thanks a lot for the PR!

Yes, we definitely want a `--skip-mnemonic-verification` flag.

I made a lot of suggestions because the deployer is as user-facing as it gets and we care a lot about UX.

As a note, we'll need to test this with Windows CMD.

packages/deployer/src/commands.ts
```diff
@@ -16,9 +16,63 @@ import {
...
+
+  console.log(mnemonic);
+  await ask(rl, `press return to start mnemonic verification, the mnemonic will disappear!`);
+  clearLine();
+  clearLine();
+
+  let correct = false;
+  while (!correct) {
+    correct = true;
+    for (const indexedMnemonic of shuffledIndexedMnemonics) {
```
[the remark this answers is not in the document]
Arrowana: Are you saying the user should enter only 3 words of the enter mnemonic? Indices to be picked randomly

packages/deployer/src/commands.ts
Arrowana: The goal here is to make the line from 
```
  console.log(mnemonic);
  await ask(rl, `press return to start mnemonic verification, the mnemonic will disappear!`);
  ```
disappear, so that the mnemonic is not left on stdout otherwise the verification is a bit pointless :)

File: packages/deployer/src/commands.ts (hunk 9 above)
Comment (Arrowana) [reply packages/deployer/src/commands.ts:65]: Don't we want to ensure we always erase user input so pieces of the mnemonic are not left in stdout?

File: packages/deployer/src/commands.ts [hunk 12]
```diff
@@ -16,9 +16,63 @@ import {
...
+    rl.question(question, (answer) => resolve(answer));
+  });
+}
+
+function clearLine() {
+  readline.moveCursor(process.stdout, 0, -1);
+  readline.clearLine(process.stdout, 1);
+}
+
+async function verifyMnemonic(mnemonic) {
```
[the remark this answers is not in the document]
Arrowana: I completely agree with the clutter problem, i just wasn't sure where to put that. Will move it to util.ts or io.ts

packages/deployer/src/commands.ts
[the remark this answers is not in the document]
bbenligiray: Yeah, that would pretty much guarantee that they have actually noted down the mnemonic. We're testing for that, and not for if they have done that correctly.

packages/deployer/src/commands.ts
bbenligiray: This comes with the suggestion above that moves the instructions above the mnemonic so a single `clearLine()` is enough to remove the mnemonic

packages/deployer/src/commands.ts
bbenligiray: I didn't consider it as an issue but you're probably right, it's better to remove them

packages/deployer/src/commands.ts
andreogle: Lodash (which we already include) has a function for shuffling arrays:

```ts
import shuffle from 'lodash/shuffle';
```

Arrowana:
Finalised on ubuntu, also gave it a go on windows 10. Looks identical over there.

packages/deployer/src/io.ts
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  ora().info(
+    'When you press Enter, the mnemonic below will disappear and you will be asked to provide 3 of the words selected at random.\n'
+  );
+  await ask(rl, mnemonic);
+  clearLine();
+
+  for (const indexedMnemonic of shuffledIndexedMnemonics) {
+    let word = await ask(rl, `Enter word #${indexedMnemonic.index}`);
```
bbenligiray, line 36:
```suggestion
    let word = await ask(rl, `Enter word #${indexedMnemonic.index}: `);
```

packages/deployer/src/io.ts
```diff
@@ -0,0 +1,45 @@
...
+    clearLine();
+    while (word != indexedMnemonic.mnemonic) {
+      word = await ask(rl, `Enter word #${indexedMnemonic.index} again, or exit and start over`);
```
bbenligiray, line 39:
```suggestion
      word = await ask(rl, `Enter word #${indexedMnemonic.index} again, or exit and start over: `);
```

bbenligiray (approved):
Hey, @Arrowana , sorry for the late review. Everything looks good. Let me know if you will implement `--skip-mnemonic-verification` or if I should merge this and create an issue for that.

packages/deployer/src/io.ts
bbenligiray: Otherwise these look ugly

andreogle (approved):
Awesome. Looks good to me (pending the last suggestions)

Thanks again for this!

Arrowana:
@bbenligiray I will do `--skip-mnemonic-verification`. However, not certain when... Feel free to merge this i can create another PR with the extra bit.

bbenligiray:
New issue -> https://github.com/api3dao/airnode/issues/219

PR #193, opened 2021-02-06
Client specific nonces

bbenligiray:
We were using global nonces for requests and withdrawal requests. This resulted in the request IDs being unpredictable.

**Requests**

Switched to client-specific nonces. We are doing this instead of a requester-specific nonce for example because a requester can have multiple client contracts and they wouldn't know if each other are making requests so the unpredictability would have remained.

Since nonces are client-specific, now we have to also use the client address while creating the `requestId`
https://github.com/api3dao/airnode/blob/client-specific-nonces/packages/protocol/contracts/Airnode.sol#L53
This required updates at the `requestId` verification parts of the node.

The client-specific nonces need to be initialized at 1 for consistent gas usage. This is done while anyone has ever endorsed the client
https://github.com/api3dao/airnode/blob/client-specific-nonces/packages/protocol/contracts/RequesterStore.sol#L73-L78
This works because you can't make a request with an unendorsed client
https://github.com/api3dao/airnode/blob/client-specific-nonces/packages/protocol/contracts/Airnode.sol#L46-L49

There is an additional minor change. `providerId` used to be used for creating full request `requestId`s
https://github.com/api3dao/airnode/blob/master/packages/protocol/contracts/Airnode.sol#L158
This is redundant because we are validating `providerId` during fulfillment
https://github.com/api3dao/airnode/blob/master/packages/protocol/contracts/Airnode.sol#L322
so it's no longer there (not that this provides any improvement)
https://github.com/api3dao/airnode/blob/client-specific-nonces/packages/protocol/contracts/Airnode.sol#L160-L165

**Withdrawal requests**

Switched to requester-specific nonces. This wasn't nearly as critical as the one above but it's easy enough to implement and doesn't require any changes at the node side because withdrawals request IDs don't need to be validated
https://github.com/api3dao/airnode/blob/client-specific-nonces/packages/protocol/contracts/ProviderStore.sol#L77-L79

Similar to using the client address in the request nonces, we now need to also use the `requesterIndex` in the withdrawal request nonces
https://github.com/api3dao/airnode/blob/client-specific-nonces/packages/protocol/contracts/ProviderStore.sol#L97
In case you're wondering about the `this` here
https://github.com/api3dao/airnode/blob/client-specific-nonces/packages/protocol/contracts/ProviderStore.sol#L98
I figured the hash of `(1, 2)` etc. would be used commonly so added the address of the contract in there to scramble things a bit.

The requester-specific nonces need to be initialized at 1 for consistent gas usage. This is done during requester creation
https://github.com/api3dao/airnode/blob/client-specific-nonces/packages/protocol/contracts/RequesterStore.sol#L32


bbenligiray:
This is probably pointless, let's keep this on hold

andreogle (approved):
Looks great

PR #194, opened 2021-02-06
Authorization fetching fallback

andreogle:
I'm targeting the template-fallback branch for now. There are still a few outstanding comments so I haven't merged it yet

### Changes

1. Falls back to individual authorization calls when the authorization batch call fails twice.

### Addresses

https://github.com/api3dao/airnode/issues/64


bbenligiray (approved):
:shipit:

PR #195, opened 2021-02-07
Convertor

### LINKED ISSUE #152: OpenAPI specifications to OIS converter
We need an OIS tool to convert [OpenAPI specifications](https://github.com/OAI/OpenAPI-Specification) to [Oracle Integration Specifications](https://github.com/api3dao/api3-docs/blob/master/airnode/ois.md) automatically (see [this](https://github.com/api3dao/api3-docs/blob/master/provider-guides/api-integration.md#step-1-specifying-the-api)). Please comment below if you are interested in building this.

### LINKED ISSUE #153: OIS endpoint generator
We need an OIS tool to generate `endpoints` that map one-to-one to `apiSpecifications.paths` (see [this](https://github.com/api3dao/api3-docs/blob/master/provider-guides/api-integration.md#conclusion)). Please comment below if you are interested in building this.

Drgy:
Adds conversion functionality into validator, **should be merged after #132** !

Convertor commands are described in the [readme](https://github.com/api3dao/airnode/blob/converter/packages/validator/README.md#Usage), new conversion related keywords used in templates are described [at the bottom](https://github.com/api3dao/airnode/blob/converter/packages/validator/README.md#Actions). 

I tried to make up the conversion from available resources, but there might be stuff that needs to be added/modified/removed. One example conversion from OAS to config & security can be found in the [convertor.test.ts](https://github.com/api3dao/airnode/blob/converter/packages/validator/src/convertor.test.ts), in which [this OAS](https://github.com/api3dao/airnode/blob/converter/packages/validator/exampleSpecs/OAS.specs.json) is transformed into [OIS](https://github.com/api3dao/airnode/blob/converter/packages/validator/exampleSpecs/OIS2.specs.json), based on which [config & security](https://github.com/api3dao/airnode/blob/de37d5978fc7a93546d85a0ac0dbe972e36e1ed2/packages/validator/src/convertor.test.ts#L6) is generated.

Few manual checks of complex OAS, would be very helpful to discover parts that aren't converted correctly, as I really doubt I got everything correct right away.

closes #152
closes #153


packages/validator/README.md
```diff
@@ -208,177 +289,628 @@ Invalid input:
...
+	"messages": [
+		{ "level": "warning", "message": "Extra field: notIgnored" }
+	]
+}
+```
+---
+
+## Actions
+
+If any action is specified in template, object `output` will be returned when processing a specification. This object is at first empty, by inserting or copying parameters from specification it can construct desired specification.
```
sassmilic, line 708: I would add something here that makes it clear that Actions are unique to Converter/conversions. Maybe something as simple as: "Actions are required for conversions". Or maybe add a header "# Convertor specs" above "## Actions".

Drgy: I think issues in the readme are not directly related to convertor functionality, also there are quite few of these problems, I opened whole [new issue](https://github.com/api3dao/airnode/issues/203) just for this, feel free to add anything that you have encountered while going through it.

packages/validator/templates/OAS2OIS.json
```diff
@@ -0,0 +1,288 @@
...
+					}
+				}
+			]
+		},
+		"__ignore": {}
+	},
+	"servers": {
+		"__arrayItem": {
+			"url": {
+				"__regexp": "^(https?|ftp)://[^\\s/$.?#].[^\\s]*$",
```
sassmilic, line 64: I don't see anything immediately wrong with the regex, but I checked the OAS specs and they say:
> Server URL format follows RFC 3986 and usually looks like this:
> ` scheme://host[:port][/path]`

And the RFC 3986 docs actually contain a regular expression:
https://tools.ietf.org/html/rfc3986#appendix-B

although we would ignore:
`
query     = $7
fragment  = $9
`

Maybe it would be safer to use the regex from the protocol document itself. What do you think, @bbenligiray, @Drgy?

sassmilic: @bbenligiray are we to support the exact same server url formats as OAS?

E.g. OAS allows relative URLs as well as `localhost` (for example) and neither of these would be captured by Jakub's regex.

https://swagger.io/docs/specification/api-host-and-base-path/
^ see: "Server URL Format"

sassmilic:
What version of OAS are we using/referring to? (I couldn't find this anywhere in the code ...) Seems like the most recent is 3.1.0? https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md

packages/validator/templates/OAS2OIS.json
```diff
@@ -0,0 +1,288 @@
...
+						"__copy": {
+							"__target": "apiSpecifications.servers[0].url"
+						}
+					}
+				]
+			}
+		}
+	},
+	"paths": {
+		"__keyRegexp": "^\\/[^\\s'\"\\\\]+$",
```
sassmilic, line 76: What the reason for the substring `\\\\` in the regex? Doesn't everything in `[^...]` get treated as a single character? i.e. `[^\\]` has the same effect as `[^\\\\]`.

(Exclaimer: I haven't written regex in Javascript, although I do imagine it's mostly the same as Python, which I'm most comfortable with.)

packages/validator/templates/OAS2OIS.json
```diff
@@ -0,0 +1,288 @@
...
+					},
+					{
+						"__insert": {
+							"__target": "endpoints[_].operation.path",
+							"__value": "{{1}}"
+						}
+					},
+					{
+						"__insert": {
+							"__target": "endpoints[_].reservedParameters",
```
sassmilic, line 107: I think this should be initialized to an empty array. AFAIK, reserved parameters *do not* map to operation parameters (i.e. the parameters of the actual API call(s)). So, if we're converting from OAS to OIS, I think it would make sense to initialize the reserved params to an empty array, because by definition, the OAS wouldn't have any reserved params to specify.

https://github.com/api3dao/api3-docs/blob/8e41ab2597cb19cb75e4904d6e9b214287c1d738/airnode/ois.md#54-reservedparameters

sassmilic: Or, you can make `reservedParameters` an optional key.

packages/validator/templates/OAS2OIS.json
```diff
@@ -0,0 +1,288 @@
...
+								},
+								{
+									"name": "_times"
+								}
+							]
+						}
+					},
+					{
+						"__insert": {
+							"__target": "endpoints[_].fixedOperationParameters",
```
sassmilic, line 123: Nit: similar to my previous comment. Is there a good reason for including `fixedOperationParameters` and not just leaving it out? Since OAS won't specify `fixedOperationParameters` since it's something that's defined on the oracle side.

packages/validator/templates/OAS2OIS.json
```diff
@@ -0,0 +1,288 @@
...
+							"__value": []
+						}
+					}
+				],
+				"parameters": {
+					"__arrayItem": {
+						"name": {
+							"__actions": [
```
sassmilic, line 131: General question about sequence of operations: is the list of actions intended to be executed sequentially **per array element** in outer array? 

More specific to this example: are the actions in `name`, `in`, and `schema` intended to be performed for a given element in `parameters` **before** moving on to perform the same set of actions (in the same order) on the next element in `parameters`?

packages/validator/templates/OAS2OIS.json
```diff
@@ -0,0 +1,288 @@
...
+						"__ignore": {}
+					}
+				},
+				"__ignore": {}
+			}
+		}
+	},
+	"components": {
+		"schemas": {
+			"securitySchemes": {
```
sassmilic, line 194: I think this might be an error?

According to OAS specs, `schemas` and ` securitySchemes` should be on the same level in the Components Object.

packages/validator/templates/OAS2OIS.json
```diff
@@ -0,0 +1,288 @@
...
+								}
+							}
+						},
+						{
+							"__if": {
+								"type": "^http$"
+							},
+							"__then": {
+								"scheme": {
+									"__regexp": "^(Basic|Bearer)$",
```
sassmilic, line 263: Looks like OAS uses lower case: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#basic-authentication-sample

packages/validator/templates/OAS2OIS.json
```diff
@@ -0,0 +1,288 @@
...
+										{
+											"__copy": {
+												"__target": "apiSpecifications.components.securitySchemes.{{3}}.name"
+											}
+										},
+										{
+											"__insert": {
+												"__target": "apiSpecifications.security.{{3}}",
+												"__value": []
+											}
```
sassmilic, line 251: I believe this can be placed in the if-statement above:
```
"__if": {
   "type": "^(apiKey|http)$"
}
```

packages/validator/templates/OAS2OIS.json
```diff
@@ -0,0 +1,288 @@
...
+										{
+											"__copy": {
+												"__target": "apiSpecifications.components.securitySchemes.{{3}}.scheme"
+											}
+										},
+										{
+											"__insert": {
+												"__target": "apiSpecifications.security.{{3}}",
+												"__value": []
+											}
```
sassmilic, line 274: Same as comment: https://github.com/api3dao/airnode/pull/195/files#r578923631

packages/validator/templates/OAS2OIS.json
```diff
@@ -0,0 +1,288 @@
...
+						"__regexp": "^(apiKey|http)$"
+					},
+					"in": {
+						"__regexp": "^(query|header|cookie)$"
+					},
+					"__conditions": [
+						{
+							"__if": {
+								"type": "^(apiKey|http)$"
+							},
```
sassmilic, line 206: This should be more strict than an if-condition. `apiKey` and `http` are the only allowed values, so you should first make sure the key exactly matches one of these. See: https://github.com/api3dao/api3-docs/blob/8e41ab2597cb19cb75e4904d6e9b214287c1d738/airnode/ois.md#421-type

packages/validator/templates/OAS2OIS.json
sassmilic: See: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#components-object

packages/validator/templates/apiSpecifications.json
```diff
@@ -12,7 +12,7 @@
     "__conditions": [
       {
         "__if": {
-          "__this": "(?<={)[^\\/{}]+(?=})"
+          "__this_name": "(?<={)[^\\/{}]+(?=})"
```
sassmilic, line 15: What's this checking for?

sassmilic:
General question: if an *optional* parameter is in the specs, the behaviour should be that its formatting **is** checked, correct? Re: https://github.com/api3dao/airnode/tree/master/packages/validator#optional-and-level

packages/validator/templates/configSecurity.json
```diff
@@ -0,0 +1,98 @@
+{
+  "config": {
+    "ois": {
+      "__arrayItem": {
+        "__oisSpecs": {}
+      }
+    },
+    "triggers": {
+      "__optional": {
```
sassmilic, line 9: Why are triggers optional? @bbenligiray

packages/validator/templates/configSecurity.json
```diff
@@ -0,0 +1,98 @@
...
+        "request": {
+          "__arrayItem": {
+            "endpointId": {
+              "__regexp": "^0x[a-fA-F0-9]{64}$"
```
sassmilic, line 13: I'm not 100% sure of this, but addresses are 42 characters (including `0x`). https://kb.myetherwallet.com/en/getting-started/glossary-of-common-terms/ @bbenligiray can you confirm?

packages/validator/templates/configSecurity.json
```diff
@@ -0,0 +1,98 @@
...
+            },
+            "oisTitle": {},
```
sassmilic, line 15: Is `oisTitle` supposed to map to some object in the `ois` array? If so, I think there should be a `__require` statement like this, no?

Hmm ... although I'm not sure the `__require` statement currently has this functionality: to check if there exists some condition in *any* element in an array.

Speaking of: @bbenligiray why is `ois` in `config.json` an array instead of an object with `oisTitle` as a key?

sassmilic: Because, if `triggers.oisTitle` does in fact map to some `ois[].title` (pls confirm), then that title should be easily index-able.

packages/validator/templates/configSecurity.json
```diff
@@ -0,0 +1,98 @@
...
+          }
+        }
+      }
+    },
+    "nodeSettings": {
+      "providerIdShort": {
+        "__regexp": "^[a-fA-F0-9]{7}$"
+      },
+      "nodeVersion": {
+        "__regexp": "^([0-9]*\\.){2}([0-9]*)$"
```
sassmilic, line 26: nit: this regex matches strings like `..1`.

@bbenligiray are we doing something like this: https://www.geeksforgeeks.org/introduction-semantic-versioning/

If so, I think it would be as simple as this: "^[0-9]+\\.[0-9]+\\.[0-9]+$"

assuming we don't have trailing identifiers, e.g. `1.0.0-alpha.1`

sassmilic:
Another general comment: since regexes are hard to debug, I think it would be a good idea to put them all in a separate file.

packages/validator/templates/configSecurity.json
```diff
@@ -0,0 +1,98 @@
...
+      },
+      "cloudProvider": {
+        "__regexp": "^aws$"
```
sassmilic, line 29: Only aws? @bbenligiray

packages/validator/templates/configSecurity.json
```diff
@@ -0,0 +1,98 @@
...
+      },
+      "region": {},
+      "stage": {},
+      "logFormat": {
+        "__regexp": "^(json|plain)$"
+      },
+      "chains": {
+        "__arrayItem": {
+          "id": {},
```
sassmilic, line 38: nit: Chain IDs are all integers, see: https://chainid.network/

so this can be checked for via a regex

packages/validator/templates/configSecurity.json
```diff
@@ -0,0 +1,98 @@
...
+                "blockHistoryLimit": {},
+                "minConfirmations": {}
+              }
+            }
+          },
+          "__optional": {
+            "providerAdminForRecordCreation": {
+              "__regexp": "^0x[a-fA-F0-9]{40}$"
+            },
+            "contracts": {
```
sassmilic, line 58: nit: `contracts` is a required parameter, so it should be above `_optional`.

packages/validator/templates/configSecurity.json
```diff
@@ -0,0 +1,98 @@
...
+              "__keyRegexp": "^(Airnode|Convenience)$",
```
sassmilic, line 59: According to the specs, it must include *both* `Airnode` and `Convenience`, so this should be checked for separately, see: https://github.com/api3dao/api3-docs/blob/8e41ab2597cb19cb75e4904d6e9b214287c1d738/airnode/config-json.md#nodesettingschains

packages/validator/templates/configSecurity.json
```diff
@@ -0,0 +1,98 @@
...
+            "security": {
+              "id": {
+                "__regexp": "^__match$",
+                "__level": "error"
+              }
+            }
+          }
+        }
+      ]
+    }
```
sassmilic, line 84: Since `id` is a required field in `config.json`, I think it would make more sense to use your `__require` condition. Would it be difficult to add a functionality that looks like this?
```
"__require": {
    "/security.id": __this
}
```

or something like that ...

packages/validator/templates/configSecurity.json
```diff
@@ -0,0 +1,98 @@
...
+  },
+  "security": {
+    "__any": {
```
sassmilic, line 87: I'm not sure how `__any` works without a `then` statement. I think this would imply the README should be updated to include an example of this functionality: https://github.com/api3dao/airnode/tree/master/packages/validator#any

packages/validator/templates/configSecurity.json
```diff
@@ -0,0 +1,98 @@
...
+                "__regexp": "^__match$",
+                "__level": "error"
+              }
+            }
+          }
+        }
+      ]
+    }
+  },
+  "security": {
```
sassmilic, line 86: There seems to be some discrepancy between this template and the `security.json` spec ... I think.
https://github.com/api3dao/api3-docs/blob/8e41ab2597cb19cb75e4904d6e9b214287c1d738/airnode/security-json.md#securityjson

It seems like `security.json` should have only two keys: `id` and `apiCredentials`.

I'm not sure what `"__keyRegexp": "^(?!id).*$"` refers to ... maybe some outdated spec?

sassmilic (changes requested):
Not done reviewing yet. I think it'll be a good idea for me to review all the templates first (which I've started doing already) in order that we all agree on the format/specs/desired output (I tagged Burak a few times) in case it changes any of the actual parsing/implementation code. Then I'll review the core implementation.

packages/validator/templates/OAS2OIS.json
bbenligiray: I agree with using  https://tools.ietf.org/html/rfc3986#appendix-B omitting the part starting from query.
Let's only support absolute URLs
https://github.com/api3dao/api3-docs/commit/75ebe672addd1c6e4cc2e68bf47f06515161c0ed

bbenligiray:
> Question: what version of OAS are we using/referring to?

While coming up with OIS 1.0.0, the latest OAS version was 3.0.3 so I took that as reference (though IIRC 3.1.0 is not significantly different). The convertor template should specify that (oas 3.0.3 to ois 1.0.0).

packages/validator/templates/OAS2OIS.json
bbenligiray: I think I told Jakub to do this due to practical reasons. Adding all three reserved parameters means that the requester will be able to specify these, so it's similarly non-opinionated as not adding them at all. However, the requester will want to use these parameters in practice, so adding them here makes more sense.

We require the provider to explicitly add the reserved parameters for them to be enabled because we will have parameters in the future that the provider may want to disable (for example specifying off-chain computation).

Btw `reservedParameters` is a required field so would have to exist even if as an empty array.

packages/validator/templates/OAS2OIS.json
bbenligiray: It's a required field so added as an empty array.

packages/validator/templates/OAS2OIS.json
bbenligiray: Good catch, no need for `schemas` there

packages/validator/templates/OAS2OIS.json
bbenligiray: I was going to say that https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml starts with uppercase so we did the same but Andre actually used lower case https://github.com/api3dao/airnode/blob/116450562b7ee39134af78afaf09ad8af2478f5b/packages/ois/src/types.ts#L31

I'll update the docs, let's use lowercase

packages/validator/templates/configSecurity.json
bbenligiray: The whole format is pretty redundant so `triggers.request` should be required here for consistency.

packages/validator/templates/configSecurity.json
bbenligiray: `endpointId` is a `bytes32` type so it's 64 hexadecimals

packages/validator/templates/configSecurity.json
bbenligiray: Because `oisTitle` is already a field in the OIS object and using the title both as the key and a field of the object that is indexed by that key is weird. Same argument also applies for `triggers` for example, we should use `endpointId`s as keys but we just keep a list of objects and loop through them to find what we're looking for (but that's not any more difficult to implement and results in a neater format).

packages/validator/templates/configSecurity.json
bbenligiray: Actually we probably want to support trailing identifiers, especially for node versions customized for some other chain for example.

packages/validator/templates/configSecurity.json
bbenligiray: Actually aws and local

packages/validator/templates/configSecurity.json
bbenligiray: We decided to treat this as an arbitrary string recently to support non-EVM chains more easily in the future. Also chain IDs are also commonly represented in hexadecimals.

packages/validator/templates/configSecurity.json
bbenligiray: Yup

packages/validator/templates/configSecurity.json
bbenligiray: Also entries here need to be matched to security schemes of OISes in `config`. We will get rid of `security.json`, but it will be replaced by a similar field in `config` so the work here can be reused.

packages/validator/templates/configSecurity.json
sassmilic: You could create a regex from the grammar from the original specs: https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions

Although it looks like a bit of pain in the ass.

sassmilic: There's a regex in the official specs for semantic versioning: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

sassmilic:
> but templates being standalone files is also nice

What's your reasoning? I don't see the downside of replacing raw regexes with something like `{SERVER_URL_PATTERN}` or smth.

packages/validator/templates/configSecurity.json
bbenligiray: Looks good (I guess we would use the second one)

bbenligiray:
@sassmilic I'm imagining a folder structure like this
```
templates/
├─ ois/
│  ├─ ois-1.0.0.json
│  ├─ ois-1.0.1.json
├─ config/
   ├─ config-0.1.1.json
   ├─ config-0.1.0.json
```
where we would add a new file to `templates/ois` with each OIS version and a new file to `templates/config` with each node version. Each of these files being completely independent from each other neat.

packages/validator/templates/OAS2OIS.json
Drgy: Since this should be updated in all templates I moved it into [appropriate issue](https://github.com/api3dao/airnode/issues/197)

packages/validator/templates/OAS2OIS.json
Drgy: Thank you for this, yes regular expressions in js work exactly as you say, however this is another case of validator's insufficient documentation. All `\` must be doubled in regular expressions, the reason for this is one `\` gets swallowed by template conversion from string to JSON object, so in this particular case `[^\\\\]` after template conversion into JSON object looks like this: `[^\\]`, which is the regular expression that we want. I added this into the [readme issue](https://github.com/api3dao/airnode/issues/203)

sassmilic:
> @sassmilic I'm imagining a folder structure like this
> 
> ```
> templates/
> ├─ ois/
> │  ├─ ois-1.0.0.json
> │  ├─ ois-1.0.1.json
> ├─ config/
>    ├─ config-0.1.1.json
>    ├─ config-0.1.0.json
> ```
> 
> where we would add a new file to `templates/ois` with each OIS version and a new file to `templates/config` with each node version. Each of these files being completely independent from each other neat.

@bbenligiray this makes sense, json doesn't really deal with variables natively. YAML is much better at this.

packages/validator/src/convertor.ts
```diff
@@ -0,0 +1,98 @@
...
+  if (!specsPath || !templatePath) {
+    return { valid: false, messages: [logger.error('Specification and template file must be provided')], output: {} };
+  }
+
+  let template, specs;
+
+  try {
+    template = fs.readFileSync(templatePath);
+  } catch (e) {
+    return { valid: false, messages: [logger.error(`Unable to read file ${templatePath}`)], output: {} };
```
sassmilic, line 29: nit: is there a reason you don't just propagate the error message `e`?

packages/validator/templates/OIS2C&S.json
```diff
@@ -0,0 +1,136 @@
+{
+	"__actions": [
+		{
+			"__insert": {
+				"__target": "config",
+				"__value": {
+					"id": "{FILL}",
```
sassmilic, line 7: I'm not sure what `{FILL}` is ... is this a placeholder for you? i.e. like a `TODO`?

packages/validator/src/validators/regexpValidator.ts
```diff
@@ -2,19 +2,19 @@ import * as logger from '../utils/logger';
...
-export function validateRegexp(specs: any, specsStruct: any, paramPath: string, isKeyRegexp = false): Log[] {
+export function validateRegexp(specs: any, template: any, paramPath: string, isKeyRegexp = false): Log[] {
   const messages: Log[] = [];
 
   if (isKeyRegexp) {
     for (const item of Object.keys(specs)) {
-      if (!item.match(new RegExp(specsStruct['__keyRegexp']))) {
+      if (!item.match(new RegExp(template['__keyRegexp']))) {
         messages.push(
           logger.error(`Key ${item} in ${paramPath}${paramPath ? '.' : ''}${item} is formatted incorrectly`)
```
sassmilic, line 19: nit: I'd change the error message to "... doesn't match regex" instead of "is formatted incorrectly."

packages/validator/src/validators/regexpValidator.ts
```diff
@@ -2,19 +2,19 @@ import * as logger from '../utils/logger';
...
+ * @param template - template containing the regular expression
  * @param paramPath - string of parameters separated by ".", representing path to current specs location
  * @param isKeyRegexp - determines if key or value should be checked
  * @returns incorrect formatting message if the parameter key or value did not match regular expression
  */
-export function validateRegexp(specs: any, specsStruct: any, paramPath: string, isKeyRegexp = false): Log[] {
+export function validateRegexp(specs: any, template: any, paramPath: string, isKeyRegexp = false): Log[] {
   const messages: Log[] = [];
 
   if (isKeyRegexp) {
```
sassmilic, line 15: It seems like you're missing the case for when `isKeyRegexp` is false?

sassmilic: Actually, on second look, it appears that when `isKeyRegexp` is false, none of the if-statements are executed, only the final `return messages;` which should be empty, correct? 

If so, could you make the code more explicit in that regard? E.g. just changing the structure to if-else I think would make the code easier to follow.

packages/validator/src/convertor.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  try {
+    specs = fs.readFileSync(specsPath);
+  } catch (e) {
+    return { valid: false, messages: [logger.error(`Unable to read file ${specsPath}`)], output: {} };
+  }
+
+  return convertJson(specs, template);
```
sassmilic, line 38: nit: since the function is called `convertJson`, I think it would make more sense to convert `specs` and `template` to json before passing them to this function.

packages/validator/src/convertor.ts
```diff
@@ -0,0 +1,98 @@
...
+    const parsedSpecs = JSON.parse(specs);
+
+    const result = processSpecs(parsedSpecs, parsedTemplate, '', nonRedundant, {
+      specs: parsedSpecs,
+      nonRedundantParams: nonRedundant,
+      output,
+    });
+
+    return { valid: result.valid, messages: result.messages, output: JSON.parse(JSON.stringify(output)) };
+  } catch (e) {
```
sassmilic, line 61: The only potential errors would be from `JSON.parse`, since you catch all errors in `processSpecs` and put them in `messages`, correct? If so, I think the catch statement should be closer to the code that might produce the error. Better yet, put `JSON.parse` and the `try-catch` in the `convert` function, like mentioned in my comment above.

packages/validator/src/convertor.ts
```diff
@@ -0,0 +1,98 @@
...
+        return { valid: ois.valid, messages: ois.messages, output: cs.output };
+      default:
+        return { valid: false, messages: [invalidConversionMessage(from, to)] };
+    }
+  } else if (from === 'ois' && (to === 'cs' || to === 'configsecurity' || to === 'configandsecurity')) {
+    return convert(specs, ois2cs);
+  }
+
+  return { valid: false, messages: [invalidConversionMessage(from, to)] };
+}
```
sassmilic, line 98: I find the structure of this kind of awkward and hard to follow. 

Correct me if I'm wrong, but these are all possible cases (using my own informal notation):

```
oas -> ois
oas -> cs|configsecurity|configandsecurity
ois -> cs|configsecurity|configandsecurity
```
otherwise, return `invalidConversionMessage`.

If so, I would make these possible options more explicit in the structure of this code, i.e.:

```
if (from === 'oas' && to === 'ois') {
    ...
} else if (from === 'oas' && (to === 'cs' || to === 'configsecurity' || to === 'configandsecurity')) {
    ...
} else if (from === 'ois' && (to === 'cs' || to === 'configsecurity' || to === 'configandsecurity')) {
} else {
    return { valid: false, messages: [invalidConversionMessage(from, to)] 
}
```

packages/validator/src/processor.ts
```diff
@@ -0,0 +1,204 @@
...
+
+const apiTemplate = JSON.parse(fs.readFileSync('templates/apiSpecifications.json', 'utf8'));
+const oisTemplate = JSON.parse(fs.readFileSync('templates/ois.json', 'utf8'));
+const endpointsTemplate = JSON.parse(fs.readFileSync('templates/endpoints.json', 'utf8'));
+
+/**
+ * Recursion validating provided specification against template
+ * @param specs - specification that is being validated
+ * @param template - template the specification is validated against
+ * @param paramPath - string of parameters separated by ".", representing path to current specs location (empty string in root)
```
sassmilic, line 19: nit: spelling error: "empty string _is_ root"

packages/validator/src/processor.ts
```diff
@@ -0,0 +1,204 @@
...
+        break;
+
+      // determines level of message for single parameter, currently used only in regexp validation
+      case '__level':
+        break;
+
+      case '__any':
+        if (!isAnyParamValid(specs, template[key], paramPath, nonRedundantParams, roots)) {
+          messages.push(logger.error(`Required conditions not met in ${paramPath}`));
+        }
```
sassmilic, line 126: For consistency, since all of your other validation subroutines return a message, I think this one should too; i.e. change the name to `validateAny` and push the output of that function to `messages`, so that all of your `validate...` subroutines have similar inputs/outputs.

sassmilic:
General question: can you have multiple `__optional` keys? This example makes it appear as though you can only have one optional key per level: https://github.com/api3dao/airnode/tree/master/packages/validator#optional-and-level

packages/validator/src/processor.ts
```diff
@@ -0,0 +1,204 @@
...
+const apiTemplate = JSON.parse(fs.readFileSync('templates/apiSpecifications.json', 'utf8'));
+const oisTemplate = JSON.parse(fs.readFileSync('templates/ois.json', 'utf8'));
+const endpointsTemplate = JSON.parse(fs.readFileSync('templates/endpoints.json', 'utf8'));
+
+/**
+ * Recursion validating provided specification against template
+ * @param specs - specification that is being validated
+ * @param template - template the specification is validated against
+ * @param paramPath - string of parameters separated by ".", representing path to current specs location (empty string in root)
+ * @param nonRedundantParams - object containing all required and optional parameters that are being used
```
sassmilic, line 20: I'm having trouble understanding what `nonRedundantParams` is used for. My best guess: this is used to keep track of extra fields (and nothing else), correct?

If so, I think it adds too much complexity to pass around `nonRedundantParams` between practically all subroutines.

What I would do to check for extra fields (and missing params) is: _before_ the for-loop in `processSpecs` do something like this:
```
let specKeys = Object.keys(specs);
// filter out "special" keys from template; i.e. those starting with "__"
let templateKeys = Object.keys(template).filter(key => !key.startsWith('__'));
 
// convert Arrays to Sets for easier set operations
specKeys = new Set(specKeys);
templateKeys = new Set(templateKeys);
 
// check for missing params
let keysNotInSpec = new Set([...templateKeys].filter(x => !specKeys.has(x)));
if (keysNotInSpec.size > 0) {
    messages.push(...) // push missing param error message
}
 
// check for extra fields
if (!Object.keys(template).includes("__ignore")) {
    let keysNotinTemplate = new Set([...specKeys].filter(x => !TemplateKeys.has(x)));
    if (keysNotInTemplate.size > 0) {
        messages.push(...) // push extra fields error message
    }
}
```

This way, you check for extra fields and missing specs in a more explicit and isolated way. (Apologies for any potential syntax errors; I rarely use js/ts.)

packages/validator/src/commands/convertCmd.ts
```diff
@@ -0,0 +1,13 @@
...
+  console.log(convert(process.env.npm_config_specs, process.env.npm_config_template));
+} else if (process.env.npm_config_from && process.env.npm_config_to && process.env.npm_config_specs) {
+  console.log(convertFromTo(process.env.npm_config_from, process.env.npm_config_to, process.env.npm_config_specs));
+} else {
+  if (process.argv.length === 5) {
+    console.log(convertFromTo(process.argv[2], process.argv[3], process.argv[4]));
+  } else {
+    console.log(convert(process.argv[3], process.argv[2]));
+  }
+}
```
sassmilic, line 13: Please add some comments here; it's really hard to understand what's happening with the `argv` indexing.

packages/validator/src/commands/validateCmd.ts
```diff
@@ -0,0 +1,7 @@
+import { validate } from '../validator';
+
+if (process.env.npm_config_specs || process.env.npm_config_template) {
+  console.log(validate(process.env.npm_config_specs, process.env.npm_config_template));
+} else {
+  console.log(validate(process.argv[3], process.argv[2]));
```
sassmilic, line 6: Please add some comments explaining what the indices into `argv` refer to.

sassmilic:
Okay, done my initial review (i.e. actually reviewed all files). @Drgy let me know if you have any questions/concerns with any of my comments. I'll take another look once changes are made.

packages/validator/src/processor.ts
sassmilic: Also: if there are any missing params and/or extra fields, I think it's safe to just return at this point with an error message.

bbenligiray:
I think a decent middle ground is documenting all regex entries in a Markdown file. Someone can use a regex in a template without documenting it, but so can they hardcode regex into templates instead of pulling it from the regex store.

sassmilic:
> I think a decent middle ground is documenting all regex entries in a Markdown file. Someone can use a regex in a template without documenting it, but so can they hardcode regex into templates instead of pulling it from the regex store.

I believe Jakub is already putting regexes in the README, but it sounds like you're suggesting an entirely new file.

packages/validator/templates/OAS2OIS.json
Drgy: Replaced in 1b7124c

packages/validator/templates/OAS2OIS.json
Drgy: @bbenligiray should this be lowercase only in OAS or does it apply to all specs, like [in API specs](https://github.com/api3dao/airnode/blob/1b7124caa2d44b555d360f5f152f5260663b8821/packages/validator/templates/apiSpecifications.json#L67) for example?

packages/validator/templates/OAS2OIS.json
Drgy: Moved in 8e97859

packages/validator/templates/OAS2OIS.json
Drgy: Moved in 8e97859

packages/validator/templates/OAS2OIS.json
Drgy: Not sure if im missing something, but i believe [this](https://github.com/api3dao/airnode/blob/8e97859e8b919e271101f137e24aaba0fa7c8108/packages/validator/templates/OAS2OIS.json#L196) should be already checking it

packages/validator/templates/apiSpecifications.json
Drgy: There can be parameters in the path, something like this for example `/path/{param1}/{param2}`, in this case `param1` and `param2` must be present in `parameters` array of the path. The regular expression is a bit more complicated because it uses lookahead and lookbehind to check for `{}`, so that `__match` in the condition gets replaced for `param1` and not `{param1}`.

packages/validator/templates/configSecurity.json
Drgy: Added to [template issue](https://github.com/api3dao/airnode/issues/197)

packages/validator/templates/configSecurity.json
Drgy: I assume this should be replaced across all the templates -> added to [template issue](https://github.com/api3dao/airnode/issues/197)

packages/validator/templates/configSecurity.json
Drgy: Added in 9b58195

packages/validator/templates/configSecurity.json
Drgy: Moved in c052d91

packages/validator/templates/configSecurity.json
Drgy: Updated in 1d40c07

packages/validator/templates/OAS2OIS.json
sassmilic: but after template conversion into JSON object, wouldn't you want it to look like this: `[^\]` i.e. "not backslash", since everything in `[^...]` is treated as a single character, no?

Drgy: Yes sorry, I should have continued with the explanation. The process has one more step, after conversion from string into JSON object, it would look something like this: `{ __regexp: "[^\\]" }`, you can notice that the value is still not js regular expression, but just a normal string. So one more `\` is consumed by creation of js regular expression from string.

packages/validator/templates/configSecurity.json
Drgy: It most certainly wouldn't be something difficult, however im deciding if I should remove `__require` altogether. During the development of validator `__if` conditions became quite powerful and can also do everything that `__require` can do, which made `__require` redundant. Currently `__require` is useful just because of it's compact syntax and has way too many downsides. Few of them would be:

- Its another keyword with its own rules that needs to be documented and remembered by whoever reads/writes templates
- If new feature is implemented into `__if`, for consistency sake it must be implemented, documented, tested in `__require` as well
- Conditions are generally buggy spots in validator, I fixed multiple nightmare bugs in them, this adds more potential buggy places

There are some more minor reasons, but you get the idea, so for now I would recommend not to use `__require` and after the issues in templates are resolved, i will analyze how much could `__require` potentially help and decide to keep/remove it.

packages/validator/templates/configSecurity.json
Drgy: Added to https://github.com/api3dao/airnode/issues/203
But in short it works exactly the same as in `__then`, if at least one nested object matches the structure specified by `__any`, it is valid specification, otherwise it will print an error.

packages/validator/templates/configSecurity.json
Drgy: Yes, this seems like some outdated part of template, what it means is basically there must be something that doesn't have `id` as a key and have the structure specified below, which doesn't match with current structure of `security.json`.
I added this to #197, matching with security schemes of OISes is already planned in #197.

packages/validator/src/convertor.ts
Drgy: I dont really know how will be validator used/invoked, I was focusing on scenario in which a nice web UI printing all standard validator messages is made, user selects corrupted file, validator will return message in standard format, UI will display this as any other message, instead of printing 'Something went wrong' or some stacktrace as it didnt receive response in standard format, also no additional work will be needed on frontend.

packages/validator/templates/OIS2C&S.json
Drgy: Some values are not present in OAS, and need to be filled manually, Burak suggested to use `{FILL}` for these values.

packages/validator/src/validators/regexpValidator.ts
Drgy: If I understood correctly how validator will be used, the end user who is checking his specification json, doesn't have to necessarily know anything about some regex or even a validator template. So I don't think mentioning regex is a good idea, but I certainly agree in many cases the messages are way too general and have no real informative value for the user. I already opened #215, in which the idea is to implement the ability of telling user what exactly is wrong and to what it should be changed.

packages/validator/src/validators/regexpValidator.ts
Drgy: Yes you are right, I use this style based on [Andre's recommendation](https://github.com/api3dao/airnode/pull/69#discussion_r495812979), so this is on many places, im not sure if I want to refactor this, but if you would think on some places it causes serious problems, let me know.

packages/validator/src/convertor.ts
Drgy: That really makes much more sense, updated in 77fdfd2
Just a side note about why to call `JSON.parse` on variables in `readmeExamples.test.ts`, instead of directly making the variables object instead of string. This is so they match regular expression from `README.md` one to one (the `\` problem)

packages/validator/src/convertor.ts
Drgy: I think i managed to sneak this in 77fdfd2, if thats what you mean

packages/validator/src/convertor.ts
Drgy: Your possible cases seem correct, there is a table in the readme describing this:
| From | To |
| ----- | -----|
| oas | ois |
| oas | cs / ConfigSecurity / ConfigAndSecurity |
| ois | cs / ConfigSecurity / ConfigAndSecurity |
Simplified in 7e8d08e

packages/validator/src/processor.ts
Drgy: Fixed in 2b2777b

packages/validator/src/processor.ts
Drgy: Changed in 87d7be6

Drgy:
> General question: can you have multiple `__optional` keys? This example makes it appear as though you can only have one optional key per level: https://github.com/api3dao/airnode/tree/master/packages/validator#optional-and-level
> 
> If so, it would make more sense to me to have '__optional' be an array; e.g.:
> 
> ```
> "__optional": [
>     "optionalKey1": {},
>     "optionalKey2": {},
>     "optionalKey3": {
>         "requiredKey1": {},
>         "requiredKey2": {},
>     }
> ]
> ```

@sassmilic Yes there can be multiple objects in `__optional` (added to #203), I imagined `__optional` being more similar to `__any`, rather than to `__actions` or `__conditions`. So even though the syntax of these keywords is different, there are just 2 standards, which I believe are not hard to remember, but mainly what you showcased is not valid JSON object in js it would have to look like this:
```
"__optional": [
    { "optionalKey1": {} },
    { "optionalKey2": {} },
    { "optionalKey3": {
        "requiredKey1": {},
        "requiredKey2": {},
    } }
]
```
which im not really a fan of.

packages/validator/src/processor.ts
Drgy: Yes this seems simple enough, however it's not going to cover way too many cases. Im aware that passing around "useless" variable is less than ideal, but it truly is the best solution that i came up with. Back to your solution, if i understand it correctly there are 2 fairly big issues right on top of my head (its going to be probably more though). First one should be what about `__keyRegexp`? Some parameters are not extra even though they are not explicitly given in template. Another one would be `__conditions`. Parameters that are not in template but used in valid `__condition` are also not extra, keep in mind `__rootThen` is evaluated all the way from root, which complicates things even more.
Im opened to any suggestions about this, as im not happy with my solution that much, but i dont really see any easy way out of this.

packages/validator/src/commands/convertCmd.ts
Drgy: Added in fb01cd9

packages/validator/src/commands/validateCmd.ts
Drgy: Added in fb01cd9

Drgy:
> I believe Jakub is already putting regexes in the README, but it sounds like you're suggesting an entirely new file.

There are few regexes in readme, that are frequently used when writing templates, I will add one or two more in the new readme issue. It might seem that there are so many regexes in the templates, but they all are just some combination of regexes from the readme. They might differ in a character or two, that are allowed/forbidden in the specific parameter, but thats pretty much it, thats one reason why i think creating extra file just for regexes wouldnt be worth it. Another reason is which version of specification will the regexes be taken from? If it will be just latest, then there is a good chance whoever wanted to find specific regex will have to go to template of the specific version anyways. If it will be all versions, i think finding specific regex in there would be as difficult as finding it in its template. In these scenarios im assuming that the [file structure](https://github.com/api3dao/airnode/pull/195#issuecomment-782890567) that Burak mentioned above is already in place.

sassmilic:
> Another reason is which version of specification will the regexes be taken from? If it will be just latest, then there is a good chance whoever wanted to find specific regex will have to go to template of the specific version anyways. If it will be all versions, i think finding specific regex in there would be as difficult as finding it in its template.

Yea, this is a good point. We can just treat these templates as "definitive" since I don't imagine new versions being that common of an occurrence. In which case, the regexes in the templates are treated as the "definitive regexes", like I think you're implying. That is, I think I agree that just keeping them in the templates is probably fine.

Although I think it would be nice to potentially have commenting enables in the templates. I know that json files don't natively support comments, and it's definitely not urgent, just something to consider.

packages/validator/src/validators/regexpValidator.ts
sassmilic: Ahh, I see. Thanks.

packages/validator/src/validators/regexpValidator.ts
sassmilic: Haha, yea I see me and Andre have opposite opinions on the issue. Yea, I'd leave it as-is.

packages/validator/src/processor.ts
sassmilic: Yea, you're right. I didn't think about `__rootThen`; that really complicates things since I thought you could do missing/extra param check at each "level", but `__rootThen` wouldn't allow that. Okay, leave it as-is.

sassmilic (approved):
Looks really good! I responded to some of your comments, but I think all "issues" are "resolved". Great work :)

packages/validator/templates/OAS2OIS.json
bbenligiray: It should be lowercase everywhere

Drgy: added to #197

PR #198, opened 2021-02-07
Provider authorizers

bbenligiray:
- Moved authorizers from endpoints to providers ~~(shouldn't affect the node)~~
- Removed `updateProvider()`. The user can only update the provider admin through their `config.json`
- Removed `fulfill()`, renamed `fulfillWithBytes()` as `fulfill()`

Once this branch is fixed, I'll branch from it to combine `Airnode.sol` and `Convenience.sol`


bbenligiray:
@andreogle The node needs to be updated to fulfill requests with
```js
ethers.utils.defaultAbiCoder.encode([_type], [typecastedValueFromApi])
```
instead of `typecastedValueFromApi` directly (though I suspect the result is exactly the same)

The event `ClientRequestFulfilledWithBytes` is removed so the node no longer needs to expect it.

Also a test is broken and I'm not sure why.

packages/protocol/hardhat.config.js
```diff
@@ -1,6 +1,7 @@
 require('@nomiclabs/hardhat-waffle');
 require('solidity-coverage');
 require('hardhat-deploy');
+require('hardhat-gas-reporter');
```
andreogle, line 4: 👍🏻

packages/protocol/contracts/interfaces/IProviderStore.sol
```diff
@@ -35,18 +31,13 @@ interface IProviderStore is IRequesterStore {
...
         address admin,
-        string calldata xpub
+        string calldata xpub,
+        address[] calldata authorizers
         )
         external
         payable
         returns (bytes32 providerId);
 
-    function updateProvider(
```
andreogle, line 44: Not sure I understand this. Does this mean that providers can't update their list of authorizers?

packages/protocol/contracts/ProviderStore.sol
```diff
@@ -154,27 +141,20 @@ contract ProviderStore is RequesterStore, IProviderStore {
     /// @param providerId Provider ID
     /// @return admin Provider admin
     /// @return xpub Master public key of the provider node
+    /// @return authorizers Authorizer contract addresses
     function getProvider(bytes32 providerId)
```
andreogle, line 145: I see this doesn't have unit tests. Worth testing too?

andreogle (approved):

andreogle:
Just realised that there are some broken functions in the operation package. I'm working on it now

packages/protocol/contracts/interfaces/IProviderStore.sol
bbenligiray: The provider can't update the provider admin or the authorizers. Those two can only be updated by updating `config.json`.

packages/protocol/contracts/Airnode.sol
```diff
@@ -2,18 +2,18 @@
...
 /// @title The contract used to make and fulfill requests
 /// @notice Clients use this contract to make requests that follow a
 /// request-response scheme. In addition, it inherits from contracts that keep
 /// records of providers, requesters, endpoints, etc.
-contract Airnode is EndpointStore, TemplateStore, IAirnode {
+contract Airnode is ProviderStore, TemplateStore, IAirnode {
     mapping(bytes32 => bytes32) private requestIdToFulfillmentParameters;
     mapping(bytes32 => bool) public requestWithIdHasFailed;
-    uint256 private noRequests = 0;
+    uint256 private noRequests = 1;
```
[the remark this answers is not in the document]
bbenligiray: Incrementing 0 costs more gas than incrementing 1. Starting the counter from 0 resulted in the first request costing more gas than the rest, which is annoying for tests.

packages/protocol/contracts/ProviderStore.sol
bbenligiray: The unit tests are not complete in general, but okay leaving this as a note

bbenligiray:
> Moved authorizers from endpoints to providers (shouldn't affect the node)

I was wrong actually, this affects how the node creates a provider and the `config.json` format. Authorizers should now be an array of contracts in `chains` similar to the `providerAdmin`

packages/protocol/contracts/Airnode.sol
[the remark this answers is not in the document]
bbenligiray: This will be irrelevant with client-specific nonces anyway

packages/protocol/contracts/interfaces/IProviderStore.sol
bbenligiray: The idea is that `providerAdmin` is not for updating the `authorizers` of a node, but only for interacting authorizer contracts (i.e. the ChainAPI one)

packages/node/src/evm/initialization.test.ts
```diff
@@ -16,26 +16,67 @@ jest.mock('ethers', () => ({
...
+
+  const validData = {
+    authorizers: [ethers.constants.AddressZero],
+    blockNumber: 12,
+    providerAdmin: '0x5e0051B74bb4006480A1b548af9F1F0e0954F410',
+    xpub:
+      'xpub661MyMwAqRbcGeCE1g3KTUVGZsFDE3jMNinRPGCQGQsAp1nwinB9Pi16ihKPJw7qtaaTFuBHbRPeSc6w3AcMjxiHkAPfyp1hqQRbthv4Ryx',
+  };
+
+  it('is false if the provider onchain data matches the expected data', () => {
```
bbenligiray, line 36:
```suggestion
  it('is true if the provider onchain data matches the expected data', () => {
```

bbenligiray (commented):
What do you think about this: Here
https://github.com/api3dao/airnode/blob/03fc4b8dddacd9802b17656e88e94cd9ad7dc184/packages/node/src/evm/initialization.ts#L138
if `txCost > masterWalletBalance` and on-chain `xpub` matches the node's, skip the provider creation tx (log a warning) and continue serving requests. I know we already talked about this, but I think we can be almost 100% sure that if the user has changed their `providerAdmin`/`authorizers` on `config.json` without funding their master wallet, they didn't do this to stop their node serving requests.

I think `providerAdmin` being optional no longer makes sense (or `authorizers` should also be optional, probably not).

andreogle:
Sure so if that was to happen (the node can't create the new provider record and a warning gets logged), then any changes to `authorizers` wouldn't come into effect? They would need to fund the master wallet again for the new `authorizers` to kick in and the warning to go away.

I'd prefer to make `providerAdmin` required. I think it's confusing having it optional

bbenligiray:
> Sure so if that was to happen (the node can't create the new provider record and a warning gets logged), then any changes to authorizers wouldn't come into effect? They would need to fund the master wallet again for the new authorizers to kick in and the warning to go away.

Yeah, and the same thing for any changes in `providerAdmin`

> I'd prefer to make providerAdmin required. I think it's confusing having it optional

:+1:  I think there was an argument for making it optional with how it previously worked but now it makes sense for it to be required.

andreogle:
I'll work on implementing these changes

packages/node/src/evm/initialization.ts
```diff
@@ -1,4 +1,5 @@
 import { ethers } from 'ethers';
```
andreogle, line 1: This file is a bit of a mess now. I think the larger functions could be broken up into smaller functions

bbenligiray:
I thought the provider update implementation was going to be a lot messier, I like this :+1:

packages/node/src/evm/initialization.ts
bbenligiray: Feel free to merge now or after you do this

PR #202, opened 2021-02-17
Adding explanation to last example in README.md

sassmilic:
I had a bit of confusion with the last example, so this explanation is meant to clarify my understanding as well as be something I can reference in case I forget :P

### DIFF
--- a/packages/validator/README.md
+++ b/packages/validator/README.md
@@ -805,7 +805,7 @@ Keyword `__insert` works similarly to `__copy`, except it doesn't copy a value o
 
 ### Target path
 
-`__target` in actions specifies path to parameter that will be modified, there can be an array in this path, to deal with this path can contain `[]` which will push back new item into the array and use it in the target, or path can use `[_]` which points to last item in the array. Parameter from path where the action is located can be accessed via `{{X}}`, where X is position of the parameter in the path numbered from 0. Target path can also include keyword `__all`, in which the action will be performed on every child of previous element.
+`__target` in actions specifies path to parameter that will be modified, there can be an array in this path, to deal with this path can contain `[]` which will push back new item into the array and use it in the target, or path can use `[_]` which points to last item in the object. Parameter from path where the action is located can be accessed via `{{X}}`, where X is position of the parameter in the path numbered from 0. Target path can also include keyword `__all`, in which the action will be performed on every child of previous element.
 
 #### Template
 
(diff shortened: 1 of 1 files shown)


packages/validator/README.md
```diff
@@ -805,7 +805,7 @@ Keyword `__insert` works similarly to `__copy`, except it doesn't copy a value o
 
 ### Target path
 
-`__target` in actions specifies path to parameter that will be modified, there can be an array in this path, to deal with this path can contain `[]` which will push back new item into the array and use it in the target, or path can use `[_]` which points to last item in the array. Parameter from path where the action is located can be accessed via `{{X}}`, where X is position of the parameter in the path numbered from 0. Target path can also include keyword `__all`, in which the action will be performed on every child of previous element.
+`__target` in actions specifies path to parameter that will be modified, there can be an array in this path, to deal with this path can contain `[]` which will push back new item into the array and use it in the target, or path can use `[_]` which points to last item in the object. Parameter from path where the action is located can be accessed via `{{X}}`, where X is position of the parameter in the path numbered from 0. Target path can also include keyword `__all`, in which the action will be performed on every child of previous element.
```
Drgy, line 808: I'm not sure about this change, because in this case it is literally referring to last object in array. So if in the specification would be some thing like `[ { param1: "value" }, { param2: "value" } ]`, then `[_]` would refer to `{ param2: "value" }`, which is last item in the array not object.

Drgy (changes requested):
Thanks for nice addition, it is a nice start that I can expand on in [here](https://github.com/api3dao/airnode/issues/203)

packages/validator/README.md
sassmilic: So `[_]` can refer to both the last object in an array or an object?

Drgy: It should be array only, is there any inconsistency somewhere, that implies it can be also an object?

sassmilic: nooo ... now i'm back at my original confusion from yesterday: if `[_]` refers/points to the last index in an array, why does this action in the last example:
```
"__insert": {
	"__target": "array[_].parameter",
	"__value": {}
}
```
result in `parameter` being inserted into *every* object in `array` instead of just the last object (i.e. the object containing `param2`)?

sassmilic:
updated

Drgy (approved):
If this explanation helps you @sassmilic we can merge it, but it will be soon heavily edited by [readme issue](https://github.com/api3dao/airnode/issues/203). I see few inaccuracies in your explanation namely:

> values get "copied" back to their original positions

suggests that this copy action is useless and the values will be in the same place, so the result would be the same without performing this action. Which is not true, since convertor creates the specification from ground up, so whatever is not copied or inserted will be not there.

The explanation of `[_]` is also not sufficient imho, i think we should go into a bit more detail on why `parameter` gets inserted into each item in array, not just the last one.

So yeah, if it helps you we can keep it for now, but it will be certainly rewritten in above mentioned issue.

bbenligiray:
Either merge or close imo

PR #220, opened 2021-02-23
Remove short requests

bbenligiray:
Removed short requests due to not being gas-efficient based on earlier tests.

I will open another PR to remove `requesterIndex`, `designatedWallet`, `fulfillAddress`, `fulfillFunctionId` from templates because regular requests use the ones provided at request-time. I'm opening this separately to avoid a larger PR.

@andreogle I realized that I didn't have to come up with new transaction hashes, etc. only replace the short ones with regular/full ones, so I didn't have to run the operations package.


packages/deployer/yarn.lock
```diff
@@ -1237,11 +1237,6 @@
   resolved "https://registry.yarnpkg.com/@types/node/-/node-13.13.32.tgz#f0edd0fb57b3c9f6e64a0b3ddb1e0f729b6f71ce"
   integrity sha512-sPBvDnrwZE1uePhkCEyI/qQlgZM5kePPAhHIFDWNsOrWBFRBOk3LKJYmVCLeLZlL9Ub/FzMJb31OTWCg2F+06g==
 
-"@types/node@^14.14.22":
```
bbenligiray, line 1240: Probably was remaining from Arrowana's PR

andreogle: Cool. I wonder if it's worth us upgrading to Node.js v14 at some point? It's the active LTS version at the moment, but v12 still has another year of support

andreogle (approved):
short-requests


south-park-nice-small

packages/deployer/yarn.lock
bbenligiray: Makes sense to me

Issue #216, opened 2021-02-20
Support _type of type uint256


bbenligiray:
`_type` only supports `int256`, `bool`, `bytes32`. No reason not to support `uint256`.

bbenligiray:
Implemented in https://github.com/api3dao/airnode/pull/227

Issue #214, opened 2021-02-19
Implement separate createProvider and createProviderAndRefund methods


bbenligiray:
Currently, `ProviderStore.col` has a `createProvider` method that creates a provider and refunds the provider admin. `createProvider` should only create a provider, and the aforementioned method should be renamed as `createProviderAndRefund`. The node will use the `createProviderAndRefund` method as before.

bbenligiray:
Implemented in https://github.com/api3dao/airnode/pull/227

PR #227, opened 2021-02-25
Finalize protocol

bbenligiray:
You may want to review commit by commit.

Implemented https://github.com/api3dao/airnode/issues/214, https://github.com/api3dao/airnode/issues/216
Updated airnode-admin according to the recent contract updates
Updated the endpoint ID convention from `hash("oisTitle/endpointName")` to `hash("oisTitle_endpointName")`
Updated contract docstrings, moved some methods around

I will update the contract tests next, then update the `config.json` format


packages/deployer/src/config-data/config.json.example
```diff
@@ -1,42 +1,26 @@
 {
   "id": "e08f5610-5772-4986-92fb-c8aef5aee6a0",
   "nodeSettings": {
-    "providerIdShort": "9e5a89d",
-    "nodeVersion": "0.1.0",
     "cloudProvider": "aws",
-    "stage": "dev",
-    "region": "us-east-1",
     "logFormat": "plain",
+    "nodeVersion": "1.0.0",
```
andreogle, line 6: It would nice to eventually have a single config.json example, otherwise they get out of date/sync. Not really sure how we could do that though

packages/protocol/contracts/RequesterStore.sol
```diff
@@ -69,9 +79,8 @@ contract RequesterStore is IRequesterStore {
         override
         onlyRequesterAdmin(requesterIndex)
     {
-        // Initialize the client nonce if it is being endorsed for the first
-        // time for consistent request gas cost
-        if (endorsementStatus && clientAddressToNoRequests[clientAddress] == 0)
+        // Initialize the client nonce for consistent request gas cost
+        if (clientAddressToNoRequests[clientAddress] == 0)
```
andreogle, line 83: Does the `endorsementStatus` matter here?

andreogle (approved):
Awesome. Excited to get this merged! 🎉

packages/deployer/src/config-data/config.json.example
bbenligiray: I still think that for the deployer we should pre-build with an arbitrary `config.json` and replace that at deployment-time with the user's validated `config.json`. I can use any `config.json` for that, it wouldn't matter. I'm waiting for the validator to mature and stop using the Serverless plugin for packing so there is still time.

packages/protocol/contracts/RequesterStore.sol
bbenligiray: Previously if the initial `endorsementStatus` was `false` and it was being updated as `false`, the nonce wasn't being initialized. I figured `endorsementStatus` is completely unrelated to the nonce initialization so it should be initialized anyway.

Here one view is that `updateClientEndorsementStatus` should revert if the `endorsementStatus` is the same with the current endorsement status (because it wouldn't be "updating" the status). Then this wouldn't be an issue (`endorsementStatus` would always be `true` for the first non-reverting call) but I'm not a fan of reverting that much (though maybe the function should be renamed as `setClientEndorsementStatus`).

bbenligiray:
I will also push the updated contract tests but it won't require a review imo.

packages/protocol/contracts/RequesterStore.sol
andreogle: > I figured endorsementStatus is completely unrelated to the nonce initialization so it should be initialized anyway.
This is what I thought and it seems better without the additional check. Just wanted to check

> maybe the function should be renamed as `setClientEndorsementStatus`
I like this name more

bbenligiray: What do you think about `setProvider` vs. `createProvider`, since that's the same thing

andreogle: I like `setProviderParameters` - seems pretty clear in what it's doing.

Calling `createProvider` multiple times implies that it creates multiple providers, when in fact it's updating/overwriting on subsequent calls

PR #237, opened 2021-03-14
info field is not valid in apiSpecifications

MrAngus666:
Discussed with @bbenligiray that info field is not valid in apiSpecifications


bbenligiray (approved):
Thanks, looks good

Issue #209, opened 2021-02-19
Unknown endpointIds should be ignored


andreogle:
## Description
If a request is submitted with an `endpointId` that is not in `triggers`, the request currently gets set to `Blocked`. This means subsequent requests cannot be actioned until the `BLOCK_COUNT_IGNORE_LIMIT` (20 blocks is reached). 

As Burak mentioned below, it is also worth reviewing all of the current blocking conditions

## Version
Master

## Present Behaviour
These requests become blocked

## Expected Behaviour
These requests should be ignored

bbenligiray:
Also a review of the blocking conditions would help. I feel like I remember that there was some other cases where the node couldn't find something in `config` and blocked the request.

bbenligiray:
Closing this, discuss the reviewing of the blocking conditions in https://github.com/api3dao/airnode/issues/296

PR #238, opened 2021-03-18
Verify unknown endpoints earlier

andreogle:
## Changes

1. Previously we verified that API calls had a valid `endpointId` and linked to a valid OIS **after** aggregation happens. This change moves this verification to immediately after templates are applied to individual requests. This is better as there we have all of the required information to do this after template application so there is no reason to delay it and potentially do unnecessary work
2. Previously we only checked that API calls contained an `endpointId` present in `triggers.request` and that they were linked to a known OIS. I've added an additional check to that the API call is linked to a known `endpoint` within the OIS 
3. A bit of logging cleanup

## Addresses

https://github.com/api3dao/airnode/issues/209

Thanks to Marco for the investigation and initial work!


packages/node/src/evm/verification/request-verification.ts
```diff
@@ -46,14 +41,78 @@ export function verifyDesignatedWallets<T>(
...
+
+    const trigger = triggers.find((t) => t.endpointId === apiCall.endpointId);
+    // If the request is for an unknown endpointId, the problem is with the requesting client contract
+    if (!trigger) {
+      const message = `Request:${apiCall.id} has no matching endpointId:${apiCall.endpointId} in Airnode config`;
+      const log = logger.pend('WARN', message);
+      const updatedApiCall = {
+        ...apiCall,
+        status: RequestStatus.Errored,
+        errorCode: RequestErrorCode.UnknownEndpoint,
```
bbenligiray, line 74: I think this should be `UnknownEndpointId`

packages/node/src/evm/verification/request-verification.ts
```diff
@@ -46,14 +41,78 @@ export function verifyDesignatedWallets<T>(
...
+    const ois = oises.find((o) => o.title === oisTitle);
+    // Although unlikely, it is possible that the Airnode instance was configured with a trigger
+    // with an 'oisTitle' that does not match an OIS in config.json. The responsibility to fix this
+    // lies with the API provider
+    if (!ois) {
+      const log = logger.pend('ERROR', `Unknown OIS:${oisTitle} received for Request:${apiCall.id}`);
+      const updatedApiCall = {
+        ...apiCall,
+        status: RequestStatus.Errored,
+        errorCode: RequestErrorCode.UnknownOIS,
```
bbenligiray, line 89: and these two can be `UnknownEndpoint`

packages/node/src/evm/verification/request-verification.ts
```diff
@@ -46,14 +41,78 @@ export function verifyDesignatedWallets<T>(
...
+    // The responsibility to fix this lies with the API provider
+    if (!endpoint) {
+      const log = logger.pend(
+        'ERROR',
+        `Unknown Endpoint:${endpointName} for OIS:${oisTitle} received for Request:${apiCall.id}`
+      );
+      const updatedApiCall = {
+        ...apiCall,
+        status: RequestStatus.Errored,
+        errorCode: RequestErrorCode.UnknownOIS,
```
bbenligiray, line 106: and these two can be `UnknownEndpoint`

packages/node/src/evm/verification/request-verification.ts
bbenligiray: This can also be `UnknownOIS` but I see the point of not wanting to have too many error codes

packages/node/src/handlers/start-coordinator.ts
```diff
@@ -46,30 +46,18 @@ export async function startCoordinator(config: Config) {
   }
 
   // =================================================================
-  // STEP 3: Group unique API calls and validate
+  // STEP 3: Group unique API calls (for the same chain & networkId)
```
bbenligiray, line 49: Does this mean that if we receive requests with the same ID from different chains, we make separate API calls for them and fulfill independently?

bbenligiray: The implementation looks to me like they would have been aggregated.

bbenligiray (approved):
Let me know if requests from different chains are being aggregated. If so, I think the best solution is to add the chain ID to the `requestId` hash instead of trying to solve it in the node.

packages/node/src/evm/verification/request-verification.ts
andreogle: Yeah agreed

packages/node/src/evm/verification/request-verification.ts
andreogle: I think a single error code should be fine to indicate to the API provider that something is wrong with their config and that they need to look into it

packages/node/src/handlers/start-coordinator.ts
andreogle: Yeah good point. The current implementation would group all requests from different chains together if they have the same request ID (so this code comment is wrong).

Would this ever happen in practice though?

bbenligiray: I think the real question is if it can be used to demonstrate a hypothetical replay attack and the answer is yes. We have two options:
- Don't aggregate requests with identical IDs coming from different chains on the node
- Include the chain ID in the request ID generation like https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md so that request IDs are unique across chains
I think the latter option is much cleaner. It will increase the gas cost a tiny bit and will use a relatively new (Istanbul so not that much) EVM opcode that may require an extra step in porting in some cases. Neither is that important.

bbenligiray: We don't need to do the same for withdrawals because we don't aggregate them at the node and you can't have the node misbehave (because the main problem is making 1 API call instead of 2)

andreogle: Sure, I'm fine with including the chain ID in the request ID. It does sound a bit better to me too.

Do you want to log a separate issue for it? I'll leave this part of the node as is then (including the code comment)

bbenligiray: I created the issue (https://github.com/api3dao/airnode/issues/244) and am working on it

I think "Group API calls with respect to request IDs" or something along that line would be the most accurate. This sounds like extra stuff is being done to keep chains/networks separate.

PR #243, opened 2021-03-18
Standardize the .env file used by airnode-client and implements healthcheck

### LINKED ISSUE #224: Implement HEALTHCHECK for the airnode-client container
Implement [HEALTHCHECK](https://docs.docker.com/engine/reference/builder/#healthcheck) for the [`airnode-client` image](https://github.com/api3dao/airnode/blob/pre-alpha/Dockerfile-client), update the [docs](https://github.com/api3dao/airnode/blob/pre-alpha/Docker-client.md) to restart the container as discussed [here](https://stackoverflow.com/a/61956475/14558682) (or feel free to come up with a better way).

### LINKED ISSUE #228: Standardize the .env file used by airnode-client
## Description

The client container expects secrets to be provided in a `.env` file. If the contents of this `.env` file are
```sh
ENV_NAME="ENV_VALUE"
```
Docker sets `ENV_NAME` as `"ENV_VALUE"` (and not `ENV_VALUE`). See [this](https://docs.docker.com/compose/env-file/#syntax-rules). 

## Version

The `api3/airnode-client:pre-alpha` image built from the [pre-alpha branch](https://github.com/api3dao/airnode/tree/pre-alpha)

## Present Behaviour

The client container treats quotation marks in `.env` as a part of the environment variable value.

## Expected Behaviour

The client container should omit the quotation marks in `.env`. See [this](https://github.com/motdotla/dotenv#rules).

## Steps to reproduce

Launch airnode-client with a `.env` file with quotation marks. It will fail with `invalid mnemonic` because quotation marks are included in the mnemonic.

---------------------------------------------------------------------------------------------------------------------------------------------------------------------------

The easiest way to fix this would be to copy `.env` in the container along with `config.json` and have the node `dotenv` pick it up.

Ashar2shahid:
Tested it using a config.json and .env in the root directly and it was successfully fulfilling requests.

Closes #228 
and 
Closes #224

### DIFF
--- a/Docker-client.md
+++ b/Docker-client.md
@@ -22,5 +22,44 @@ Note that `nodeSettings.cloudProvider` should be `local`.
 ```sh
 docker run -it --rm \
     -v $(pwd):/airnode/out \
+    --restart unless-stopped
     api3/airnode-client:pre-alpha
+```
+
+5. The docker image comes with a [healthcheck status](https://docs.docker.com/engine/reference/builder/#healthcheck) that can be used to monitor the health of the `airnode-client`. If the container becomes unhealthy due to any reason, you should view the logs(in case of any errors) and then restart the container. (Note: healthy status of the container does not indicate that there is no error while running the `airnode-client`, it just ensures that the `airnode-client` is being executed every minute and not halting. It is advisable to view the logs to check the actual status).
+
+6. If you want the `airnode-client` to auto-restart incase of unhealthy status then you can run the following [image](https://github.com/willfarrell/docker-autoheal) that checks for unhealthy containers and restarts them.
+
+```
+docker run -d \
+    --name autoheal \
+    --restart=always \
+    -e AUTOHEAL_CONTAINER_LABEL=all \
+    -v /var/run/docker.sock:/var/run/docker.sock \
+    willfarrell/autoheal
+```
+
+# Logs
+
+To view the logs you can use the following command 
+
+```
+docker inspect --format "{{json .State.Health }}" <containter_name> | jq
+```
+
+Expected Output
+
+```
+{
+  "Status": "healthy",
+  "FailingStreak": 0,
+  "Log": [
+    {
+      "Start": "2021-03-19T02:40:22.1717628Z",
+      "End": "2021-03-19T02:40:50.9162687Z",
+      "ExitCode": 0,
+      "Output": "yarn run v1.22.5\n$ (cd packages/node && yarn run dev:invoke)\n$ (cd __dev__ && ts-node invoke.ts)\n[2021-03-19 02:40:50.214] INFO Coordinator starting...                                                           Coordinator-ID:a82f50c2736cf3db\n[2021-03-19 02:40:50.788] INFO Fetching current block and provider admin details...                              Coordinator-ID:a82f50c2736cf3db, Provider:evm-local, Chain:EVM, Chain-ID:31337\n[2021-03-19 02:40:50.788] INFO Current block:14                                                                  Coordinator-ID:a82f50c2736cf3db, Provider:evm-local, Chain:EVM, Chain-ID:31337\n[2021-03-19 02:40:50.788] INFO Admin address:0x2886De6bbd66DB353C5Ce2e91359e7C39C962fd7                          Coordinator-ID:a82f50c2736cf3db, Provider:evm-local, Chain:EVM, Chain-ID:31337\n[2021-03-19 02:40:50.788] INFO Provider extended public key:xpub661MyMwAqRbcGeCE1g3KTUVGZsFDE3jMNinRPGCQGQsAp1nwinB9Pi16ihKPJw7qtaaTFuBHbRPeSc6w3AcMjxiHkAPfyp1hqQRbthv4Ryx  Coordinator-ID:a82f50c2736cf3db, Provider:evm-local, Chain:EVM, Chain-ID:31337\n[2021-03-19 02:40:50.816] INFO Pending requests: 0 API call(s), 0 withdrawal(s)                                  Coordinator-ID:a82f50c2736cf3db, Provider:evm-local, Chain:EVM, Chain-ID:31337\n[2021-03-19 02:40:50.830] INFO Initialized EVM provider:evm-local                                                Coordinator-ID:a82f50c2736cf3db\n[2021-03-19 02:40:50.830] INFO Forking to initialize providers complete                                          Coordinator-ID:a82f50c2736cf3db\n[2021-03-19 02:40:50.831] INFO No actionable requests detected. Returning...                                     Coordinator-ID:a82f50c2736cf3db\nDone in 28.21s.\n"
+    },
+  ]
+}
 ```
--- a/Docker-client.md
+++ b/Docker-client.md
@@ -10,7 +10,6 @@ mv Dockerfile Dockerfile-deployer && \
 ```
 
 2. Ensure that your `.env` file looks like [`.env.example`](https://github.com/api3dao/airnode/blob/pre-alpha/packages/node/__dev__/.env.example) and is the current working directory.
-Also, **do not** use quotations around your mnemonic or other variables.
 
 3. Also ensure that [`config.json`](https://github.com/api3dao/airnode/blob/pre-alpha/packages/node/__dev__/config.json.example) is also in the current working directory.
 Note that `nodeSettings.cloudProvider` should be `local`.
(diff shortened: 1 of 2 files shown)


Docker-client.md
```diff
@@ -21,7 +21,45 @@ Note that `nodeSettings.cloudProvider` should be `local`.
 
 ```sh
 docker run -it --rm \
-    --env-file .env \
     -v $(pwd):/airnode/out \
+    --restart unless-stopped
```
bbenligiray, line 25: Missing the backslash. Also `--restart` conflicts with `--rm`.

Docker-client.md
```diff
@@ -21,7 +21,45 @@ Note that `nodeSettings.cloudProvider` should be `local`.
...
     api3/airnode-client:pre-alpha
 ```
+
+5. The docker image comes with a [healthcheck status](https://docs.docker.com/engine/reference/builder/#healthcheck) that can be used to monitor the health of the `airnode-client`. If the container becomes unhealthy due to any reason, you should view the logs(in case of any errors) and then restart the container. (Note: healthy status of the container does not indicate that there is no error while running the `airnode-client`, it just ensures that the `airnode-client` is being executed every minute and not halting. It is advisable to view the logs to check the actual status).
```
bbenligiray, line 29: 5 and 6 are not steps following 1–4 so it's not appropriate to enumerate them as such

Dockerfile-client
```diff
@@ -14,8 +14,12 @@ RUN yarn add dotenv
 RUN yarn run bootstrap
 RUN yarn run build
 
-RUN echo '*/1 * * * * cd /airnode && yarn run dev:invoke' > local-cron
+RUN echo '*/1 * * * * echo "AirNode Running"' > local-cron
```
bbenligiray, line 17: Don't capitalize the N in general

Docker-client.md
```diff
@@ -21,7 +21,45 @@ Note that `nodeSettings.cloudProvider` should be `local`.
...
+    --name autoheal \
+    --restart=always \
+    -e AUTOHEAL_CONTAINER_LABEL=all \
+    -v /var/run/docker.sock:/var/run/docker.sock \
+    willfarrell/autoheal
+```
+
+# Logs
+
+To view the logs you can use the following command
```
bbenligiray, line 44: The whole point of running the node locally is to be able to read the logs easily. This is pretty annoying in this form.

Docker-client.md
```diff
@@ -21,7 +21,45 @@ Note that `nodeSettings.cloudProvider` should be `local`.
...
 docker run -it --rm \
-    --env-file .env \
     -v $(pwd):/airnode/out \
+    --restart unless-stopped
     api3/airnode-client:pre-alpha
 ```
+
+5. The docker image comes with a [healthcheck status](https://docs.docker.com/engine/reference/builder/#healthcheck) that can be used to monitor the health of the `airnode-client`. If the container becomes unhealthy due to any reason, you should view the logs(in case of any errors) and then restart the container. (Note: healthy status of the container does not indicate that there is no error while running the `airnode-client`, it just ensures that the `airnode-client` is being executed every minute and not halting. It is advisable to view the logs to check the actual status).
+
+6. If you want the `airnode-client` to auto-restart incase of unhealthy status then you can run the following [image](https://github.com/willfarrell/docker-autoheal) that checks for unhealthy containers and restarts them.
```
bbenligiray, line 31: I thought automatic restarts could be implemented without an external dependency with something like https://stackoverflow.com/a/61956475/14558682
but this probably is not the right approach because whatever orchestrates the container is supposed to restart it.

bbenligiray:
https://github.com/api3dao/airnode/issues/228#issuecomment-801944926

Here I was referring to

> Also, do not use quotations around your mnemonic or other variables.

Dockerfile-client
```diff
@@ -14,8 +14,12 @@ RUN yarn add dotenv
...
 RUN /usr/bin/crontab /airnode/local-cron 
 
+HEALTHCHECK --interval=1m --timeout=50s --start-period=10s \
```
bbenligiray, line 20: This is pretty weird. The cron job is doing nothing, all the work is being done here, and it gives 10 seconds to the node to finish the job even though we assume the node can take 1 minute serving requests.

bbenligiray (changes requested):
- Revert it to the state where the cron job invokes the node (and remove stuff about logs because we will be able to see them on the terminal again)
- Have healthcheck check if the cron job is still working with something like https://askubuntu.com/questions/85558/verify-if-crontab-works
- Remove the autoheal recommendation, let people handle it however they want
- Remove `--restart unless-stopped`
- Remove the docs sentence referring to mnemonics being enclosed in quotation marks

Also add `--init` to the `docker run` command so that the container can be killed with `Ctrl-C`

Ashar2shahid:
> * Revert it to the state where the cron job invokes the node (and remove stuff about logs because we will be able to see them on the terminal again)
> * Have healthcheck check if the cron job is still working with something like https://askubuntu.com/questions/85558/verify-if-crontab-works
> * Remove the autoheal recommendation, let people handle it however they want
> * Remove `--restart unless-stopped`
> * Remove the docs sentence referring to mnemonics being enclosed in quotation marks
> 
> Also add `--init` to the `docker run` command so that the container can be killed with `Ctrl-C`

Got it, will fix these issues.
                
Regarding the automatic restart, we will need to use the restart-policy if we want to restart the container on an error. If we don't plan on using the restart-policy should we not aim to push automatic restart in this PR? because other than the restart-policy docker does not have the ability to restart containers on its own.

bbenligiray:
Yes, let's have the healthcheck and leave it to the user to decide what to do with it

Dockerfile-client
```diff
@@ -17,5 +18,10 @@ RUN yarn run build
 RUN echo '*/1 * * * * cd /airnode && yarn run dev:invoke' > local-cron
 RUN /usr/bin/crontab /airnode/local-cron 
 
+HEALTHCHECK --interval=5m --timeout=2m \
+    CMD (strace -p $(pgrep crond) -s9999 -o cronLog.txt) & sleep 70 ; kill $! \
+        && cat cronLog.txt | grep crond > /dev/null; if [ 0 != $? ]; then kill 1; fi && rm cronLog.txt
```
Ashar2shahid, line 23: This HEALTHCHECK passes if the cron command has executed atleast once in the following 70 seconds. If the cron command is not found to be executed then the container exits .If we don't want the container to exit we can replace kill 1 with sleep 70 which will cause the HEALTHCHECK to exceed the timeout and change the status to `unhealthy` .

bbenligiray:
@MrAngus666 Can you check the healthcheck please?

javimaravillas:
Sure, I will run locally and review

Dockerfile-client
MrAngus666: I think -s9000 should be -s 9000 based on https://man7.org/linux/man-pages/man1/strace.1.html (still trying to test locally)

MrAngus666 (commented):
Change request on the healtcheck command

Dockerfile-client
Ashar2shahid: it should work even without the space, in fact -p should also work without the space ut I added it for clarity, You can have a look [here](https://itectec.com/unixlinux/bash-how-to-view-the-output-of-a-running-process-in-another-bash-session/)

Dockerfile-client
```diff
@@ -17,5 +18,10 @@ RUN yarn run build
...
+
 CMD cp out/config.json /airnode/packages/node/__dev__/config.json | true \
+    && cp out/.env /airnode/packages/node/__dev__/.env | true \
```
MrAngus666, line 26: Storing the config.json and .env into /airnode/out creates this error:
```
cp: can't stat 'out/config.json': No such file or directory
cp: can't stat 'out/.env': No such file or directory
```
Finally to make it working, I stored the .env and the config.json in the root directory of airnode, so I think this command need to be changed if we want to use a different directory than the root

MrAngus666 (changes requested):

Dockerfile-client
Ashar2shahid: the original [dockerfile ](https://github.com/api3dao/airnode/blob/master/Dockerfile)uses the config.json in the root so I followed the same design choice.

MrAngus666: My question is why the "out" folder when the docker file read then from the root.
I would remove from both if the docker finally expect them in the root folder.

Ashar2shahid: when you run the
```
docker run -it --init --rm \
    -v $(pwd):/airnode/out \
    api3/airnode-client:pre-alpha
```

it mounts the current working directory to airnode/out . So essentially when you run it from the root, it mounts the root folder to airnode/out in the docker container. So now config.json is in the airnode/out folder in the docker container. 

> Have you run it locally? Are you storing the files in the our folder or in the root?

 Yes I have runned it locally, I was storing the files in the root. If you wanted to save it in a out folder you could modify the run command like so 
 ```
 docker run -it --init --rm \
    -v $(pwd)/out:/airnode/out \
    api3/airnode-client:pre-alpha
 ```
 
and that should work, but I haven't tested it.

MrAngus666: I would change the docker run command. It will me more clear to the user that has to store the config.json and the .env file in the out folder, but if you don't think/want to, we can merge this PR as it is and create another issue to address the out directory thing

MrAngus666 (approved):

MrAngus666:
Approved. Feel free to merge :)

PR #246, opened 2021-03-20
Unique request ids

bbenligiray:
Includes the `block.chainid` to the API/withdrawal request ID hashes to ensure that requests across chains don't get assigned the same ID.

For API requests, the chain ID is also included in the event for the node to be able to validate the request ID. The node already knows the chain ID and doesn't actually need this, but by the same logic it's not an issue if request IDs are identical across chains because the node would also know which request came from which chain. This adds an additional ~500 gas cost to a total of ~90,000.

For withdrawal requests, the chain ID is included in the hash but not in the event because the node doesn't need to validate withdrawal request IDs.


packages/node/src/evm/handlers/fetch-pending-requests.test.ts
```diff
@@ -20,19 +20,20 @@ describe('fetchPendingRequests', () => {
       apiCalls: [
         {
           airnodeId: '0x19255a4ec31e89cea54d1f125db7536e874ab4a96b4d4f6438668b6bb10a6adb',
-          clientAddress: '0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0',
```
bbenligiray, line 23: Had to recreate all this because find and replace wasn't going to cut it this time

packages/node/src/evm/requests/api-calls.ts
```diff
@@ -31,6 +31,7 @@ export function initialize(logWithMetadata: EVMEventLogWithMetadata): ClientRequ
 
   const request: ClientRequest<ApiCall> = {
     airnodeId: parsedLog.args.airnodeId,
+    chainId: parsedLog.args.chainId.toString(),
```
bbenligiray, line 34: Treated similarly to `requestCount`

packages/protocol/contracts/AirnodeRrp.sol
```diff
@@ -14,17 +14,6 @@ contract AirnodeRrp is Convenience, IAirnodeRrp {
     mapping(bytes32 => bytes32) private requestIdToFulfillmentParameters;
     mapping(bytes32 => bool) public requestWithIdHasFailed;
 
-    /// @dev Reverts if the caller is not endorsed by the requester
```
bbenligiray, line 17: Had to inline this because I was getting "Stack too deep" error

packages/protocol/contracts/AirnodeParameterStore.sol
```diff
@@ -77,8 +77,8 @@ contract AirnodeParameterStore is RequesterStore, IAirnodeParameterStore {
     {
         bytes32 withdrawalRequestId = keccak256(abi.encodePacked(
             requesterIndexToNoWithdrawalRequests[requesterIndex]++,
-            requesterIndex,
-            address(this)
```
bbenligiray, line 81: I was using the contract address in the hash just to scramble the ID a bit more. I figured it's no longer needed with the chain ID (and it wasn't needed in the first place anyway).

andreogle (approved):
Look good. How much of a pain was it to update the fixture values in the node tests? This is something I find really painful and something we need to improve on (if possible)

bbenligiray:
I ended up writing a few scripts that use the protocol package to make requests and print the logs. Fortunately I didn't have to touch the fulfillments. Each time I deal with this I'm pretty confident I'll never have to again so don't come up with a permanent solution.

bbenligiray:
I created an issue to improve this. I think doing this at the protocol package would be the simplest.
https://github.com/api3dao/airnode/issues/248

Issue #207, opened 2021-02-19
Consolidate config.json and security.json


bbenligiray:
Currently, the user provider three files during deployment:
- `config.json`
- `security.json`
- `.env`

`config.json` was aimed to not include secrets, but it keeps the provider URLs, which are in some cases secret. The formats need to be revised so that only two files are needed:
- `config.json` (not secret)
- `secrets.env` (secret)

bbenligiray:
Closed by https://github.com/api3dao/airnode/pull/250

PR #250, opened 2021-03-22
Consolidate config security

bbenligiray:
Implements https://github.com/api3dao/airnode/issues/207
based on these specs
https://api3dao.github.io/api3-docs/next/technology/deployment-files/config-json.html

- `chains` is moved to `config.json` root (I think it already was there)
- `chains.*.id` type changed from `number` to `string`
- Security scheme values and chain provider URLs are read from environment variables based on the new `config.environment` field
- `logLevel` is a `config.nodeSettings` field instead of a environment variable (I considered making the values lowercase and only turning them to uppercase while printing the log but it felt too inconsistent)

Also this breaks the deployer checking if the Airnode parameters are set at the respective chain (it will skip). I didn't bother fixing it because the deployer will be redone very soon.


packages/node/package.json
```diff
@@ -15,10 +15,10 @@
     "prepublishOnly": "yarn run build",
     "test": "SILENCE_LOGGER=true jest --coverage --selectProjects unit",
     "test:e2e": "SILENCE_LOGGER=true jest --selectProjects e2e",
-    "test:e2e:debug": "LOG_LEVEL=debug jest \"withdrawals.feature\" --selectProjects e2e",
+    "test:e2e:debug": "jest \"withdrawals.feature\" --selectProjects e2e",
```
bbenligiray, line 18: Default fixture `logLevel` is `DEBUG`

packages/node/src/providers/state.ts
```diff
@@ -1,24 +1,24 @@
...
+  chainProviderName: string,
   config: Config
 ): ProviderState<EVMProviderState> {
   const masterHDNode = evm.getMasterHDNode();
-  const provider = evm.buildEVMProvider(chainProvider.url, chain.id);
+  const chainProviderEnvironmentConfig = config.environment.chainProviders.find(
+    (c) => c.chainType === chain.type && c.chainId === chain.id && c.name === chainProviderName
+  );
+  if (!chainProviderEnvironmentConfig) {
+    throw new Error('Chain provider environment variable name must be defined in the provided config');
```
bbenligiray, line 18: Not sure if I should be throwing an error or doing something else here

packages/node/src/providers/initialize.test.ts
```diff
@@ -41,16 +43,35 @@ const chains: ChainConfig[] = [
...
+    },
+  ],
+};
+
 describe('initializeProviders', () => {
   it('sets the initial state for each provider', async () => {
-    const nodeSettings = fixtures.buildNodeSettings({ chains });
-    const config = fixtures.buildConfig({ nodeSettings });
+    const config = fixtures.buildConfig({ chains, environment: environmentConfig });
+    process.env[config.environment.chainProviders[0].envName] = 'https://mainnet.infura.io/v3/<key>';
```
bbenligiray, line 73: These probably persist between tests right? I wonder if we should delete them at the end of the test.

packages/node/test/fixtures/config/config.ts
```diff
@@ -12,13 +12,51 @@ export function buildTrigger(overrides?: Partial<RequestTrigger>): RequestTrigge
...
 
 export function buildConfig(overrides?: Partial<Config>): Config {
+  const oisTitle = 'currency-converter-ois';
+  const securitySchemeName = 'mySecurityScheme';
+  const securitySchemeEnvName = `ss_${oisTitle}_${securitySchemeName}`;
+  const chainType = 'evm';
+  const chainId = '31337';
+  const chainProviderName = 'evm-local';
+  const chainProviderEnvName = `cp_${chainType}_${chainId}_${chainProviderName}`;
+  process.env[securitySchemeEnvName] = 'supersecret';
```
andreogle, line 22: I don't think we should be setting environment variables (or any side effects) from fixtures. The job of these functions is only to build and return an object with a specific structure & type. 

These should either be set in the test, a `beforeEach` block or in `test/setup/init/set-env-vars.ts`. (I don't like the nested structure of the `test/setup/` directory and want to change it)

packages/node/test/fixtures/provider-states/evm.ts
```diff
@@ -1,26 +1,40 @@
...
 export function buildEVMProviderState(
   overrides?: Partial<ProviderState<EVMProviderState>>
 ): ProviderState<EVMProviderState> {
   const coordinatorId = '837daEf231';
-  const chainProvider: ChainProvider = { name: 'ganache-test', url: 'http://localhost:4111' };
+  const chainType = 'evm';
+  const chainId = '1337';
+  const chainProviderName = 'ganache-test';
+  const chainProviderEnvName = `cp_${chainType}_${chainId}_${chainProviderName}`;
+  process.env[chainProviderEnvName] = 'http://localhost:4111';
```
andreogle, line 14: Same here

packages/node/src/providers/state.ts
andreogle: I'm fine with throwing but I think the error message should include the context we already have (chain type, ID and provider name)

andreogle (changes requested):

packages/node/src/providers/initialize.test.ts
bbenligiray: I think we would have needed to worry about this if it was the node itself adding/removing environment variables

packages/node/test/fixtures/config/config.ts
bbenligiray: Agreed. I think setting them all in `test/setup/init/set-env-vars.ts` is good enough here (because of https://github.com/api3dao/airnode/pull/250#discussion_r599534099) and consistent with how the mnemonic is treated.
Came across this btw, but the current implementation is more obvious, to me at least
https://jestjs.io/docs/environment-variables

packages/node/jest-e2e.config.js
```diff
@@ -5,6 +5,6 @@ module.exports = {
   // Add custom settings below
   name: 'e2e',
   displayName: 'e2e',
-  setupFiles: ['<rootDir>/test/setup/init/set-define-property.ts'],
+  setupFiles: ['<rootDir>/test/setup/init/set-define-property.ts', '<rootDir>/test/setup/init/set-env-vars.ts'],
```
bbenligiray, line 8: ^^^

packages/node/src/providers/initialize.test.ts
javimaravillas: It will persist during the test. You can clean in the beforeEach to make sure you don't have an issue there

packages/node/test/fixtures/config/config.ts
MrAngus666: I agree that `test/setup/init/set-env-vars.ts` is good enough now but I think we need to change it to the future to beforeEach(), specially due to concurrency running the tests. Specially if we do this kind of initialisation: 
https://github.com/api3dao/airnode/blob/master/packages/node/src/handlers/call-api.test.ts#L11, but agree this is good enough here.

MrAngus666 (approved):
Looks good to me and requested changes (by @andreogle) has been addressed

bbenligiray:
Created an issue about test setup, merging and closing this

PR #253, opened 2021-03-27
Update to Node 14

### LINKED ISSUE #222: Upgrade to Node.js v14
[Active LTS](https://nodejs.org/en/about/releases/) for Node.js is v14. We are currently using v12, which is also being supported, but we should upgrade at some point.

MrAngus666:
Closes #222

Using 14.16.0-alpine3.13 rather than 14.16.0-alpine3.12 because 14.16.0-alpine3.12 was giving errors and was not able to run the docker.

No breaking changes in the codebase.

Tested:
- Building and running all the repo tests.
- Running dev:invoke locally
- Running dockerfile-client (pushed a branch with node14 so the git clone of the docker was getting the new version)

After: 
https://github.com/api3dao/airnode/pull/243 is merged

We should merge pre-alpha into master for having #222 #228 and #224 into master (will create an issue for tracking)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12.19.0-alpine3.12
+FROM node:14.16.0-alpine3.13
 
 ENV NODE_ENV production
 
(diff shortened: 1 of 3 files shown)


andreogle (approved):
LGTM :shipit:

PR #256, opened 2021-03-30
Add initial `version` 0.1.0 to `package.json`

sitch:
Can't install without a valid "version" in `package.json`

```cli
❯ yarn add https://github.com/api3dao/airnode                                           ─╯
yarn add v1.22.10
[1/5] Validating package.json...
[2/5] Resolving packages...
error Can't add "airnode": invalid package version undefined.
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "airnode",
+  "version": "0.1.0",
   "license": "MIT",
   "engines": {
     "node": "^12.13.1"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "airnode",
-  "version": "0.1.0",
+  "version": "0.0.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "airnode",
+  "version": "0.1.0",
```
MrAngus666, line 3: We are still working to have the 0.1.0 version, so do you mind to change to 0.0.1?

MrAngus666 (changes requested):

andreogle:
I'm just a bit curious, what were you trying to do with `yarn add ...` here?

package.json
sitch: Totally! I'll modify right now

PR #260, opened 2021-03-31
updated .env example and config.json

MrAngus666:
Updating .env.example and config.example so we can run start-coordinator locally add-hoc using them without need any changes.

### DIFF
--- a/packages/node/__dev__/.env.example
+++ b/packages/node/__dev__/.env.example
@@ -3,7 +3,7 @@
 # =================================================================
 # This mnemonic is also used in the operation package
 MASTER_KEY_MNEMONIC="achieve climb couple wait accident symbol spy blouse reduce foil echo label"
-# This matches the config.json environments->chain providers -> name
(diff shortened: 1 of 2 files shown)


packages/node/__dev__/.env.example
```diff
@@ -1,8 +1,10 @@
 # =================================================================
 # Master key
 # =================================================================
-MASTER_KEY_MNEMONIC="enter your mnemonic phrase here inside the quotes"
-
+# This mnemonic is also used in the operation package
+MASTER_KEY_MNEMONIC="achieve climb couple wait accident symbol spy blouse reduce foil echo label"
+# This matches the config.json environments->chain providers -> name
```
andreogle, line 6: Just for consistency:
```suggestion
# This matches the config.json environment -> chain providers -> envName
```

andreogle (approved):
Nice catch. Thanks for fixing this 👍🏻

PR #265, opened 2021-04-05
Median contract & tests

sassmilic:
High-level idea:
- Arrays of length <=16 have their median computed "directly" via min number of comparisons, see: http://www.cs.hut.fi/~cessu/selection/
- Arrays of length >16 use quickselect (https://en.wikipedia.org/wiki/Quickselect)

See the notebook for analysis on best sorting algo re: minimizing gas costs. Also, see discussion below.


Drgy (commented):
I'm new into the smart contracts, im worried i might be overlooking something very obvious so please bear with me. Im not entirely sure what the function `_partiton`  is supposed to do. If im not mistaken in the quicksort you are using it as a Hoare partition, however the current implementation is giving quite strange results and I wasn't able to find out what was the goal just from the code.

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,359 @@
...
+      uint256 greatestLessThan = 0;
+      uint256 leastGreaterThan = 2**256 - 1; // effectively acts as "infinity"
+      for (uint i=0; i<array.length; i++) {
+        if (array[i] < median && array[i] > greatestLessThan) {
+          greatestLessThan = array[i];
+        } else if (array[i] > median && array[i] < leastGreaterThan) {
+          leastGreaterThan = array[i];
+        }
+      }
+      return median == (greatestLessThan + leastGreaterThan) / 2;
```
Drgy, line 49: Perhaps it would be easier to have `greatestLessThan` and `leastGreaterThan` in the previous cycle so that we don't iterate twice?

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,359 @@
...
+        } else if (array[i] > median) {
+          greaterthan++;
+        }
+      }
+      if (lessthan != greaterthan) {
+        return false;
+      }
+      if (array.length % 2 == 1) {
+        return true;
+      }
```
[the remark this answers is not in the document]
sassmilic: Here's an example where it fails:
```
array := [1, 2, 3, 4]
median := 2.1
```
Then `lessthan == 2` and `greaterthan == 2`. So your proposed code would return `true` even though the true median is 2.5.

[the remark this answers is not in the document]
sassmilic: Yes, that works too, but I wanted the code to exit early *if* the array is even-length and the "first check" doesn't pass. That is, if I were to replace my code with your code, the second for-loop would execute even though I know prior to that execution that `median` isn't the true median.

packages/protocol/contracts/Median.sol
sassmilic: Yep, Burak suggested that too. So, yea, there's a way to avoid doing two scans of the array.

sassmilic:
>If im not mistaken in the quicksort you are using it as a Hoare partition, however the current implementation is giving quite strange results and I wasn't able to find out what was the goal just from the code.

@Drgy what do you mean by "strange results"? Are you running the code on some test cases or "manually" testing it (i.e. going through some possible scenarios in your head)?

Drgy:
> > If im not mistaken in the quicksort you are using it as a Hoare partition, however the current implementation is giving quite strange results and I wasn't able to find out what was the goal just from the code.
> 
> @Drgy what do you mean by "strange results"? Are you running the code on some test cases or "manually" testing it (i.e. going through some possible scenarios in your head)?

Good old debugging on the paper :sweat_smile:

I had trouble of understanding what's the idea behind this algorithm, but now I see, it makes sense. I just wrongly interpreted the results, the function is correct, I messed up. Do you plan on keeping this implementation, or replacing it with Horae's partition algorithm?

sassmilic:
> > > If im not mistaken in the quicksort you are using it as a Hoare partition, however the current implementation is giving quite strange results and I wasn't able to find out what was the goal just from the code.
> > 
> > 
> > @Drgy what do you mean by "strange results"? Are you running the code on some test cases or "manually" testing it (i.e. going through some possible scenarios in your head)?
> 
> Good old debugging on the paper 😅
> 
> I had trouble of understanding what's the idea behind this algorithm, but now I see, it makes sense. I just wrongly interpreted the results, the function is correct, I messed up. Do you plan on keeping this implementation, or replacing it with Horae's partition algorithm?

@Drgy Hoare's is better since it results in fewer `_swaps` in majority of cases ... if only I can figure out how to amend it to work in this context :)

sassmilic:
Interesting: (known) minimum number of comparisons needed to find a median: http://www.cs.hut.fi/~cessu/selection/

This would imply, I think, that lists upwards of length ~10 would be more (gas) efficient to solve with direct comparison.

sassmilic:
Lots of refactoring. @drgy @bbenligiray ready for another looksie 👀

sassmilic:
copy-pasting what i wrote in slack

- direct comparison method is: using the minimum number of comparisons to find a median, see: http://www.cs.hut.fi/~cessu/selection/
- we only do a direct comparison for arrays <= 7. it gets quite infeasible to implement more than that because of the shear amount of code (you start reaching your contract size limit). also, since we're dealing with first-party oracles (and thus 'independent' data points), we likely won't be aggregating more than ~10 responses, for the most part
- interestingly: quickselect without using a "medians of medians" heuristic for the pivot performs better than quickselect with the heuristic. turns out the gas cost overhead of computing "median of medians" isn't worth it.
- this shows the importance of doing these sorts of tests, because the results, for the most part, are rather unintuitive & hard to predict
sorting_gas_costs_zoomed
sorting_gas_costs

sassmilic:
another thing: feel free to ignore the files in the `notebooks` directory. that's just there for reference, in case you wanna know how the graphs were computed and in case i want to re-run tests in the future.

sassmilic:
small note: the tests i ran were on randomized arrays, thus quickselect behaves identically to quickselect with a random pivot. in reality, responses aren't random, and they might actually display a pattern since responses are recorded as they "come in". e.g. prices reported by honest oracles as a price is going up is going to display a sorted-ish pattern when saved to the response array.

sassmilic:
important! the code currently modifies the passed array. we would want to copy the array before needing to modify it, correct? @bbenligiray

sassmilic:
comparing to chainlink as a baseline, since they've been implementing medians for the last 3 years or so
chainlink-comparison

bbenligiray:
> important! the code currently modifies the passed array. we would want to copy the array before needing to modify it, correct? @bbenligiray

I was hoping to review this today but didn't get around to so I'll just reply to this. I don't see us needing not to modify the passed array. Either provide both options or calculate the median in-place, name the method accordingly and document it as a `@dev` string.

sassmilic:
> Either provide both options
Hah, that's what I was thinking. @bbenligiray

sassmilic:
@bbenligiray done. Please review when you get the chance :)

Final results, fyi:
final

sassmilic:
fyi & for logging purposes, i believe most of the gas cost savings (for array length > 16) come from a more efficient implementation of hoare's partition algo: https://stackoverflow.com/questions/58331986/quickselect-with-hoare-partition-scheme/67097378#67097378

sassmilic:
> @bbenligiray done. Please review when you get the chance :)
> 
> Final results, fyi:
> final

So for arrays <= 16 (which is probably what we'll be dealing with) the cost of aggregation is ~2x a normal transaction, which I think is reasonably cheap.

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,441 @@
...
+  uint256 constant SMALL_ARRAY_MAX_LENGTH = 16;
+    
+  // EXTERNAL FUNCTIONS
+  
+  /* Returns the kth smallest element in an array of signed ints without
+   * modifying the input array. 
+   * @param arr An array of signed integers.
+   * @param k
+   * @return the kth smallest element in `arr`
+   */
```
bbenligiray, line 18: We shouldn't use different styles in a single package

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,441 @@
...
+    }
+    (uint256 idx1, uint256 idx2) = _quickSelect(arr, 0, arr.length - 1, k, true);
+    return (arr[idx1], arr[idx2]);
+  }
+
+  // PRIVATE FUNCTIONS
+
+  /* Select the index of the kth element in an array.
+   * @dev This function may modify array.
+   * @param arr an array of uints.
```
bbenligiray, line 110:
```suggestion
   * @param arr an array of ints.
```

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,441 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.2;
+
+/// @title A contract for seelecting the kth smallest element in an array of uints.
```
bbenligiray, line 4:
```suggestion
/// @title A contract for selecting the kth smallest element in an array of ints.
 ```

File: packages/protocol/contracts/Median.sol [hunk 6]
```diff
@@ -0,0 +1,441 @@
...
+   *         (k+1)st elements of `arr`, respectively. In the case where
+   *         `selectKplusOne` is false, the tuple returned is (i, 0).
+   */
+   function _quickSelect
+  (
+    int256[] memory arr,
+    uint256 left,
+    uint256 right,
+    uint256 k,
+    bool selectKplusOne
```
bbenligiray, line 126: It's weird how these are sometimes implemented as different functions (`compute`/`computeInPlace`) and sometimes parameterized by a flag

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,441 @@
...
+  )
+    private
+    pure
+    returns (uint256, uint256)
+  {
+    assert(left <= k && k <= right);
```
bbenligiray, line 132: `require()` (if you consider this as a standalone method, which I think we should)

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,441 @@
...
+   * @param left the left index to begin search.
+   * @param right the right index to begin search.
+   * @param k
+   * @param selectKplusOne a bool representing whether the function should
+   *                       return the (k+1)st element or not.
+   * @return a tuple (i, j) where i and j are the indices of the kth and
+   *         (k+1)st elements of `arr`, respectively. In the case where
+   *         `selectKplusOne` is false, the tuple returned is (i, 0).
+   */
+   function _quickSelect
```
bbenligiray, line 120: Underscore prefixes are not in the Solidity style guide

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,441 @@
...
+  (
+    int256[] memory arr,
+    uint256 k
+  )
+      private
+      pure
+      returns (int256)
+  {
+    assert(arr.length <= SMALL_ARRAY_MAX_LENGTH);
+    assert(0 <= k && k <= arr.length - 1);
```
bbenligiray, line 221: `require()`

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,441 @@
...
+
+    if (arr.length == 1) {
+    }
+    else if (arr.length == 2) {
```
bbenligiray, line 225:
```suggestion
    // if (arr.length == 1) {} else
    if (arr.length == 2) {
```

bbenligiray: If you want to be explicit about it

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,441 @@
...
+    return arr[k];
+  }
+
+  /* Swap two elements of an array iff the first element
+   * is greater than the second.
+   * @param arr an array of unsigned integers
+   * @param i the first index
+   * @param j the second index
+   */
+  function _swap
```
bbenligiray, line 409: The name is misleading. Maybe `conditionalSwap`?

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,441 @@
...
+  {
+    assert(i < j);
+    if (arr[i] > arr[j]) {(arr[i], arr[j]) = (arr[j], arr[i]);}
+  }
+
+  /**
+   * Make an in-memory copy of an array
+   * @param arr The array to be copied.
+   */
+  function copy
```
bbenligiray, line 426:
```suggestion
  function copy(int256[] memory arr)
```

packages/protocol/notebooks/ChainlinkMedian.sol
```diff
@@ -0,0 +1,242 @@
+// SPDX-License-Identifier: MIT
```
bbenligiray, line 1: We probably want to extract this to another repo, it will clutter things up

packages/protocol/test/Median.sol.js
```diff
@@ -0,0 +1,144 @@
+const { expect } = require('chai');
```
bbenligiray, line 1: We may want to seed the RNG for the `test:coverage` script

bbenligiray (changes requested):
Good job, requesting changes
- Some minor suggestions
- Move `Median.sol` to `dapis/`
- Extract the notebook stuff to another repo, could be a personal or api3 one. You can link it in the package README
- There should be two kinds of tests for this, one runs in 2 seconds, the other is the current one. The current one should only run with the `test:full` script. Let me know if you want me to implement this. We can't merge this to master in this state.
- I was going to check test coverage but the tests didn't end so I couldn't. I don't think the`require()`s are tested. We can check after the test fix.

The inplace/non-inplace and even/odd length generality bloated the contract and the interface. I'd like to merge this and open a PR about that.

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,441 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.2;
+
+/// @title A contract for seelecting the kth smallest element in an array of uints.
+/// @author Sasa Milic - <sasa@api3.org>
+contract SelectK {
+  
+  // maximum length of an array that `_medianSmallArray` can handle
+  uint256 constant SMALL_ARRAY_MAX_LENGTH = 16;
```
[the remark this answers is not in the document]
sassmilic: Why not `private`?

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,441 @@
...
+  )
+    public
+    pure
+    returns (int256)
+  {
+    assert(0 <= k && k <= arr.length - 1);
+    if (arr.length <= SMALL_ARRAY_MAX_LENGTH) {
+      return _selectK_smallArray(arr, k);
+    }
+    (uint256 idx1, uint256 idx2) = _quickSelect(arr, 0, arr.length - 1, k, false);
```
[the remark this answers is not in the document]
sassmilic: didn't know you could do that ...

packages/protocol/contracts/Median.sol
sassmilic: `compute`/`computeInPlace` are public functions. `quickselect` is a private function so it's optimized for gas cost savings without caring how one interfaces with it. So there's just one function that handles many cases.

packages/protocol/contracts/Median.sol
sassmilic: Then `arr.length == 1` falls under the `else` case. So it's a matter of having a `arr.length == 1` check or having an `arr.length == 16` check. And, on second thought, I think it might be better to do this check in the topmost functions, and get it "out of the way" immediately.

packages/protocol/contracts/Median.sol
sassmilic: good call

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,441 @@
...
+      } 
+      (arr[i], arr[j]) = (arr[j], arr[i]);
+    }
+  }
+
+  /* Return the kth element of a small array (at most length 16).
+   * @param arr an array of signed integers (at most length 16)
+   * @return the kth smallest element in `arr`
+   */ 
+  function _selectK_smallArray
```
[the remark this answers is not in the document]
sassmilic: It's a private function. Do you want me just to add it to the docstring (like with the other functions)?

sassmilic:
@bbenligiray 

> Extract the notebook stuff to another repo
Could you suggest me a repo or directory to put it in?

sassmilic:
> The inplace/non-inplace and even/odd length generality bloated the contract and the interface. I'd like to merge this and open a PR about that.

I wanted to have just one function that you pass parameters to, but you seem to have a problem with users passing parameters to a function @bbenligiray

packages/protocol/contracts/Median.sol
```diff
@@ -0,0 +1,441 @@
...
+  function computeInPlace
+  (
+    int256[] memory arr,
+    uint256 k
+  )
+    public
+    pure
+    returns (int256)
+  {
+    assert(0 <= k && k <= arr.length - 1);
```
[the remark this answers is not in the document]
sassmilic: This is an internal invariant. It should never evaluate to false. `assert` here essentially acts as a really loud comment, also: https://medium.com/blockchannel/the-use-of-revert-assert-and-require-in-solidity-and-the-new-revert-opcode-in-the-evm-1a3a7990e06e

packages/protocol/contracts/Median.sol
sassmilic: You said smth about not wanting users to pass too many arguments/flags to the main/external functions.

packages/protocol/contracts/Median.sol
sassmilic: What do you mean by "standalone method"?

packages/protocol/contracts/Median.sol
sassmilic: That's gonna make those above functions uglier -- is the trade-off worth it?

sassmilic:
re: `require` statements. I only see the use if it's checking user input. If it's a condition/invariant in the internal code that should *never* happen, it's the context for an `assert`. I consider `asserts` like loud comments in code that make you aware of certain invariants. They can be removed all-together, but I _like_ having them.

sassmilic:
> The current one should only run with the test:full script.

Yea, I asked about this earlier re: continuous testing. I'm not sure how to set this up, so it might be quicker if you do it.

packages/protocol/contracts/Median.sol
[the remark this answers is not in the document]
bbenligiray: Actually `private` makes more sense since `SMALL_ARRAY_MAX_LENGTH` doesn't affect the behavior

packages/protocol/contracts/Median.sol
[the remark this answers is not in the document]
bbenligiray: Also since `k` is a `uint`, `0 <= k` is redundant

packages/protocol/contracts/Median.sol
bbenligiray: Got it. I think these flags are bad in general, not only for external functions. Solidity is strongly typed, you're working around that. When `selectKplusOne` is `true`, both the returned type becomes a tuple and what the function does changes so `quickSelect` is not a valid name anymore. The two should be implemented as separate functions.

packages/protocol/contracts/Median.sol
bbenligiray: Referring to this
https://github.com/api3dao/airnode/pull/265#discussion_r617247627
It could be argued that this can be `assert()` because this is `private` and in the methods that this is called it's ensured that this will never evaluate to `true`. However this is prone to error because one could add a method to the contract that doesn't satisfy that later on and forget to update this `assert()` as `require()`.
I'm 100% sure that the other `assert()`s should be `require()`, I'm 51% on this one.

packages/protocol/contracts/Median.sol
bbenligiray: Ah, right sorry. It's just that the linter complains about empty blocks, so maybe replace `else` with `else if (arr.length == 16)`

packages/protocol/contracts/Median.sol
bbenligiray: I think so because I never realized that those were not regular swaps before this review. Maybe `conSwap()` or something like that.

bbenligiray:
Can you update the `extended_test/Median.sol.js` with the long running test from before? If you run `yarn run test:coverage-extended`, you can see the coverage in `coverage/index.html`. It's pretty low at the moment.

packages/protocol/contracts/Median.sol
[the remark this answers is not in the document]
sassmilic: Sorry that shouldn't be public.

PR #271, opened 2021-04-06
Prefer `const assertions` over `type unions`

sitch:
Const assertions will automatically resolve string literals without the need to explicitly and manually typecast

For instance:
```ts
import { OIS } from "./types";
import json from "./ois.json";

const example: OIS = json;
```
Will raise a type error as a string is not automatically assignable to either a product union or an enum.

### DIFF
--- a/packages/ois/src/types.ts
+++ b/packages/ois/src/types.ts
@@ -1,8 +1,17 @@
 // ===========================================
 // General
 // ===========================================
-export type Method = 'get' | 'post';
-export type ParameterTarget = 'path' | 'query' | 'header' | 'cookie';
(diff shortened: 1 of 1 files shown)


sitch:
See: https://mariusschulz.com/blog/const-assertions-in-literal-expressions-in-typescript#:~:text=As%20of%20TypeScript%203.4%2C%20we,assertion%20to%20a%20literal%20expression%3A&text=Using%20a%20const%20assertion%20on,literals%20will%20get%20readonly%20properties.

andreogle:
Nice one, thanks. I wasn't aware of `const assertions` before 👍🏻

PR #288, opened 2021-04-14
Rename the withdrawal counter

bbenligiray:


Siegrift (commented):
I still not understand why the variable can't start from 0

bbenligiray:
Incrementing 0 costs more gas than incrementing 1

Siegrift:
Wow. Do you have explanation for that? That seems very interesting :)

Siegrift (approved):

bbenligiray:
https://ethereum.github.io/yellowpaper/paper.pdf
Appendix G
Gsset    20000    Paid for an SSTORE operation when the storage value is set to non-zero from zero.
Gsreset   5000    Paid for an SSTORE operation when the storage value’s zeroness remains unchanged or is set to zero.

Whoever creates the requester record also initializes the index at 1 and pays the 20,000 gas upfront. Then, all the following withdrawal requests will pay an additional 5,000 gas to increment the index (which is used for unique withdrawal IDs).

PR #294, opened 2021-04-21
Update deployer CLI to work in a more common way

amarthadan:
Deployer CLI can now be run as a common CLI even from the Docker container, e.g. `docker run -it --rm -e AWS_ACCESS_KEY_ID=xxx -e AWS_SECRET_KEY=xxx api3/airnode --help`

I didn't update the Docker readme much because the image building will be reworked entirely in the following PRs.


packages/deployer/package.json
```diff
@@ -2,10 +2,15 @@
...
   "version": "0.1.0",
+  "bin": {
+    "deployer": "./bin/deployer"
+  },
+  "main": "bin/deployer",
   "scripts": {
     "_build": "yarn run clean && yarn run compile",
     "clean": "rimraf -rf ./dist",
     "compile": "tsc -p tsconfig.build.json",
+    "cli": "ts-node bin/deployer.ts",
```
bbenligiray, line 13: Nice, I was wanting to use ts-node for this

bbenligiray (commented):
The Docker image build on master was broken anyway so I don't have any problems with merging this to be used as reference later on. I think it's a good idea to leave Docker to last in general.

I think we should build the image inside the deployer package. As long as we have the bundled node in `packages/deployer/.webpack`, we shouldn't need the rest of the monorepo.

The user needs to pass `config.json` and `secrets.env` to the image. I'm not sure how that will be done here.

amarthadan:
> The Docker image build on master was broken anyway so I don't have any problems with merging this to be used as reference later on. I think it's a good idea to leave Docker to last in general.

Current solution should work even though it's a bit cumbersome as it bundles everything.

> I think we should build the image inside the deployer package. As long as we have the bundled node in `packages/deployer/.webpack`, we shouldn't need the rest of the monorepo.

I agree, it's in the roadmap :slightly_smiling_face: 

> The user needs to pass `config.json` and `secrets.env` to the image. I'm not sure how that will be done here.

Via volumes, you don't really have any other option when using Docker anyway :confused:. It would look something like

```sh
docker run -it --rm -v $(pwd)/config.json:/airnode/config.json -v $(pwd)/secrets.env:/airnode/secrets.env -e AWS_ACCESS_KEY_ID=xxx -e AWS_SECRET_KEY=xxx api3/airnode deploy
```
Or mounting just one configuration directory where all the neccessary files are to make it simpler.

package.json
```diff
@@ -46,7 +46,8 @@
     "test:protocol": "(cd packages/protocol && yarn run test)",
     "test:node": "(cd packages/node && yarn run test)",
     "test:node:watch": "(cd packages/node && yarn run test:watch)",
-    "test:node:watch:debug": "(cd packages/node && yarn run test:watch:debug)"
+    "test:node:watch:debug": "(cd packages/node && yarn run test:watch:debug)",
+    "cli:deployer": "lerna run --scope @airnode/deployer cli -- --"
```
andreogle, line 50: Small thing, but could we keep these ordered alphabetically please?

amarthadan: Of course, I didn't realize that's the case :slightly_smiling_face:

bbenligiray (approved):
Okay, let's merge and come back to it later

package.json
sitch: There are a few eslint package-json sorting plugins as well as a stand-alone script called sort-package-json

PR #298, opened 2021-04-24
Aggregator contracts with a standardized interface

bbenligiray:
Added three aggregator contracts with a standardized interface
- Mean
- Median
- MeanMedianHybrid
and a stub dAPI contract that illustrates how the user selects the specific aggregation method


packages/protocol/contracts/dapis/aggregators/Aggregator.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.2;
+
+/// @title The abstract contract that generalizes aggregation methods
+/// @notice Aggregation method implementations should inherit this contract
+abstract contract Aggregator {
```
bbenligiray, line 6: I would much rather `Reducer`/`Reduce` instead of `Aggregator`/`Aggregate`
https://en.wikipedia.org/wiki/Reduction_Operator
WDYT

packages/protocol/contracts/dapis/aggregators/MeanMedianHybrid.sol
```diff
@@ -0,0 +1,98 @@
...
+
+    /// @notice Called by the owner to set the tolerance
+    /// @dev This should be set high enough that the aggregation does not fall
+    /// back to median unless there is a misreport among the values to be
+    /// aggregated
+    /// @param _toleranceInPercentages Tolerance in percentages
+    function setTolerance (int256 _toleranceInPercentages)
+        external
+        onlyValidPercentage(_toleranceInPercentages)
+        // onlyOwner
```
bbenligiray, line 60: It's awkward that some aggregator contracts are ownable and some are not

packages/protocol/contracts/dapis/aggregators/MeanMedianHybrid.sol
```diff
@@ -0,0 +1,98 @@
...
+    {
+        toleranceInPercentages = _toleranceInPercentages;
+    }
+
+    /// @notice Aggregates the array of values by computing their mean,
+    /// checking if the mean is similar enough to median, and computing the
+    /// median instead if it is not
+    /// @param values Values to be aggregated
+    function aggregateInplace(int256[] memory values)
```
bbenligiray, line 69: To not make `MeanMedianHybrid.sol` ownable I'd rather pass `toleranceInPercentages` as an argument from the dAPI contract. The problem then is that `aggregateInplace` interface is not consistent among aggregators so you can't switch them in and out modularly.

packages/protocol/contracts/dapis/aggregators/Aggregator.sol
```diff
@@ -0,0 +1,17 @@
...
+    /// @notice Called internally by the dAPI contract to aggregate an array of
+    /// values
+    /// @dev It is not guaranteed that `values` will not be changed after this
+    /// method runs
+    /// @param values Values to be aggregated
+    function aggregateInplace(int256[] memory values)
+        internal
+        view
```
bbenligiray, line 14: Had to switch from `pure` to `view` because `MeanMedianHybrid.sol` reads `toleranceInPercentages` from storage. A lot of the potential aggregator contracts such as weighted ones will also do that so this is probably fine.

packages/protocol/contracts/dapis/aggregators/Aggregator.sol
sassmilic: I agree

packages/protocol/contracts/dapis/aggregators/Aggregator.sol
sassmilic: What if you set the parameter to a constant?

packages/protocol/contracts/dapis/aggregators/MeanMedianHybrid.sol
```diff
@@ -0,0 +1,98 @@
...
+/// and median. Since the calculation of mean and its validation takes O(n)
+/// time, it is more scalable than bare median under normal circumstances
+/// (where mean will be similar to median).
+/// @dev Even though the types are defined as `int256` for consistency, this
+/// method is more suitable for unsigned values. For example, consider the
+/// values [-100, -4, 1, 1, 1, 1, 100]. The mean of these values is 0 and the
+/// median is 1. Here, even though the mean is very close to the median, upper
+/// and lower tolerances will be zero, which means the aggregation will fall
+/// back to median. Therefore, consider using Median.sol instead for values
+/// with potentially alternating signs.
```
sassmilic, line 21: It seems to me this is an edge case when mean == 0? If so, I would just say this explicitly.

packages/protocol/contracts/dapis/aggregators/MeanMedianHybrid.sol
sassmilic: why is that?

sassmilic (commented):
Looks good. I made a few comments

packages/protocol/contracts/dapis/aggregators/MeanMedianHybrid.sol
bbenligiray: I mean the hybrid aggregator contract has an interaction that needs to be put behind `onlyOwner` while mean and median don't

packages/protocol/contracts/dapis/aggregators/Aggregator.sol
bbenligiray: I think tolerance should be modifiable (by the owner)

packages/protocol/contracts/dapis/aggregators/MeanMedianHybrid.sol
bbenligiray: It's not only when mean == 0 but it becomes more prominent when mean gets closer to 0

packages/protocol/contracts/dapis/aggregators/MeanMedianHybrid.sol
sassmilic: Why can't the user have access to that parameter?

packages/protocol/contracts/dapis/aggregators/MeanMedianHybrid.sol
sassmilic: I don't think so. Obviously it depends on the data, but if you're dealing with small data or data that averages to ~0, then small changes are important, so I think % difference would work in these cases as well. The only real exception is 0. That's why I thought it should be more explicit in the docstring.

sassmilic: E.g. consider PNK ;) @ $0.1501, a 1% changes is only fractions of a penny, but it's still significant since the price is so low to begin with.

sassmilic: nvm, i agree with you on second thought :) it matters when the number representation is fewer than the max bits stored by the `int`

packages/protocol/contracts/dapis/aggregators/MeanMedianHybrid.sol
bbenligiray: Random people can set the tolerance to 0% or 100%

packages/protocol/contracts/dapis/aggregators/MeanMedianHybrid.sol
bbenligiray: This works well when all values are positive yet small. The problem is when the mean is close to 0 and the values have mixed signs. So it wouldn't be a problem with price data.

bbenligiray:
I'm merging this but I'm still not 100% about some things. I think they will be more clear with the aggregator.

bbenligiray:
@Siegrift Sure, no idea what I was thinking

PR #301, opened 2021-04-26
Memoryless aggregator

sassmilic:
After implementing this, I think a "memoryless" aggregator is actually a bad idea lulz. In particular, when you create a new request, you have to delete all existing responses (from the previous request) and ignore any incoming responses that aren't for the current request. Idk, maybe it works in certain contexts.

Questions:
- Do we want to make the aggregator “type” changeable?
- Were you intending to make the dAPI inherit from `Aggregator`? I'm asking since you made the aggregator functions internal. What are the pros/cons of inheritance vs just calling an on-chain aggregator? The latter makes more sense to me, but I'm not an expert on this.
  - This is sorta related to the first question: if it inherits, you can't just 'switch out' aggregation methods, right?
- I just realized that we'll be recomputing the median (if that's the chosen method) multiple times and we would have modified the array (essentially sorting it) the first time. This means aggregation calls after the first should be cheaper *if* I change the default pivot to the middle element and not the first (like it is now) -- just autistic tingz.

### DIFF
--- a/packages/protocol/contracts/dapis/aggregators/MemorylessDApi.sol
+++ b/packages/protocol/contracts/dapis/aggregators/MemorylessDApi.sol
@@ -55,7 +55,7 @@ contract MemorylessDApi is DApi, AirnodeRrpClient {
     }
 
     modifier onlyOwner(){
-        require(msg.sender == _owner, 'only the owner can call this function');
+        require(msg.sender == _owner, "only owner can call this function");
         _;
     }
 
--- a/packages/protocol/contracts/dapis/aggregators/MemorylessDApi.sol
+++ b/packages/protocol/contracts/dapis/aggregators/MemorylessDApi.sol
@@ -1,0 +1,142 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.2;
+
+import "../../AirnodeRrpClient.sol";
+import "./Median.sol";
+import "./DApi.sol";
+
+/**
+ * @title A sample dAPI/aggregator contract
+ * @dev only owner can trigger request
+ */
+contract MemorylessDApi is DApi, AirnodeRrpClient {
+    Aggregator public aggregator;
+
+    bytes32[] public templateIds;
+    uint256 public requesterIndex;
+    address public designatedWallet;
+    uint128 public minimumResponses;
+
+    int256 public aggregatedAnswer;
+
+    address immutable _owner;
+    address immutable fulfillAddress;
+    bytes4 immutable fulfillFunctionId;
+    mapping (bytes32 => bool) private _pendingRequests;
+    // individual answers from Airnodes are stored in an array
+    int256[] _individualAnswers;
+
+    event NewUpdateRequested(uint256 timestamp);
+    event RequestFulfilled(uint256 requestId);
+    event AggregatedAnswerUpdated(uint256 timestamp, uint256 aggregatedAnswer);
+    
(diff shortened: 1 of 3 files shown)


bbenligiray:
> Do we want to make the aggregator “type” changeable?

Solidity doesn't have [templates](https://en.cppreference.com/w/cpp/language/templates) yet so we will have to duplicate contracts with different types. We probably only need `int256` and maybe `bytes32` anyway.

> Were you intending to make the dAPI inherit from Aggregator? I'm asking since you made the aggregator functions internal. 

Yes, [this](https://github.com/api3dao/airnode/blob/master/packages/protocol/contracts/dapis/rrpDapi.sol) was supposed to be an example

> What are the pros/cons of inheritance vs just calling an on-chain aggregator? The latter makes more sense to me, but I'm not an expert on this. 

It's cheaper to use the inherited reducer vs. making an external call. Also having a lot of reducer contracts floating around and being plugged into dAPIs will result in a rat's nest, this being an example https://github.com/daostack/arc and we'll end up having 50+ versions of the same contracts deployed https://github.com/daostack/alchemy-monorepo/tree/dev/packages/migration/contracts
It may be a lot more manageable with fewer and more static contracts, I'm just being cautious.

> This is sorta related to the first question: if it inherits, you can't just 'switch out' aggregation methods, right?

No. Even if the reducer is an external contract, you can't switch them in and out if their interfaces are different. So external reducers provide limited modularity (you can swap mean with median, you can't swap mean with weighted mean--though there may be ways)

> ...we would have modified the array (essentially sorting it) the first time

You would have modified the array in memory, not storage (and writing back the modified array to storage would cost a lot). So the median will operate on the initial unsorted values each time.

packages/protocol/contracts/dapis/aggregators/MemorylessDApi.sol
```diff
@@ -0,0 +1,142 @@
...
+                templateIds[i],
+                requesterIndex,
+                designatedWallet,
+                fulfillAddress,
+                fulfillFunctionId,
+                "" // no additional parameters passed
+            );
+        }
+        // must empty existing responses since contract is "memoryless"
+        _emptyAnswerArray();
```
bbenligiray, line 87: You should also remove old `_pendingRequests` entries because there are incoming fulfillments for the old request

bbenligiray:
I agree that this is not a very good idea, let's have a call about this

packages/protocol/contracts/dapis/aggregators/MemorylessDApi.sol
```diff
@@ -0,0 +1,142 @@
...
+        templateIds = _templateIds;
+        requesterIndex = _requesterIndex;
+        designatedWallet = _designatedWallet;
+        minimumResponses = _minimumResponses;
+    }
+
+    function _emptyAnswerArray() private
+    {
+        for (uint i=0; i<_individualAnswers.length; i++) {
+            delete _individualAnswers[i];
```
bbenligiray, line 138: Won't even refund :(
https://eips.ethereum.org/EIPS/eip-3300

packages/protocol/contracts/dapis/aggregators/MemorylessDApi.sol
sassmilic: They just get ignored if the requestID doesn't match. But, yea, you're right, it should be deleted.

PR #302, opened 2021-04-26
Enable TS strict mode

andreogle:
Following on from some recent work in other projects, I figured we should enable TS strict mode on Airnode. I've implemented this for all packages _except the deployer package_ which has still has a few errors.

Once the deployer package has been fixed, we can enable `"strict": true` in the root tsconfigs


packages/adapter/src/request-building/cookies.ts
```diff
@@ -12,7 +12,7 @@ export function buildHeader(cookies: Parameters): CookieHeader | object {
 
   const keys = Object.keys(cookies);
 
-  const cookieValues = keys.reduce((values, key) => {
+  const cookieValues = keys.reduce((values: string[], key) => {
```
Siegrift, line 15: Nit: I'd type the initial accumulator instead: `[] as string[]` - rest gets infered

packages/adapter/tsconfig.build.json
```diff
@@ -2,8 +2,10 @@
   "extends": "../../tsconfig.build.json",
   "compilerOptions": {
     "baseUrl": "./",
+    "outDir": "./dist",
     "rootDir": "src",
-    "outDir": "./dist"
+    // Enable at root tsconfig.json once all packages have been fixed
```
Siegrift, line 7: Github should learn how to highlight this better :)

Edit: Actually, he understands tsconfig, but this is tsconfig.build so it's probably confused whether comments are allowed here.

packages/adapter/tsconfig.json
```diff
@@ -2,7 +2,9 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "baseUrl": "./",
-    "outDir": "dist"
+    "outDir": "dist",
+    // Enable at root tsconfig.json once all packages have been fixed
```
Siegrift, line 6: Maybe just add `TODO: Set to true once all of the packages have migrated to strict mode` to the root `tsconfig.json` with `strict: false` instead of a comment to every pckg?

packages/airnode-abi/src/decoding.ts
```diff
@@ -51,13 +51,13 @@ export function decode(encodedData: string): DecodedMap {
...
+    (type) => PARAMETER_SHORT_TYPES[type as ABIParameterTypeShort]
   );
 
   // The first `bytes32` is the type encoding
   const initialDecodedTypes: ABIParameterType[] = ['bytes32'];
 
-  const decodingTypes = fullParameterTypes.reduce((acc, type) => {
+  const decodingTypes = fullParameterTypes.reduce((acc: string[], type) => {
     // Each parameter is expected to have a `bytes32` name
     return [...acc, 'bytes32', type];
```
Siegrift, line 62: Use `'bytes32' as const` instead.

packages/node/src/evm/requests/api-calls.ts
```diff
@@ -80,12 +81,17 @@ export function applyParameters(request: ClientRequest<ApiCall>): LogsData<Clien
...
+  requests: ClientRequest<ApiCall>[];
+}
+
 export function updateFulfilledRequests(
   apiCalls: ClientRequest<ApiCall>[],
   fulfilledRequestIds: string[]
 ): LogsData<ClientRequest<ApiCall>[]> {
   const { logs, requests } = apiCalls.reduce(
-    (acc, apiCall) => {
+    (acc: UpdatedFulfilledRequests, apiCall) => {
```
Siegrift, line 94: Again, I slightly prefer when the initial value is typed instead (you also get a bit more type checking that way).

Siegrift (changes requested):

packages/adapter/tsconfig.build.json
andreogle: Yeah this is something I really don't like, but hoping to remove it soon

packages/adapter/tsconfig.json
andreogle: Yeah good idea

Siegrift (approved):
@andreogle Can we setup "up to date with master" check for this airnode repo as well?

LGTM, :shipit:

andreogle:
@Siegrift I've enabled it now 👍🏻

PR #303, opened 2021-04-26
Add RRP dAPI server

bbenligiray:
A dAPI server POC, essentially a contract that everyone can use to create dAPIs. It's highly opinionated regarding the aggregation because I couldn't come up with an actually modular way (for example how do you reduce all fulfillments with mean but every other fulfillment with median)


packages/protocol/contracts/dapis/CustomReducer.sol
```diff
@@ -0,0 +1,255 @@
+// SPDX-License-Identifier: MIT
```
bbenligiray, line 1: Just copy pasted stuff to see how large it was going to be

packages/protocol/contracts/dapis/rrpDapiServer.sol
```diff
@@ -0,0 +1,76 @@
...
+        airnodeRrp = AirnodeRrp(_airnodeRrp);
+    }
+
+    function makeDapiRequest(bytes32 dapiId)
+        external
+    {
+        DapiParameters storage dapiParameters = dapiIdToParameters[dapiId];
+        require(dapiParameters.reduceAddress == msg.sender);
+        for (uint i = 0; i < dapiParameters.templateIds.length; i++) {
+            bytes32 requestId = airnodeRrp.makeRequest(
```
bbenligiray, line 34: This should probably use `delegatecall` to allow dAPI requesters to use their own designated wallets

packages/protocol/contracts/dapis/rrpDapiServer.sol
```diff
@@ -0,0 +1,76 @@
...
+            dapiRequestIndexToResponses[dapiRequestIndex].push(decodedData);
+            uint256 responsesLength = dapiRequestIndexToResponses[dapiRequestIndex].length;
+            DapiParameters storage dapiParameters = dapiIdToParameters[dapiRequestIndexToDapiId[dapiRequestIndex]];
+            if (
+                responsesLength % 2 == 1
+                    && responsesLength >= dapiParameters.minResponsesToReduce
+                )
+            {
+                int256 reducedValue = reduceInPlace(dapiRequestIndexToResponses[dapiRequestIndex]);
+                dapiParameters.reduceAddress.call(  // solhint-disable-line
```
bbenligiray, line 70: dAPI callbacks :ok_hand:

packages/protocol/contracts/dapis/rrpDapiServer.sol
```diff
@@ -0,0 +1,76 @@
...
+
+    constructor(address _airnodeRrp) {
+        airnodeRrp = AirnodeRrp(_airnodeRrp);
+    }
+
+    function makeDapiRequest(bytes32 dapiId)
+        external
+    {
+        DapiParameters storage dapiParameters = dapiIdToParameters[dapiId];
+        require(dapiParameters.reduceAddress == msg.sender);
```
sassmilic, line 32: So this contract is called by the dAPI/aggregator contract?

sassmilic:
So, just so I understand: `rrpDapiServer.sol` functions are called by the dAPI/aggregator when it wants to make requests?

This means the dAPI/aggregator is in charge of:
- the implementation of the aggregation function
- storing the result of the aggregation

Anything else?

packages/protocol/contracts/dapis/rrpDapiServer.sol
bbenligiray: This contract is called by a price data feed contract or a prediction market for example, so essentially like an Airnode client contract

bbenligiray:
`rrpDapiServer.sol` is called by the client. Implementation of the aggregation is implemented inside `rrpDapiServer.sol` under `CustomReducer.sol`. `rrpDapiServer.sol` only keeps individual responses (and will overwrite them after a period of time), the client is responsible with storing the result of the aggregation.

sassmilic:
@bbenligiray check my latest commit. It's some pretty simple modulo arithmetic + an added mapping to handle limited storage/memory. If you have any questions, lemme know.

packages/protocol/contracts/RequesterStore.sol
```diff
@@ -25,6 +25,14 @@ contract RequesterStore is IRequesterStore {
         _;
     }
 
+    function isEndorsed(uint256 requesterIndex, address addr)
```
bbenligiray, line 28: This is not needed because `requesterIndexToClientAddressToEndorsementStatus` is `public`

packages/protocol/contracts/dapis/rrpDapi.sol
```diff
@@ -2,13 +2,91 @@
...
 
 // A stub for an RRP dAPI contract
 // When the user wants to deploy a dAPI contract, they inherit the specific
 // reducer contract such as:
 //       contract myRrpDapi is rrpDapi, Median {
 //           // Any specific implementation required
 //       }
-abstract contract rrpDapi is Reducer {
-    // TODO: Implement
+abstract contract RrpDapi is Reducer {
```
bbenligiray, line 13: I initially created this `RrpDapi` contract, then implemented `RrpDapiServer` instead. We don't need a proxy contract to store the reduced answers in storage because the user may not need that. If they do, they implement this in their contract. This overwriting scheme should be implemented at `RrpDapiServer` to overwrite individual responses, as that is the main source of storage gas costs, not the reduced response.

packages/protocol/contracts/dapis/rrpDapiServer.sol
```diff
@@ -0,0 +1,117 @@
...
+        uint256 requesterIndex,
+        bytes32[] calldata templateIds,
+        address[] calldata designatedWallets,
+        address reduceAddress,
+        bytes4 reduceFunctionId
+    )
+        external
+        returns (uint256 dapiId)
+    {
+      dapiIdToParameters[nextDapiID] = DapiParameters(
```
bbenligiray, line 43: I wonder how expensive it would be to compute `dapiId` by taking a has of its parameters. Otherwise, you will have to make a transaction to register a dAPI and then wait enough confirmations to know that your dAPI index will not change.

packages/protocol/contracts/dapis/rrpDapiServer.sol
```diff
@@ -0,0 +1,117 @@
...
+        uint256 requesterIndex;
+        bytes32[] templateIds;
+        address[] designatedWallets;
+        address reduceAddress;
+        bytes4 reduceFunctionId;
+        }
+    
+    AirnodeRrp public airnodeRrp;
+    mapping(uint256 => DapiParameters) private dapiIdToParameters;
+    uint256 public nextDapiID = 1;
```
bbenligiray, line 20: I tend to refer to `bytes32` identifiers as "ID" and number-type identifiers as "index"

packages/protocol/contracts/dapis/rrpDapi.sol
```diff
@@ -2,13 +2,91 @@
...
+
+    RrpDapiServer public rrpDapiServer;
+
+    uint256 public immutable maxAnswersInStorage;
```
bbenligiray, line 17: This is not practical. If I'm a prediction market, I don't want any kind of upper limit to my active requests if I can avoid it. This becomes for of a problem when you're overwriting individual responses in `RrpDapiServer` because then you would be sharing this limit with everyone else.

packages/protocol/contracts/dapis/rrpDapi.sol
```diff
@@ -2,13 +2,91 @@
...
+      // +1 is added to the modulo to prevent 0 values
+      uint256 answerMod = nextAnswerId % maxAnswersInStorage + 1;
+      
+      // delete stale answer in storage
+      delete requestIndexToAnswerIdMod[answerIdModToRequestIndex[answerMod]];
+      
+      requestIndexToAnswerIdMod[requestIndex] = answerMod;
+      answerIdModToRequestIndex[answerMod] = requestIndex;
+      
+      answerId = nextAnswerId++;
```
bbenligiray, line 78: I'm having difficulty understanding what's happening here because names are confusing

packages/protocol/contracts/dapis/rrpDapi.sol
```diff
@@ -2,13 +2,91 @@
...
+      // reverts if requester is not endorsed
+      uint256 requestIndex = rrpDapiServer.makeDapiRequest(dapiId, requesterIndex);
+
+      // +1 is added to the modulo to prevent 0 values
+      uint256 answerMod = nextAnswerId % maxAnswersInStorage + 1;
+      
+      // delete stale answer in storage
+      delete requestIndexToAnswerIdMod[answerIdModToRequestIndex[answerMod]];
+      
+      requestIndexToAnswerIdMod[requestIndex] = answerMod;
```
bbenligiray, line 75: This writes to an unwritten storage location for each request. This is in general is not more efficient than keeping the answers in an infinitely growing array.

packages/protocol/contracts/RequesterStore.sol
sassmilic: oh i thought it wasn't for some reason ...

packages/protocol/contracts/dapis/rrpDapiServer.sol
sassmilic: I thought it would be too expensive without any upside, that's why I changed it.

> to know that your dAPI index will not change

Why would it change?

packages/protocol/contracts/dapis/rrpDapi.sol
sassmilic: Why not? You can estimate how many requests you'll make in, say, a day, and then use that to create a limit (if desired).

packages/protocol/contracts/dapis/rrpDapi.sol
sassmilic: I think the only confusing thing is `answerIdModToRequestIndex` which is the _modulo_ of the answer ID, which maps to the corresponding request ID.

packages/protocol/contracts/dapis/rrpDapi.sol
sassmilic: Even if you delete the previous one? Are you 100% sure?

sassmilic: If so, I can still implement a constant storage, it's similar to this algorithm: https://www.geeksforgeeks.org/how-to-delete-an-element-from-the-set-by-passing-its-value-in-c/

packages/protocol/contracts/dapis/rrpDapi.sol
sassmilic: Also, there is the possibility for an expiry date implementation. But I think it has similar usefulness but is more complicated to implement.

packages/protocol/contracts/RequesterStore.sol
bbenligiray: I think it needs to be added to the interface as an external method

packages/protocol/contracts/dapis/rrpDapiServer.sol
bbenligiray: The transactions may be reordered if multiple dAPIs are being created in a short period. I'm not sure how much more expensive it would be, probably not that much.

packages/protocol/contracts/dapis/rrpDapi.sol
bbenligiray: What if you need to exceed the limit, you close up shop? That's not nice. I think expiry date is required and it's complexity would be justified if it was being used to overwrite individual reponses and not only the reduced response.

packages/protocol/contracts/dapis/rrpDapi.sol
bbenligiray: What can I say I'm easily confused ¯\_(ツ)_/¯

packages/protocol/contracts/dapis/rrpDapi.sol
bbenligiray: `requestIndex` is constantly being incremented so yeah. The main point is that overwriting the reduced answer is not worth it due to these overheads, this needs to be done for individual responses.

bbenligiray: Also this should be implemented with the assumption that `delete` won't refund gas due to EIP-3300

sassmilic: I'm not sure I understand your point about overhead? Writing to allocated, fixed-sized storage will be more gas efficient then an ever-increasing storage array.

packages/protocol/contracts/dapis/rrpDapiServer.sol
```diff
@@ -0,0 +1,117 @@
...
+        templateIds,
+        designatedWallets,
+        reduceAddress,
+        reduceFunctionId
+      );
+      dapiId = nextDapiID++;
+    }
+
+
+    function makeDapiRequest(uint256 dapiId, uint256 requesterIndex)
```
bbenligiray, line 56: This should include a `bytes parameters` argument that gets used in all `makeRequest`s

sassmilic: yea, i think it was empty as a TODO

sassmilic:
@bbenligiray Okay, I've thought about this for several hours (long, deep think).

There's no way to overwrite request indices; i.e. `dapiRequestIndexToResponses` and `dapiRequestIndexToDapiId` will grow indefinitely.

(Actually: to be technical, there is, but it requires an additional array in storage and then reading from that array in ways that are most likely more expensive than simple writing a value to storage in the first place).

**However**: if the dAPI contracts were to implement the server functionality on their own (or inherit from an abstract contract that does that) then you would considerably save gas on storage because all mappings would be limited in size.

This is mainly due to the fact that the current answer ID and what's referred to as the `dapiRequestIndex` in the server contract, would be the exact same number. I would be able to utilize the same modulo arithmetic technique for maintaining fixed-sized storage (by way of re-writing to a fixed-size array).

Further, the need for `dapiRequestIndexToDapiId` would be eliminated.

---

So, to summarize, under this proposed change (i.e. central dapi server contract -> dapis implementing/inheriting server functionality), you would:
- eliminate `dapiIdToParameters`
- eliminate `dapiRequestIndexToDapiId`
- change `requestIdToDapiRequestIndex` to fixed-sized / overwrite-able storage
- change `dapiRequestIndexToResponses` to fixed-sized / overwrite-able storage

If my explanation doesn't make sense, I can explain in a call.

packages/protocol/contracts/dapis/rrpDapiServer.sol
sassmilic: Is it reasonable to assume all templates would use the same parameters?

packages/protocol/contracts/dapis/rrpDapi.sol
bbenligiray: > Writing to allocated, fixed-sized storage will be more gas efficient then an ever-increasing storage array.

Only if you don't have to write additional stuff (timestamp to check expiry for example) to storage to make sure that you're not overwriting answers you shouldn't overwrite. If it's a data feed and you're overwriting an `int256` then it will be very gas efficient.

packages/protocol/contracts/dapis/rrpDapiServer.sol
bbenligiray: You are asking if it should be `bytes` or `bytes[]` right? Even if different endpoints may require differently formatted parameters I don't think it's feasible to specify parameters for each request. Then, the Airnodes should be expected to accept request-time parameters in the same format (so `bytes`).

bbenligiray:
I can do a call tomorrow. My initial thought is that the UX hit of inheriting the server contract is not worth it to be able to overwrite the reduction result.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "license": "MIT",
   "engines": {
-    "node": "^12.13.1"
+    "node": "^14.16.0"
```
bbenligiray, line 6: Run these to do this change on your machine
```
nvm install 14.16.1
nvm alias default 14.16.1
```

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.4;
+
+import "../AirnodeRrp.sol";
+import "./reducers/MeanMedianHybrid.sol";
```
bbenligiray, line 5: I removed `CustomReducer.sol` because reducers will only be deployed once so no need to optimize contract size

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
...
+
+contract RrpDapiServer is MeanMedianHybrid {
+    struct Dapi {
+        uint256 noResponsesToReduce;
```
bbenligiray, line 9: dAPI reduces responses only once, when there are this many responses

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
...
+        int256 toleranceInPercentages;
+        uint256 requesterIndex;
+        bytes32[] templateIds;
+        address[] designatedWallets;
+        address reduceAddress;
+        bytes4 reduceFunctionId;
+        mapping(uint256 => int256[]) requestIndexToResponses;
+        mapping(uint256 => uint256) requestIndexToResponsesLength;
```
bbenligiray, line 17: Before, we were using `.push()`, using `.length` for the array length and calling `delete` when we no longer needed it. Now, we're keeping a separate length variable for each dAPI request and incrementing/resetting it manually.

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
...
+        uint64 nextRequestIndex;
+        uint64 requestIndexResetCount;
+        address requestIndexResetter;
```
bbenligiray, line 20: This is the "owner" of the dAPI, authorized to reset its request indices, which invalidates all previous requests. That didn't sound right when I wrote it.

bbenligiray: Yeah this will need a rework

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
...
+        }
+    
+    struct DapiRequestIdentifiers {
+        bytes16 dapiId;
+        uint64 dapiRequestIndex;
+        uint64 dapiRequestIndexResetCountAtRequestTime;
+        }
```
bbenligiray, line 27: This was what I was referring to by optimizing storage by packing them into structs. The `Dapi` struct above has a lot of room for optimization for example.

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
...
+      dapiId = bytes16(keccak256(abi.encodePacked(
+          noResponsesToReduce,
+          toleranceInPercentages,
+          requesterIndex,
+          templateIds,
+          designatedWallets,
+          reduceAddress,
+          reduceFunctionId,
+          requestIndexResetter
+          )));
```
bbenligiray, line 68: I'm pretty sure this is very cheap. Note that I'm using half of the hash for identification because I want to pack into a 32 byte struct.

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
...
+    function resetDapiRequestIndex(bytes16 dapiId)
+        external
+    {
+        Dapi storage dapi = dapis[dapiId];
+        require(
+            msg.sender == dapi.requestIndexResetter,
+            "Caller not resetter"
+            );
+        dapi.nextRequestIndex = 1;
+        dapi.requestIndexResetCount++;
```
bbenligiray, line 89: This invalidates all dAPI requests made when `requestIndexResetCount` was lower

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
...
+        )
+        external
+        returns (uint64 currDapiRequestIndex)
+    {
+        Dapi storage dapi = dapis[dapiId];
+        currDapiRequestIndex = dapi.nextRequestIndex++;
+        dapi.requestIndexToResponsesLength[currDapiRequestIndex] = 0;
+        if (dapi.requestIndexToResponses[currDapiRequestIndex].length == 0) {
+            dapi.requestIndexToResponses[currDapiRequestIndex] = new int256[](dapi.noResponsesToReduce);
+        }
```
bbenligiray, line 104: We allocate the memory once, then reuse it with each dAPI request index reset

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
...
+        require(
+            airnodeRrp.requesterIndexToClientAddressToEndorsementStatus(dapi.requesterIndex, msg.sender),
+            "Caller not endorsed"
+            );
+        for (uint i = 0; i < dapi.templateIds.length; i++) {
+            (bool success, bytes memory returnedData) = address(airnodeRrp).delegatecall(abi.encodeWithSelector( // solhint-disable-line
```
bbenligiray, line 110: Muh delegatecall

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
...
+                this.fulfill.selector,
+                parameters
+                ));
+            require(success, "Request unsuccessful"); // This will never happen if AirnodeRrp is valid
+            bytes32 requestId = abi.decode(returnedData, (bytes32));
+            airnodeRequestIdToDapiRequestIdentifiers[requestId] = DapiRequestIdentifiers({
+                dapiId: dapiId,
+                dapiRequestIndex: uint64(currDapiRequestIndex),
+                dapiRequestIndexResetCountAtRequestTime: dapi.requestIndexResetCount
+                });
```
bbenligiray, line 125: This is where we're storing gas by packing the struct

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
...
+        require(
+            dapiRequestIdentifiers.dapiRequestIndex != 0,
+            "Request ID invalid"
+            );
+        if (statusCode == 0) {
+            Dapi storage dapi = dapis[dapiRequestIdentifiers.dapiId];
+            require(
+                dapiRequestIdentifiers.dapiRequestIndexResetCountAtRequestTime == dapi.requestIndexResetCount,
+                "Request stale"
+                );
```
bbenligiray, line 150: dAPI owner has reset the dAPI request indices after this dAPI request has been made

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
...
+            uint256 responsesLength = dapi.requestIndexToResponsesLength[dapiRequestIdentifiers.dapiRequestIndex];
+            
+            if (responsesLength < dapi.noResponsesToReduce) {
+                int256 decodedData = abi.decode(data, (int256));
+                dapi.requestIndexToResponses[dapiRequestIdentifiers.dapiRequestIndex][responsesLength] = decodedData;
+                dapi.requestIndexToResponsesLength[dapiRequestIdentifiers.dapiRequestIndex] = ++responsesLength;
```
bbenligiray, line 156: This is what I mean by "manually keeping response array length"

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
...
+                    else {
+                        reducedValue = computeMeanMedianHybridInPlace(dapi.requestIndexToResponses[dapiRequestIdentifiers.dapiRequestIndex], dapi.toleranceInPercentages);
+                    }
+                    (bool success, ) = dapi.reduceAddress.call(abi.encodeWithSelector( // solhint-disable-line
+                        dapi.reduceFunctionId,
+                        dapiRequestIdentifiers.dapiId,
+                        dapiRequestIdentifiers.dapiRequestIndex,
+                        reducedValue
+                        ));
+                    require(success, "Reduce callback unsuccessful");
```
bbenligiray, line 172: I'm not sure if we should `require()` this. Probably.

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
```
bbenligiray, line 1: ezpz lemon squeezy data feed

packages/protocol/contracts/dapis/reducers/Mean.sol
```diff
@@ -1,22 +1,18 @@
...
-import "./Reducer.sol";
-
 /// @title Mean reduction contract
 /// @notice Mean reduction result can be skewed even with a single misreport.
 /// Therefore, it should either be used in special circumstances or as a part
 /// of a more complex reduction method such as MeanMedianHybrid.sol.
 /// @dev Note that there is a potential for overflow while summing large enough
 /// values. If that is a probability, use Median.sol instead.
-contract Mean is Reducer {
+contract Mean {
```
bbenligiray, line 10: Decided against implementing these as a general `Reducer` contract

packages/protocol/contracts/dapis/RrpDapiServer.sol
sassmilic: nit: when I first read this variable name, i thought it said "no responses to reduce" and that it was a boolean. why not opt for something like `requiredMin`

packages/protocol/contracts/dapis/RrpDapiServer.sol
sassmilic: Does it have to be as granular as `requestIndex`?

Won't dAPIs always require the same number of responses for aggregation? And this is a parameter in the `Dapi` struct.

packages/protocol/contracts/dapis/RrpDapiServer.sol
sassmilic: Yea, I was thinking of discussing this with you. Figuring out things like: max number of dAPIs, max number of requests, etc. In order to potentially change ints/uints to smaller types.

packages/protocol/contracts/dapis/RrpDapiServer.sol
sassmilic: What do you mean by "I'm using half of the hash for identification"?

packages/protocol/contracts/dapis/RrpDapiServer.sol
sassmilic: I'm not sure what you mean ... like, what happens to pending requests when the index is reset?

packages/protocol/contracts/dapis/RrpDapiServer.sol
sassmilic: You could instead track the last index before reset, then you could reconstruct the index of the previous requests (for a short period of time), if that makes sense.

packages/protocol/contracts/dapis/RrpDapiServer.sol
sassmilic: nice

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
sassmilic: yea, i like this. it makes sense to have the requester reset storage, cuz they know when requests get made/fulfilled

sassmilic: what happens if `makeDapiRequest` is called again before the last request gets processed? it just gets "overwritten", and that doesn't matter since the price feed is concerned with the most up-to-date responses?

packages/protocol/contracts/dapis/RrpDapiServer.sol
bbenligiray: Yeah I know you don't like no- as in "number of". I actually scratched my head a bit looking for an alternative. `requiredMin` is definitely worse. I think we can keep it as placeholder and replace it when we come up with a better name.

packages/protocol/contracts/dapis/RrpDapiServer.sol
bbenligiray: So if `noResponsesToReduce` for a dAPI is `7`, `requestIndexToResponses[dapiId][*].length` will always be `7`. Then, we need a separate counter to keep how many responses we have received because we no longer can use the response length as the counter as we did before. Refer to how it is used in `fulfill()`.

packages/protocol/contracts/dapis/RrpDapiServer.sol
bbenligiray: Any fulfillments by Airnodes will revert. I thought about this a bit and maybe it's fine. Do a full review and let's have a call about it for real this time.

packages/protocol/contracts/dapis/RrpDapiServer.sol
bbenligiray: Can be left to last imo

packages/protocol/contracts/dapis/RrpDapiServer.sol
bbenligiray: `keccak256()` returns a `bytes32`. I typecast it as `bytes16`, essentially throwing half of it away. Now it is a lot more likely for dAPI IDs to collide, but still unlikely enough.

packages/protocol/contracts/dapis/RrpDapiServer.sol
bbenligiray: Would need to see implementation. I couldn't come up with a feasible alternative.

packages/protocol/contracts/dapis/RrpDapiServer.sol
bbenligiray: https://github.com/api3dao/airnode/pull/303/files#r624727056

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
bbenligiray: The Airnode fulfillments to the previous request will get reverted, meaning that we will not receive a reduce callback from them.
> it just gets "overwritten", and that doesn't matter since the price feed is concerned with the most up-to-date responses?

Yes

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.4;
+
+import "../AirnodeRrp.sol";
+import "./reducers/MeanMedianHybrid.sol";
+
+contract RrpDapiServer is MeanMedianHybrid {
+    struct Dapi {
+        uint256 noResponsesToReduce;
+        int256 toleranceInPercentages;
```
sassmilic, line 10: This can be negative?

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
...
+        bytes16 dapiId,
+        bytes calldata parameters
+        )
+        external
+        returns (uint64 currDapiRequestIndex)
+    {
+        Dapi storage dapi = dapis[dapiId];
+        currDapiRequestIndex = dapi.nextRequestIndex++;
+        dapi.requestIndexToResponsesLength[currDapiRequestIndex] = 0;
+        if (dapi.requestIndexToResponses[currDapiRequestIndex].length == 0) {
```
sassmilic, line 102: When would this not be 0?

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -0,0 +1,178 @@
...
+            dapi.requestIndexToResponses[currDapiRequestIndex] = new int256[](dapi.noResponsesToReduce);
```
sassmilic, line 103: Do you have to create a new array? Can't you just reinsert into the same array?

packages/protocol/contracts/dapis/RrpDapi.sol
```diff
@@ -0,0 +1,85 @@
+// SPDX-License-Identifier: MIT
```
sassmilic, line 1: I think this file needs to be deleted.

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
```diff
@@ -0,0 +1,55 @@
...
+// A sample price data feed implemented as a dAPI client
+contract SamplePriceDataFeed {
+    RrpDapiServer public rrpDapiServer;
+    bytes16 public latestDapiId;
+    int256 public latestAnswer;
+
+    constructor (address _rrpDapiServer)
+    {
+        rrpDapiServer = RrpDapiServer(_rrpDapiServer);
+    }
```
sassmilic, line 15: I believe `registerDapi` should go in here as well.

packages/protocol/contracts/dapis/RrpDapiServer.sol
sassmilic: Oh, wait, that is what's happening right? You allocate an array of length `dapi.noResponsesToReduce` and then just rewrite to it?

sassmilic (commented):
Looks pretty good to me! I left a couple of comments/questions.

packages/protocol/contracts/dapis/RrpDapiServer.sol
bbenligiray: No, fixed.

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
bbenligiray: Let's cover that stuff in following PRs

packages/protocol/contracts/dapis/RrpDapiServer.sol
bbenligiray: Yep

packages/protocol/contracts/dapis/RrpDapiServer.sol
bbenligiray: If `currDapiRequestIndex` is being used for the first time so the memory is not allocated yet

PR #304, opened 2021-04-26
Add TypeChain integration to node package

Siegrift:
Review commit by commit.

There are some TODOs in the code which I am not sure how to resolve. Specifically:
- Make sure that `endpointId` is not null when passed to contract call. I think we should create a type with `endpointId` non nullable after it's been merged with template. Resolved in https://api3workspace.slack.com/archives/C01N6001YCE/p1619425785068000 and the properties have been made required.
- Better type `event-logs.ts`, but I am not sure what log does it return - This has been implemented in last 3 commits.


packages/node/src/evm/verification/request-verification.test.ts
```diff
@@ -41,19 +41,6 @@ describe('verifyDesignatedWallets', () => {
     }
   });
 
-  it('ignores API calls that have no requester index', () => {
```
Siegrift, line 44: This is duplicated (same test name with same code) so I removed one copy.
@andreogle I see many tests with the same name in code, maybe there was some bad rebase along the way?

packages/node/src/evm/verification/request-verification.ts
```diff
@@ -16,19 +16,6 @@ export function verifyDesignatedWallets<T>(
       return [[log], request];
     }
 
-    // We can't validate the wallet if there is no requester index. There should always be one
-    // at this point but just in case there isn't.
-    if (!request.requesterIndex) {
```
Siegrift, line 21: Since the `requesterIndex` is now non-nullable this check seems unnecessary so I removed it.

packages/node/src/evm/fulfillments/api-calls.test.ts
```diff
@@ -227,11 +230,11 @@ describe('submitApiCall', () => {
 
     it('returns an error if everything fails', async () => {
       const provider = new ethers.providers.JsonRpcProvider();
-      const contract = new ethers.Contract('address', ['ABI']);
+      const contract = createAirnodeRrpFake();
       (contract.callStatic.fulfill as jest.Mock).mockRejectedValueOnce(new Error('Server did not respond'));
```
Siegrift, line 234: @andreogle Any specific reason why mocking the values through the contract variable?
You could just as well do `staticFulfillMock.mockRejectedValueOnce(...)`

packages/node/src/evm/verification/request-verification.test.ts
andreogle: Could be yeah, or maybe just bad-copy pasting

packages/node/src/evm/fulfillments/api-calls.test.ts
andreogle: No reason. I wasn't really sure how to mock ethers for a while and there are a few leftovers from that

packages/node/src/evm/verification/request-verification.test.ts
Siegrift: I'll check that in another PR

packages/node/src/evm/requests/events.ts
```diff
@@ -15,18 +23,26 @@ export const WITHDRAWAL_REQUEST_TOPICS = [airnodeRrpTopics.WithdrawalRequested];
...
 }
 
-export function isWithdrawalFulfillment(log: ethers.utils.LogDescription) {
-  return WITHDRAWAL_FULFILLED_TOPICS.includes(log.topic);
+export function isApiCallFulfillment(log: EVMEventLog): log is EVMRequestFulfilledLog {
+  return API_CALL_FULFILLED_TOPICS.includes(log.parsedLog.topic);
+}
+
+export function isWithdrawalRequest(log: EVMEventLog): log is EVMWithdrawalRequestLog {
+  return WITHDRAWAL_REQUEST_TOPICS.includes(log.parsedLog.topic);
```
Siegrift, line 43: @andreogle This is a bit confusing to me. You check for the request type using the event topics and in tests you are checking this e.g. by `log.name === 'ClientRequestCreated'`. Is there a specific reason why not to differentiate between the type of logs using only the name?

We could simplify the code by removing https://github.com/api3dao/airnode/blob/master/packages/node/src/evm/contracts/airnodeRrp.ts#L21

packages/node/src/types.ts
```diff
@@ -209,14 +212,60 @@ export interface WorkerResponse {
...
+
+export interface EVMWithdrawalRequestLog extends EVMEventLogMetadata {
+  parsedLog: AirnodeLogDescription<AirnodeRrpLog<'WithdrawalRequested'>>;
+}
+
+export interface EVMWithdrawalFulfilledLog extends EVMEventLogMetadata {
+  parsedLog: AirnodeLogDescription<AirnodeRrpLog<'WithdrawalFulfilled'>>;
+}
+
+export type EVMEventLog =
```
Siegrift, line 264: The idea is to have the events which are used in node package typed. This is useful for many things - autocomplete, refactorings, static guarantees imposed by TS

packages/node/test/fixtures/evm/event-logs.test.ts
```diff
@@ -0,0 +1,36 @@
+import { buildClientRequest } from './event-logs';
```
Siegrift: This was a dev-only file and will be removed later.

packages/protocol/package.json
```diff
@@ -8,7 +8,7 @@
     "build:contracts": "hardhat compile",
     "build:tsc": "rimraf -rf ./dist && yarn build:contract-dts && yarn build:fix-contract-dts && yarn build:copy-contract-dts && tsc -p tsconfig.build.json",
     "build:contract-dts": "yarn typechain --target ethers-v5 --outDir ./src/contracts './artifacts/contracts/**/!(*.dbg).json'",
-    "build:fix-contract-dts": "grep -rl 'extends ethers.utils.Interface' src/contracts | xargs sed -i 's/ethers\\.utils\\.Interface/Interface/g' && grep -rl 'Result ' src/contracts | xargs sed -i 's/Result /Result, Interface /g'",
+    "build:fix-contract-dts": "echo \"export { TypedEventFilter } from './commons'\n\" >> src/contracts/index.ts && grep -rl 'extends ethers.utils.Interface' src/contracts | xargs sed -i 's/ethers\\.utils\\.Interface/Interface/g' && grep -rl 'Result ' src/contracts | xargs sed -i 's/Result /Result, Interface /g'",
```
Siegrift, line 11: Unfortunately, I had to append an export to make the TS inference work with typechain.
I'd like TypeChain to make this more developer friendly and remove this low-level handling afterwards.

packages/node/src/evm/verification/request-verification.test.ts
Siegrift: Actually, this is my bad - this test is NOT duplicated and shouldn't be removed. I'll revert back.

packages/node/src/evm/initialization.test.ts
```diff
@@ -445,7 +443,7 @@ describe('setAirnodeParameters', () => {
       expect(setAirnodeParametersMock).toHaveBeenCalledTimes(2);
     });
 
-    it('does not warn if the Airnode parameters match', async () => {
+    it.only('does not warn if the Airnode parameters match', async () => {
```
andreogle, line 446:
```suggestion
    it('does not warn if the Airnode parameters match', async () => {
```

packages/node/src/evm/requests/api-calls.ts
```diff
@@ -1,56 +1,57 @@
...
       return 'full';
     // This should never be reached
     default:
       throw new Error(`Unknown topic:${topic} during API call initialization`);
   }
 }
 
-export function initialize(logWithMetadata: EVMEventLogWithMetadata): ClientRequest<ApiCall> {
-  const { parsedLog } = logWithMetadata;
+export function initialize(log: EVMRequestCreatedLog): ClientRequest<ApiCall> {
```
andreogle, line 30: I like this name more 👍🏻

packages/node/src/evm/requests/api-calls.ts
```diff
@@ -1,56 +1,57 @@
...
     chainId: parsedLog.args.chainId.toString(),
     clientAddress: parsedLog.args.clientAddress,
-    designatedWallet: parsedLog.args.designatedWallet || null,
+    designatedWallet: parsedLog.args.designatedWallet,
     encodedParameters: parsedLog.args.parameters,
     id: parsedLog.args.requestId,
-    endpointId: parsedLog.args.endpointId || null,
-    fulfillAddress: parsedLog.args.fulfillAddress || null,
-    fulfillFunctionId: parsedLog.args.fulfillFunctionId || null,
+    endpointId: events.isFullApiRequest(log) ? log.parsedLog.args.endpointId : null,
```
andreogle, line 40: Do we need to check the event type like this?

packages/node/src/evm/requests/events.ts
andreogle: Good point. I'm fine with using just the name

packages/node/src/evm/templates/template-fetching.test.ts
```diff
@@ -223,10 +214,11 @@ describe('fetch (templates)', () => {
 });
 
 describe('fetchTemplate', () => {
-  let airnodeRrp: ethers.Contract;
+  let airnodeRrp: AirnodeRrp;
 
   beforeEach(() => {
-    airnodeRrp = new ethers.Contract('address', ['ABI']);
+    // mock AirnodeRrp contract
```
andreogle, line 220: and here

packages/protocol/src/index.ts
```diff
@@ -1,3 +1,15 @@
+/**
```
andreogle, line 1: Thanks for this explanation 👍🏻

andreogle (changes requested):
A few minor comments, otherwise looks great!

packages/node/src/evm/requests/events.ts
Siegrift: Actually maybe it's because theoretically you can have log with same name but different params? I tried that it remox in the contract compiled without errors. 

Anyway, that is probably antipattern and we will never use it in airnode?
I definitely do'n't want to fix that in this PR

packages/node/src/evm/authorization/authorization-fetching.test.ts
```diff
@@ -156,10 +152,11 @@ describe('fetch (authorizations)', () => {
 
 describe('fetchAuthorizationStatus', () => {
   const airnodeId = '0xairnodeId';
-  let airnodeRrp: ethers.Contract;
+  let airnodeRrp: AirnodeRrp;
 
   beforeEach(() => {
-    airnodeRrp = new ethers.Contract('address', ['ABI']);
+    // create mock of AirnodeRrp contract
```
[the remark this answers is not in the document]
Siegrift: :zero:

packages/node/src/evm/requests/api-calls.ts
Siegrift: `endpointId` is only defined in full request. This was the cleanest way I found to make this readable and satisfy TS static checkings.

packages/node/src/evm/requests/events.ts
Siegrift: Tracked in https://github.com/api3dao/airnode/issues/331

andreogle (approved):
🚀

amarthadan (approved):

PR #326, opened 2021-05-02
Simplify tests

Siegrift:
Took a look at the tests and:
- Corrected test name to `fetches group authorization status if it can be fetched`
- Remove duplicated `does nothing if unable to verify or set Airnode parameters` test
- Tests which are copy pasted can be written in a for each pattern.


andreogle (approved):
Nice one 👍🏻

Issue #286, opened 2021-04-13
Add examples to admin CLI


Siegrift:
For now we have examples of commands in README.

It would be nice if they were printed out from the CLI as well. We can use: http://yargs.js.org/docs/#api-reference-examplecmd1-desc1-cmd2-desc2

jeebster:
@Siegrift addressed in #327

Siegrift:
Unfortunately, it is not supported by https://yargs.js.org/docs/ 
See more in: https://github.com/api3dao/airnode/pull/327

PR #327, opened 2021-05-02
Admin CLI Examples

jeebster:
Ref: #286 

NOTE: [set-airnode-parameters](https://github.com/api3dao/airnode/tree/master/packages/admin#set-airnode-parameters) omitted as the `README` notes it should not be invoked

### DIFF
--- a/packages/admin/src/cli.ts
+++ b/packages/admin/src/cli.ts
@@ -4,6 +4,10 @@ import * as evm from './evm';
 import * as admin from '.';
 import { exit } from 'process';
 
+type CmdExample = {
+  [key: string]: [string, (string | undefined)?];
+};
+
 const COMMON_COMMAND_ARGUMENTS = {
   airnodeRrpCommands: {
     providerUrl: {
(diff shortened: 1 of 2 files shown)


packages/admin/src/cli.ts
```diff
@@ -343,6 +467,20 @@ yargs
       console.log(`Endpoint ID: ${endpointId}`);
     }
   )
+  .example([
```
Siegrift, line 470: I'd prefer if the example would show only on corresponding subcommand (e.g. only show `createRequester` when running the `create-requester` subcommand)

Siegrift (changes requested):

packages/admin/src/cli.ts
Siegrift: Mhh, doesn't simple `yarn test:e2e-admin -u` work?

>I cannot regenerate the failing snapshot or build the package; lots of typescript errors and configuration might need to change relative to how other packages in the repository are being imported.
Did you run `yarn bootstrap` in the root dir?

jeebster: After invoking `yarn bootstrap` and running the e2e admin tests:

```
$ cd packages/admin && yarn run test:e2e -u
$ jest --selectProjects e2e -u
Running one project: e2e
 FAIL   e2e  test/e2e/sdk.feature.ts
  ● Test suite failed to run

    test/e2e/sdk.feature.ts:1:35 - error TS2307: Cannot find module '@airnode/protocol' or its corresponding type declarations.

    1 import { AirnodeRrpFactory } from '@airnode/protocol';
                                        ~~~~~~~~~~~~~~~~~~~

 FAIL   e2e  test/e2e/cli.feature.ts
  ● Test suite failed to run

    test/e2e/cli.feature.ts:1:47 - error TS2307: Cannot find module '@airnode/protocol' or its corresponding type declarations.

    1 import { AirnodeRrp, AirnodeRrpFactory } from '@airnode/protocol';
                                                    ~~~~~~~~~~~~~~~~~~~

Test Suites: 2 failed, 2 total
Tests:       0 total
Snapshots:   0 total
Time:        2.354 s, estimated 12 s
Ran all test suites.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Yesterday I ran some bypasses around those imports and ran into lots of typescript compilation errors so I disabled strict mode locally as I noticed a few other packages do. I'll take a look into this again later

Siegrift: You also need to use `yarn build`. Those packages are internal airnode pckgs and need to be built first.

jeebster: @Siegrift ah I missed the root `build` command to build the internal packages. Thank you! 🙏

jeebster: @Siegrift I spent some time thumbing through the [Yargs documentation](https://yargs.js.org/docs/) and I do not think the library provides fine-grained control of examples relative to sub-commands. The [example() implementation](https://github.com/yargs/yargs/blob/395bb67749787d269cabe80ffc3133c2f6958aeb/lib/usage.ts) simply collects and prints out all examples without any level of hierarchy.

If you think there's possibility to achieve what you're describing, any direction towards documentation would be helpful - otherwise I think these are the constraints with the existing library

Siegrift: That is a bit disappointing. That seems to me like a basic feature, but I didn't find such option as well.
I think it's better to leave this as is - so many examples would just clutter the `--help` output of CLI.

jeebster: Agreed - a bit of a bummer that's not supported and the `--help` output does seem pretty cluttered with this addition. Feel free to close this issue/PR accordingly

Siegrift: Anyway, thanks for trying this out. Even closing some doors is important :).

Siegrift:
Closing because https://yargs.js.org/docs/ doesn't support example per command (only supports global examples).

PR #329, opened 2021-05-03
Update deps

Siegrift:
The ethers library must be explicetely pinned down such that npm dependency resolution won't load multiple different versions of (ethers) library. That caused problems with TypeChain code generation, because some part of the generated code was using one version and the other part of code a different one.

### DIFF
--- a/packages/protocol/package.json
+++ b/packages/protocol/package.json
@@ -30,11 +30,14 @@
     "@typechain/ethers-v5": "^6.0.5",
     "chai": "^4.2.0",
     "copyfiles": "^2.4.1",
-    "ethereum-waffle": "^3.2.1",
-    "hardhat": "^2.0.5",
+    "ethereum-waffle": "^3.3.0",
+    "hardhat": "^2.2.1",
     "hardhat-deploy": "^0.7.0-beta.38",
     "hardhat-gas-reporter": "^1.0.4",
-    "solidity-coverage": "^0.7.13",
+    "solidity-coverage": "^0.7.16",
     "typechain": "^4.0.3"
+  },
+  "dependencies": {
+    "ethers": "^5.1.4"
   }
 }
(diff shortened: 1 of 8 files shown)


Siegrift:
I closed the dependabot PRs and instead updated manually.

packages/protocol/package.json
```diff
@@ -30,11 +30,14 @@
...
-    "hardhat": "^2.0.5",
+    "ethereum-waffle": "^3.3.0",
+    "hardhat": "^2.2.1",
     "hardhat-deploy": "^0.7.0-beta.38",
     "hardhat-gas-reporter": "^1.0.4",
-    "solidity-coverage": "^0.7.13",
+    "solidity-coverage": "^0.7.16",
     "typechain": "^4.0.3"
+  },
+  "dependencies": {
```
andreogle, line 40: Do we need ethers here if it's already in the root package.json with the same version?

andreogle: Small thing if we do, but could we move the dependencies block above the `devDependencies` block just for consistency?

andreogle (approved):
Thanks for fixing this 👍🏻

packages/protocol/package.json
Siegrift: > Do we need ethers here if it's already in the root package.json with the same version?
We do, if it is only in root `package.json` then it will use multiple versions of ethers library - it's need by ethereum-waffle and a few more (you can check with `yarn why ethers`). When we pin the version in package.json explicitly it make it change the node_modules hierarchy and the generated code is using just a single ethers lib (again you can check with `yarn why ethers`).

> Small thing if we do, but could we move the dependencies block above the devDependencies block just for consistency?
Sure.

andreogle: Ah right, didn't realise ethereum-waffle was including it too 👍🏻

PR #349, opened 2021-05-12
Replace Bluebird with native promise implementation

andreogle:
### Description

We only use Bluebird for promise timeouts. After some investigation and comments by @Siegrift it seemed like it could be replaced with native JS promises (using `Promise.race`). The initial comments are no longer applicable

### DIFF
--- a/packages/node/__dev__/invoke.ts
+++ b/packages/node/__dev__/invoke.ts
@@ -7,6 +7,4 @@ async function invoke() {
   await local.startCoordinator();
 }
 
-invoke()
-  .then(() => process.exit(0))
-  .catch(() => process.exit(1));
+invoke();
--- a/packages/node/__dev__/invoke.ts
+++ b/packages/node/__dev__/invoke.ts
@@ -7,4 +7,6 @@ async function invoke() {
   await local.startCoordinator();
 }
 
-invoke();
+invoke()
+  .then(() => process.exit(0))
+  .catch(() => process.exit(1));
--- a/packages/node/src/utils/promise-utils.ts
+++ b/packages/node/src/utils/promise-utils.ts
@@ -1,4 +1,3 @@
-import Bluebird from 'bluebird';
 import { DEFAULT_RETRY_OPERATION_TIMEOUT } from '../constants';
 
 // Adapted from:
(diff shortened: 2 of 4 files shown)


packages/node/__dev__/invoke.ts
```diff
@@ -7,4 +7,6 @@ async function invoke() {
   await local.startCoordinator();
 }
 
-invoke();
+invoke()
+  .then(() => process.exit(0))
```
Siegrift, line 11: @andreogle Maybe add a comment why this is needed

packages/node/src/utils/promise-utils.ts
```diff
@@ -25,22 +24,15 @@ export function goTimeout<T>(ms: number, fn: Promise<T>): Promise<Response<T>> {
...
-//
-// The problem with this is that that the slow promise still runs until it resolves.
-// This means that the serverless function will not exit until the entire timeout
-// duration has been reached which is a problem.
 export function promiseTimeout<T>(ms: number, promise: Promise<T>): Promise<T> {
-  return Bluebird.resolve(promise).timeout(ms);
+  const timeout = new Promise((_res, reject) => {
+    const id = setTimeout(() => {
+      clearTimeout(id);
+      reject(new Error(`operation timed out in ${ms} ms.`));
```
Siegrift, line 31:
```suggestion
      reject(new Error(Operation timed out in ${ms} ms.`));
```

packages/node/src/utils/promise-utils.ts
```diff
@@ -25,22 +24,15 @@ export function goTimeout<T>(ms: number, fn: Promise<T>): Promise<Response<T>> {
...
-//   }
-//
-// The problem with this is that that the slow promise still runs until it resolves.
-// This means that the serverless function will not exit until the entire timeout
-// duration has been reached which is a problem.
 export function promiseTimeout<T>(ms: number, promise: Promise<T>): Promise<T> {
-  return Bluebird.resolve(promise).timeout(ms);
+  const timeout = new Promise((_res, reject) => {
+    const id = setTimeout(() => {
+      clearTimeout(id);
```
Siegrift, line 30: `clearTimeout` is useless here.

You want to call it other way around -- when the original promise resolves/rejects.
(Maybe that's the reason you need the process.exit calls?)

Siegrift (changes requested):
You forgot to remove bluebird from one of the test files.

packages/node/src/utils/promise-utils.ts
andreogle: Thanks, how about this?

```ts
let timeoutId: NodeJS.Timeout;
const timeout = new Promise((_res, reject) => {
  timeoutId = setTimeout(() => {
    reject(new Error(`Operation timed out in ${ms} ms.`));
  }, ms);
});

const wrappedPromise = promise.finally(() => {
  if (timeoutId) {
    clearTimeout(timeoutId);
  }
});

return Promise.race([wrappedPromise, timeout]) as Promise<T>;
```

Siegrift: Yep, lgtm. Did it fix the issue with explicit process.exit call?

Siegrift (approved):

packages/node/src/utils/promise-utils.ts
andreogle: It did actually fix it 👍🏻

PR #350, opened 2021-05-18
Rework packaging and publishing

amarthadan:
After some research into how publishing works with Lerna I ended up with this solution.

In order for Lerna and Yarn working together you have to enable Yarn workspaces - Lerna is internally using them so that dependencies are stored in one common `node_modules` directory at the root and only linked for the rest of the packages (this wasn't the case before). You also have to configure independent versioning for Lerna in order to version the underlying packages independently.

Dependencies in `package.json` are explicitly listed in all the packages, so there is no need for custom`package.publish.json`. It's a bit cumbersome but the dependencies can be managed by Lerna from the root package (e.g. `lerna add <dependency>`).

So, to publish a package you just run `yarn publish` (an alias for `lerna publish`) and Lerna will publish all the packages that have some unpublished changes. Lerna will also automatically create a git tag with published version. This can be properly configured later when we decide on some releasing flow.

As for the naming clash, I simply renamed all packages from `@airnode` to `@api3`. There was no need for directory changes or anything more than just a simple string replace. We can keep directories as they are and possibly rename the packages back to `@airnode` once the name is available.

Currently, few tests are not passing (most likely due to some version mismatch) but the solution itself should be sound. I'll try to resolve this ASAP.


andreogle (approved):

packages/adapter/package.json
```diff
@@ -16,8 +16,17 @@
     "test:watch": "SILENCE_LOGGER=true jest --coverage --watch"
   },
   "dependencies": {
-    "@airnode/ois": "^0.1.0",
+    "@api3/ois": "^0.1.0",
     "axios": "^0.21.1",
-    "bignumber.js": "^9.0.0"
+    "bignumber.js": "^9.0.0",
+    "ethers": "^5.1.4",
+    "lodash": "^4.17.21"
```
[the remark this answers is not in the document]
amarthadan: Damn it, I thought I united them :sweat_smile:
I don't like it much either but this was probably just my mistake and result of the transition process. Once the dependencies are managed via Lerna from the root this should not happen :crossed_fingers:

amarthadan:
@bbenligiray @andreogle Do you guys agree with this concept? Can I merge this?

package.json
```diff
@@ -1,24 +1,29 @@
 {
-  "name": "airnode",
+  "name": "api3",
```
Siegrift, line 2: Are you sure this simple renaming works?

packages/admin/README.md
```diff
@@ -1,4 +1,4 @@
-# Monorepo version: `@airnode/admin`
+# Monorepo version: `@api3/admin`
 # Stand-alone version: `@api3/airnode-admin`
```
Siegrift, line 2: What about these monorepo vs standalone version README. 
This will no longer be the case, right?

packages/admin/package.json
```diff
@@ -8,16 +8,24 @@
     "build": "yarn run clean && yarn run compile",
     "clean": "rimraf -rf ./dist",
     "compile": "tsc -p tsconfig.build.json",
+    "prepublishOnly": "yarn run build",
```
Siegrift, line 11: I believe lerna will automatically build for you, no?
In case no, why do we need it only in admin package?

Siegrift: Actually, other packages have that too, but for example `node` package doesn't.

package.json
amarthadan: Haven't tested all the packages extensively but the deployer (which is using 2-3 others) was working fine.

packages/admin/README.md
amarthadan: Multiple READMEs are out of date, I'm planning to go through them once all the base concepts are settled.

packages/admin/package.json
amarthadan: All of them has it. Without it the package is not build prior the publishing.

Siegrift: Maybe I missed that, I tried ctrl+f on the PR page..

Siegrift (approved):
LGTM

PR #351, opened 2021-05-18
 Update deployer Dockerization process

### LINKED ISSUE #335: Update deployer Dockerization process
## Description
There are actually multiple issues here:

* Prepare Docker image to contain everything necessary for deployment (Terraform, ...)
* Dockerize only necessary parts (deployer build, Terraform recipes, ...)
* Allow passing necessary data to container (cloud provider credentials, configuration and secrets files, airnode build, ...)

amarthadan:
At first, I wanted to install the deployer from registry but decided against it. I think it would complicate the development and container testing process. So, I decided to build the deployer during the container build. But just the deployer, its dependencies are pulled from the registry. It felt like a nice compromise but I'm open to suggestions.

The resulting Docker image is around 500MB big and that's as low I was able to get due to all the dependencies.

I'm using [`su-exec`](https://github.com/ncopa/su-exec) to run deployer as user/group specified via `USER_ID`/`GROUP_ID` environment variables in order to set the correct access rights for the `receipt.json` output.

I'm providing a build argumment `yarnRegistry` that can be set to a custom Yarn respository URL for testing purposes.

As for the usage, it's a bit cumbersome mainly due to the defaults CLI is currently using and directory placement it's expecting. So, the command to run the deployment migh look something like:
```bash
docker run -it --rm --env-file aws.env -e USER_ID=1000 -e GROUP_ID=1000 -v $(pwd)/config.json:/config.json -v $(pwd)/secrets.env:/secrets.env -v $(pwd)/output:/output airnode-deployer:latest deploy -c /config.json -s /secrets.env -o /output/receipt.json
```
I would very much like to get rid of the duplicate path specification (once for the mount and once for the deployer cli itself) but I'm not sure it would be easily possible. I'll do some testing and I'll see.

Close #335

### DIFF
--- a/packages/deployer/terraform/airnode/modules/function/main.tf
+++ b/packages/deployer/terraform/airnode/modules/function/main.tf
@@ -12,8 +12,8 @@ EOC
   }
 
   triggers = {
-    source_file_hash        = fileexists("${local.tmp_source_dir}/${var.source_file}") ? filesha256(var.source_file) : ""
-    configuration_file_hash = fileexists("${local.tmp_configuration_dir}/${var.configuration_file}") ? filesha256(var.configuration_file) : ""
+    // Run always
+    trigger = uuid()
   }
 }
 
@@ -55,7 +55,7 @@ resource "aws_lambda_function" "lambda" {
   ]
 
   environment {
-    variables = jsondecode(file(var.secrets_file))
+    variables = fileexists(var.secrets_file) ? jsondecode(file(var.secrets_file)) : {}
   }
 }
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,15 +1,61 @@
-FROM node:12.19.0-alpine3.12
+# Build preparation
+FROM node:12-alpine as buildPrep
 
-# ENV NODE_ENV production
+ARG yarnRegistry
 
-WORKDIR /airnode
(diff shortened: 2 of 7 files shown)


packages/deployer/terraform/airnode/modules/function/main.tf
```diff
@@ -12,8 +12,8 @@ EOC
   }
 
   triggers = {
-    source_file_hash        = fileexists("${local.tmp_source_dir}/${var.source_file}") ? filesha256(var.source_file) : ""
-    configuration_file_hash = fileexists("${local.tmp_configuration_dir}/${var.configuration_file}") ? filesha256(var.configuration_file) : ""
+    // Run always
+    trigger = uuid()
```
andreogle, line 16: What is this for?

andreogle (approved):

packages/deployer/terraform/airnode/modules/function/main.tf
amarthadan: I needed a way for that `null_resource` to run every time recipes are run. Well, at first I wanted to do it in a clever way, so it will run only if something changes, but that was not working for the Docker version in a scenario where the state already exists but the temp directories don't. So I decided to run it every time. It does so little that it should not be a problem.

PR #352, opened 2021-05-18
SamplePriceDataFeed adminship and dapi update

acenolaza:
* Adminship in SamplePriceDataFeed should be pretty much completed and used the same code that Api3Authorizer is using.
* Cooldown period implemented by admin to prevent users from spamming dapi changes.
* Added single functions for updating each individual member of a dapi. So far I've added addTemplate, removeTemplate and updateTemplate. 

I'm still trying to get a happy path test working for the addTemplate function. There are a lot of TODOs in the code that need to be reviewed.


packages/protocol/contracts/dapis/Api3Adminship.sol
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.4;
+
+import "./interfaces/IApi3Adminship.sol";
+
+/// @title Adminship contract controlled by the API3 DAO
+contract Api3Adminship is IApi3Adminship {
+  string internal constant ERROR_UNAUTHORIZED = "Unauthorized";
+  string internal constant ERROR_ZERO_ADDRESS = "Zero address";
+  string internal constant ERROR_EXPIRATION_NOT_EXTENDED = "Expiration not extended";
```
sassmilic, line 10: What's the reason for storing the error messages to constants? Is it more gas cost efficient in case you use the same error message several times?

packages/protocol/contracts/dapis/Api3Adminship.sol
```diff
@@ -0,0 +1,60 @@
...
+
+  /// @dev Meta admin sets the admin statuses of addresses and has super
+  /// admin privileges
+  address public metaAdmin;
+  mapping(address => AdminStatus) public adminStatuses;
+
+  /// @dev Reverts if the caller is not the meta admin or has admin/superadmin priviliges
+  modifier onlyAdminOrMetaAdmin() {
+    require(adminStatuses[msg.sender] >= AdminStatus.Admin || msg.sender == metaAdmin, ERROR_UNAUTHORIZED);
```
sassmilic, line 19: I think this may be redundant? Since `SuperAdmin` is in the `AdminStatus` struct.

packages/protocol/contracts/dapis/Api3Adminship.sol
```diff
@@ -0,0 +1,60 @@
...
+    _;
+  }
+
+  /// @param _metaAdmin Address that will be set as the meta admin
+  constructor(address _metaAdmin) {
+    require(_metaAdmin != address(0), ERROR_ZERO_ADDRESS);
```
sassmilic, line 25: Do we have to check for this? Since `_metaAdmin` is only provided by the user who deploys the contract, and they likely won't be providing the zero address.

packages/protocol/contracts/dapis/Api3Adminship.sol
```diff
@@ -0,0 +1,60 @@
...
+    metaAdmin = _metaAdmin;
+  }
+
+  /// @notice Called by the meta admin to set the meta admin
+  /// @param _metaAdmin Address that will be set as the meta admin
+  function setMetaAdmin(address _metaAdmin) external override {
+    require(msg.sender == metaAdmin, ERROR_UNAUTHORIZED);
+    require(_metaAdmin != address(0), ERROR_ZERO_ADDRESS);
```
sassmilic, line 33: same with this one

packages/protocol/contracts/dapis/Api3Adminship.sol
```diff
@@ -0,0 +1,60 @@
...
+
+  /// @notice Called by an admin to renounce their admin status
+  /// @dev To minimize the number of transactions the meta admin will have
+  /// to make, the contract is implemented optimistically, i.e., the admins
+  /// are expected to renounce their admin status when they are needed to.
+  /// If this is not the case, the meta admin can always revoke their
+  /// adminship.
+  /// This method cannot be used by the meta admin to renounce their meta
+  /// adminship.
+  function renounceAdminStatus() external override {
```
sassmilic, line 55: Do you have any idea when this would be called? What's the use case? Maybe a better question for @bbenligiray

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -28,13 +30,41 @@ contract RrpDapiServer is MeanMedianHybrid {
 
   AirnodeRrp public airnodeRrp;
   mapping(bytes16 => Dapi) private dapis;
+  // TODO: are we missing a dapiIds bytes16 array for getting all the dapi ids?
```
sassmilic, line 33: Is there a situation where we would need to get all `dapiIds`? Or is it just good practice to do so?

sassmilic: Ahh, I think there isn't a good way to "take a look" and see which dAPIs we've registered, right? Unless we have an array?

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
```diff
@@ -2,15 +2,33 @@
...
 // A sample price data feed implemented as a dAPI client
-contract SamplePriceDataFeed {
+contract SamplePriceDataFeed is Api3Adminship {
   RrpDapiServer public rrpDapiServer;
+
   bytes16 public latestDapiId;
   int256 public latestAnswer;
 
-  constructor(address _rrpDapiServer) {
+  uint256 public cooldownTime = 1 days; // TODO: is this the right value?
```
sassmilic, line 14: @bbenligiray

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
```diff
@@ -2,15 +2,33 @@
...
+        block.timestamp >= (adminToDapiLastUpdated[msg.sender] + cooldownTime),
+      "Cooldown period has not finished"
+    );
+    _;
+  }
+
+  constructor(address _rrpDapiServer, address _metaAdmin) Api3Adminship(_metaAdmin) {
+    require(_rrpDapiServer != address(0), ERROR_ZERO_ADDRESS);
     rrpDapiServer = RrpDapiServer(_rrpDapiServer);
+    // TODO: should we set superadmin status to technical team deploying the contract?
```
sassmilic, line 30: You mean, don't supply `_metaAdmin` and just set to the address of the deployer? Yea, I agree with you, I like that better.

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
```diff
@@ -2,15 +2,33 @@
...
+    // this.setAdminStatus(msg.sender, AdminStatus.SuperAdmin);
   }
 
   function setDapi(
```
sassmilic, line 34: Do you know when this would be used? I think this is Burak's "reset" function or smth. Can you confirm?

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
```diff
@@ -19,6 +37,170 @@ contract SamplePriceDataFeed {
...
+        requestIndexResetter
+      );
+
+    // Update dapiLastUpdated timestamp
+    adminToDapiLastUpdated[msg.sender] = uint64(block.timestamp);
+
+    // Record the returned dapiId
+    latestDapiId = newDapiId;
+
+    // TODO: publish TemplateAdded event?
```
sassmilic, line 88: What do you see the benefit being?

sassmilic: (Actually wondering, not being smarmy haha)

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
```diff
@@ -19,6 +37,170 @@ contract SamplePriceDataFeed {
...
+    (
+      uint256 noResponsesToReduce,
+      uint256 toleranceInPercentages,
+      uint256 requesterIndex,
+      bytes32[] memory templateIds,
+      address[] memory designatedWallets,
+      address reduceAddress,
+      bytes4 reduceFunctionId,
+      address requestIndexResetter
+    ) = rrpDapiServer.getDapi(_dapiId);
```
sassmilic, line 102: I'm actually not familiar with this notation ...

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
```diff
@@ -19,6 +37,170 @@ contract SamplePriceDataFeed {
...
+    bool found = false;
+    bytes32[] memory newTemplateIds = new bytes32[](templateIds.length - 1);
+    address[] memory newDesignatedWallets = new address[](designatedWallets.length - 1);
+    for (uint256 i = 0; i < templateIds.length; i++) {
+      // copy array without _templateId
+      if (templateIds[i] == _templateId) {
+        found = true;
+      }
+      uint256 j = found ? i + 1 : i;
+      if (i != templateIds.length - 1) {
```
sassmilic, line 116: This is at risk of being really gas cost inefficient. I think we might wanna consider a different data structure to prevent looping.

packages/protocol/contracts/dapis/Api3Adminship.sol
acenolaza: The truth is that I copied them over from [here](https://github.com/api3dao/airnode/blob/871f17741045d4ff6c47dfb4f81f897e6beaf1f2/packages/protocol/contracts/authorizers/Api3Authorizer.sol#L8).

I don't really know if it costs less than using the same string error message several times but at least the [docs](https://docs.soliditylang.org/en/latest/contracts.html#constants) say that constants state cost is lower than regular state variables. 

> Compared to regular state variables, the gas costs of constant and immutable variables are much lower. For a constant variable, the expression assigned to it is copied to all the places where it is accessed and also re-evaluated each time. 

Maybe this is a question better suited for Burak and I also wonder if we should try to force their length to 32 bytes

sassmilic: I'm currently asking some friends. The benefits of having smart contract auditor friends!

packages/protocol/contracts/dapis/Api3Adminship.sol
acenolaza: Also copied that over from API3Authorizer [contract](https://github.com/api3dao/airnode/blob/871f17741045d4ff6c47dfb4f81f897e6beaf1f2/packages/protocol/contracts/authorizers/Api3Authorizer.sol#L23) but I can't really make a strong case why it is needed other than being a good practice to check for "null" on arguments.

packages/protocol/contracts/dapis/Api3Adminship.sol
acenolaza: Maybe it's just following a similar pattern that openzeppelin uses for the [Ownable.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/406c83649bd6169fc1b578e08506d78f0873b276/contracts/access/Ownable.sol#L54) contract? I'm just guessing here

packages/protocol/contracts/dapis/Api3Adminship.sol
sassmilic: I think it's unnecessary, but I'll let @bbenligiray share his take.

packages/protocol/contracts/dapis/Api3Adminship.sol
sassmilic: Ah, okay. Thanks for the explanation. That makes sense. I still don't really get when it would be used, but I trust openzepplin has a good justification.

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
acenolaza: Basically I see this aggregator contract being used by possible many client apps and they might want to know what changed on the underlying dAPI everytime an admin calls any of these functions that cause a mutation on the dAPI. Not sure tho and that's why there is a TODO :smile:

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
acenolaza: I see this notation to something similar to [destructuring](https://docs.soliditylang.org/en/v0.8.4/control-structures.html#destructuring-assignments-and-returning-multiple-values) in Typescript. 

I would have preferred to be able to just return a struct or an some other data structure from RrpDapiServer.getDapi() but I found out that it was not possible therefore that's why I ended up returning all the individual values.

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
acenolaza: Not really sure why but that's a good observation. I'll research more on why and how to improve this.

sassmilic: Yea, this is very inefficient, although it's not your implementation, it's the dAPI data structure.

If we must be accessing individual templates in the dAPI struct, then it should be a mapping @bbenligiray

sassmilic (changes requested):
Okay, done! Mostly minor things except the really inefficient for-loop search. That requires changing the dAPI struct. I tagged @bbenligiray on comments/issues that I don't think can be resolved between you (Santiago, not Frederico) and me.

acenolaza:
Thanks again Sasa for taking the time to look at this :clap:

acenolaza (commented):
Thanks again Sasa for taking the time to look at this :clap:

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
sassmilic: True, but it also makes sense that the team would also deploy the contract. @bbenligiray thoughts?

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
sassmilic: That makes sense

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
sassmilic: ah, yes yes. same thing in python ("unpacking" values). Yea, this looks nice.

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
acenolaza: Something kind of related is also the designatedWallets array. It seems that there's a 1 to 1 relationship between designated wallets and template ids and that makes me think that all this info might be better stored together using a struct.

packages/protocol/contracts/dapis/Api3Adminship.sol
sassmilic: Screen Shot 2021-05-24 at 6 36 21 PM

sassmilic: yea, looks like there isn't really a difference (only minor one during deploy). that's why i suggest removing the constant declarations and just putting string literals in the `require` statements. it makes the code simpler, since the error codes are quite simple to begin with.

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
acenolaza: @sassmilic @bbenligiray I've been thinking and maybe the best way to avoid having to iterate over the templateIds/designatedWallets arrays is to have the addTemplate, updateTemplate and removeTemplate methods also in the RrpDapiServer contract. This way we can access the storage array directly and use push/pop methods. Maybe for update we'll still need to iterate tho.

acenolaza: Even though we can add another mapping to store the index in the array for each templateId (`mapping(bytes32 => uint)`) I'm still struggling to completely get rid of the for loops since we must make a copy of both arrays (templateIds and designatedWallets) before calling `registerDapi()` function.

sassmilic: Ahh, this is what I said this morning's meeting. I didn't realize you already mentioned it!

sassmilic: Can you show me where in the code you're referring to
> we must make a copy of both arrays (templateIds and designatedWallets) before calling registerDapi() function.

acenolaza: Sure, I don't think I was clear enough on this point. Since we are not actually changing the current dapi templateIds but in fact what we need to do is register a new dapi based on the values of the current dapi. This means we need to make a copy of the templateIds and designatedWallets arrays in order to call `registerDapi()`.

For the `addTemplate() `function I was not able to create a copy of the arrays (plus the new id) without using a for loop. Basically because memory arrays in solidty cannot be resized anymore.

I also thought about using [this](https://solidity-by-example.org/app/iterable-mapping/) IterableMapping library and even tough it is pretty neat I still couldn't get rid of all the for loops.

Like you said this is a very common problem in solidity (need to access an element by key and also iterate over the list) but what makes this specific case somewhat special is that we need to make copies of the arrays and not actually modify the arrays stored in state.

packages/protocol/contracts/dapis/Api3Adminship.sol
bbenligiray: It's commonly considered as being good practice

packages/protocol/contracts/dapis/Api3Adminship.sol
bbenligiray: Inlining error strings is not an issue if they are only used once but if they are used in multiple places, you want to use variables so that you don't change one and forget to change the other. Once you use a variable for some you may as well use variables for all. But all in all it's just a pattern that I liked and picked up, it's not critical.

packages/protocol/contracts/dapis/Api3Adminship.sol
bbenligiray: The justification is explained here
https://github.com/api3dao/airnode/blob/8658431d252f7442906af61f1e8c6ec27d1ec134/packages/protocol/contracts/authorizers/Api3Authorizer.sol#L46
I'm more thinking about the day where the DAO assigns adminship to a multisig controlled by a team. Once that multisig becomes obsolete, we don't want to require a proposal from the DAO to pass, it would be nice if the multisig could just renounce its adminship.

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
bbenligiray: I don't know, coming up with the specs is the hardest part of this and depends on what you envision a dAPI operation would look like. The main concern is the admins attacking, so I guess erring on the side of a longer cooldown period is better (1 week?).

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
bbenligiray: I think a dev would deploy the contract using their personal wallet, setting themselves as the meta admin. Then they would give the multisig adminship, then they would transfer the meta adminship to the DAO. Then it's verified that it's set up correctly and then it starts being used.

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
bbenligiray: Yeah this was a sort of a placeholder

packages/protocol/contracts/dapis/SamplePriceDataFeed.sol
bbenligiray: An alternative way to implement this is that the caller gives the new template ID array, the function checks that the new template ID array is sufficiently similar to the old one, then the new array gets used to create the new dAPI

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -70,6 +101,13 @@ contract RrpDapiServer is MeanMedianHybrid {
     dapis[dapiId].reduceFunctionId = reduceFunctionId;
     dapis[dapiId].requestIndexResetter = requestIndexResetter;
     dapis[dapiId].nextRequestIndex = 1;
+
+    nextDapiIndex++;
+
+    // TODO: move this into an interface? IRrpDapiServer.sol?
```
bbenligiray, line 107: Although I keep events in interfaces people hate that, be warned

bbenligiray:
One potential architecture is to have a single contract that serves multiple data feeds. So you would not have
```solidity
int256 public latestAnswer;
```
but
```solidity
mapping(string/bytes32 => int256) public latestAnswer;
```
Not sure if it's worth the gas overhead, feel free to explore that as well.

acenolaza:
@sassmilic On my last commit I pushed a POC of adding the `templateIdToIndex` mapping and these are my conclusions:
1. it really helps with updating templateIds since we know exactly which index in the array to update with the new value. We can also use the memory array that we get from calling `getDapi()` and pass it to `registerDapi()` with the new values.
2. it kind of helps when removing templateIds but the tricky part here is that we cannot actually delete the id from the array but instead all we can do is clear the value at the given index (I mean without using a for loop).
3. it doesn't help at all when adding a new templateId since we need to pass a new array which is bigger than the one we get back from getDapi() and we must loop thru the array to copy all ids to the new array anyways.

With that said I think I'm now starting to lean towards @bbenligiray [suggestion](https://github.com/api3dao/airnode/pull/352#discussion_r642602661) of just receive the new templateIds array in an `updateTemplateIds()` method.  Thoughs?

sassmilic:
sorry i haven't looked at the code yet, but re:
> it doesn't help at all when adding a new templateId since we need to pass a new array which is bigger than the one we get back from getDapi() and we must loop thru the array to copy all ids to the new array anyways.
why can't you just `push` to the array?

acenolaza:
> why can't you just `push` to the array?

Because push and pop are only available on state arrays. Also I think we need a copy of the array (memory) here because we don't want to modify the array of the current dapi (state). 

I think the idea is that we register a new dapi with the values of the current dapi plus changes. The new dapiId is a hash of all its values.

packages/protocol/contracts/dapis/RrpDapiServer.sol
```diff
@@ -28,13 +38,45 @@ contract RrpDapiServer is MeanMedianHybrid {
...
+      uint256,
+      uint256,
+      bytes32[] memory,
+      address[] memory,
+      address,
+      bytes4,
+      address
+    )
+  {
+    Dapi storage dapi = dapis[dapiId];
```
sassmilic, line 67: pardon my ignorance, by why does this need to be storage?

sassmilic:
> I think the idea is that we register a new dapi with the values of the current dapi plus changes. The new dapiId is a hash of all its values.

So just create a new dAPI when a change needs to be implemented? I think that's simple enough. I don't think changes will need to be made too often, so the gas costs of re-registering might not be that bad. But I'm willing to hear other opinions ...

packages/protocol/contracts/dapis/RrpDapiServer.sol
acenolaza: My understanding is that storage means "pointer to" so in this case I am creating a pointer to the dapi struct in state instead of making a copy because I just want to read it's values and return them.

acenolaza:
@sassmilic I've just pushed a new branch with a single updateDapi() method to change the templateIds array using the approach suggested by @bbenligiray ([here](https://github.com/api3dao/airnode/pull/352#discussion_r642602661))

> An alternative way to implement this is that the caller gives the new template ID array, the function checks that the new template ID array is sufficiently similar to the old one, then the new array gets used to create the new dAPI

I have to say the code looks much simpler and it completely avoids having to loop thru arrays in order to find a specific key or create a new copy of the array to then pass it as argument to `registerDapi()`

Here is the [diff](https://github.com/api3dao/airnode/compare/data-feed-adminship...data-feed-adminship-update-dapi) between both branches

sassmilic:
@acenolaza I agree, looks good! Just to clarify re:

> checks that the new template ID array is sufficiently similar to the old one

This isn't actually implemented (checking for "sufficient similarity"), correct? You just check to make sure the ID is different? I think it's fine the way it is, just want to clarify.

acenolaza:
> @acenolaza I agree, looks good! Just to clarify re:
> 
> > checks that the new template ID array is sufficiently similar to the old one
> 
> This isn't actually implemented (checking for "sufficient similarity"), correct? You just check to make sure the ID is different? I think it's fine the way it is, just want to clarify.

You're completely right. This is just checking that the arrays of IDs and designatedWallets are different before actually registering a new dapi. I don't fully remember why but when I was working on this implementation it sounded like a better check.

sassmilic:
@acenolaza looks good! Since this is a draft PR, I think the next step is making a real PR? So go ahead and do that, and put me and @bbenligiray as reviewers.

Issue #354, opened 2021-05-20
Fix enviroment variables in node and deployer


Ashar2shahid:
## Description
AWS Lambda has stopped allowing "-" in the environment variable name. Which causes the following error: 

![image](https://user-images.githubusercontent.com/31223740/118907970-7cc5d800-b8ee-11eb-836d-548d331c7c6f.png)

## Version
Airnode pre-alpa

## Present Behaviour
When "-" is present in environment variables or if "-" is present in config.json/security.json that maps to environment variables on AWS the deployer fails. In the following config.json and security.json, the key "finage-apikey" causes an error in deployment 

### Config.json
![image (1)](https://user-images.githubusercontent.com/31223740/118908928-42f5d100-b8f0-11eb-9aca-296a3fb42377.png)

### Security.json
![image (2)](https://user-images.githubusercontent.com/31223740/118909151-9405c500-b8f0-11eb-9016-73760f1791e2.png)

## Expected Behaviour
It should deploy even with the presence of "-" in config.json/security.json

## Steps to reproduce
Using above config.json and security.json and deploying.

PR #355, opened 2021-05-21
fix for env variables in pre-alpha

Ashar2shahid:
This is the fix for #354


andreogle:
Is this the right fix? I don't think that the node should be replacing "-" with "_" when fetching secrets.

Say someone doesn't know about this rule, has deployed with Docker and in their config.json there is a security scheme that contains a "-", e.g. "finage-securityscheme". The node will then try to find a secret that looks like "finage_securityscheme" which won't exist (?)

I would say it should rather be validated in the validator. The deployer should throw an error if they try to deploy any secrets with a "-".  Something we output ourselves that is more readable than the AWS error.

Ashar2shahid:
Noted, I've changed to it showing an error message instead, so you can have a look.

acenolaza (approved):
LGTM :+1:

PR #373, opened 2021-06-07
Metadata reserved parameter

acenolaza:


andreogle (changes requested):
Looks great so far, just a few minor change requests 🚀

packages/node/src/handlers/call-api.ts
```diff
@@ -1,27 +1,48 @@
...
+import { removeKeys } from '../utils/object-utils';
+import { go, retryOnTimeout } from '../utils/promise-utils';
 
 function buildOptions(
+  chain: ChainConfig,
   ois: OIS,
   securitySchemeEnvironmentConfigs: SecuritySchemeEnvironmentConfig[],
   aggregatedApiCall: AggregatedApiCall
 ): adapter.BuildRequestOptions {
+  let parameters = aggregatedApiCall.parameters;
```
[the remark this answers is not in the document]
acenolaza: That's a really good pattern to follow :+1:

packages/node/src/handlers/call-api.ts
```diff
@@ -1,27 +1,48 @@
...
+  // Include airnode metadata based on _relay_metadata version number
+  if (
+    ReservedParameterName.RelayMetadata in aggregatedApiCall.parameters &&
```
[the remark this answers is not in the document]
acenolaza: Right, it might be enough to just do the second check or change it to a switch statement when we have newer versions. I'll double check this tho

packages/node/src/handlers/call-api.ts
```diff
@@ -1,27 +1,48 @@
...
+    aggregatedApiCall.parameters[ReservedParameterName.RelayMetadata] == 'v1'
```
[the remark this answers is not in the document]
acenolaza: Good catch!

packages/node/src/handlers/call-api.ts
```diff
@@ -1,27 +1,59 @@
...
 function buildOptions(
+  chain: ChainConfig,
   ois: OIS,
   securitySchemeEnvironmentConfigs: SecuritySchemeEnvironmentConfig[],
   aggregatedApiCall: AggregatedApiCall
 ): adapter.BuildRequestOptions {
+  const parameters = aggregatedApiCall.parameters;
+
+  // Include airnode metadata based on _relay_metadata version number
+  const parametersWithMetadata = addMetadataParameters(parameters, aggregatedApiCall, chain);
```
andreogle, line 53: Why pass in `parameters` and `aggregatedApiCall` separately? We can access parameters in `addMetadataParameters()` through the `aggregatedApiCall`?

andreogle (changes requested):

packages/node/src/handlers/call-api.ts
```diff
@@ -1,43 +1,77 @@
...
-function buildOptions(
-  ois: OIS,
-  securitySchemeEnvironmentConfigs: SecuritySchemeEnvironmentConfig[],
-  aggregatedApiCall: AggregatedApiCall
-): adapter.BuildRequestOptions {
-  // Don't submit the reserved parameters to the API
-  const parameters = removeKeys(aggregatedApiCall.parameters || {}, RESERVED_PARAMETERS);
+function addMetadataParameters(aggregatedApiCall: AggregatedApiCall, chain: ChainConfig): ApiCallParameters {
+  const parameters = aggregatedApiCall.parameters;
+  switch (aggregatedApiCall.parameters[ReservedParameterName.RelayMetadata]) {
```
andreogle, line 22: I was thinking that we probably want this to be case insensitive. So both `v1` and `V1` would give the same result. Maybe needs a `?.toLowerCase()` or something similar?

acenolaza: Another good catch

.eslintrc.js
```diff
@@ -44,5 +44,6 @@ module.exports = {
     'no-console': 0,
     'no-useless-escape': 0,
     semi: 2,
+    eqeqeq: ['error', 'smart'],
```
andreogle, line 47: 🙌🏻

packages/node/src/handlers/call-api.ts
```diff
@@ -1,43 +1,77 @@
...
-function buildOptions(
-  ois: OIS,
-  securitySchemeEnvironmentConfigs: SecuritySchemeEnvironmentConfig[],
-  aggregatedApiCall: AggregatedApiCall
-): adapter.BuildRequestOptions {
-  // Don't submit the reserved parameters to the API
-  const parameters = removeKeys(aggregatedApiCall.parameters || {}, RESERVED_PARAMETERS);
+function addMetadataParameters(aggregatedApiCall: AggregatedApiCall, chain: ChainConfig): ApiCallParameters {
+  const parameters = aggregatedApiCall.parameters;
+  switch (aggregatedApiCall.parameters[ReservedParameterName.RelayMetadata]?.toLocaleLowerCase()) {
```
andreogle, line 22: I don't think we should be using locale as it might lead to inconsistent results
```suggestion
  switch (parameters[ReservedParameterName.RelayMetadata]?.toLowerCase()) {
```

packages/node/src/handlers/call-api.ts
```diff
@@ -1,43 +1,77 @@
...
+  aggregatedApiCall: AggregatedApiCall
+): adapter.BuildRequestOptions {
+  // Include airnode metadata based on _relay_metadata version number
+  const parametersWithMetadata = addMetadataParameters(aggregatedApiCall, chain);
+
+  // Don't submit the reserved parameters to the API
+  const sanitizedParameters = removeKeys(parametersWithMetadata || {}, RESERVED_PARAMETERS);
+
+  // Fetch secrets and build a list of security schemes
+  const securitySchemes = buildSecuritySchemes(ois, securitySchemeEnvironmentConfigs);
```
andreogle, line 70: I think the naming makes this the most confusing. So just to clarify, here we only create objects with a `name` and `value`. We send these with the OIS and then the `adapter` package matches the new objects up with the relevant security scheme to build the "full" security scheme/parameters.

Maybe the type should be called `SecuritySecrets` or `SecuritySchemeSecrets` instead (and the variables changed accordingly)? What we're building here isn't actually a security scheme. I wrote I lot of the original code and I'm still confusing by it 😄

packages/adapter/src/request-building/authentication.ts
```diff
@@ -75,24 +75,27 @@ export function buildParameters(options: CachedBuildRequestOptions): Authenticat
   };
 
   const { securitySchemes } = options;
-  // Security schemes originate from 'security.json' and contain all of the authentication details
+  // Security schemes originate from 'config.json' and contain all of the authentication details
```
andreogle, line 78: I think changing the name of this will help a lot with the confusion (like I mentioned below). These objects are not actually `securitySchemes`

andreogle (changes requested):

packages/node/src/handlers/call-api.test.ts
```diff
@@ -31,10 +31,108 @@ describe('callApi', () => {
...
-            type: 'apiKey',
+            value: 'supersecret',
+          },
+        ],
+      },
+      { timeout: 20000 }
+    );
+  });
+
+  it('calls the adapter with the given parameters (with _relay_metadata set to v1 lowercase)', async () => {
```
andreogle, line 43: There's a lot of duplication in these tests just to test different values. Can we rather use `.forEach` (like [here](https://github.com/api3dao/airnode/blob/master/packages/node/src/providers/state.test.ts#L153)) or `test.each`. I wasn't aware of `test.each` until recently which is why I didn't use it in the past, but also `.forEach` seems a bit more intuitive.

You probably need some object to loop over like:

```ts
const relayPermutations = [
  { value: 'v1', metadataExpected: true },
  { value: 'V1', metadataExpected: true },
  { value: 'false', metadataExpected: false },
  { value: undefined, metadataExpected: false },
];
```

packages/node/src/handlers/call-api.ts
acenolaza: :smile: 

> There are only two hard things in Computer Science: cache invalidation and naming things.
> Phil Karlton

Your description is correct. 
1. in call-api.ts we read the secret from secrets.env and create a key/value object (security scheme name / secret). Then we pass this object plus the ois object to the adapter
2. in authentication.ts we match them by securty scheme name and extract the security scheme from ois object in order to add the auth headers to the request.

I think SecuritySchemeSecrets is a much more descriptive name for this key/value array.

packages/node/src/handlers/call-api.test.ts
acenolaza: I have used the .forEach approach in the past but I'll dig more into the test.each too. Thanks :+1:

packages/node/src/handlers/call-api.ts
acenolaza: @andreogle should we also consider changing the config.json key under `enviroment`?

andreogle: Maybe... What do you think the key should be then? We'll probably want to get Burak involved if you want to change `config.json` keys

acenolaza: SecuritySchemeValues maybe? I started wondering about this name change in the config.json when you (accurately) said that the secret was not actually a security scheme but maybe I'm just over complicating things.

andreogle: So in config.json, it would look like this?

```json
{
  "environment": {
    "securitySchemeValues": [
      { ... }
    ]
  }
}
```

I'm not sure... I think let's leave config.json changes out of this PR for now. I think it's fine to change the variable and type names though

packages/node/src/handlers/call-api.test.ts
```diff
@@ -43,6 +40,55 @@ describe('callApi', () => {
     );
   });
 
+  describe('with _relay_metadata set', () => {
```
andreogle, line 43: 👌🏻

andreogle (approved):
Nice one! 🥇 Keen to get this merged

PR #388, opened 2021-06-20
Validator

drgy:
Bringing validator and convertor to usable stage, most notable changes:

* Reworked readme & documentation
* Updated tests
* Many bug fixes
* Support multiple versions of specification validation/conversion
* Parameter paths were strings with `.` as delimiter, this was causing many issues, so it was changed to array of strings (inside validator, but also in templates - [example](https://github.com/api3dao/airnode/blob/validator/packages/validator/docs/dynamic_params.md#retrieving-value-of-any-parameter-in-specification))
* New keywords like [`__catch`](https://github.com/api3dao/airnode/blob/validator/packages/validator/docs/catch.md) or [`__template`](https://github.com/api3dao/airnode/blob/validator/packages/validator/docs/template.md) added


Siegrift:
@drgy can you make sure the PR passes the CI checks?

drgy:
@Siegrift the only failing check should be non existing link (https://github.com/api3dao/airnode/tree/master/packages/validator), which will be created by merging into master...

packages/validator/README.md
```diff
@@ -0,0 +1,96 @@
+# `@airnode/validator`
+
+Package for json specifications validation, validation process can be configured with validator templates.
```
sassmilic, line 3: nit: two sentences instead of one

packages/validator/docs/actions.md
```diff
@@ -0,0 +1,292 @@
+# Actions
+
+If any action is specified in template, object `output` will be returned when processing a specification. This object is at first empty, by inserting or copying parameters from specification it can construct desired specification.
```
sassmilic, line 3: suggested change:
>This object is at first empty. By inserting or copying parameters from the given specification, it can construct the desired specification.

sassmilic:
@drgy I'm making several suggested edits re: grammar and word choice. If you feel like your original wording is better, feel free to keep it as-is.

sassmilic:
@drgy really good job on the documentation!

packages/validator/docs/dynamic_params.md
```diff
@@ -0,0 +1,183 @@
+# Dynamic values and keys
+
+## Retrieving key from parameter path
+
+Any key or value can contain a number surrounded by `{{}}`, which will be replaced by key located on given index in parameter path. For example `{{0}}` located in value with parameter path `outer.arr[2].inner` will be replaced with `outer`, similarly `{{1}}` will be replaced with `arr[2]` and so on.
```
sassmilic, line 5: suggested change:
> Any key or value can contain a number surrounded by {{}} -- `{{x}}` where `x` is an integer. This indicates that `{{x}}` will be replaced by the key located on the `x`-th index in the parameter path.

packages/validator/docs/dynamic_params.md
```diff
@@ -0,0 +1,183 @@
+# Dynamic values and keys
```
sassmilic, line 1: nit: 
> Dynamic keys and values

packages/validator/docs/dynamic_params.md
```diff
@@ -0,0 +1,183 @@
...
+
```
sassmilic, line 2: Add something like:
> The validator template accepts both dynamic keys and dynamic values. Dynamic parameters contain placeholders that are filled with strings contained in other parts of the template.

packages/validator/docs/optional.md
```diff
@@ -0,0 +1,47 @@
+
+# Optional
+
+Some parameters in specification can be present but in case they are not there, it shouldn't be an issue as well. For this `__optional` is used, parameters inside it will not cause any extra parameter messages or missing parameter messages to show.
```
sassmilic, line 4: Suggested:
> By default, parameters in the specification are required. However, there is the option of creating optional parameters, in case certain parameters are present but _not_ required.

packages/validator/docs/template.md
```diff
@@ -0,0 +1,64 @@
+# Nested templates
+
+In case a template contains another template, the nested template doesn't have to be copied, relative path to the template can be provided in `__template` and parameter will be evaluated using provided template.
```
sassmilic, line 3: Suggestion:

> Templates can be recursive. That is, a template can contain a pointer to another template in the form of a relative path provided in `__template`.

sassmilic:
Please fix linting errors :)

packages/validator/src/commands/convertCmd.ts
```diff
@@ -0,0 +1,49 @@
...
+      convert(process.env.npm_config_specs || process.argv[3], process.env.npm_config_template),
+      null,
+      '\t'
+    )
+  );
+} else {
+  const from = (process.env.npm_config_from || process.argv[2]).toLowerCase();
+  const to = (process.env.npm_config_to || process.argv[3]).toLowerCase();
+  const specs = process.env.npm_config_specs || process.argv[4];
+  const version = process.env.npm_config_fromVersion || process.argv[5];
```
sassmilic, line 22: Is it expected that the user *either* supply environment variables *or* commandline variables? If so, I find this a bit confusing because it seems like the user can do a mix of both.

sassmilic:
@drgy can you please confirm: most of this code is the same as the code I reviewed in https://github.com/api3dao/airnode/pull/195 ?

(i wanna make sure i'm not crazy 🙃)

packages/validator/docs/regex.md
```diff
@@ -0,0 +1,69 @@
...
+  ]
+}
+```
+---
+
+## Useful regular expressions
+
+`^(one|two)$` - all the valid strings are `one` and `two`
+
+`(?<={)[^\/{}]+(?=})` - anything thats between `{` and `}` (usage example at parameters in paths)
```
sassmilic, line 67: which `{` and `}` ?

packages/validator/docs/dynamic_params.md
drgy: That seems a bit confusing, I added this:

> The validator can access both, keys from the parameter path and values from the specification. These dynamic parameters consist of placeholders that are on evaluation replaced with strings contained in other parts of the specification.

packages/validator/src/commands/convertCmd.ts
drgy: Validator can process mix of both variables, it is a bit of useless feature, but I don't see a reason why not to allow it.

packages/validator/docs/regex.md
drgy: > anything thats between characters `{` and `}`

I'm not sure if I understand, is this better?

drgy:
@sassmilic yes, most of the code is same, summary of significant changes is in the description of this PR, however there are also many small changes in multiple places, those are bug fixes that were discovered by more validator testing.

packages/validator/docs/dynamic_params.md
sassmilic: sure! just one thing: i think the first comma is unnecessary.

packages/validator/docs/regex.md
sassmilic: i think i didn't properly understand/explain my confusion: why are you using a lookbehind? i.e. `(?<=...)`

sassmilic:
> @sassmilic yes, most of the code is same, summary of significant changes is in the description of this PR, however there are also many small changes in multiple places, those are bug fixes that were discovered by more validator testing.

thanks! (in that case, it makes more sense for me to check individual commits vs going file-by-file)

packages/validator/src/processor.ts
```diff
@@ -43,7 +43,7 @@ export function processSpecs(
         }
       }
 
-      break;
```
sassmilic: Let me try and understand this fix: is this for when you have multiple `__optional` keys in the format file? If so: is it legal to have duplicate keys in a json file? i.e. it's valid?

packages/validator/src/utils/utils.ts
```diff
@@ -0,0 +1,332 @@
...
+ * Replaces paths inside "[[]]" with value of the parameter in the path
+ * @param specs - specification of parameter which is being validated
+ * @param rootSpecs - root of the validated specification
+ * @param template - template in which paths will be replaced
+ */
+export function replacePathsWithValues(specs: any, rootSpecs: any, template: any): any {
+  const substitute = (toReplace: string) => {
+    const matches = toReplace.match(/(?<=\[)\[.+?\](?=\])/g);
+
+    if (!matches) {
```
sassmilic, line 30: When would this happen?

packages/validator/templates/1.0.0/endpoints.json
```diff
@@ -0,0 +1,71 @@
+{
+  "__arrayItem": {
+    "name": {},
+    "operation": {
+      "path": {
+        "__regexp": "^\\/[^\\s'\"\\\\]+$"
```
sassmilic, line 6: OAS/Swagger allows for path templating. Is this something the current validator can handle? https://swagger.io/docs/specification/paths-and-operations/

packages/validator/templates/1.0.0/endpoints.json
```diff
@@ -0,0 +1,71 @@
...
+            "__catch": {
+              "__message": "Allowed values in __fullPath are: 'path', 'query', 'header' or 'cookie'"
+            }
+          }
+        },
+        "__optional": {
+          "default": {},
+          "description": {},
+          "required": {},
+          "example": {}
```
sassmilic, line 61: It seems like OAS/Swagger allows for several other field names:
`deprecated` and `allowEmptyValue`

packages/validator/templates/1.0.0/ois.json
```diff
@@ -0,0 +1,184 @@
+{
+  "oisFormat": {
+    "__regexp": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
+  },
+  "title": {
+    "__regexp": "^[a-zA-Z0-9-_\\s]{1,64}$"
```
sassmilic, line 6: spaces are allowed?

sassmilic: and where does the 64 come from?

packages/validator/templates/0.1.0/config.json
```diff
@@ -0,0 +1,253 @@
+{
+	"__arrayItem": {
+		"ois": {
+			"__arrayItem": {
+				"__template": "../1.0.0/ois.json",
+				"apiSpecifications": {
```
sassmilic, line 6: `apiSpecifications` are already contained in `"./1.0.0/ois.json`

packages/validator/templates/1.0.0/apiSpecifications.json
```diff
@@ -0,0 +1,186 @@
+{
+  "servers": {
+    "__maxSize": 1,
+    "__arrayItem": {
+      "url": {
+        "__regexp": "^(([^:\\/?#]+):)?(\\/\\/([^\\/?#]*))?([^?#]*)$"
```
sassmilic, line 6: Was this copied from somewhere or did you write this?

packages/validator/src/processor.ts
drgy: `__ignore` is a bit hidden keyword, that validator is using internally when working with nested templates. I guess I accidentally submitted it in this commit, it should have been separate bug fix commit.
This section will ensure, that validator won't print any extra parameter messages, we want this because that part of specification should be already validated by nested template. However if we `break` and don't check any further parameters in template, then validation wouldn't be complete. Which means there must be a `continue` so that everything is validated.
Templates are normal json, which does not allow duplicated keywords, but this change is not about that.

packages/validator/src/utils/utils.ts
drgy: This function is called few lines below in `recursiveSubstitute` on every key and value of the specification, it is up to `substitute` to figure out if there is anything to be replaced or not, whatever it returns will replace what was previously there.

packages/validator/templates/1.0.0/endpoints.json
drgy: Absolutely, in api specifications path it even warns about missing parameter if it is not present in list of params of the path.

packages/validator/templates/1.0.0/endpoints.json
drgy: I was following a bit different version of [this document](https://docs.api3.org/pre-alpha/airnode/specifications/ois.html), which doesn't seem to be available anymore, but those fields don't seem to be there as well. I guess we should wait for a more stable version of specifications and then fix these minor inconsistencies.

packages/validator/templates/1.0.0/ois.json
drgy: I don't quite remember, there were many changes like this during last few months, but this just seems like something I either read in documentation or consulted with @bbenligiray

packages/validator/templates/0.1.0/config.json
drgy: Please correct me if I didnt understand whats the problem here, I think you are asking why there is `apiSpecifications` in config template, but also in OIS template. I believe its pretty obvious why is it in OIS, in config it is needed to check if all the security schemes from api specifications are also in `environment.securitySchemes`. This check can't be done in OIS, since it doesnt have full specification so it must be done here.

packages/validator/templates/1.0.0/apiSpecifications.json
drgy: This should be copied from what was suggested in [previous PR](https://github.com/api3dao/airnode/pull/195#discussion_r578886984)

packages/validator/docs/regex.md
drgy: I see, thats pretty confusing would something like this help:
> anything thats between `{` and `}` without the surrounding curly brackets (usage example at parameters in paths)

packages/validator/templates/1.0.0/ois.json
```diff
@@ -0,0 +1,184 @@
...
+  },
+  "version": {
+    "__regexp": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
+  },
+  "apiSpecifications": {
+    "__template": "apiSpecifications.json",
+    "paths": {
```
sassmilic, line 13: Why aren't these params in `apiSpecifications.json`?

packages/validator/templates/1.0.0/ois.json
```diff
@@ -0,0 +1,184 @@
...
+              ]
+            }
+          }
+        }
+      }
+    }
+  },
+  "endpoints": {
+    "__template": "endpoints.json",
+    "__arrayItem": {
```
sassmilic, line 84: Same question.

packages/validator/templates/1.0.0/endpoints.json
```diff
@@ -0,0 +1,71 @@
+{
+  "__arrayItem": {
+    "name": {},
```
sassmilic, line 3: Could anything bad happen if we don't have regex to check for certain characters?

sassmilic (changes requested):
I made a few minor comments.

Can you get at least one other person to check the json files (e.g. `ois.json`, `endpoints.json`, etc)

Also: it seems like the best way to test this is just to feed it a bunch of OAS files, run them through oas-to-ois converter, and then the validator. How many real OAS files did you test on? (Since I believe you did it for Open Banking Project.)

Also, have you seen some of the validators implemented for OAS? https://github.com/OAI/OpenAPI-Specification/blob/main/IMPLEMENTATIONS.md

I'm asking because it might be a good resource (for future features and potential refactoring.)

packages/validator/templates/1.0.0/endpoints.json
drgy: There is no requirement in the documentation, but it might be worth to ask people working on integrations to try submit bunch of invalid specifications to see if validator covers all edge cases.

packages/validator/src/commands/convertCmd.ts
```diff
@@ -0,0 +1,50 @@
+import fs from 'fs';
```
amarthadan, line 1: Also, if I'm converting something I would expect the output (not the field `output` but the output of the command) to be the converted JSON. Or, at least to have an option output the result into a file.

packages/validator/src/commands/validateCmd.ts
```diff
@@ -0,0 +1,24 @@
+import * as utils from './utils';
+import { validate } from '../validator';
+import { Log, templates } from '../types';
+
+let template = process.env.npm_config_template || process.argv[2];
```
amarthadan, line 5: Using `let` and reassigning variables is usually not a good practice, especially when the content is different (like here). Code is hard to follow that way.

packages/validator/src/commands/validateCmd.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const messages: Log[] = [];
+
+if (templates[template.toLowerCase() as keyof typeof templates]) {
+  template = utils.getPath(templates[template.toLowerCase() as keyof typeof templates], messages, version);
+} else if (version) {
+  messages.push({
+    level: 'warning',
+    message: 'Version argument will be ignored when validating provided template file',
+  });
```
amarthadan, line 17: You're mixing using `logger` functions and writing the messages yourself. Please, stick to one.

packages/validator/src/processor.ts
```diff
@@ -0,0 +1,183 @@
...
+
+        if (!paramPath.length) {
+          paramPath.push('[0]');
+        }
+
+        // validate each item in specs
+        for (let i = 0; i < specs.length; i++) {
+          nonRedundantParams.push({});
+
+          paramPath[paramPath.length - 1] = paramPath[paramPath.length - 1].replace(/\[[0-9]+\]$/, '') + `[${i}]`;
```
amarthadan, line 86: I would prefer keeping regular expressions at one place (so the same regular expressions don't have to be initialized multiple times) and add a comment explaining what each regular expression does.

packages/validator/src/validator.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  try {
+    specs = JSON.parse(specs);
+  } catch (e) {
+    return { valid: false, messages: [logger.error(`${specsPath} is not valid JSON: ${e}`)] };
+  }
+
+  return validateJsonWithTemplate(specs, templateName, returnJson);
+}
```
amarthadan, line 120: So, what are these two functions good for? As far as I can tell they're not used anywhere, are they?

amarthadan (changes requested):
I  was thinking, wouldn't be better to have `messages` initialized as a singleton at the beggining and accessible from anywhere so it doesn't have to be passed around? Basically the same way as logging usually works but instead printing messages they will be stored and printed at the end. I think it would make the code much cleaner.

packages/validator/README.md
```diff
@@ -0,0 +1,96 @@
+# `@airnode/validator`
+
+Package for json specifications validation. Validation process can be configured with validator templates.
+
+## Usage
+
+The validator can be run as an NPM script, by providing template of specification, and the path to the JSON file of specification that will be validated:
+```sh
+npm run validate --template="[template]" --specs="[specsFile]"
```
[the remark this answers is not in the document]
drgy: changed in 9e2fcc5

packages/validator/docs/actions.test.ts
```diff
@@ -0,0 +1,234 @@
+import * as convertor from '../src/convertor';
```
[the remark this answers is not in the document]
drgy: They are not connected to specific .ts file, however they are connected to .md file, so it makes sense to keep them in the same directory, as they contain examples from the documentation.

packages/validator/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "files": [
+    "src"
+  ],
+  "scripts": {
+    "test": "jest",
+    "validate": "ts-node src/commands/validateCmd.ts",
+    "convert": "ts-node src/commands/convertCmd.ts"
+  },
+  "dependencies": {
+    "ts-node": "^9.0.0"
```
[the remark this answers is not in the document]
drgy: Moved in 3dc19ea

packages/validator/src/commands/validateCmd.ts
```diff
@@ -0,0 +1,24 @@
+import * as utils from './utils';
+import { validate } from '../validator';
+import { Log, templates } from '../types';
+
+let template = process.env.npm_config_template || process.argv[2];
+const specs = process.env.npm_config_specs || process.argv[3];
+const version = process.env.npm_config_templateVersion || process.argv[4];
```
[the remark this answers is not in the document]
drgy: The commands are very simple and yargs just seem just like overkill and not worth the effort in this case. Is it even possible for the commands to stay `yarn run validate ...`/`yarn run convert ...` instead of something like `yarn run validator validate ...`/`yarn run validator convert ...`? Either way in 962e3d8 this issue is fixed.

packages/validator/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "version": "1.0.0",
+  "main": "src/validator.ts",
+  "directories": {
+    "lib": "src",
+    "test": "__tests__"
+  },
+  "files": [
+    "src"
+  ],
+  "scripts": {
```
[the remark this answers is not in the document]
drgy: I hope I didnt forget anything else in 26fcf31, could you please check that?

packages/validator/package.json
```diff
@@ -0,0 +1,21 @@
+{
```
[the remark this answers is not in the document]
drgy: Added in 26fcf31

packages/validator/docs/catch.md
```diff
@@ -0,0 +1,196 @@
...
+---
+
+## Special keywords
+
+Some keywords in `__message` will be replaced with certain values:
+
+- `__value` - if containing parameter is string in specification, `__value` will be replaced with its value
+- `__path` - path of the parameter in the current specification
+- `__prefix` - path to root parameter of the current template in case it is nested inside another template
+- `__path` - combination of `__prefix` and `__path`
```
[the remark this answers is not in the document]
drgy: Fixed in 1509129

packages/validator/src/commands/utils.ts
```diff
@@ -0,0 +1,52 @@
+import { Log } from '../types';
+
+const fs = require('fs');
+const path = require('path');
```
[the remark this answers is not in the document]
drgy: f99609f should be all

packages/validator/src/commands/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @param template
+ */
+function getLatestPath(template: string): string {
+  for (const version of templateVersions) {
+    if (fs.existsSync(path.resolve(__dirname, `../../templates/${version}/${template}`))) {
+      return path.resolve(__dirname, `../../templates/${version}/${template}`);
+    }
+  }
+
+  return '';
```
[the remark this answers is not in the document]
drgy: Fixed in 345f489

packages/validator/src/commands/convertCmd.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const oas2ois = 'OAS2OIS.json';
+const ois2config = 'OIS2Config.json';
+
+if (process.env.npm_config_template) {
+  console.log(
+    JSON.stringify(
+      convert(process.env.npm_config_specs || process.argv[3], process.env.npm_config_template),
+      null,
+      '\t'
```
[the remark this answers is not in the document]
drgy: Seems like specifications in the OIS repository use 2 spaces, so it is indeed better, replaced in d6d453b

packages/validator/src/commands/validateCmd.ts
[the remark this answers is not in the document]
drgy: I just took a look at yargs, it seems unnecessarily complicated to switch to yarn tbh. Using yargs might be useful in more complicated cli program, but validator is just a command with few arguments, which npm parses automatically...

packages/validator/src/commands/convertCmd.ts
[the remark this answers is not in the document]
drgy: Under the hood they are the same thing, there was a need for a conversion tool, rather to build it from ground up it was more efficient to extend already existing validator. It is pretending to be 2 different things on the surface not to cause confusion, it just makes more sense to separate validation and conversion as it might be quite counter-intuitive to run validation command when you want to convert your OAS to OIS. Keep in mind users that have no idea how it works and what validator templates are might be using this tool.
Validator should not return `output`, this was bug and it was fixed in a70ef65.

drgy: This is a definitely useful feature, added in aea2e9b

packages/validator/src/commands/validateCmd.ts
drgy: I would usually agree, however in this case, there is very few uses of the variable and it might be even harder to follow when rewritten using multiple variables.

packages/validator/src/commands/validateCmd.ts
drgy: This place got fixed somewhere, but some parts of code are older than `logger` and even though I'm trying to replace everything with `logger` it is possible that it is still not used here and there...

packages/validator/src/processor.ts
drgy: I dont think thats something worth doing, as most of the regular expressions are very simple and its obvious what they are doing, there is maybe 4 of them that are a bit trickier. They also don't really repeat, so it wouldn't help with repeated initialization.

packages/validator/src/validator.ts
drgy: They become useful when validator is used as a package, check the [deployer integration PR](https://github.com/api3dao/airnode/pull/391) for example.

packages/validator/docs/actions.md
```diff
@@ -0,0 +1,292 @@
+# Actions
```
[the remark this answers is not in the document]
drgy: This should be definitely done, but its a bit tricky, I will have to check what options do we have and it might take a bit time, so I created [new issue](https://github.com/api3dao/airnode/issues/423) specifically for this.

drgy:
> I was thinking, wouldn't be better to have `messages` initialized as a singleton at the beggining and accessible from anywhere so it doesn't have to be passed around? Basically the same way as logging usually works but instead printing messages they will be stored and printed at the end. I think it would make the code much cleaner.

The code would be definitely way cleaner, this was how I originally tried to write the validator actually. However this design pattern doesn't work well with conditions / `__catch`. The issue is while validating the template the context must be often temporarily switched, functional design pattern just seems most viable for this problem.

packages/validator/src/commands/convertCmd.ts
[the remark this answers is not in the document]
drgy: > so that our own templates will be used instead of some provided custom ones, right?


`from`/`to` in convertor is just to help users, so they don't have to look for common conversions, but ofc they can make their own conversion templates and run it as mentioned in the readme: 
`npm run convert --template="templates/3.0.0/OAS2OIS.json" --specs="exampleSpecs/OAS.specs.json"`

sassmilic (approved):
lgtm, but i only reviewed changes since i last reviewed

packages/validator/src/processor.ts
```diff
@@ -0,0 +1,183 @@
...
+import { validateOptional } from './validators/optionalValidator';
+import { validateAny } from './validators/anyValidator';
+import { Log, Result, Roots } from './types';
+import { execute } from './utils/action';
+import { validateParameter } from './validators/parameterValidator';
+import { validateCatch } from './validators/catchValidator';
+import { validateTemplate } from './validators/templateValidator';
+
+/**
+ * Recursion validating provided specification against template
```
Siegrift, line 14:
```suggestion
 * Validates provided specification against template
```

packages/validator/src/processor.ts
```diff
@@ -0,0 +1,183 @@
...
+ * @param specs - specification that is being validated
+ * @param template - template the specification is validated against
```
Siegrift, line 16:
```suggestion
 * @param template - template against which the specification is validated
```

packages/validator/src/processor.ts
Siegrift: I do agree with @amarthadan. When reading the code I have to understand what it does. If this was instead a constant imported from somewhere, than the variable name would "carry" the meaning. It will also make a refactor easier (e.g. change the format of indexing).

This is the only regex in this file, but there are many other regexes in other files.

packages/validator/src/utils/action.ts
```diff
@@ -0,0 +1,36 @@
+import { insertValue, replaceParamIndexWithName, replacePathsWithValues } from './utils';
+import { Roots } from '../types';
+
+export function execute(specs: any, template: any, currentPath: string[], roots: Roots) {
```
Siegrift, line 4: Can we replace `any` with `object` ? (It will at least carry a better type name)

packages/validator/src/utils/action.ts
```diff
@@ -0,0 +1,36 @@
...
+  for (const action of template) {
+    let targetPath;
+
+    try {
+      targetPath = JSON.parse(
+        (action['__insert'] || action['__copy'])['__target'].replace(/(?<!\\)[']/g, '"').replace(/\\'/g, "'")
```
Siegrift, line 10: I agree with @amarthadan previous suggestion to make constants out of the regexes. If the regex was a constant variable It would be easier to understand what it does.

packages/validator/src/utils/action.ts
Siegrift: But I think we can type this even better no?

packages/validator/src/utils/action.ts
Siegrift: Also, a comment above the regex would be nice.

packages/validator/src/validator.test.ts
```diff
@@ -0,0 +1,991 @@
...
+
+const apiTemplate = JSON.parse(fs.readFileSync(getPath('apiSpecifications.json', messages), 'utf8'));
+const endpointsTemplate = JSON.parse(fs.readFileSync(getPath('endpoints.json', messages), 'utf8'));
+const oisTemplate = JSON.parse(fs.readFileSync(getPath('ois.json', messages), 'utf8'));
+
+describe('validator templates loaded', () => {
+  expect(messages).toStrictEqual([]);
+});
+
+const validAPISpecification = `{
```
Siegrift, line 25: Nit: consider having some fixtures folder with the contents of the specifications and import them instead of inlining.

packages/validator/src/validators/catchValidator.ts
```diff
@@ -0,0 +1,49 @@
...
+  messages: Log[],
+  paramPath: string[],
+  paramPathPrefix: string[],
+  rootSpecs: any
+): Log[] {
+  if (!messages.length || !template['__catch']) {
+    return messages;
+  }
+
+  let message = utils.replaceParamIndexWithName(template['__catch'], paramPath);
```
Siegrift, line 26: I strongly think the special commands should be listed in one place and imported to avoid using `__catch` in many places, but rather import it as constant. E.g.

```
import {CATCH} from '../directives'
```

(similarly for others)

packages/validator/src/validators/anyValidator.ts
```diff
@@ -0,0 +1,88 @@
...
+ * @param specs - specification that is being validated
+ * @param template - must be on the same level as specs
+ * @param paramPath - string of parameters separated by ".", representing path to current specs location
+ * @param nonRedundantParams - object containing all required and optional parameters that are being used
+ * @param roots - roots of specs and nonRedundantParams
+ * @param templatePath - path to current validator template file
+ * @returns empty array if at least one child of specs satisfies template, otherwise returns array with error message
+ */
+export function validateAny(
+  specs: any,
```
Siegrift, line 17: Can we type this better?

packages/validator/src/validators/anyValidator.ts
```diff
@@ -0,0 +1,88 @@
...
+  template: any,
+  paramPath: string[],
+  nonRedundantParams: any,
+  roots: Roots,
+  templatePath: string
+): Log[] {
+  if (!specs || typeof specs !== 'object') {
+    return [logger.error(`Required conditions not met in ${paramPath.join('.')}`)];
```
Siegrift, line 25: You have a function for this message in `messages.ts`

packages/validator/src/validators/anyValidator.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    return [logger.error(`Required conditions not met in ${paramPath.join('.')}`)];
+  }
+
+  for (const paramKey of Object.keys(specs)) {
+    let nonRedundantParamsCopy = {};
+
+    if (nonRedundantParams[paramKey]) {
+      nonRedundantParamsCopy = JSON.parse(JSON.stringify(nonRedundantParams[paramKey]));
```
Siegrift, line 61: Slightly better is to use deepCopy from lodash.

Siegrift (changes requested):
@drgy I really love all the effort put in the PR, especially tests and documentation as those made it much easier to review. Unfortunately the PR is extremely big :/

packages/validator/README.md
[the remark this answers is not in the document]
drgy: Switching to yarn would bering quite a bit of overhaul to this, would it have some considerable benefits? Because otherwise I would keep it as npm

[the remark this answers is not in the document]
drgy: > What is that commit part of? I still see this unchanged in this PR.

Yeah I had to switch it back, because I realized there are issues with arguments. It will stay as npm at least for this PR, if it would be causing issues, it can be rewritten to be using yarn in different PR.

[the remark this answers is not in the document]
Siegrift: Mhh, strange. What was those issues?

packages/validator/docs/actions.md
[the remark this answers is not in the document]
drgy: Yeah there shouldnt be many onconsistencies, they can stay for now. I will look into "autodocumentation" right after we get this merged.
> Also, will the validator README be part of API3 docs?
Good question, I will bring it up on the next dev call.

packages/validator/README.md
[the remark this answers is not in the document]
drgy: npm automatically parses arguments, which are then accessible as globals. It doesnt seem yarn does this, so there is need to parse it manually or add another library...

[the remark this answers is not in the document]
Siegrift: I am pretty sure it work (I have used yarn to run scripts with arguments) and it's also what they say in https://classic.yarnpkg.com/en/docs/cli/run/#toc-yarn-run-script

packages/validator/src/commands/validateCmd.ts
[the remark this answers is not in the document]
amarthadan: I'm pretty sure you won't be able to use named arguments instead of possitional ones without it. You're currently using npm for that but when the package is build and installed you're not using npm (nor yarn for that matter) to run it. It's run via `node` binary and that doesn't provide those variables. So, using yargs would also solve your npm vs. yarn problems.

Other than that, it provides the CLI with automatic help and a descriptive way to define arguments which is IMO better. It doesn't need to be in this PR but I would definitelly go this way since `npm_config_*` is a no go anyway.

packages/validator/docs/actions.test.ts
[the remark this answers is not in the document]
amarthadan: I don't agree because if I see a `docs` directory in some project and I'm looking for some code that's the last place where I would be looking.

packages/validator/src/commands/convertCmd.ts
```diff
@@ -11,11 +11,7 @@ const ois2config = 'OIS2Config.json';
 
 if (process.env.npm_config_template) {
   console.log(
-    JSON.stringify(
-      convert(process.env.npm_config_specs || process.argv[3], process.env.npm_config_template),
-      null,
-      '\t'
-    )
+    JSON.stringify(convert(process.env.npm_config_specs || process.argv[3], process.env.npm_config_template), null, 2)
```
amarthadan, line 14: If calling `JSON.stringify(something, null, 2)` is so common maybe we should have a helper method for that.

packages/validator/README.md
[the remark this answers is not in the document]
drgy: Not sure what I should look for in the link, but let me explain the issue in more detail. If I run following command `npm run validate --template="dir/template.json"`, all I have to do to access the `template` argument in the validator is `process.env.npm_config_template`. When running `yarn run validate --template="dir/template.json"`, the value of `template` argument obviosly won't be in npm global variable, but afaik needs to be parsed from `process.argv` either manually or using a library. Thats why using npm in this package is more viable, but if consistency is a problem I can later look for a simpler library then yargs that would do the job.

packages/validator/src/commands/validateCmd.ts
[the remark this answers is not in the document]
drgy: When used as a package it will call appropriate functions in `validator.ts` and `convertor.ts`, you can check how its used [here](https://github.com/api3dao/airnode/blob/f21eeb28b986360c39e7232a90bd49de53ae6d6d/packages/deployer/src/utils/files.ts#L11). Scripts in `commands/` directory are there only to call those functions based on what is in the arguments and in validator package they won't be used at all.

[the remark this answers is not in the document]
amarthadan: Sorry, I didn't mean package as an included dependency I meant as a separate package installed basically as an executable with a CLI.

packages/validator/src/processor.ts
drgy: Added in fae0353b9739d359b219e25b9be72f72960c1dde

packages/validator/README.md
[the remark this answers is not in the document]
Siegrift: Oh, I see. Npm parses the named parameters for you and yarn does not. 

We definitely should not rely on people using npm. Most people use npm and yarn interchangeably, so I agree with what @amarthadan was saying initially and migrate to yargs. (But it's OK to be done in a separate PR)

packages/validator/docs/actions.test.ts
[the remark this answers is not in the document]
Siegrift: I mean it's fine to keep them in here now. When #423 is implemented this will probably change anyway...

packages/validator/README.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+Templates are case-insensitive, valid templates are: `config`, `OIS`, `apiSpecifications`/`apiSpecs` and `endpoints`. If arguments are in order `template`, `specs` command can be simplified: `npm run validate [template] [specsFile]`:
+```sh
+npm run validate config exampleSpecs/config.specs.json
+```
+
+Validator will automatically validate the latest available version of provided template, in case specific version should be used in validation, it can be passed as `--templateVersion` argument or as the last argument if they are ordered:
```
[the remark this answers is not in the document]
drgy: Thats quite good suggestion, the versioning needs a bit of work though, it will be probably modified on [this branch](https://github.com/api3dao/airnode/tree/validator-templates), where I will try to include this.

README.md
```diff
@@ -24,6 +24,8 @@ It houses the following packages:
 
 [**protocol**:](https://github.com/api3dao/airnode/tree/master/packages/protocol) Contracts that implement Airnode RRP (request–response protocol)
 
+[**validator**:](https://github.com/api3dao/airnode/tree/master/packages/validator) A package that can be used to validate [OIS](https://github.com/api3dao/api3-docs/blob/master/airnode/ois.md), [config.json](https://github.com/api3dao/api3-docs/blob/master/airnode/config-json.md) and [security.json](https://github.com/api3dao/api3-docs/blob/master/airnode/security-json.md) files
```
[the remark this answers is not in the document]
drgy: Changed in 6300bb5f9aa8811cb00520d9c0a557d68c66f1f0, as it might be impractical to change links every time new version is released.

packages/validator/README.md
```diff
@@ -0,0 +1,98 @@
...
+```sh
+npm run validate config exampleSpecs/config.specs.json 0.1.0
+```
+
+Custom templates can be used, by providing path to the validator template file in place of `template`:
+```sh
+npm run validate templates/1.0.0/config.json exampleSpecs/config.specs.json
```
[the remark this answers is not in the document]
drgy: Generally the default behavior is that relative path starts in the directory where the command is ran from, so I dont think it needs to be explicitly stated, otherwise it would have to be mentioned everywhere including the `specs` argument which is quite redundant tbh.

packages/validator/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+# `@airnode/convertor`
+
+Built-in validator extension capable of conversions between various specifications, conversions are configured using validator templates with conversion actions.
+
+## Usage
+
+Convertor works the same way as validator and can be invoked with the `convert` command, for example:
+```sh
+npm run convert --template="templates/3.0.0/OAS2OIS.json" --specs="exampleSpecs/OAS.specs.json"
```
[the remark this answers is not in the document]
drgy: Conversion is configured in a single template, which transforms provided specification json to target specification. `from`, `to` arguments are there to simplify this command and will be probably used much more. However this version of the command is necessary for creating new conversions.

packages/validator/README.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+Conversions can be invoked without providing any template, specifying which format provided specification is in and to which format it should be converted into, is enough:
```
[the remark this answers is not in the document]
drgy: Yes for known specifications `from`, `to` arguments can be used, for unknown specifications command described above must be used.

packages/validator/README.md
```diff
@@ -0,0 +1,98 @@
...
+```sh
+npm run convert --from="OAS" --to="Config" --specs="exampleSpecs/OAS.specs.json"
```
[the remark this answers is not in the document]
drgy: 1 - 3: this will be reworked in another PR, probably using [your suggestion](https://github.com/api3dao/airnode/pull/388#discussion_r674639458)
4:
> From/to formats are case-insensitive

packages/validator/README.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+From/to formats are case-insensitive, these are valid conversions:
```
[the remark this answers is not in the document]
drgy: > Why case insensitive? Any reason for that?

it might be annoying to type OIS, so lets say we set it to ois, now every time you accidentally type OIS it will be not recognized, so why to force one specific way when it doesnt actually matter as there are no collisions....

> Also, what is actually format? It's still a template no?

Templates only specify the conversion/validation of specification not how the specification should be actually formatted. In this context format is a well defined structure of how provided json will look like so the conversion to different well defined structure can be performed, not sure how else to call it as I don't want to use "specification" nor "template" there because it would break the definition in [here](https://github.com/api3dao/airnode/blob/validator/packages/validator/docs/basics.md).

> The part From/to doesn't look good in github README. Maybe reword it to avoid the slash?

I think it looks better than "From and to", not sure what are the other options.

packages/validator/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+| From | To |
+| ----- | -----|
+| oas | ois |
+| oas | config |
+| ois | config |
+
+Another example of `convert` command:
+```sh
+npm run convert oas config "exampleSpecs/OAS.specs.json"
```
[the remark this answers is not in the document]
drgy: I strongly disagree with removing the shorthand version as it proved to be very useful when doing multiple validations/conversions. It will stay for now and if it would cause significant difficulties for users it can be removed.

packages/validator/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+```
+{
+  valid: boolean,
+  messages: { level: "error" | "warning", message: string }[],
+  output: object
+}
+```
+
+Alternatively command can be ran with argument `--specs_only` or just `--so`, which will return only the converted specification.
```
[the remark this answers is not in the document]
drgy: Removed in f8013c9bee42ed1262fafd71c2c18a243e0d27c2

packages/validator/docs/actions.md
```diff
@@ -0,0 +1,292 @@
...
+      "quantity": 10
+    },
+    {
+      "name": "item2",
+      "quantity": 3
+    }
+  ]
+}
+```
+### Expected output
```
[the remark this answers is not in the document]
drgy: 1. These are supposed to be only examples showcasing various validator features, there is no command to run them. If you want to you can copy the template and specification into json files and run it with commands described in the readme.
2. Is there something inherently bad about `Expected output`?

packages/validator/docs/actions.md
```diff
@@ -0,0 +1,292 @@
...
+      }
+    }
+  }
+}
+```
+---
+
+## Arrays in `__target`
+
+Arrays in `__target` can be accessed as regular items in arrays with `[x]`, where `x` is index of the item that will be accessed. However, in arrays inside `__target` last item of the array can be accessed with `[_]` and new item can be inserted into array with `[]`.
```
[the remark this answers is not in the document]
drgy: Because this part is quite difficult to comprehend, when Sasa was going through this it was causing confusion about what the last element of the array is as it is not the last item of resulting array rather then last item of intermediary that is currently build. Bringing there something that already has a unified meaning to certain degree would cause even more confusion, also I can't really imagine a scenario where anything else than `[-1]` would be used, so perhaps not giving the possibility to access other elements might be a way to avoid undesired behavior.

packages/validator/docs/actions.md
```diff
@@ -0,0 +1,292 @@
...
+      }
+    ]
+  }
+}
+```
+---
+
+## All
+
+Action can be also performed on every parameter somewhere in the parameter path, this is achieved by using `__all`.
```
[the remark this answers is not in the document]
drgy: > Can I use multiple alls like: ['fieldA', '__all', '__all', 'fieldB']?

all keywords can be used in all possible combinations how many times you want

> Does it work for objects?

this is in the example right below it

packages/validator/docs/actions.test.ts
```diff
@@ -0,0 +1,234 @@
...
+      company: {},
+      inventory: {
+        __arrayItem: {
+          name: {},
+          quantity: {},
+        },
+        __actions: [
+          {
+            __copy: {
+              __target: "[ 'backups', '[[ \\'/\\', \\'company\\' ]]' ]",
```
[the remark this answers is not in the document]
drgy: > __target: `[ "backups", "[[ '/', 'company' ]]" ]`,

is not valid json

> Why can't I just write [["/company"]]?

some keys in specifications might be endpoints like this one: `/endpt/path`, which would break it

packages/validator/docs/actions.test.ts
```diff
@@ -0,0 +1,234 @@
+import * as convertor from '../src/convertor';
+
+describe('actions (docs)', () => {
+  it('copy action', () => {
+    const template = {
+      company: {},
```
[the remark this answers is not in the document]
drgy: Good question, not quite those values can have only primitive types (no object/array), added this to readme as it wasnt mentioned anywhere (881cb5706a1617e5ded4714a266d4be90066129a)

packages/validator/docs/actions.test.ts
```diff
@@ -0,0 +1,234 @@
...
+          },
+        },
+      ],
+    };
+
+    const specs = {
+      original: {},
+    };
+
+    expect(convertor.convertJson(specs, template)).toMatchObject({
```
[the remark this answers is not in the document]
drgy: I think there were some cases in which `.toEqual` didnt catch something, but I cant remember what it was. Now it seems to be working just fine so replaced in 7659a7f906b611da566fb935849aaadb67293904

packages/validator/docs/actions.test.ts
```diff
@@ -0,0 +1,234 @@
...
+            value: 'inserted',
+          },
+        },
+      },
+    });
+  });
+
+  it('arrays', () => {
+    const template = {
+      __objectItem: {
```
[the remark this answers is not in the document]
drgy: [yes](https://github.com/api3dao/airnode/blob/validator/packages/validator/docs/regex.md)

packages/validator/docs/actions.test.ts
```diff
@@ -0,0 +1,234 @@
...
+            {
+              __copy: {
+                __target: "[ 'vehicles[]' ]",
+              },
+            },
+            {
+              __insert: {
+                __target: "[ 'vehicles[_]' ]",
+                __value: {
+                  type: '{{0}}',
```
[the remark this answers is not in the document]
drgy: it is key on specific position in the path, it is all explained in [the readme](https://github.com/api3dao/airnode/blob/validator/packages/validator/docs/dynamic_params.md#retrieving-key-from-parameter-path)

packages/validator/docs/basics.md
```diff
@@ -0,0 +1,232 @@
+# Basics
+
+## Terminology
+
+In the whole documentation unified terminology will be used, here is the definition of some important keywords:
+
+- **specification** - json structure, that will be checked for compliance with well-defined format
```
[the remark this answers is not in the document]
drgy: not quite format is something different then template. Template just describes format for validator, note that not whole format has to be described, just the part important for validation.

packages/validator/docs/basics.md
```diff
@@ -0,0 +1,232 @@
...
+- **template** - json structure, which defines format of valid specification
```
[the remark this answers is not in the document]
drgy: this sounds more like its okay if its format of any valid specification not the one that we are validating necessarilly.

packages/validator/docs/conditions.md
```diff
@@ -0,0 +1,443 @@
...
+            }
+          }
+        }
+      ]
+    }
+  }
+}
+```
+---
+### Valid specification
```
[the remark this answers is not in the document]
drgy: It is missing intentionally, as output of successful validation is specified in the basics, since it looks the same everywhere, here it would be just redundant and make the readme much longer without providing any value.

packages/validator/docs/conditions.md
```diff
@@ -0,0 +1,443 @@
+# Conditions
```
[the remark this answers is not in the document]
drgy: It might seem a bit tricky, but it is quite intuitive, if you attempt to write one validation its very easy to use.

packages/validator/docs/template.md
```diff
@@ -0,0 +1,64 @@
+# Nested templates
+
+Templates can be recursive. That is, a template can contain a pointer to another template in the form of a relative path provided in `__template`.
```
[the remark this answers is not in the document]
drgy: we are not providing any validation for correctness of templates and honestly i dont think its something worth implementing. If you think about it no scripting language bans you from creating infinite loops, I dont see a reason why we should limit some functionalities of the templates.

packages/validator/exampleSpecs/config.specs.json
```diff
@@ -0,0 +1,209 @@
+[
+	{
+		"ois": [
```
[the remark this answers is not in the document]
drgy: not sure what do you mean, but it should match [this](https://docs.api3.org/next/reference/deployment-files/config-json.html)

packages/validator/jest.config.js
```diff
@@ -0,0 +1,6 @@
+const config = require('../../jest.config.base');
+
+module.exports = {
+  ...config,
+  // Add custom settings below
```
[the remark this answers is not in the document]
drgy: Its there for consistency across packages :laughing: 
Removed in bb5dd9ed7457f95e1d4bf7c31a817ea174caab0d

packages/validator/src/commands/convertCmd.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const oas2ois = 'OAS2OIS.json';
+const ois2config = 'OIS2Config.json';
+
+if (process.env.npm_config_template) {
+  console.log(
+    JSON.stringify(convert(process.env.npm_config_specs || process.argv[3], process.env.npm_config_template), null, 2)
+  );
+} else {
+  const from = (process.env.npm_config_from || process.argv[2] || '').toLowerCase();
```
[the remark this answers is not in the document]
drgy: It is only the specification names in which it is a bad idea to allow only lowercase/uppercase. If we allow lowercase only, then it might not be intuitive to type `ois`, but if we allow only `OIS` then it doesn't look consistent with `config`. So the solution for this is to allow user to type it as they feel comfortable, since it can't cause any trouble on our end.

packages/validator/src/commands/utils.ts
```diff
@@ -0,0 +1,51 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import { Log } from '../types';
+
+export const templateVersions = fs
+  .readdirSync(path.resolve(__dirname, '../../templates'), { withFileTypes: true })
+  .filter((dirent: { isDirectory: () => any }) => dirent.isDirectory())
```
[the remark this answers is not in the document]
drgy: generally it is a structure holding inode and other useful information of an entry in the filesystem

packages/validator/src/convertor.test.ts
```diff
@@ -0,0 +1,71 @@
...
+          endpointName: '/simple/token_price/{id}',
+          oisTitle: 'CoinGecko API V3',
+        },
+      ],
+    },
+  },
+];
+
+describe('convertor', () => {
+  it('OAS2OIS', () => {
```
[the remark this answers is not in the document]
drgy: Changed in 7d9a7cf75b9b89412b4baedc2c2201c3321ab2de

packages/validator/src/utils/action.ts
drgy: > Can we replace any with object ? (It will at least carry a better type name)

no, action can be executed on primitives as well

> But I think we can type this even better no?

I don't think so, any suggestions?

packages/validator/src/utils/action.ts
drgy: Done in fae0353b9739d359b219e25b9be72f72960c1dde

packages/validator/src/validator.test.ts
drgy: I dont think it would look good, also it might be difficult to detect the test that is failing, but it is true this file is way too big, I rather split it into two in 6cda8a15af3e479132b7d9d6f9378b8578c8f274

packages/validator/src/validators/anyValidator.ts
drgy: I dont think so, any suggestions?

packages/validator/src/validators/anyValidator.ts
drgy: Fixed in 0529c86051a72a666d1be1f8107b3a3fb80d3cff

packages/validator/src/validators/anyValidator.ts
drgy: it might be slightly better, however it is not worth importing a whole library just for this

packages/validator/src/validators/parameterValidator.ts
```diff
@@ -0,0 +1,102 @@
...
+        template,
+        nonRedundantParams,
+        specs[processedParam]
+      );
+      nonRedundantParams = nonRedundantParams[processedParam];
+    }
+
+    if (!Array.isArray(currentSpecs)) {
+      return [
+        logger.error(
```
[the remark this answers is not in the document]
drgy: Done in c31e36654447f6c2ff987ce3dab3d168cf216f05

packages/validator/templates/0.1.0/config.json
```diff
@@ -0,0 +1,253 @@
+{
```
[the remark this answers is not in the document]
drgy: It seems to be formatted, but i dont think eslint formats json out of the box

packages/validator/src/commands/convertCmd.ts
drgy: Seems to be used only 3 times, I don't think its worth to clutter it with another function

drgy:
@Siegrift thanks for the review, I think I reacted to all the comments, here are the ones that didnt have comment option for some reason:

> [What is that commit part of? I still see this unchanged in this PR.](https://github.com/api3dao/airnode/pull/388#discussion_r674635595)

it was changed back to npm, as yarn has some issues with arguments, which is explained more [here](https://github.com/api3dao/airnode/pull/388#discussion_r675410067)

> [Also, shouldn't it say: --specs="[specFile] instead of --specs="[specsFile] (singular)](https://github.com/api3dao/airnode/pull/388#discussion_r674637457)

if you are pedantic perhaps there should be spec, however for consistency sake there is specs, i dont think it matters that much and it can stay

> [Not sure I understand, why not just {(.*)} ?](https://github.com/api3dao/airnode/pull/388#discussion_r674713040)

It uses lookbehind and lookahead so it matches only the contents between the braces not the braces itself`

packages/validator/src/commands/validateCmd.ts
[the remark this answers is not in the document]
drgy: I see, if yargs supports this then it should be integrated into validator, but in another PR perhaps

amarthadan (approved):
I'd say let's merge this and deal with any other issues with pretty, small PRs, one issue at a time :pray: :sweat_smile:

drgy:
> I'd say let's merge this and deal with any other issues with pretty, small PRs, one issue at a time

I already fixed the minor issues from the reviews here, I dont think its a problem to finish up the minor ones here.
To be able to merge I still need approval from @Siegrift

Issue #367, opened 2021-06-02
Building Docker container


joshcaskie:
## Description

Building the docker container for Airnode is giving me errors.

## Version

OS: Windows 10 Pro 64-bit
Docker Desktop: 3.3.1
Node.js version: 12.13.1
Yarn version: 1.22.10
npm: 6.12.1

I'm building from the master branch cloned at [https://github.com/api3dao/airnode.git](url)

## Present Behaviour

I'm attempting to build the Airnode container by following the steps in **airnode/docker/Docker.md**. However, upon running 
`docker build -f docker/Dockerfile -t api3/airnode:latest .` 
I'm getting an error that looks like this:
![image](https://user-images.githubusercontent.com/57299645/120520438-ca881900-c3a1-11eb-9365-32aafe92a446.png)

## Expected Behaviour

I'm assuming that the command should run normally and build the container properly.

## Steps to reproduce

I'm simply cloning the repo, running `yarn run bootstrap` then running `docker build -f docker/Dockerfile -t api3/airnode:latest .`

After doing some troubleshooting, I believe it has something to do with the capitalized names in the Dockerfile as Docker might not like them (buildPrep, buildBuild, buildDependencies). 
![image](https://user-images.githubusercontent.com/57299645/120522956-55b5de80-c3a3-11eb-8adc-e2e6ec79b4e2.png)

I changed them all to fully lowercase variants and the `build` command started to execute, but then I reached another error:
![image](https://user-images.githubusercontent.com/57299645/120522879-4f276700-c3a3-11eb-9a2b-1736ffc42c0f.png)

amarthadan:
Hi @joshcaskie

> After doing some troubleshooting, I believe it has something to do with the capitalized names in the Dockerfile as Docker might not like them (buildPrep, buildBuild, buildDependencies).

I'm guessing this would be a Windows-specific issue. But if changing the image aliases to lowercase works that's an easy fix. I created an issue (https://github.com/api3dao/airnode/issues/369) to track it. Thanks for pointing it out.

 > I changed them all to fully lowercase variants and the build command started to execute, but then I reached another error

Yes, that's an expected behavior at the moment. The Docker image building process was reworked recently and the new version is expecting the packages to be available in the NPM registry which is not yet the case. We're working on releasing them but it will still take some time until we can do that. Also the documentation is outdated as well, we're working on updating it at the moment.

You can still try the deployer CLI, just not via Docker at the moment. You can do the following:
```bash
[...]

PR #389, opened 2021-06-21
Fix deployer Docker build on Windows

### LINKED ISSUE #369: Docker container for deployer not building on Windows
It looks like Windows (or rather Docker on Windows) allows aliasing images with only lowercase names.
https://github.com/api3dao/airnode/issues/367

amarthadan:
Close #369

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,5 @@
 # Build preparation
-FROM node:14.17.1-alpine3.13 as buildPrep
+FROM node:14.17.1-alpine3.13 as preparation
 
 ARG yarnRegistry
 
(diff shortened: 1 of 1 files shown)


Siegrift (commented):
Looks to me you just renamed the Dockerfile variables. How does this fix the windows docker issue?

amarthadan:
> Looks to me you just renamed the Dockerfile variables. How does this fix the windows docker issue?

Based on what was reported that should be enough https://github.com/api3dao/airnode/issues/367
It looks like Docker for Windows doesn't like capital letters in aliases.

Siegrift:
Ah, I see now.

Siegrift (approved):

PR #391, opened 2021-06-25
config.json validation in deployer

### LINKED ISSUE #136: Deployer doesn't validate config.json
The deployer should validate the `config.json` and `security.json` provided by the user using the validator package.

Tracked by: https://api3dao.atlassian.net/browse/AN-159

drgy:
1. Added validation when [parsing config](https://github.com/api3dao/airnode/blob/32c114dbcf38910a40ca676b81eb3bc1908b0ea2/packages/deployer/src/utils/files.ts#L7)
2. If deployer is launched in debug mode, deployment will be attempted even with incorrect `config.json`. Is it enough or should there be special hidden argument for this?
3. Validator checks entire `config.json`, as it seems config is parsed only once and after that it is passed between functions, any other checks [here](https://github.com/api3dao/airnode/blob/master/packages/deployer/src/evm/config.ts), would be redundant if im not missing anything.
4. Updated [config.json example](https://github.com/api3dao/airnode/blob/master/packages/deployer/config/config.json.example) to pass validation, however validator is based on [these docs](https://docs.api3.org/next/technology/deployment-files/config-json.html) and im not sure if they are up to date, are there any deployer test suites so i would be able to verify what works and what doesnt?

closes #136

### DIFF
--- a/packages/deployer/config/config.json.example
+++ b/packages/deployer/config/config.json.example
@@ -7,7 +7,7 @@
           "0x0000000000000000000000000000000000000000"
         ],
         "contracts": {
-          "AirnodeRrp": "0x5FbDB2315678afecb367f032d93F642f64180aa3"
+          "AirnodeRRP": "0x5FbDB2315678afecb367f032d93F642f64180aa3"
         },
         "id": "31337",
         "providerNames": [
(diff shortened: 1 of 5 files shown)


amarthadan:
I totally forgot about this PR, sorry about that :confused: 

After all the changes in deployer I would suggest closing this one without merging and opening a new one. What do you think?
Also, we need integration with both deployer and node itself. At least I think that was the latest consensus.

drgy:
I labeled this as blocked, the agreement was that this should wait for now at least until the templates are updated to 0.1.0 version. But I will leave the PR opened, so it can be easily found, because most of the modifications can still be just copy pasted into correct places.

Siegrift:
@drgy What is the status of this?

amarthadan:
Yes, I'm working on a different branch. Closing.

Issue #330, opened 2021-05-03
Improve developer README


Siegrift:
## Description

Improve the developer README, because it's missing `yarn build` to build the lerna packages.
I think I wasted some time until I understood what was hapenning.
Also happened to @jeebster in https://github.com/api3dao/airnode/pull/327

PR #403, opened 2021-07-01
update root documentation

Ashar2shahid:
The amount of times I fail to do is embarrassing.
 Also @Siegrift encountered this in #330 so will close that as well


Siegrift (approved):

README.md
```diff
@@ -32,6 +32,11 @@ To install dependencies, run this at the repository root:
 yarn run bootstrap
 ```
 
+To build the project, run this at the repository root:
```
andreogle, line 35: I would include something about this being a required step for many (all?) commands

andreogle (changes requested):

README.md
bbenligiray: Good point, the reader may decide to be a smartass about it

sitch:
Can you just add this as a `postinstall`?

So add to `package.json`:
```json
{
  "scripts": {
    "postinstall": "lerna bootstrap",
    ...
  }
}
```
Now after you call `yarn install` it will trigger the `yarn postinstall` automatically.  Not sure if this does exactly whats needed but worth a shot.  You can read more about pre/post scripts https://docs.npmjs.com/cli/v7/using-npm/scripts#pre--post-scripts

andreogle:
I think ^ is probably a good idea. The `postinstall` should just be:

```json
{
  "scripts": {
    "postinstall": "yarn run build",
    ...
  }
}
```

Ashar2shahid:
Yeah the post install script looks like the best way.

Siegrift (commented):
@Ashar2shahid can you finish this? :)

Ashar2shahid:
> @Ashar2shahid can you finish this? :)

just pushed a commit, tested it and it works well. You can have a look and then I'll merge it

andreogle (approved):

PR #405, opened 2021-07-02
update authorizer contract and tests according to new specs

Ashar2shahid:
This new Authorizer has a bool `airnodeIdToClientAddressToWhitelist`  that can be set along with the expiration. The client is only authorized if both the bool and the expiration satisfy. The `setWhitelistExpiration` can both set and unset `airnodeIdToClientAddressToWhitelist` thereby granting/revoking the whitelisting.  

The  old isAuthorized tests seem to be worded wrong, I've updated them but they should be changed

### DIFF
--- a/packages/protocol/contracts/authorizers/Api3Authorizer.sol
+++ b/packages/protocol/contracts/authorizers/Api3Authorizer.sol
@@ -16,7 +16,7 @@ contract Api3Authorizer is IApi3Authorizer {
   address public metaAdmin;
   mapping(address => AdminStatus) public adminStatuses;
   mapping(bytes32 => mapping(address => uint256)) public airnodeIdToClientAddressToWhitelistExpiration;
-  mapping(bytes32 => mapping(address => bool)) public airnodeIdToClientAddressToWhitelist;
+  mapping(bytes32 => mapping(address => bool)) public airnodeIdToClientAddressToWhitelistStatus;
 
   /// @param _metaAdmin Address that will be set as the meta admin
   constructor(address _metaAdmin) {
@@ -72,7 +72,6 @@ contract Api3Authorizer is IApi3Authorizer {
       expiration > airnodeIdToClientAddressToWhitelistExpiration[airnodeId][clientAddress],
       ERROR_EXPIRATION_NOT_EXTENDED
     );
-    airnodeIdToClientAddressToWhitelist[airnodeId][clientAddress] = true;
     airnodeIdToClientAddressToWhitelistExpiration[airnodeId][clientAddress] = expiration;
     emit ExtendedWhitelistExpiration(airnodeId, clientAddress, expiration, msg.sender);
   }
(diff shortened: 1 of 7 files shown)


packages/protocol/contracts/authorizers/Api3Authorizer.sol
```diff
@@ -125,7 +130,7 @@ contract Api3Authorizer is IApi3Authorizer {
   ) external view override returns (bool) {
     return
       designatedWallet.balance != 0 &&
-      !clientAddressToBlacklistStatus[clientAddress] &&
-      airnodeIdToClientAddressToWhitelistExpiration[airnodeId][clientAddress] > block.timestamp;
+      (airnodeIdToClientAddressToWhitelistStatus[airnodeId][clientAddress] ||
+        airnodeIdToClientAddressToWhitelistExpiration[airnodeId][clientAddress] > block.timestamp);
```
bbenligiray, line 134: An `airnodeId`-`clientAddress` pair is authorized if its status is true or its expiration is in the future. I know this is a bit counterintuitive but this is the functionality we want and we can potentially work around that by some remaining.
Say the integrations team has an admin (not super-admin) wallet and set the whitelist expiration time as next month. Authorizer should return true. During that, the user locked some tokens and the token lock contract, which is a super-admin, set the whitelist status to be true. Then, the authorizer should still remain true. Then, the user unlocked their tokens and the token lock contract set their whitelist status as false. Since their whitelisting set by the authorizer team hasn't still expired, the authorizer should still remain true. So this is the main reason we're keeping the status and the expiration separately and we need to give access if either of them satisfies the condition.

packages/protocol/contracts/authorizers/SelfAuthorizer.sol
```diff
@@ -138,7 +138,7 @@ contract SelfAuthorizer is ISelfAuthorizer {
   ) external view override returns (bool) {
     return
       designatedWallet.balance != 0 &&
-      !airnodeIdToClientAddressToBlacklistStatus[airnodeId][clientAddress] &&
-      airnodeIdToClientAddressToWhitelistExpiration[airnodeId][clientAddress] > block.timestamp;
+      (airnodeIdToClientAddressToWhitelistStatus[airnodeId][clientAddress] ||
+        airnodeIdToClientAddressToWhitelistExpiration[airnodeId][clientAddress] > block.timestamp);
```
bbenligiray, line 142: Same

packages/protocol/contracts/authorizers/SelfAuthorizer.sol
```diff
@@ -15,7 +15,7 @@ contract SelfAuthorizer is ISelfAuthorizer {
   IAirnodeRrp public airnodeRrp;
   mapping(bytes32 => mapping(address => AdminStatus)) public airnodeIdToAdminStatuses;
   mapping(bytes32 => mapping(address => uint256)) public airnodeIdToClientAddressToWhitelistExpiration;
-  mapping(bytes32 => mapping(address => bool)) public airnodeIdToClientAddressToBlacklistStatus;
+  mapping(bytes32 => mapping(address => bool)) public airnodeIdToClientAddressToWhitelistStatus;
```
bbenligiray, line 18: SelfAuthorizer also doesn't need blacklisting anymore, blacklisting should be implemented at the super-admin level

packages/protocol/contracts/authorizers/Api3Authorizer.sol
Ashar2shahid: Got it, updated the tests to reflect this

bbenligiray (approved):
Looks good :ok_hand:

PR #407, opened 2021-07-03
Add eslint-plugin-import

andreogle:
### Description

Adds [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import) to the Airnode monorepo and enables the recommended ruleset.

### Notable Changes

1. Sets a required order for imports.
2. Cleans up some existing imports
3. The `protocol` package now uses the same `.eslintrc.js` as the rest of the application
4. Drops `context()` in favour of `describe()` in `protocol` tests


packages/adapter/src/request-building/build-request.test.ts
```diff
@@ -1,5 +1,5 @@
 import * as build from './build-request';
-import * as fixtures from 'test/fixtures';
+import * as fixtures from '../../test/fixtures';
```
andreogle, line 2: I was previously mixing these kinds of imports. I think it's better to have them all consistent

.eslintrc.js
```diff
@@ -38,6 +44,13 @@ module.exports = {
     '@typescript-eslint/no-use-before-define': 0,
     '@typescript-eslint/no-var-requires': 0,
 
+    // eslint-plugin-import
+    'import/namespace': [2, { allowComputed: true }],
+    'import/order': ['error', {
+      groups: ['builtin', 'external', 'internal', 'sibling', 'parent', 'index', 'object', 'type'],
+      pathGroups: [{ pattern: 'mock-utils', group: 'builtin', patternOptions: { matchBase: true, nocomment: true } }],
```
andreogle, line 51: `mockEthers` needs to be at the top of some test files so this adds an exception for the `mock-utils.ts` file

amarthadan (approved):
LGTM :slightly_smiling_face:

bbenligiray:
@andreogle This replaced my `context`s in the contract tests with `describe`s :thinking:

andreogle:
Sorry should have given you a heads up. `context` hasn't been a thing in Jest for a while now: https://jestjs.io/docs/api https://github.com/facebook/jest/issues/2468 and I'm not even sure why it was still working.

I can add back the workaround (`global.context = describe`) and undo this specific change if you want though, then we can keep using `context`

bbenligiray:
They are Mocha tests though. Actually the main reason I use `context` is that I like it semantically and also it prints nice on the terminal. I don't know what nested `describe`s look like but I can live with this if it's decent.

andreogle:
I reverted the `context()` to `describe()` change here: https://github.com/api3dao/airnode/pull/410

PR #408, opened 2021-07-03
Remove dAPI-related contracts and tests

bbenligiray:
Simply removes dAPI-related contracts so that they don't clutter things up while we schedule audits. The contracts are in the following branches
https://github.com/api3dao/airnode/tree/dapi
https://github.com/api3dao/airnode/tree/data-feed-adminship


acenolaza (approved):
LGTM 👍🏻

PR #409, opened 2021-07-03
Refactor authorizers

bbenligiray:
The PR refactors the two authorizers as

```

       IRrpAuthorizer                             AirnodeCwRrpAuthorizer (old SelfAuthorizer)
                      \                          /
RankedAdminnable ---> ClientWhitelistRrpAuthorizer 
                                                 \
                                                  Api3CwRrpAuthorizer (old Api3Authorizer)

```

I started using "Rrp" in all authorizer names because I expect to have a lot of authorizers for PSP.

RankedAdminnable is very flexible and once we have this audited we can use it in many other places (but I'm mostly thinking about dAPIs). Note that there is a slight difference in functionality for these authorizers: SuperAdmins can assign and remove Admins (which will probably come in handy).

**ClientWhitelist**RrpAuthorizer is called as such because it whitelists at the client level. We could have also whitelisted at the requester level, etc.

AirnodeCwRrpAuthorizer and Api3CwRrpAuthorizer override a function from RankedAdminnable to implement the difference.

We were already working on/talking about this with @Ashar2shahid. @acenolaza was working on similar stuff for data feeds, so I figured would provide good input. Please review the whole thing and let me know if any of you want to take care of the tests.


packages/protocol/contracts/authorizers/ClientWhitelistRrpAuthorizer.sol
```diff
@@ -0,0 +1,90 @@
...
+  /// @notice Called by an admin to extend the whitelist expiration of a
+  /// client
+  /// @param airnodeId Airnode ID from `AirnodeParameterStore.sol`
+  /// @param clientAddress Client address
+  /// @param expirationTimestamp Timestamp at which the client will no longer
+  /// be whitelisted
+  function extendWhitelistExpiration(
+    bytes32 airnodeId,
+    address clientAddress,
+    uint64 expirationTimestamp
```
acenolaza, line 22: Wouldn't it be more semantic to receive the number of seconds to increase current `expirationTimestamp`? A calling contract could use solidity values like `7 days`, `2 weeks`, etc and we could make the type even smaller (uint32 maybe?).

packages/protocol/contracts/authorizers/RankedAdminnable.sol
```diff
@@ -0,0 +1,62 @@
...
+import "./interfaces/IRankedAdminnable.sol";
+
+/// @title Contract that implements multiple ranks of admins
+contract RankedAdminnable is IRankedAdminnable {
+  /// @notice Keeps the ranks of admins for each individual adminned entity
+  /// @dev This contract implements adminship for a mapping of entities, rather
+  /// than a single entity. However, the logic at the inheriting contract can
+  /// easily be adapted to use this contract to admin a single entity (see
+  /// Api3CwRrpAuthorizer.sol)
+  mapping(bytes32 => mapping(address => uint256)) public adminnedIdToAdminToRank;
```
acenolaza, line 13: I'm wondering if we really need uint256 for holding ranks 🤔 it seems to me that even with an uint8 or uint16 might have enough ranks for an entity (I might be missing something here). Also setting MAX_RANK in this contract seems to make more sense to me.

acenolaza (approved):
Just a couple of comments/questions but LGTM an I can totally see us using the RankedAdminnable.sol contract in the dapis

packages/protocol/contracts/authorizers/ClientWhitelistRrpAuthorizer.sol
bbenligiray: Hmm, good point, and then we wouldn't need the `require()` below. That being said, I expect this to used with a definite deadline, such as "we want to hold a hackathon for this API, so we authorize an admin contract that allows the participant addresses to whitelist any number of their clients, but only until July 30". But then that contract can get the current expiration time and increase it by an amount that clips at July 30. So I suppose the argument should indeed specify the amount to increase the expiration timestamp. Let me think a bit about this and let's also see what @Ashar2shahid thinks.

bbenligiray: Actually it's fine to keep unix epochs in 32 bits (the only issue being https://en.wikipedia.org/wiki/Year_2038_problem). That being said, using 32/64 bits doesn't matter in this context (wouldn't be cheaper in gas).

packages/protocol/contracts/authorizers/RankedAdminnable.sol
bbenligiray: Yeah 256 bits is overkill, but using a smaller type doesn't save any gas (when you read a single rank one full word will be read from the storage, when you read a single rank, one full word will be written to the storage). Future forks may easily change this but I don't find it a good idea to try to predict non-existent EIPs and using a smaller type than 256 bits is additional complexity so I'd rather avoid it unless there is no advantage.
I also thought about including `MAX_RANK` in this contract. Obviously the reason I didn't is that it's not used in it but then it looks like all the inheriting contracts will need to use it in some form so maybe it should be added. Again, let me think and let's get @Ashar2shahid 's opinion.

bbenligiray:
I'm busy with "other stuff" so I'm in no rush to merge this, I'll do so after Ashar's review. That being said, I decided to merge this without the tests to open PRs to implement all the other updates to the main protocol. Tests for this can be implemented in parallel with that.

Ashar2shahid (commented):
I actually added these yesterday but forgot to click on submit review :/

packages/protocol/contracts/authorizers/ClientWhitelistRrpAuthorizer.sol
Ashar2shahid: @bbenligiray wouldn't the clipping have to be done at the dAPP end? 
I think its fine either way, it does make more sense to receive the number of second to extend but then its a different "parameter" from `setWhitelistExpiration`. Could lead to confusion if someone doesn't read what each are.

packages/protocol/contracts/authorizers/RankedAdminnable.sol
Ashar2shahid: Yeah there probably should be `MAX_RANK`, also there might be a need for a modifier that is onlyMaxRank, because I can see it being reused alot.

packages/protocol/contracts/authorizers/Api3CwRrpAuthorizer.sol
```diff
@@ -0,0 +1,64 @@
...
+  /// Airnode ID-specific admin ranks (meaning that all admins are kept for ID
+  /// 0 and there will be no point in setting admin ranks for Airnode IDs other
+  /// than 0)
+  /// @param adminnedId ID of the entity being adminned (obsolete, left to
+  /// override the existing `getRank()`)
+  /// @param admin Admin address whose rank will be returned
+  /// @return Admin rank
+  // solhint-disable-next-line no-unused-vars
+  function getRank(bytes32 adminnedId, address admin) public override returns (uint256) {
+    if (msg.sender == api3Admin) return MAX_RANK;
```
Ashar2shahid, line 61: correct me if I'm wrong but isn't the point of this function to get the rank of a particular admin, if so why would we want the api3Admin to receive its own rank when its calling this function to view the rank of some other admin?

packages/protocol/contracts/authorizers/RankedAdminnable.sol
```diff
@@ -0,0 +1,62 @@
...
+
+  /// @dev Reverts if the caller's rank is not greater than or equal to `rank`
+  /// @param adminnedId ID of the entity being adminned
+  /// @param rank Rank caller's rank will be compared to
+  modifier onlyWithRank(bytes32 adminnedId, uint256 rank) {
+    require(getRank(adminnedId, msg.sender) >= rank, "Caller is ranked low");
```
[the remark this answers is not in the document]
Ashar2shahid: having thought about it, due to the way numbers are initialized  from 0 its probably harder to do lower rank number - higher precedence because everything will have been initialized with highest rank which will be a pain to workaround.

Ashar2shahid:
Some key comments from me apart from the reviews:

- File renaming, first glance the file names were really confusing maybe change to:
```
Api3RrpClientAuthorizer.sol
AirnodeRrpClientAuthorizer.sol
```
- The constructor in `AirnodeCwRrpAuthorizer.sol` and `Api3CwRrpAuthorizer.sol ` should be calling setRank  with `MAX_RANK` and getRank should be fetching this instead returning 'MAX_RANK' by default.

- I like the flexibility of `RankedAdminnable.sol ` and having the ability to describe any hierarchy depending on the use case. I'm not sure if this is needed but would a predefined enum structure of 5 ranks or so be useful if we are going to keep repeating the same pattern everywhere?

```
unauthorized
authorized
admin
superAdmin
metaAdmin
```

Ashar2shahid:
ill be merging this and working on a seprate PR

PR #410, opened 2021-07-05
Use context instead of describe

andreogle:
### Description

In a recent PR, I removed some `context()` functions as I mistakenly thought the `protocol` package was using Jest. The `protocol` package uses `hardhat` which in turn uses Mocha. This PR reverts this change.

### DIFF
--- a/packages/protocol/test/AirnodeParameterStore.js
+++ b/packages/protocol/test/AirnodeParameterStore.js
@@ -1,4 +1,4 @@
-/* globals ethers */
+/* globals context ethers */
 
 const { expect } = require('chai');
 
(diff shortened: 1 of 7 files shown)


Siegrift:
Wait, but is this really a regression? I would say the `describe` would have the same effect, no?

andreogle:
I think `context` is a bit nicer semantically than `describe`. I don't really mind either way

bbenligiray (approved):
Thanks, turns out Mocha treats the two the same as far as how thing are printed on the terminal but as Andre has said I like it slightly better due to semantics (use `describe` to specify the thing being tested and `context` to specify the exact conditions it's being tested under).

PR #412, opened 2021-07-09
Remove Typechain hacks, avoid using sed

Siegrift:
Fixes https://api3workspace.slack.com/archives/C01N6001YCE/p1621963547042900

The hacks I needed to use for TypeChain were probably caused by dependency clashed, which now seems to be resolved, but I had to add the generated files to eslintignore (not sure why).

Also avoided the using `sed`, because it is not cross platform compatible (although for that specific use case it probably was).


packages/protocol/dev-scripts/delete-flattened-licenses.js
```diff
@@ -0,0 +1,9 @@
+const replace = require('replace-in-file');
+
+const results = replace.sync({
+  files: '*.flat.sol',
+  from: /.*SPDX-License-Identifier.*\n/g,
```
andreogle, line 5: Why are we removing this "license-identifier"? I'm just curious

andreogle (approved):

packages/protocol/dev-scripts/delete-flattened-licenses.js
Siegrift: I don't know :)

andreogle: But things break without removing it?

Siegrift: No, I guess this is unrelated to core development. Maybe if you want to copy and modify the contract? Not really sure, I think @bbenligiray would know the answer.

PR #413, opened 2021-07-09
Add functional rules for ESLint

andreogle:
### Description

Adds [eslint-plugin-functional](https://github.com/jonaskello/eslint-plugin-functional) to the Airnode codebase. I think there's a lot of value in trying to keep the code base as functional as possible. However, I also don't want this to get in the way of things so I've disabled some of the default rules.

Each package can opt-in to the functional rules by adding a `.eslintrc.js` rule in it's root and extending the base config and the new "fp" config. If they do not implement their own `.eslintrc.js`, they will use the base ESLint config. Currently only the `deployer` and `protocol` packages do not extend the new functional packages.

Some notable new rules:
1. `functional/immutable-data` - This means that any new variable must be explicitly prefixed with "mutable". I like this as it makes the intention clear that a variable is going to be mutated. I don't think mutability should be the default.
2. `functional/prefer-readonly-types` - Types should include `readonly` by default
3. `functional/no-try-statement` - This one is more of a personal preference. I think try/catch should be avoided unless necessary.

The rules can be opted out of in the package or at the specific line.

Let me know what you guys think


andreogle:
Merging this as some work I'm doing at the moment will be affected by it

PR #419, opened 2021-07-14
Fixes beta protocol tests

acenolaza:
It was not easy to only fix airnodeAdmin changes so I went ahead and fixed all tests (requesterAdmin changes too). 

* I kept the name `requesterAdmin` for the requester account but I not sure if I should just used `requester` instead.
* Deleted all tests that were expecting a revert by a "only..." modifier since they were also removed in the solidity code BUT we might want to add tests around the fact that anyone can now call those methods (not only airnode or request admins).
* ~~I hardcoded the wallet path `m/0/1` and left a TODO to handle the address to path conversion in a later commit.~~ Implemented in [utils.js](https://github.com/api3dao/airnode/blob/1a36428d382bc39a602c1b518f0ba5bd32933371/packages/protocol/test/utils.js) (Thanks Burak!)
* Didn't fix authorizer related tests since those contracts no longer exists in master (we need to rebase beta-protocol on top of master first)


bbenligiray:
> I kept the name requesterAdmin for the requester account but I not sure if I should just used requester instead.

I can do that as a part of the renaming task

> Deleted all tests that were expecting a revert by a "only..." modifier since they were also removed in the solidity code BUT we might want to add tests around the fact that anyone can now call those methods (not only airnode or request admins).

I don't think that's needed, it's a bit confusing

Looks good overall, @Ashar2shahid can you review please?

packages/protocol/test/RequesterStore.sol.js
```diff
@@ -0,0 +1,49 @@
...
+
+beforeEach(async () => {
+  const accounts = await ethers.getSigners();
+  roles = {
+    deployer: accounts[0],
+    requesterAdmin: accounts[1],
+    client: accounts[2],
+    randomPerson: accounts[9],
+  };
+  const airnodeRrpFactory = await ethers.getContractFactory('AirnodeRrp', roles.deployer);
```
Ashar2shahid, line 16: Wouldn't it be better to just deploy RequesterStore, instead of the whole AirnodeRrp?

Ashar2shahid (approved):
So I referenced ed7d62c0258b3c48b65f1e07bed9d1c120f60814 commit for this PR. I highlighted 2 minor things that I think could be changed for better clarity but other than that everything looks good

packages/protocol/test/AirnodeParameterStore.js
```diff
@@ -14,9 +14,7 @@ beforeEach(async () => {
   const accounts = await ethers.getSigners();
   roles = {
     deployer: accounts[0],
-    airnodeAdmin: accounts[1],
-    updatedAirnodeAdmin: accounts[2],
-    requesterAdmin: accounts[3],
+    requesterAdmin: accounts[1],
```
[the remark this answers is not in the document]
Ashar2shahid: I think burak mentioned he will address this later, so this is resolved.

packages/protocol/test/RequesterStore.sol.js
acenolaza: > Wouldn't it be better to just deploy RequesterStore, instead of the whole AirnodeRrp?

hmmm, maybe? 🤔  If yes then this would also apply to all other contracts except fro AirnodeRrp.

The thing is that my understanding is that only AirnodeRrp is deployable and other parent contracts should not be deployed on their own. It's just a way to split up the code to make the contract more readable and maintainable.

@bbenligiray any thoughts on this?

Ashar2shahid: Actually, yes thats what I wanted to point out for TemplateStore, Convenience etc as well. My thinking is that shouldn't you be able to test these contracts on their own and if there is a functionality to test that requires inheriting this contract then that should be done separately. In the case of RequesterStore no extra functionality needs to be tested if AirnodeRrp is deployed.

Ashar2shahid:
I guess its alright to merge this then

PR #420, opened 2021-07-14
add metaAdminnable contract and integrate with Api3CwRrpAuthorizer

Ashar2shahid:
Before Reviewing this please have a look at #409. There are few key things that I am still confused about.

1. The current implementation only makes it possible to have a metaAdmin for a single Adminned entity at the time of contract creation. How would we go about setting the metaAdmin for other adminned entites(does id 0 have the ability to set metaAdmin for other adminned entites)? If that isn't possible would it make sense to have just a single metaAdmin for all adminnedEntitites? 

2. I've updated the getRank function in `Api3CwRrpAuthorizer` but why are we returning the max rank for a query to get the rank of a specific adminned entity @bbenligiray ? 

3. The `transferMetaAdminStatus` needs to be overriden for `AirnodeCwRrpAuthorizer`  any ideas?


bbenligiray:
I refactored all the authorizers. The inheritance is a bit convoluted so let's make sure that everything works as intended in the unit tests.

packages/protocol/contracts/authorizers/ClientWhitelister.sol
```diff
@@ -0,0 +1,84 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+import "./RankedAdminnable.sol";
+import "./interfaces/IClientWhitelister.sol";
+
+/// @title Authorizer contract where clients are whitelisted until an
+/// expiration time or indefinitely (until the whitelisting is revoked)
+abstract contract ClientWhitelister is RankedAdminnable, IClientWhitelister {
```
acenolaza, line 9: Why is this contract abstract? I only ask because my understanding is that contracts are identified as abstract contracts when at least one of their functions lacks an implementation. Maybe the idea is that this contract never gets deployed and be always used as a base contract for others to inherit.

packages/protocol/contracts/authorizers/Api3ClientRrpAuthorizer.sol
```diff
@@ -0,0 +1,34 @@
...
+
+/// @title Authorizer contract that whitelists clients where the API3 DAO is
+/// the metaAdmin
+contract Api3ClientRrpAuthorizer is
+    ClientRrpAuthorizer,
+    MetaAdminnable,
+    IApi3ClientRrpAuthorizer
+{
+    /// @notice Authorizer contracts use `authorizerType` to signal their type
+    uint256 public constant override AUTHORIZER_TYPE = 2;
```
acenolaza, line 16: I seemed to keep forgetting what's the use case for the AUTHORIZER_TYPE value and does this value map to some enum or string constant somewhere?

acenolaza (approved):
phew! not an easy PR I have to be honest :) I really like the naming refactor but I still have a hard time following all the hierarchies between contracts. Overall it looks good to me but I left a couple of questions for me to better understand some specifics.

packages/protocol/contracts/authorizers/interfaces/IClientWhitelister.sol
```diff
@@ -1,9 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.6;
 
-import "./IRrpAuthorizer.sol";
-
-interface IClientWhitelistRrpAuthorizer is IRrpAuthorizer {
+interface IClientWhitelister {
     enum AdminRank {
```
acenolaza, line 5: One other thing, isn't it a bad practice to define enums and structs in interfaces?

packages/protocol/contracts/authorizers/Api3ClientRrpAuthorizer.sol
```diff
@@ -0,0 +1,34 @@
...
+
+    /// @notice Called to get the rank of an admin for an adminned entity
+    /// @dev Overriden to use metaAdminned ranks
+    /// @param adminnedId ID of the entity being adminned
+    /// @param admin Admin address whose rank will be returned
+    /// @return Admin rank
+    function getRank(bytes32 adminnedId, address admin)
+        public
+        view
+        override(MetaAdminnable, RankedAdminnable)
```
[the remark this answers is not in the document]
dextracker: So to me this seems to be more of a safety measure than anything. I found a medium article detailing some usecases for the override(Base1, Base2) inheritance, but basically if you're inheriting from multiple base contracts _that are unrelated_ then you must explicitly state which contracts you are overriding. In our case, the contracts are in a multiple inheritance structure so they know which functions to override, however it is better from a security perspective to directly assign them in the override call. So that there's no possibility of someone inheriting this class and being able to modify other contracts they should not be able to.

dextracker (commented):
Overall I think this looks good i dont really see any issues with it!

packages/protocol/contracts/authorizers/Api3ClientRrpAuthorizer.sol
Ashar2shahid: I'm guessing since this is public variable, that the node will call the contract to fetch this variable to indicate which type of authorizer it is.

packages/protocol/contracts/authorizers/MetaAdminnable.sol
```diff
@@ -0,0 +1,47 @@
...
+/// @title Contract that implements a metaAdmin that has full authority over
+/// all the entities being adminned by RankedAdminnable
+contract MetaAdminnable is RankedAdminnable, IMetaAdminnable {
+    /// @notice The metaAdmin has the maximum possible rank
+    address public override metaAdmin;
+
+    uint256 private constant MAX_RANK = 2**256 - 1;
+
+    /// @param metaAdmin_ Initial metaAdmin
+    constructor(address metaAdmin_) {
```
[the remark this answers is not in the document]
bbenligiray: https://docs.soliditylang.org/en/v0.8.6/style-guide.html#avoiding-naming-collisions

packages/protocol/contracts/authorizers/Api3ClientRrpAuthorizer.sol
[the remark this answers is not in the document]
bbenligiray: Yeah I tried to be overly explicit

packages/protocol/contracts/authorizers/interfaces/IClientWhitelister.sol
Ashar2shahid: I'm not sure if its a bad practice or not, but It was added a while back and no one really pointed any issues in the pull request: https://github.com/ethereum/solidity/issues/4087 . Care to share the link where its pointed out as a bad practice?

packages/protocol/contracts/authorizers/Api3ClientRrpAuthorizer.sol
bbenligiray: It's not really needed tbh, maybe we should remove it

packages/protocol/contracts/authorizers/ClientWhitelister.sol
bbenligiray: Actually it is `ClientRrpAuthorizer` that needs to be abstract because it doesn't implement `AUTHORIZER_TYPE`. I just copy pasted it to create `ClientWhitelisted` so that ended up being an abstract as well. Fixed that now.

packages/protocol/contracts/authorizers/Api3ClientRrpAuthorizer.sol
bbenligiray: (decided to be kept)

acenolaza: You are right, I was reading a medium post which was using an older version of solidity as reference.

packages/protocol/contracts/authorizers/interfaces/IClientWhitelister.sol
acenolaza: https://medium.com/upstate-interactive/solidity-how-to-know-when-to-use-abstract-contracts-vs-interfaces-874cab860c56

acenolaza: You are right, I was reading a medium post which was using an older version of solidity as reference. https://medium.com/upstate-interactive/solidity-how-to-know-when-to-use-abstract-contracts-vs-interfaces-874cab860c56

PR #422, opened 2021-07-16
Fixes operation package after removing airnodeAdmin and requesterAdmin from the protocol contracts

acenolaza:
Had to disable `functional/no-let` es-lint rule because I wanted to use a for loop. Only disabled the rule in that particular line and not in the local es-lint file because I get the feeling that this function might be moved to a shared package since it's being used in multiple places.

Tests:
ran successfully:
1. `dev:background`
2. `dev:eth-deploy`
3. `dev:eth-requests`

`dev:invoke` failed because node package needs to be fixed first.

### DIFF
--- a/packages/operation/src/evm/utils.ts
+++ b/packages/operation/src/evm/utils.ts
@@ -25,6 +25,21 @@ export function getDesignatedWallet(mnemonic: string, requester: string, provide
   return deriveWalletFromPath(mnemonic, `m/0/${addressToDerivationPath(requester)}`, provider);
 }
 
+/**
+ * HD wallets allow us to create multiple accounts from a single mnemonic.
+ * Each requester creates a designated wallet for each provider to use
+ * in order for them to be able to respond to the requests their clients make.
+ *
+ * By convention derivation paths start with a master index
+ * followed by child indexes that can be any integer up to 2^31.
+ *
+ * Since addresses can be represented as 160bits (20bytes) we can then
+ * split it in chunks of 31bits and create a path with the following pattern:
+ * m/0/1st31bits/2nd31bits/3rd31bits/4th31bits/5th31bits/6th31bits.
+ *
+ * @param address A string representing a 20bytes hex address
+ * @returns The path derived from the address
+ */
 export function addressToDerivationPath(address: string): string {
   const requesterBN = ethers.BigNumber.from(address);
   const paths = [];
--- a/packages/operation/src/evm/utils.ts
+++ b/packages/operation/src/evm/utils.ts
@@ -21,10 +21,17 @@ export function deriveWalletFromPath(mnemonic: string, path: string, provider: e
   return new ethers.Wallet(designatorHdNode.privateKey, provider);
 }
 
-export function getDesignatedWallet(
-  mnemonic: string,
-  requesterIndex: string,
-  provider: ethers.providers.JsonRpcProvider
-) {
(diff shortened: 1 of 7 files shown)


packages/operation/src/evm/utils.ts
```diff
@@ -21,10 +21,17 @@ export function deriveWalletFromPath(mnemonic: string, path: string, provider: e
...
-  requesterIndex: string,
-  provider: ethers.providers.JsonRpcProvider
-) {
-  return deriveWalletFromPath(mnemonic, `m/0/${requesterIndex}`, provider);
+export function getDesignatedWallet(mnemonic: string, requester: string, provider: ethers.providers.JsonRpcProvider) {
+  return deriveWalletFromPath(mnemonic, `m/0/${addressToDerivationPath(requester)}`, provider);
+}
+
+export function addressToDerivationPath(address: string): string {
+  const requesterBN = ethers.BigNumber.from(address);
```
andreogle, line 29: Why are we creating a big number from an address? Could you add a comment here explaining what's going on here please?

andreogle (approved):

packages/operation/src/evm/utils.ts
acenolaza: Since we are no longer using the `requesterIndex` that used to force the requester to send a tx for creating a record on chain now we need another way for deriving the hd wallet path. 

Burak came up with a smart way of transforming the requester address to a series of ntegers that can be then be used as a derivation path.

Since by convention each section of the path can hold integers up to 2^31 we need to split the BigNumber we get from the address. Since we know that the BigNumber contains 160bits what we did was split it in 31bits chunks in order to get a path like this: m/0/1st31bits/2nd31bits/3rd31bits/4th31bits/5th31bits/6th31bits.

Please let me know if this explanation made it even more confusing than the code itself 🤭

andreogle: Ok I think that makes sense. It's worth adding an explanation like this in the code too for future readers (and future Andre who is going to forget 😄 )

acenolaza: I agree 👍🏻 and I even replied on a Saturday morning because future Santiago would had have a tough time explaining it on  Monday morning 😄

packages/operation/src/evm/utils.ts
```diff
@@ -25,6 +25,21 @@ export function getDesignatedWallet(mnemonic: string, requester: string, provide
...
+ * in order for them to be able to respond to the requests their clients make.
+ *
+ * By convention derivation paths start with a master index
+ * followed by child indexes that can be any integer up to 2^31.
+ *
+ * Since addresses can be represented as 160bits (20bytes) we can then
+ * split it in chunks of 31bits and create a path in with the following pattern:
+ * m/0/1st31bits/2nd31bits/3rd31bits/4th31bits/5th31bits/6th31bits.
+ *
+ * @param address A string representing a 20bytes hex address
```
andreogle, line 40: For what it's worth, we haven't really been using JSDoc comment syntax anywhere other than the Solidity contracts. I don't mind it though

andreogle (approved):
Thanks for adding the comment! 🚀

packages/operation/src/evm/utils.ts
acenolaza: Right, I saw that too but I'm thinking that this function might be refactored into it's own package or some common place because it's being used in at least 2 or 3 different packages at the moment.

PR #426, opened 2021-07-22
Authorizer tests for both SelfClientRrpAuthorizer and Api3ClientRrpAuthorizer

Ashar2shahid:
Split up the tests according to the solidity files, since there will be two different deployments. the tests are split internally for the different deployments.


packages/protocol/test/authorizers/ClientWhitelister.sol.js
```diff
@@ -0,0 +1,813 @@
...
+  Admin: 1,
+  SuperAdmin: 2,
+};
+
+let roles;
+let selfClientRrpAuthorizer;
+let api3ClientRrpAuthorizer;
+let airnodeRrp;
+let airnodeId;
+const api3AdminnedEntity = ethers.constants.HashZero;
```
acenolaza, line 16: Can we just use `airnodeId` wherever `api3AdminnedEntity` is being used? I think that closest to a real world value instead of using `HashZero`.

acenolaza (approved):
I'm ok with splitting the single test file for the authorizer in multiple files because it only adds a little bit of duplication on the beforeEach methods.

One thing I've noticed tho is that the code for each test is duplicated for each authorizer and I would really like to avoid that by using an array of authorizers and run the [`it()` blocks inside a for each block](https://blog.harveydelaney.com/running-a-test-with-multiple-test-cases-using-jest-enzyme-react/) or maybe use [test.each](https://dev.to/bgord/simplify-repetitive-jest-test-cases-with-test-each-310m).

packages/protocol/test/authorizers/ClientWhitelister.sol.js
Ashar2shahid: So, the api3Authorizer will use bytes32(0) if I recall correctly which is why I used HashZero to represent that

Ashar2shahid:
> I'm ok with splitting the single test file for the authorizer in multiple files because it only adds a little bit of duplication on the beforeEach methods.
> 
> One thing I've noticed tho is that the code for each test is duplicated for each authorizer and I would really like to avoid that by using an array of authorizers and run the [`it()` blocks inside a for each block](https://blog.harveydelaney.com/running-a-test-with-multiple-test-cases-using-jest-enzyme-react/) or maybe use [test.each](https://dev.to/bgord/simplify-repetitive-jest-test-cases-with-test-each-310m).

test cases looks like the way to go, ill update the repetitive tests based on that

packages/protocol/test/authorizers/ClientWhitelister.sol.js
acenolaza: My bad and I like that you used the constant 👍🏻 . Resolving this comment.

Issue #343, opened 2021-05-10
Add more infrastructure specific options to deployer


amarthadan:
https://github.com/api3dao/airnode/pull/333#discussion_r626508755

andreogle:
Also worth considering letting the user specify any throttling/quotas for the testing API

bbenligiray:
Should these be specified in `nodeSettings.cloudProvider` in `config.json`, assuming they will be cloud provider-specific?

amarthadan:
> Should these be specified in `nodeSettings.cloudProvider` in `config.json`, assuming they will be cloud provider-specific?

Yes, that would probably be a good place for them.

bbenligiray:
Tracked by https://api3dao.atlassian.net/browse/AN-145

PR #428, opened 2021-07-23
Add API test calls functionality

amarthadan:
What's missing:
* Tests
* Flag to select which endpoints can be tested (currently all can be tested)

But I wanted to create a PR anyway because it's getting quite large and also I want to know what you think about the approach.

You can test it locally - running `yarn dev:testApi` from the `node` package directory (correct `config.json` and `.env` must be present, same as for `yarn dev:invoke`).
Or, you can test it on AWS by deploying and calling the API gateway.


packages/deployer/src/handlers/aws/index.ts
```diff
@@ -1,13 +1,12 @@
...
 import * as path from 'path';
 import { config, handlers, logger, promiseUtils, providerState } from '@api3/node';
-import * as node from '@api3/node';
 
 const configFile = path.resolve(`${__dirname}/../../config-data/config.json`);
 const rawConfig = JSON.parse(fs.readFileSync(configFile, 'utf8'));
 const parsedConfig = config.parseConfig(rawConfig[0]);
 
-function encodeBody(data: node.WorkerResponse): string {
+function encodeBody(data: any): string {
```
andreogle, line 9: Maybe it's better to just use `JSON.stringify()` directly in the `testApi` handler and keep the type here?

packages/deployer/package.json
```diff
@@ -14,7 +14,8 @@
     "copy:terraform": "copyfiles terraform/**/**/*.tf dist/",
     "copy:webpack": "copyfiles .webpack/**/**/*.js dist/",
     "prepublishOnly": "yarn build",
-    "webpack": "webpack"
+    "webpack": "webpack",
+    "webpack:dev": "webpack --mode development"
```
andreogle, line 18: Where is this script used and what for?

packages/deployer/src/infrastructure/index.ts
```diff
@@ -80,9 +88,9 @@ async function deploy(airnodeIdShort: string, stage: string, region: string, con
 
   command = `terraform apply -var="aws_region=${region}" -var="airnode_id_short=${airnodeIdShort}" -var="stage=${stage}" -var="configuration_file=${path.resolve(
     configPath
-  )}" -var="secrets_file=${path.resolve(
-    secretsPath
-  )}" -var="handler_file=${handlerFile}" -auto-approve -input=false -no-color`;
+  )}" -var="secrets_file=${path.resolve(secretsPath)}" -var="handler_file=${handlerFile}" -var='api_keys=[${apiKeys
```
andreogle, line 91: I think this becomes harder to follow when we're doing multi-line interpolation. Maybe assign this to a variable above?

packages/deployer/terraform/airnode/main.tf
```diff
@@ -5,61 +5,80 @@
...
   depends_on = [module.initializeProvider, module.callApi, module.processProviderRequests]
 }
+
+module "testApi" {
+  source = "./modules/function"
+
+  name               = "${local.name_prefix}-testApi"
+  handler            = "handlers/aws/index.testApi"
+  source_file        = var.handler_file
+  timeout            = 60
```
andreogle, line 61: Probably worth keeping this timeout the same as the `callApi` function's timeout? i.e. 30 seconds. Otherwise, very slow requests might fail to be processed in the node, but testing the endpoint works

packages/node/__dev__/test-api.ts
```diff
@@ -0,0 +1,42 @@
...
+      'endpoint-name': {
+        alias: 'e',
+        description: 'Endpoint name',
+        type: 'string',
+        required: true,
+      },
+      body: {
+        alias: 'b',
+        description: 'Request body',
+        default: '{ "parameters": {} }',
```
andreogle, line 27: Do we need to wrap parameters here? Couldn't we just use the whole object as the parameters? Maybe even worth changing the argument to be "parameters" instead of "body"?

```suggestion
        default: '{}',
```

packages/node/src/handlers/call-api.ts
```diff
@@ -1,5 +1,6 @@
 import * as adapter from '@api3/adapter';
```
andreogle, line 1: This whole module is going to need a bit of an overhaul at some point

packages/node/src/handlers/test-api.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  const workerOpts: WorkerOptions = {
+    cloudProvider: config.nodeSettings.cloudProvider,
+    airnodeIdShort: wallet.getAirnodeIdShort(masterHDNode),
+    stage: config.nodeSettings.stage,
+    region: config.nodeSettings.region,
+  };
+
+  const aggregatedApiCall: AggregatedApiCall = {
```
andreogle, line 32: This feels weird to me but I don't have any better suggestions at the moment 😄

packages/node/src/logger.ts
```diff
@@ -8,11 +8,11 @@ const logLevels: { readonly [key in LogLevel]: number } = {
   ERROR: 3,
 };
 
-export function buildBaseOptions(config: Config, state: CoordinatorState) {
+export function buildBaseOptions(config: Config, meta: LogMetadata) {
```
andreogle, line 11: 👍🏻 much better

packages/node/src/workers/cloud-platforms/aws.ts
```diff
@@ -14,9 +14,12 @@ export function spawn(params: WorkerParameters): Promise<WorkerResponse> {
       Payload: JSON.stringify(params.payload),
     };
 
+    console.log(options);
```
andreogle, line 17: remove here and below

packages/node/src/workers/cloud-platforms/aws.ts
```diff
@@ -14,9 +14,12 @@ export function spawn(params: WorkerParameters): Promise<WorkerResponse> {
...
     lambda.invoke(options, (err, data) => {
+      console.log([err, data]);
       if (err) {
         reject(err);
+        return;
```
andreogle, line 22: not sure how I missed this before 🤦🏻

packages/node/src/handlers/test-api.ts
```diff
@@ -0,0 +1,57 @@
...
+    designatedWallet: '',
+    chainId: '',
+    endpointId: triggerRequest.endpointId,
+    endpointName,
+    oisTitle: triggerRequest.oisTitle,
+    parameters,
+  };
+
+  const [err, logData] = await go(() => spawnNewApiCall(aggregatedApiCall, logOptions, workerOpts, false), {
+    timeoutMs: WORKER_CALL_API_TIMEOUT,
```
andreogle, line 46: See comment about timeouts above. This would time out after 30 seconds but the lambda has a timeout of 60 seconds (currently)

packages/node/src/handlers/test-api.ts
```diff
@@ -0,0 +1,57 @@
...
+    clientAddress: '',
+    designatedWallet: '',
+    chainId: '',
+    endpointId: triggerRequest.endpointId,
+    endpointName,
+    oisTitle: triggerRequest.oisTitle,
+    parameters,
+  };
+
+  const [err, logData] = await go(() => spawnNewApiCall(aggregatedApiCall, logOptions, workerOpts, false), {
```
andreogle, line 45: I don't really like grouping the logs and data that much. I think maybe this should be changed to return a tuple with 3 elements? i.e. `[err, logs, data]`. (Not now though)

andreogle (changes requested):

packages/deployer/config/secrets.env.example
```diff
@@ -5,3 +5,6 @@ SS_CURRENCY_CONVERTER_API_CURRENCY_CONVERTER_SECURITY_SCHEME=""
 HEARTBEAT_API_KEY=""
 HEARTBEAT_ID=""
 HEARTBEAT_URL=""
+
+CHAINAPI_TESTING_API_KEY=""
```
andreogle, line 9: Forgot to add, I don't think Airnode should be hardcoding any references to ChainAPI. Can this be done with a single environment variable?

andreogle: I also think that just "TESTING_" by itself doesn't necessarily accurately describe the key as "testing" can mean any number of things.

Maybe `TEST_ENDPOINT_API_KEY`, `ENDPOINT_TESTING_API_KEY`  or something similar instead? Probably want to get some input from the rest of the team

andreogle (changes requested):

packages/deployer/package.json
amarthadan: It's not used anywhere. I used it manually to get a proper stacktrace in AWS logs. It skips minification so I could see what's actually causing a problem.

packages/node/__dev__/test-api.ts
amarthadan: I was thinking about that and decided that we should keep the interface as close to the normal HTTP request as possible, hence providing the whole body. But I can change it to just parameters and wrap it inside the function if you think it's better.

packages/node/src/handlers/test-api.ts
amarthadan: Yeah, I agree. But using `spawnNewApiCall` felt like the best way to reuse node's code and I didn't want to change the types too much. So, "mocking" `aggregatedApiCall` seemed like a way to go :man_shrugging:

I can make those mocked fields optional and refactor the code to accomodate that if you think that would be better.

packages/node/src/handlers/test-api.ts
amarthadan: I don't like it either I just didn't want to rework it as part of this PR.

packages/deployer/config/secrets.env.example
amarthadan: Making two env variables was @bbenligiray 's idea. One can be used by ChainAPI and one by the user deploying the airnode for som custom access if need be. Not sure if it's a use case worth supporting though. Maybe we can stick with just one for now and see if somebody actually needs multiple API keys.

packages/node/__dev__/test-api.ts
andreogle: What would other potential keys be in this object? I think headers would be a separate flag if we ever added that

packages/node/src/handlers/test-api.ts
andreogle: > I can make those mocked fields optional and refactor the code to accomodate

Nah I don't think that's necessary. I like the strictness of making fields required

packages/node/__dev__/test-api.ts
amarthadan: Yes, headers would be most likely separate. I was thinking more like some execution control options. I didn't want to provide parameters without a wrapper in case we need to send something with them. But maybe it's not really necessary.

andreogle: That also seems like it might be better suited as another flag. I still lean towards changing this flag to "parameters" instead of "request body" but I don't really very strongly either way.

packages/deployer/src/cli/commands.ts
```diff
@@ -40,6 +40,14 @@ export async function deploy(
     throw new Error('Invalid mnemonic');
   }
 
+  let apiKey: string | undefined = undefined;
+  if (config.nodeSettings.enableTestingGateway) {
+    apiKey = secrets.ENDPOINT_TESTING_API_KEY;
+    if (!apiKey) {
+      throw new Error('Unable to deploy testing gateway as the API key is missing');
```
andreogle, line 47: Maybe put the key in the message just as an additional tip to the user?
```suggestion
      throw new Error('Unable to deploy testing gateway as the ENDPOINT_TESTING_API_KEY secret is missing');
```

packages/deployer/src/cli/commands.ts
```diff
@@ -40,6 +40,14 @@ export async function deploy(
     throw new Error('Invalid mnemonic');
   }
 
+  let apiKey: string | undefined = undefined;
```
andreogle, line 43: Rename for a bit more clarity?
```suggestion
  let testingApiKey: string | undefined = undefined;
```

packages/deployer/src/infrastructure/index.ts
```diff
@@ -36,20 +36,28 @@ export async function deployAirnode(
...
     throw err;
   }
 }
 
-async function deploy(airnodeIdShort: string, stage: string, region: string, configPath: string, secretsPath: string) {
+async function deploy(
+  airnodeIdShort: string,
+  stage: string,
+  region: string,
+  apiKey: string | undefined,
```
andreogle, line 57:
```suggestion
  testingApiKey: string | undefined,
```

packages/deployer/terraform/airnode/modules/apigateway/main.tf
```diff
@@ -0,0 +1,64 @@
...
+  }
+}
+
+resource "aws_api_gateway_stage" "stage" {
+  deployment_id = aws_api_gateway_deployment.deployment.id
+  rest_api_id   = aws_api_gateway_rest_api.api_gateway.id
+  stage_name    = var.stage
+}
+
+resource "aws_api_gateway_usage_plan" "usage_plan" {
```
andreogle, line 46: Do we want to add some throttle settings here?

andreogle (approved):
A few more minor comments, but looks good otherwise 🚀

packages/deployer/terraform/airnode/modules/apigateway/main.tf
amarthadan: I think we would have to deal with adding some more customization into the infrastructure setup anyway in the future (#343). I don't think it's necessary for 0.1.0 and especially this case as it will most likely be a "hidden" feature till ChainAPI supports it. Maybe we shouldn't even merge it to `master` before 0.1.0 is out if we don't really need it :thinking:

packages/deployer/src/infrastructure/index.ts
```diff
@@ -78,11 +86,12 @@ async function deploy(airnodeIdShort: string, stage: string, region: string, con
   const options = { cwd: airnodeTmpDir };
   await runCommand(command, options);
 
+  const testingApiKeyVar = testingApiKey ? `-var='api_key=${testingApiKey}'` : '';
```
Siegrift, line 89: You can think about adding some convenience functions to simplify creation of the command. I created similar ones for the admin CLI https://github.com/api3dao/airnode/blob/master/packages/admin/test/e2e/cli.feature.ts#L114

(Not in this PR)

packages/node/__dev__/test-api.ts
```diff
@@ -0,0 +1,42 @@
+import dotenv from 'dotenv';
+import omitBy from 'lodash/omitBy';
+import yargs from 'yargs';
+import { hideBin } from 'yargs/helpers';
+import * as local from '../src/workers/local-handlers';
+
+function longArguments(args: Record<string, any>) {
+  return JSON.stringify(omitBy(args, (_, arg) => arg === '$0' || arg.length === 1));
```
Siegrift, line 8: I am not sure what this does. Can you add a small inline comment?

Siegrift (approved):
I still lack the knowledge to review the terraform changes, but code LGTM.

packages/deployer/src/handlers/aws/index.ts
```diff
@@ -54,3 +53,15 @@ export async function processProviderRequests(event: any) {
   const body = encodeBody({ ok: true, data: providerState.scrub(updatedState) });
   return { statusCode: 200, body };
 }
+
+export async function testApi(event: any) {
```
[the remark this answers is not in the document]
amarthadan: I think we can type it better, created an [issue](https://github.com/api3dao/airnode/issues/460) to do that later.

packages/deployer/src/infrastructure/index.ts
amarthadan: Created an [issue](https://github.com/api3dao/airnode/issues/461) for that.

PR #433, opened 2021-07-23
Refactor protocol contracts

bbenligiray:
- Clients are now called requesters. It was very weird how the entity making the request wasn't called a requester.
- Requesters are now called sponsors. Designated wallets are called sponsor wallets, endorsing is called sponsoring. This both more literal and reduces the number of terms.
- The request that refers to a template is now called a template request (instead of a client request/regular request)
- Sponsors, requesters, airnodes are referred to as such, we don't say sponsorAddress, requesterAddress, airnodeAddress because these already type'd 
- Airnodes are now identified by the address of the default BIP 44 wallet derived from their mnemonic (instead of an ID).
- Airnode gets its `authorizers` from `config.json`. `authorizers` on-chain is merely an announcement, like `xpub`. As a result:
  - We no longer have "default authorizers"
  - AirnodeRrp.sol is not Ownable, we don't have an OpenZeppelin dependency
  - Checking the authorization requires the caller to provide authorizer addresses as an argument
    - We still don't support authorizers with different interfaces
    - Sponsor wallet is no longer an authorizer argument (because we're hitting the stack limit by providing the authorizer addresses as an additional argument)
    
~Entire protocol is reduced to three contracts:~
~RequestUtils.sol implements the request functionality~
~WithdrawUtils.sol implements the withdrawal functionality~
~AirnodeRrp.sol inherits both of these and implements the convenience functions~
Updated with a slightly different structure that implements the main functionality in `AirnodeRrp.sol`

- Improved tests, naming, in-code documentation for the protocol contracts
- Removed redundant getter function in favor of calling the public variables directly (and added the public variables to the interfaces)


acenolaza (approved):

Ashar2shahid (approved):

packages/protocol/contracts/authorizers/interfaces/IRrpAuthorizerNew.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+interface IRrpAuthorizerNew {
```
Ashar2shahid, line 4: This is probably a temporary file, shouldn't this be renamed back to IRrpAuthorizer?

Ashar2shahid (changes requested):

packages/protocol/contracts/AirnodeRrp.sol
```diff
@@ -1,239 +1,161 @@
...
-        address fulfillAddress,
-        bytes4 fulfillFunctionId
-    )
+    /// @notice A convenience method to retrieve multiple templates with a
+    /// single call
+    /// @param templateIds Request template IDs
+    /// @return airnodes Array of Airnode addresses
+    /// @return endpointIds Array of endpoint IDs
+    /// @return parameters Array of request parameters
+    function getTemplates(bytes32[] calldata templateIds)
```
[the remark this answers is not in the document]
bbenligiray: I had it there because it's optional / a convenience function because the node may just as well use `templates()` to get the templates individually. I will reorganize everything in a way that all templates stuff is together though.

packages/protocol/contracts/authorizers/interfaces/IRrpAuthorizerNew.sol
bbenligiray: It is temporary but updating `IRrpAuthorizer.sol` breaks the contracts in the `authorizers/`directory. I didn't want to do that because you're working on their tests.

bbenligiray:
A change I forgot to mention: `abi.encodePacked()` is used to create the request IDs, so `defaultAbiCoder.encode()`s at the node-end will need to be replaced with `solidityPack()`

packages/protocol/contracts/authorizers/MetaAdminnable.sol
```diff
@@ -41,7 +41,7 @@ contract MetaAdminnable is RankedAdminnable, IMetaAdminnable {
         override(RankedAdminnable, IRankedAdminnable)
         returns (uint256)
     {
-        if (msg.sender == metaAdmin) return MAX_RANK;
+        if (admin == metaAdmin) return MAX_RANK;
```
bbenligiray, line 44: Here's a bug

packages/protocol/contracts/requesters/RrpBeaconServer.sol
```diff
@@ -38,53 +38,52 @@ contract RrpBeaconServer is
...
+        bytes32 requestId = airnodeRrp.makeTemplateRequest(
             templateId,
-            requester,
-            designatedWallet,
+            sponsor,
+            sponsorWallet,
             address(this),
             this.fulfill.selector,
-            "0x"
+            ""
```
bbenligiray, line 78: Here's another bug. `"0x"` is an empty `bytes` when making ethers calls, in Solidity and empty `bytes` is `""`.

packages/protocol/contracts/authorizers/MetaAdminnable.sol
Ashar2shahid: I actually mentioned this a while back https://github.com/api3dao/airnode/pull/409#discussion_r664677178 thought it was intended.

bbenligiray: Right, I missed that comment

PR #434, opened 2021-07-26
Revert `config.json` structure back to object

### LINKED ISSUE #432: Change `config.json` structure back into an object
Basically a revert of #291

amarthadan:
Close #432

### DIFF
--- a/packages/node/src/workers/local-handlers.ts
+++ b/packages/node/src/workers/local-handlers.ts
@@ -16,7 +16,7 @@ export interface CallApiArgs {
 }
 
 function loadConfig() {
-  const rawConfig = JSON.parse(fs.readFileSync('./config.json', 'utf8'))[0];
+  const rawConfig = JSON.parse(fs.readFileSync('./config.json', 'utf8'));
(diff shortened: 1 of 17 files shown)


packages/node/src/workers/local-handlers.ts
```diff
@@ -16,7 +16,7 @@ export interface CallApiArgs {
 }
 
 function loadConfig() {
-  const rawConfig = JSON.parse(fs.readFileSync('./config.json', 'utf8'))[0];
+  const rawConfig = JSON.parse(fs.readFileSync('./config.json', 'utf8'));
```
andreogle, line 19: So this was actually broken as it wouldn't have worked with more than config? 😄

andreogle (approved):
Nice one! This is significantly simpler with just one config

packages/node/src/workers/local-handlers.ts
amarthadan: :smile: Kinda. It wasn't broken just not finished.

PR #435, opened 2021-07-26
Splits airnodeToAnnouncement mapping into airnodeToPublicKey and airnodeToAuthorizers mappings

acenolaza:
I wanted to make both mappings `public` in order to get automagic getters for them but I ended up creating `getAirnodePublicKey` and `getAirnodeAuthorizers` functions because the automatic getter for `airnodeToAuthorizers` was expecting 2 arguments (airnode address and index for authorizer array).

I expect `setAirnodePublicKey` to be called maybe only once while `setAirnodeAuthorizers` multiple times.

### DIFF
--- a/packages/protocol/contracts/rrp/ConvenienceUtils.sol
+++ b/packages/protocol/contracts/rrp/ConvenienceUtils.sol
@@ -6,48 +6,64 @@ import "./authorizers/interfaces/IRrpAuthorizer.sol";
 
 /// @title Contract that implements convenience functions
 contract ConvenienceUtils is IConvenienceUtils {
-    struct Announcement {
-        string xpub;
-        address[] authorizers;
-    }
+    mapping(address => string) private airnodeToPublicKey;
+    mapping(address => address[]) private airnodeToAuthorizers;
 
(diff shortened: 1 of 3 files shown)


packages/protocol/contracts/rrp/ConvenienceUtils.sol
```diff
@@ -6,48 +6,62 @@ import "./authorizers/interfaces/IRrpAuthorizer.sol";
...
-    /// This announcement does not need to be made for the protocol to be used,
+    /// This public key set does not need to be made for the protocol to be used,
     /// it is mainly for convenience.
     /// @param xpub Extended public key of the Airnode
+    function setAirnodePublicKey(string calldata xpub) external override {
+        airnodeToPublicKey[msg.sender] = xpub;
+        emit SetAirnodePublicKey(msg.sender, xpub);
+    }
+
+    /// @notice Called by the Airnode operator to set authorizers
```
Ashar2shahid, line 23: So we can probably mention it here that these are announcements and not used for actual authorization

Ashar2shahid (approved):
This approach is better and has less ambiguity. Looks good

PR #436, opened 2021-07-27
Xpub fixes

bbenligiray:
Public key and extended public key are different things, so renamed references to the public key as xpub.
We only need a getter for the authorizers, so made the xpub mapping public and removed the respective getter.
Please review and merge yourselves.


acenolaza (approved):
LGTM

PR #439, opened 2021-07-29
Variable interpolation for `config.json`

### LINKED ISSUE #429: Allow variable interpolation for `config.json`
Variables defined in `secrets.env` should be interpolated in `config.json` using a pattern `${VAR}`.

amarthadan:
A bit bigger but should be manageable when reviewed commit by commit.

Close #429


packages/adapter/test/fixtures/security.ts
```diff
@@ -1,5 +1,3 @@
-import { SecuritySchemeSecret } from '@api3/ois';
+import { ApiCredentials } from '@api3/ois';
 
-export const securitySchemeSecrets: SecuritySchemeSecret[] = [
-  { securitySchemeName: 'myapiApiScheme', value: 'super-secret-key' },
-];
+export const credentials: ApiCredentials = { securityScheme: 'myapiApiScheme', value: 'super-secret-key' };
```
andreogle, line 3: While you're updating here, might as well update this to be a function instead to be consistent with other fixtures so we get a new object each time it's used

```ts
buildCredentials(overrides?: Partial<ApiCredentials>): ApiCredentials
```

packages/deployer/config/config.json.example
```diff
@@ -9,21 +9,15 @@
...
+      "providers": {
+        "local": {
+          "url": "${LOCAL_PROVIDER_URL}"
+        }
+      },
       "type": "evm"
     }
   ],
   "environment": {
-    "chainProviders": [
```
andreogle, line 19: 🎉

packages/deployer/src/utils/validation.ts
```diff
@@ -0,0 +1,13 @@
+import { Config } from '@api3/node';
+import * as logger from '../utils/logger';
+
+export function validateConfig(config: Config, nodeVersion: string) {
+  if (nodeVersion !== config.nodeSettings.nodeVersion) {
```
andreogle, line 5: You've dropped the `cloudProvider` check in the validation?

packages/deployer/src/utils/validation.ts
```diff
@@ -0,0 +1,13 @@
...
+    logger.fail(
+      `nodeVersion under nodeSettings in config.json is ${config.nodeSettings.nodeVersion} while the deployer node version is ${nodeVersion}`
+    );
+    throw new Error('Attempted to deploy node configuration with the wrong node version');
```
andreogle, line 9: This message is a bit misleading/confusing to me. I had to read through the code to actually understand what it means. I would rather phrase it as something like this. What do you think?

```suggestion
    throw new Error('Node version specified in config.json does not match the deployer's version');
```

packages/node/src/config/index.ts
```diff
@@ -1,7 +1,12 @@
+import * as fs from 'fs';
 import { OIS } from '@api3/ois';
+import template from 'lodash/template';
 import { Config } from '../types';
 import { randomString } from '../utils/string-utils';
 
+const NO_MATCH_REGEXP = /($^)/;
+const ES_MATCH_REGEXP = /\$\{([^\\}]*(?:\\.[^\\}]*)*)\}/g;
```
andreogle, line 8: Would you be able to add some comments here explaining this regex please? 😄

packages/node/src/config/index.ts
```diff
@@ -12,9 +17,17 @@ function parseOises(oises: OIS[]): OIS[] {
   });
 }
 
-export function parseConfig(config: any): Config {
-  const ois = parseOises(config.ois);
-  return { ...config, ois };
+export function parseConfig(configPath: string, secrets: Record<string, string | undefined>): Config {
+  const config = fs.readFileSync(configPath, 'utf8');
+  const interpolatedConfig = template(config, {
```
andreogle, line 22: The only thing I don't like about interpolating everything all at once and using that config, is that we might accidentally leak credentials somewhere when logging a message that includes the value or in a stack trace if something crashes.

I guess the alternative would be to check every field before it's used but that seems a bit worse.

packages/adapter/test/fixtures/ois.ts
```diff
@@ -102,6 +99,10 @@ export function buildOIS(overrides?: Partial<OIS>): OIS {
         ],
       },
     ],
+    credentials: {
+      securityScheme: 'myapiApiScheme',
+      value: 'value',
```
andreogle, line 104: Nit: I think we should try to use more descriptive values in fixtures where we can:
```suggestion
      value: 'supersecret',
```

packages/node/test/setup/e2e/utils.ts
```diff
@@ -17,7 +17,11 @@ export function buildChainConfig(contracts: Contracts): ChainConfig {
     authorizers: [ethers.constants.AddressZero],
     id: '31337',
     type: 'evm',
-    providerNames: ['EVM local'],
+    providers: {
+      ['EVM local']: {
```
andreogle, line 21:
```suggestion
      'EVM local': {
```

andreogle (changes requested):
Nice work - just a few minor comments. This significantly simplifies config.json 👍🏻

packages/deployer/src/utils/validation.ts
amarthadan: Yes I did, it will be done by validator anyway.

packages/node/src/config/index.ts
amarthadan: That's the thing, we would have to check EVERY field because the interpolation can be done anywhere in the config :confused:

I guess we could introduce a custom logging that would replace any sensitive data (those replaced from env variables) with `***` or something. But other that that, I'm not sure what else we can do here.

andreogle: Yeah exactly. I'm not sure either. I'm not sure how we would determine which variables to replace as `***` since it's already interpolated at that point. I was just pointing it out as it was something we might want to consider.

I think in almost all cases, someone who has access to the logs will have access to the environment secrets. It's the edge cases where they don't have access to the secrets that I was considering.

andreogle (approved):

packages/adapter/src/request-building/authentication.ts
```diff
@@ -74,32 +73,16 @@ export function buildParameters(options: CachedBuildRequestOptions): Authenticat
...
-  if (!securitySchemeSecrets || isEmpty(securitySchemeSecrets)) {
+  const { credentials } = options;
+  if (!credentials) {
     return initialParameters;
   }
 
   // API security schemes also originate from 'config.json' and specify which schemes should be used
-  const apiSecuritySchemes = options.ois.apiSpecifications.components.securitySchemes;
-  if (isEmpty(apiSecuritySchemes)) {
+  const apiSecurityScheme = options.ois.apiSpecifications.components.securitySchemes[credentials.securityScheme];
```
Siegrift, line 82: Will `options.ois.apiSpecifications.components.securitySchemes` be defined?

packages/adapter/src/request-building/authentication.ts
```diff
@@ -74,32 +73,16 @@ export function buildParameters(options: CachedBuildRequestOptions): Authenticat
...
+  if (!apiSecurityScheme) {
     return initialParameters;
   }
 
-  const apiSecuritySchemeNames = Object.keys(apiSecuritySchemes);
```
Siegrift, line 90: Do I understand correctly that this just merged security schemes along with the secrets and we don't need to do that for all schmes, just the one that will be used?

packages/adapter/src/request-building/authentication.ts
```diff
@@ -74,32 +73,16 @@ export function buildParameters(options: CachedBuildRequestOptions): Authenticat
...
-
-  return apiSecuritySchemeNames.reduce((authentication, apiSecuritySchemeName) => {
```
Siegrift, line 92: Btw. there are some examples in the code base (adapter and deployer pckg) where the `securitySchemes` field is an array not an object. Can you fix that? 

We could also create an issue or brainstorm how to keep the examples and docs up to date...

packages/deployer/config/config.json.example
Siegrift: This just moved directly to `providers` right?

packages/deployer/src/utils/files.ts
```diff
@@ -1,34 +1,8 @@
...
-  let config: Configuration;
-  try {
-    config = JSON.parse(fs.readFileSync(configPath, 'utf8'));
-  } catch (e) {
-    logger.fail('Failed to parse configuration file');
-    throw e;
-  }
-
-  // A more comprehensive validation should be done beforehand
-  // https://github.com/api3dao/airnode/issues/136
```
Siegrift, line 17: This issue link is lost.

packages/node/src/config/index.ts
Siegrift: Is it even a valid edge case? I can't think of such example

packages/node/src/handlers/call-api.test.ts
```diff
@@ -30,12 +30,10 @@ describe('callApi', () => {
...
         parameters: { from: 'ETH' },
-        securitySchemeSecrets: [
-          {
-            securitySchemeName: 'My Security Scheme',
-            value: 'supersecret',
-          },
-        ],
+        credentials: {
+          securityScheme: 'My Security Scheme',
+          value: 'supersecret',
```
Siegrift, line 35: What is this `value` for? I have hard time understanding it from the code.

Siegrift (approved):

packages/adapter/src/request-building/authentication.ts
amarthadan: I expect that to be checked by the validator in the (near) future.

packages/adapter/src/request-building/authentication.ts
amarthadan: We'll discuss how we want to deal with this in the end but yes, for now, we have multiple scheme definitions in `ois.components.securitySchemes` and are selecting one that is to be used with `ois.credentials`.

packages/deployer/src/utils/files.ts
amarthadan: I know, integrating the validator is planned anyway.

amarthadan:
@Siegrift Maybe [this discussion](https://api3workspace.slack.com/archives/C01N6001YCE/p1627381690003900) will help you with what's going on here. Basically we're getting rid of the `environment` section and usign variable interpolation instead. Now we have `chains.providers` section with provider URL and `ois.credentials` section with API credentials instead.

PR #441, opened 2021-07-30
Api3 requester rrp authorizer fix

Ashar2shahid:
This fix makes it sure that `Api3RequesterRrpAuthorizer` only allows a single AdminnedId at `bytes32(0)` .

There were two ways to do this:

1. Have separate functions for setRank and decreaseSelfRank, this makes it so that we don't need to override the existing functions.

2. Override the setRank and decreaseSelfRank similar to getRank.

I didn't opt for 2 because unlike getRank, setRank and decreaseSelfRank changes the state. making these functions virtual unnecessarily exposes them (could lead to incorrect implementation and security issue?).  getRank is overridden but maybe we should also create a new function for that for the sake of consistency?

### DIFF
--- a/packages/protocol/contracts/rrp/authorizers/Api3RequesterRrpAuthorizer.sol
+++ b/packages/protocol/contracts/rrp/authorizers/Api3RequesterRrpAuthorizer.sol
@@ -19,8 +19,9 @@ contract Api3RequesterRrpAuthorizer is
     constructor(address metaAdmin_) MetaAdminnable(metaAdmin_) {}
 
     /// @notice Called to get the rank of an admin for an adminned entity
-    /// @dev Overriden to use metaAdminned ranks
-    /// @param adminnedId ID of the entity being adminned
+    /// @dev Overriden to use the bytes32(0) as the only adminnedId
+    ///      Overriden to use metaAdminned ranks
+    /// @param adminnedId ID of the entity being adminned(not used)
     /// @param admin Admin address whose rank will be returned
     /// @return Admin rank
     function getRank(bytes32 adminnedId, address admin)
@@ -37,14 +38,14 @@ contract Api3RequesterRrpAuthorizer is
     /// @dev Use this instead of setRank(bytes32 adminnedId,address targetAdmin, uint256 newRank)
     /// @param targetAdmin Target admin address
     /// @param newRank Rank to be set
-    function setRank(address targetAdmin, uint256 newRank) public {
+    function setRank(address targetAdmin, uint256 newRank) public override {
         RankedAdminnable.setRank(bytes32(0), targetAdmin, newRank);
     }
 
     /// @notice Called by an admin to decrease its rank
     /// @dev Use this instead of setRank(bytes32 adminnedId, uint256 newRank)
     /// @param newRank Rank to be set
-    function decreaseSelfRank(uint256 newRank) public {
+    function decreaseSelfRank(uint256 newRank) public override {
         RankedAdminnable.decreaseSelfRank(bytes32(0), newRank);
     }
 }
--- a/packages/protocol/contracts/rrp/authorizers/Api3RequesterRrpAuthorizer.sol
+++ b/packages/protocol/contracts/rrp/authorizers/Api3RequesterRrpAuthorizer.sol
@@ -34,13 +34,15 @@ contract Api3RequesterRrpAuthorizer is
 
     /// @notice Called by an admin of higher rank to set the rank of an admin of
     /// lower rank
-    /// @param targetAdmin Target admin addressZ
+    /// @dev Use this instead of setRank(bytes32 adminnedId,address targetAdmin, uint256 newRank)
+    /// @param targetAdmin Target admin address
     /// @param newRank Rank to be set
     function setRank(address targetAdmin, uint256 newRank) public {
         RankedAdminnable.setRank(bytes32(0), targetAdmin, newRank);
     }
 
     /// @notice Called by an admin to decrease its rank
+    /// @dev Use this instead of setRank(bytes32 adminnedId, uint256 newRank)
     /// @param newRank Rank to be set
     function decreaseSelfRank(uint256 newRank) public {
         RankedAdminnable.decreaseSelfRank(bytes32(0), newRank);
--- a/packages/protocol/contracts/rrp/authorizers/Api3RequesterRrpAuthorizer.sol
+++ b/packages/protocol/contracts/rrp/authorizers/Api3RequesterRrpAuthorizer.sol
@@ -32,20 +32,4 @@ contract Api3RequesterRrpAuthorizer is
     {
         return MetaAdminnable.getRank(bytes32(0), admin);
     }
-
-    /// @notice Called by an admin of higher rank to set the rank of an admin of
-    /// lower rank
-    /// @dev Use this instead of setRank(bytes32 adminnedId,address targetAdmin, uint256 newRank)
-    /// @param targetAdmin Target admin address
-    /// @param newRank Rank to be set
-    function setRank(address targetAdmin, uint256 newRank) public override {
-        RankedAdminnable.setRank(bytes32(0), targetAdmin, newRank);
-    }
-
-    /// @notice Called by an admin to decrease its rank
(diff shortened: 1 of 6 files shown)


packages/protocol/contracts/rrp/authorizers/Api3RequesterRrpAuthorizer.sol
```diff
@@ -29,6 +29,22 @@ contract Api3RequesterRrpAuthorizer is
         override(MetaAdminnable, RankedAdminnable, IRankedAdminnable)
         returns (uint256)
     {
-        return MetaAdminnable.getRank(adminnedId, admin);
+        return MetaAdminnable.getRank(bytes32(0), admin);
```
bbenligiray, line 32: The docstring for this function needs to change

packages/protocol/contracts/rrp/authorizers/Api3RequesterRrpAuthorizer.sol
```diff
@@ -29,6 +29,22 @@ contract Api3RequesterRrpAuthorizer is
...
+    }
+
+    /// @notice Called by an admin of higher rank to set the rank of an admin of
+    /// lower rank
+    /// @dev Use this instead of setRank(bytes32 adminnedId,address targetAdmin, uint256 newRank)
+    /// @param targetAdmin Target admin address
+    /// @param newRank Rank to be set
+    function setRank(address targetAdmin, uint256 newRank) public {
```
bbenligiray, line 40: This needs to be added too `IApi3RequesterRrpAuthorizer`

packages/protocol/contracts/rrp/authorizers/Api3RequesterRrpAuthorizer.sol
```diff
@@ -29,6 +29,22 @@ contract Api3RequesterRrpAuthorizer is
...
+        RankedAdminnable.setRank(bytes32(0), targetAdmin, newRank);
+    }
+
+    /// @notice Called by an admin to decrease its rank
+    /// @dev Use this instead of setRank(bytes32 adminnedId, uint256 newRank)
+    /// @param newRank Rank to be set
+    function decreaseSelfRank(uint256 newRank) public {
```
bbenligiray, line 47: This needs to be added too `IApi3RequesterRrpAuthorizer`

bbenligiray:
We override `getRank()` because that is then used internally. So there is not much of an argument for overriding `setRank()` and `decreaseSelfRank()` similarly. I think there are two options:
1. The current implementation (It's impossible to understand why there are two different `setRank()`/`decreaseSelfRank()` and how they should be used without having complete understanding of the contracts. Also it needs an additional `getRank()` with a single argument)
2. Only override `getRank()`, expect the user to use `bytes32(0)` as the `adminnedId`` for `setRank()`/`decreaseSelfRank()` (again, this requires you to know what you're doing)
I think both are equally confusing and ugly, but 2 has fewer lines of code. The worst that can happen with it is that the user may assign adminship for a non-zero `adminnedId` expecting that to do something meaningful but it wouldn't, which is not a critical issue.

So I propose only this change to be done and the others to be reverted:
https://github.com/api3dao/airnode/pull/441/files#diff-8434efca12d424ddbd0e0497c619d8d77d135c2872145c4e48c1d7f42241c36fR32
What do you guys think?

acenolaza:
> We override `getRank()` because that is then used internally. So there is not much of an argument for overriding `setRank()` and `decreaseSelfRank()` similarly. I think there are two options:

internally here means that is called by the `RankedAdminnable.onlyWithRank` modifier? 
 
> 1. The current implementation (It's impossible to understand why there are two different `setRank()`/`decreaseSelfRank()` and how they should be used without having complete understanding of the contracts. Also it needs an additional `getRank()` with a single argument)
> 2. Only override `getRank()`, expect the user to use `bytes32(0)` as the `` adminnedId`` for  ``setRank()`/`decreaseSelfRank()`(again, this requires you to know what you're doing) I think both are equally confusing and ugly, but 2 has fewer lines of code. The worst that can happen with it is that the user may assign adminship for a non-zero`adminnedId` expecting that to do something meaningful but it wouldn't, which is not a critical issue.
> 
> So I propose only this change to be done and the others to be reverted:
> https://github.com/api3dao/airnode/pull/441/files#diff-8434efca12d424ddbd0e0497c619d8d77d135c2872145c4e48c1d7f42241c36fR32
> What do you guys think?

I think I can agree with 1 and 2 but if we plan to "trust" the user interacting and whitelisting against this contract to always use `bytes32(0)` as the `adminnedId` then we must provide extensive documentation and diagrams on how this authorizer contract works and how it should be used.
BTW going back to these contracts after a couple of days was quite challenging 😄

packages/protocol/contracts/rrp/authorizers/Api3RequesterRrpAuthorizer.sol
Ashar2shahid: updated

packages/protocol/contracts/rrp/authorizers/Api3RequesterRrpAuthorizer.sol
Ashar2shahid: My bad, tend to forget to update the interfaces and have to remind myself.I've updated it

packages/protocol/contracts/rrp/authorizers/Api3RequesterRrpAuthorizer.sol
```diff
@@ -29,6 +30,22 @@ contract Api3RequesterRrpAuthorizer is
...
-        return MetaAdminnable.getRank(adminnedId, admin);
+        return MetaAdminnable.getRank(bytes32(0), admin);
+    }
+
+    /// @notice Called by an admin of higher rank to set the rank of an admin of
+    /// lower rank
+    /// @dev Use this instead of setRank(bytes32 adminnedId,address targetAdmin, uint256 newRank)
+    /// @param targetAdmin Target admin address
+    /// @param newRank Rank to be set
+    function setRank(address targetAdmin, uint256 newRank) public override {
```
acenolaza, line 41: This could be external since no internal function is calling it right?

packages/protocol/contracts/rrp/authorizers/Api3RequesterRrpAuthorizer.sol
```diff
@@ -29,6 +30,22 @@ contract Api3RequesterRrpAuthorizer is
...
+        RankedAdminnable.setRank(bytes32(0), targetAdmin, newRank);
+    }
+
+    /// @notice Called by an admin to decrease its rank
+    /// @dev Use this instead of setRank(bytes32 adminnedId, uint256 newRank)
+    /// @param newRank Rank to be set
+    function decreaseSelfRank(uint256 newRank) public override {
```
acenolaza, line 48: Same here

acenolaza (commented):
I guess the tests will still apply even if we decide not to add these functions to the api3 authorizer contract.

Ashar2shahid:
I propose a third option of overriding the setRank and decreaseSelfRank. If that is not viable then we go with 2. The third option makes it so that the user can never make a mistake and also doesn't have the ambiguity of 1 and 2.

bbenligiray:
> internally here means that is called by the RankedAdminnable.onlyWithRank modifier?
Yeah

> I think I can agree with 1 and 2 but if we plan to "trust" the user interacting and whitelisting against this contract to always use bytes32(0) as the adminnedId then we must provide extensive documentation and diagrams on how this authorizer contract works and how it should be used. I usually don't like the idea of trusting the user with doing something.
I don't think this is too bad because the user in this case will be us (for example while implementing the API3 token lock contract)

> I propose a third option of overriding setRank and decreaseSelfRank methods. If that is not viable then we go with 2. The third option makes it so that the user can never make a mistake and also doesn't have the ambiguity of 1 and 2.
The problem here is that the original setRank/decreaseSelfRank methods still exist even if you define new ones with different number of arguments. So not only you need to implement these with one fewer arguments, you also need to override the original ones and "disable" them by doing something like
https://github.com/api3dao/api3-contracts/blob/master/packages/api3-token/contracts/Api3Token.sol#L47
which feels overengineered

Ashar2shahid:
Alright ill go with no.2 and merge it then.

PR #442, opened 2021-08-01
Api3TokenLock Integration and tests

Ashar2shahid:
Api3TokenLock_specs_3

old reference image for better understanding(old and not completely accurate but helps to get an idea.)

This integrates the token lock contracts along with all their tests

Currently the token lock sets the whitelisting `expirationTimestamp` to `type(uint64).max` would setting `setWhitelistStatusPastExpiration` true be better instead?

### DIFF
--- a/packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
+++ b/packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
@@ -1,0 +1,393 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+import "./Api3RequesterRrpAuthorizer.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+import "./interfaces/IApi3Token.sol";
+import "./interfaces/IApi3TokenLock.sol";
+
+/// @title The contract used to lock API3 Tokens in order to gain access to Airnodes
+contract Api3TokenLock is IApi3TokenLock, Ownable {
+    string private constant ERROR_UNAUTHORIZED = "Unauthorized";
+    string private constant ERROR_ZERO_ADDRESS = "Zero address";
+    string private constant ERROR_ZERO_AMOUNT = "Zero amount";
+    string private constant ERROR_INSUFFICIENT_AMOUNT = "Insufficient amount";
+    string private constant ERROR_ALREADY_LOCKED = "Already locked";
+    string private constant ERROR_NOT_LOCKED = "No amount locked";
+    string private constant ERROR_LOCK_PERIOD_NOT_EXPIRED =
+        "Locking period not expired";
+    string private constant ERROR_LOCK_PERIOD_ZERO = "Zero lock period";
+    string private constant ERROR_CLIENT_BLACKLISTED = "Client blacklisted";
+    string private constant ERROR_CLIENT_NOT_BLACKLISTED =
+        "Client not blacklisted";
+
+    /// @dev Address of api3RequesterRrpAuthorizer contract
+    address public api3RequesterRrpAuthorizer;
+    /// @dev Address of Api3Token
+    address public api3Token;
+    /// @dev The Minimum locking time (in seconds)
+    uint256 public minimumLockingTime;
+    /// @dev Lock amount for each user
+    uint256 public lockAmount;
+
+    /// @dev Represents the locked amounts, periods and whitelist counts
+    /// of an airnode-client pair
+    struct AirnodeClient {
+        mapping(address => uint256) lockAmountAt;
+        mapping(address => uint256) canUnlockAt;
+        uint256 whitelistCount;
+    }
+
+    /// @dev Meta admin sets the admin statuses of addresses and has super
+    /// admin privileges
+    address public metaAdmin;
+    /// @dev Stores information about admins
+    mapping(address => AdminStatus) public adminStatuses;
+
+    /// @dev Stores information about all token locks for airnode-client pair
+    mapping(bytes32 => mapping(address => AirnodeClient)) public tokenLocks;
+
+    /// @dev Stores information for blacklisted clients
+    mapping(address => bool) public clientAddressToBlacklistStatus;
+
+    /// @dev Sets the values for {_metaAdmin}, {_minimumLockingTime} and {_lockAmount}
+    /// @param _metaAdmin The address that will be set as meta admin
+    /// @param _minimumLockingTime The minimum lock time of API3 tokens for users
+    /// @param _lockAmount The lock amount
+    constructor(
+        address _metaAdmin,
+        uint256 _minimumLockingTime,
+        uint256 _lockAmount
+    ) {
+        require(_metaAdmin != address(0), ERROR_ZERO_ADDRESS);
+        require(_lockAmount != 0, ERROR_ZERO_AMOUNT);
+
+        minimumLockingTime = _minimumLockingTime;
+        lockAmount = _lockAmount;
+        metaAdmin = _metaAdmin;
+    }
+
+    /// @notice Called by the meta admin to set the meta admin
+    /// @param _metaAdmin Address that will be set as the meta admin
+    function setMetaAdmin(address _metaAdmin) external override {
+        require(msg.sender == metaAdmin, ERROR_UNAUTHORIZED);
+        require(_metaAdmin != address(0), ERROR_ZERO_ADDRESS);
+        metaAdmin = _metaAdmin;
+        emit SetMetaAdmin(metaAdmin);
+    }
+
+    /// @notice Called by the meta admin to set the admin status of an address
+    /// @param _admin Address whose admin status will be set
+    /// @param _status Admin status
+    function setAdminStatus(address _admin, AdminStatus _status)
+        external
+        override
+    {
+        require(msg.sender == metaAdmin, ERROR_UNAUTHORIZED);
+        adminStatuses[_admin] = _status;
+        emit SetAdminStatus(_admin, _status);
+    }
+
+    /// @notice Called by admin to set the address of the API3 Authorizer contract
+    /// @param _api3RequesterRrpAuthorizer Address of the new API3 Authorizer
+    function setApi3RequesterRrpAuthorizer(address _api3RequesterRrpAuthorizer)
+        external
+        override
+    {
+        require(
+            adminStatuses[msg.sender] == AdminStatus.Admin,
+            ERROR_UNAUTHORIZED
+        );
+        require(_api3RequesterRrpAuthorizer != address(0), ERROR_ZERO_ADDRESS);
+        api3RequesterRrpAuthorizer = _api3RequesterRrpAuthorizer;
+        emit SetApi3RequesterRrpAuthorizer(api3RequesterRrpAuthorizer);
+    }
+
+    /// @notice Called by admin to set the address of the API3 Token
+    /// @param _api3Token Address of the new API3 Token
+    function setApi3Token(address _api3Token) external override {
+        require(
+            adminStatuses[msg.sender] == AdminStatus.Admin,
+            ERROR_UNAUTHORIZED
+        );
+        require(_api3Token != address(0), ERROR_ZERO_ADDRESS);
+        api3Token = _api3Token;
+
+        bool isBurner = IApi3Token(api3Token).getBurnerStatus(address(this));
+        if (!isBurner) {
+            IApi3Token(api3Token).updateBurnerStatus(true);
+        }
+
+        emit SetApi3Token(api3Token);
+    }
+
+    /// @notice Called by owner to set the minimum locking time
+    /// @param _minimumLockingTime The new minimum locking time
+    function setMinimumLockingTime(uint256 _minimumLockingTime)
+        external
+        override
+        onlyOwner
+    {
+        minimumLockingTime = _minimumLockingTime;
+        emit SetMinimumLockingTime(_minimumLockingTime);
+    }
+
+    /// @notice Called by owner to set the locking amount
+    /// @param _lockAmount The new lock amount
+    function setLockAmount(uint256 _lockAmount) external override onlyOwner {
+        require(_lockAmount != 0, ERROR_ZERO_AMOUNT);
+        lockAmount = _lockAmount;
+        emit SetLockAmount(_lockAmount);
+    }
+
+    /// @notice Called by owner to set the blacklist status of a client
+    /// @param clientAddress Client address
+    /// @param status Blacklist status to be set
+    function setBlacklistStatus(address clientAddress, bool status)
+        external
+        override
+        onlyOwner
+    {
+        clientAddressToBlacklistStatus[clientAddress] = status;
+        emit SetBlacklistStatus(clientAddress, status, msg.sender);
+    }
+
+    /// @notice Locks API3 Tokens to gain access to Airnodes.
+    /// Airnode-client pair gets authorized as long as there is at least one token lock for given pair.
+    /// @dev The amount to be locked is determined by a memory variable set by
+    /// the owners of the contract.
+    /// @param _airnodeId The airnode id
+    /// @param _clientAddress The client address
+    function lock(bytes32 _airnodeId, address _clientAddress)
+        external
+        override
+    {
+        AirnodeClient storage airnodeClient = tokenLocks[_airnodeId][
+            _clientAddress
+        ];
+        require(
+            airnodeClient.lockAmountAt[msg.sender] == 0,
+            ERROR_ALREADY_LOCKED
+        );
+        require(
+            !clientAddressToBlacklistStatus[_clientAddress],
+            ERROR_CLIENT_BLACKLISTED
+        );
+
+        uint256 expirationTime = block.timestamp + minimumLockingTime;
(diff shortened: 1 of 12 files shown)


packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,393 @@
...
+            expirationTime
+        );
+    }
+
+    /// @notice Burns tokens for a user of an airnode-client pair.
+    /// Can only be done when the client is blacklisted.
+    /// @param _airnodeId The airnode id
+    /// @param _clientAddress The client address
+    /// @param _burnTarget The address of the user
+    function burn(
```
acenolaza, line 337: Does it make sense that anyone can call this function? I'm thinking that maybe only an Admin should be allowed to call it since it looks like a cleanup task to remove token locks for blacklisted requesters.

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,393 @@
...
+                lockAmount
+            )
+        );
+
+        if (airnodeClient.whitelistCount == 1) {
+            IApi3RequesterRrpAuthorizer(api3RequesterRrpAuthorizer)
+                .setWhitelistExpiration(
+                _airnodeId,
+                _clientAddress,
+                type(uint64).max
```
acenolaza, line 196: What about using `expirationTime` here instead? This way we make sure that authorizer will only authorize the requester while tokens are locked.

packages/protocol/contracts/rrp/authorizers/mock/Api3Token.sol
```diff
@@ -0,0 +1,110 @@
+//SPDX-License-Identifier: MIT
```
acenolaza, line 1: It is a little bit weird to see the whole implementation under the mocks folder :)

Can we just add a mocked impl for just the functions needed in the tests?

acenolaza (commented):
I feel like I've missed a call or something because I kept hearing about TokenLock contract but I didn't really knew the requirements or use case for it. Anyway after looking at it I think the implementation is correct. I still haven't reviewed the tests but since I have some questions and comments I wanted to post them to save time while I do that.

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,393 @@
...
+        require(
+            tokenLocks[_airnodeId][_clientAddress].lockAmountAt[msg.sender] >
+                lockAmount,
+            ERROR_INSUFFICIENT_AMOUNT
+        );
+        require(
+            !clientAddressToBlacklistStatus[_clientAddress],
+            ERROR_CLIENT_BLACKLISTED
+        );
+        uint256 withdrawAmount = tokenLocks[_airnodeId][_clientAddress]
```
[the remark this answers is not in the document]
Ashar2shahid: So there is a this statement that checks for that to avoid that. Even if this statement wasn't there, solidity subtractions now throw underflow error without using safemath.
```
require(
            tokenLocks[_airnodeId][_clientAddress].lockAmountAt[msg.sender] >
                lockAmount,
            ERROR_INSUFFICIENT_AMOUNT
);
 ```

File: packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol [hunk 5]
```diff
@@ -0,0 +1,393 @@
...
+    /// @dev Address of Api3Token
+    address public api3Token;
+    /// @dev The Minimum locking time (in seconds)
+    uint256 public minimumLockingTime;
+    /// @dev Lock amount for each user
+    uint256 public lockAmount;
+
+    /// @dev Represents the locked amounts, periods and whitelist counts
+    /// of an airnode-client pair
+    struct AirnodeClient {
```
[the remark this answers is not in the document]
Ashar2shahid: Agreed making the changes

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,393 @@
...
+        mapping(address => uint256) lockAmountAt;
+        mapping(address => uint256) canUnlockAt;
+        uint256 whitelistCount;
```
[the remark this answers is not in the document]
Ashar2shahid: So each struct is mapped to an airnode requester pair and any number of addresses can lock for this airnode requester pair. So we have to keep track of them. One way would be to use an array the other is this way.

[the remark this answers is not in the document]
acenolaza: oh I see, so for instance if I need 100 api3 tokens locked then I can have 10 different addresses that lock 10 api3 each in order for 1 requester to access the airnode.

packages/protocol/test/authorizers/Api3TokenLock.sol.js
```diff
@@ -0,0 +1,702 @@
...
+          timeTravel(ONE_MINUTE_IN_SECONDS + 1000);
+          // then:
+          await expect(api3TokenLock.connect(alice).unlock(airnodeId, bobClient.address)).to.revertedWith(
+            Errors.ClientBlacklisted
+          );
+        });
+      });
+    });
+
+    describe('withdraw', async () => {
```
acenolaza, line 465: Same here. Maybe we could try adding more token locks and make sure everything works as expected with more than one.

acenolaza (commented):
Nothing major to object on the tests and I really liked how they were structured and organized 🔥 

Added a few more comments and I can review them again after more changes are added to the contracts.

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
Ashar2shahid: I don't fully understand what you mean here

packages/protocol/contracts/rrp/authorizers/mock/Api3Token.sol
Ashar2shahid: yeah I've updated with only the required functions

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
acenolaza: I was trying to address this question you'd made: 

> Currently the token lock sets the whitelisting expirationTimestamp to type(uint64).max would setting setWhitelistStatusPastExpiration true be better instead?

I personally think it might be a little bit dangerous to set the `whitelistedPastExpiration` var to true and have the requester whitelisted by accident. So my thinking was to remove the check in line [200](https://github.com/api3dao/airnode/blob/57e74a00231a723ecfaedd3c28dea7655227ac88/packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol#L200) and use the value from [187](https://github.com/api3dao/airnode/blob/57e74a00231a723ecfaedd3c28dea7655227ac88/packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol#L187) everytime a token lock is added to the airnode-client pair.

Ashar2shahid: so the check in line 200 is necessary because of multiple locks of single airnode-requester pair i mentioned before.

Ah i see what you mean. So after the minimumLocking period the requester doesn't need to lock tokens and will be authorized indefinitely according to burak. That's why its been set to infinity.

acenolaza: ah! that makes a lot of sense. Thanks for the explanation 💪🏻

acenolaza (approved):

packages/protocol/test/authorizers/Api3TokenLock.sol.js
Ashar2shahid: yup done

packages/protocol/test/authorizers/Api3TokenLock.sol.js
```diff
@@ -0,0 +1,702 @@
...
+
+        it('emits event', async () => {
+          // then:
+          await expect(api3TokenLock.connect(alice).lock(airnodeId, bobClient.address))
+            .to.emit(api3TokenLock, 'Lock')
+            .withArgs(airnodeId, bobClient.address, alice.address, LOCK_AMOUNT);
+        });
+      });
+
+      context('caller has already locked', async () => {
```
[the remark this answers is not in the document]
Ashar2shahid: done

packages/protocol/test/authorizers/Api3TokenLock.sol.js
```diff
@@ -0,0 +1,702 @@
...
+      });
+
+      context('caller is first time locking', async () => {
+        beforeEach(async () => {
+          // given:
+          await api3Token.connect(alice).approve(api3TokenLock.address, await api3TokenLock.lockAmount());
+        });
+
+        it('is set', async () => {
+          const beforeBalance = await api3Token.balanceOf(alice.address);
```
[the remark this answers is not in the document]
Ashar2shahid: yes makes sense

Ashar2shahid:
@acenolaza I've updated the tests you can have one final look before I merge this tomorrow afternoon

packages/protocol/contracts/rrp/authorizers/interfaces/IApi3TokenLock.sol
```diff
@@ -0,0 +1,91 @@
...
+    event SetApi3RequesterRrpAuthorizer(address api3RequesterRrpAuthorizer);
+
+    event SetApi3Token(address api3Token);
+
+    event SetMinimumLockingTime(uint256 minimumLockingTime);
+
+    event SetLockAmount(uint256 lockAmount);
+
+    event SetBlacklistStatus(
+        address indexed clientAddress,
```
acenolaza, line 23: clientAddress -> requester

packages/protocol/contracts/rrp/authorizers/interfaces/IApi3TokenLock.sol
```diff
@@ -0,0 +1,91 @@
...
+        bool status,
+        address indexed admin
+    );
+
+    event Lock(
+        bytes32 airnodeId,
```
acenolaza, line 29: Also airnodeId -> airnode

packages/protocol/contracts/rrp/authorizers/interfaces/IApi3TokenLock.sol
```diff
@@ -0,0 +1,91 @@
...
+        address clientAddress,
+        address requester,
+        uint256 lockedAmount
+    );
+
+    event Unlock(
+        bytes32 airnodeId,
+        address clientAddress,
+        address requester,
```
acenolaza, line 38: requester -> sponsor

packages/protocol/test/authorizers/Api3TokenLock.sol.js
```diff
@@ -0,0 +1,702 @@
...
+      context('caller is not admin', async () => {
+        it('reverts', async () => {
+          await expect(
+            api3TokenLock.connect(unknown).setApi3RequesterRrpAuthorizer(api3RequesterRrpAuthorizer.address)
+          ).to.revertedWith(Errors.Unauthorized);
+        });
+      });
+    });
+
+    describe('setApi3Token', async () => {
```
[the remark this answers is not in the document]
acenolaza: Here what I meant was checking `updateBurnerStatus` is set to true.

acenolaza (approved):
LGTM 🔥 Just need to also rename params in interfaces

bbenligiray:
I wonder if the path should instead be `rrp/authorizers/admins/Api3TokenLock.sol`

Ashar2shahid:
I haven't updated the tests for the new contracts, will do so once we finalize the contracts.

- Contracts are metaAdminnable with 3 levels of heirarchy (MetaAdmin,Admin and Unauthorized).
- Blacklisting can be set by the metaAdmin and the airnodeAdmin
- Admins can burn the tokens

There are two things to keep in mind:
- Blacklisting globally (address(0)) does not globally whitelist from all airnodes, all it does is not allow the user to lock/unlock tokens anymore
- The airnode admin being able to blacklist is a nice thing to have, I've done a implementation here that is not too convoluted. Take a look and let me know if it can be improved.

packages/protocol/contracts/rrp/authorizers/interfaces/IApi3TokenLockExternal.sol
```diff
@@ -0,0 +1,110 @@
...
+    function withdraw(
+        uint256 chainId,
+        address airnode,
+        address requesterAddress
+    ) external;
+
+    function transfer(
+        uint256 chainId,
+        address airnode,
+        address fromClientAddress,
```
acenolaza, line 100: fromRequesterAddress and toRequesterAddress instead?

packages/protocol/contracts/rrp/authorizers/interfaces/IApi3TokenLockExternal.sol
```diff
@@ -0,0 +1,110 @@
...
+        address airnode,
+        address requesterAddress,
+        address sponsor,
+        uint256 lockedAmount
+    );
+
+    event Transfer(
+        uint256 chainId,
+        address airnode,
+        address fromClientAddress,
```
acenolaza, line 38: fromRequesterAddress and toRequesterAddress instead?

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,419 @@
...
+pragma solidity 0.8.6;
+
+import "./Api3RequesterRrpAuthorizer.sol";
+import "./interfaces/IApi3Token.sol";
+import "./interfaces/IApi3TokenLock.sol";
+import "../../admin/MetaAdminnable.sol";
+import "../../admin/RankedAdminnable.sol";
+
+/// @title The contract used to lock API3 Tokens in order to gain access to Airnodes
+contract Api3TokenLock is RankedAdminnable, MetaAdminnable, IApi3TokenLock {
```
acenolaza, line 11: Why do we need to inherit RankedAdminnable if we are already getting it's functionality thru MetaAdminnable? Am I missing something in the getRank() function? MetaAdminable overrides this function and then we also override it here so the Api3TokenLock.getRank() code will be the one executed I think.

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,419 @@
...
+    enum AdminRank {
```
acenolaza, line 12: I think the convention is to put enums in the interface (IApi3TokenLock in this case). Not a big deal to me personally tho.

packages/protocol/contracts/rrp/authorizers/interfaces/IApi3TokenLockExternal.sol
Ashar2shahid: yes my bad forgot about these two

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,419 @@
...
+    {
+        AirnodeRequester storage from = airnodeToRequesterToTokenLocks[
+            _airnode
+        ][_fromRequesterAddress];
+        AirnodeRequester storage to = airnodeToRequesterToTokenLocks[_airnode][
+            _toRequesterAddress
+        ];
+        require(
+            from.lockAmountAt[msg.sender] != 0,
+            "locked amount must be != 0"
```
acenolaza, line 320: Should we also put this error msg in a constant along with the others?

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
Ashar2shahid: I think you're right. It's how it was done in the Api3RequesterRrpAuthorizer but that is inherriting two different RankedAdminnable contracts which is why we explicitly overrrided there. Not the case here

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
Ashar2shahid: Its how it was done in the Whitelister.sol so I just replicated that

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
Ashar2shahid: hmm yes, that would be better.

Ashar2shahid:
> I wonder if the path should instead be `rrp/authorizers/admins/Api3TokenLock.sol`

Intially I was thinking of putting it in a seprate folder but couldn't decide on a name. Why admin could work but what about 
`rrp/authorizers/subscriptions/Api3TokenLock.sol`

acenolaza:
specs [document](https://docs.google.com/document/d/1WSq97TYHL9p62UqGC5AuF48xFtcN7Q5f5FnkLcWIuTg/edit#heading=h.4hn6gl4fxwr8)

Ashar2shahid:
Closing this and opening another one, since all the tests and contracts have changed

PR #454, opened 2021-08-02
Add `testable` flag to mark endpoints that can be tested

amarthadan:


packages/node/src/handlers/test-api.ts
```diff
@@ -9,17 +9,25 @@ import { WORKER_CALL_API_TIMEOUT } from '../constants';
...
-    return [new Error(`No such endpoint '${endpointName}'`), null];
+    return [new Error(`No such endpoint with ID '${endpointId}'`), null];
+  }
+
+  const endpoint = find(find(config.ois, ['title', triggerRequest.oisTitle])?.endpoints, [
+    'name',
+    triggerRequest.endpointName,
+  ]);
+  if (!endpoint || !endpoint.testable) {
+    return [new Error(`Endpoint with ID '${endpointId}' can't be tested`), null];
```
andreogle, line 30: I would say separate the endpoint doesn't exist error from the endpoint is not testable error

packages/node/__dev__/test-api.ts
```diff
@@ -15,9 +15,9 @@ yargs(hideBin(process.argv))
     '$0',
     'Test API calls',
     {
-      'endpoint-name': {
+      'endpoint-id': {
         alias: 'e',
-        description: 'Endpoint name',
+        description: 'Endpoint id',
```
andreogle, line 20: I would expand on this description. As a user, when I try to view the help/description of a command and it returns the same string as the key, I find it kinda annoying.

Maybe something like:

```suggestion
        description: 'The ID of the endpoint to be tested, which is derived from the OIS title and endpoint name',
```

packages/node/src/handlers/test-api.ts
```diff
@@ -9,17 +9,25 @@ import { WORKER_CALL_API_TIMEOUT } from '../constants';
...
   if (!triggerRequest) {
-    return [new Error(`No such endpoint '${endpointName}'`), null];
+    return [new Error(`No such endpoint with ID '${endpointId}'`), null];
+  }
+
+  const endpoint = find(find(config.ois, ['title', triggerRequest.oisTitle])?.endpoints, [
+    'name',
+    triggerRequest.endpointName,
+  ]);
+  if (!endpoint || !endpoint.testable) {
```
andreogle, line 29: I like that this is opt-in. i.e. if you don't specify exactly `"testable": true`, then it's not testable

andreogle (approved):
👍🏻

packages/node/src/handlers/test-api.ts
```diff
@@ -36,8 +44,8 @@ export async function testApi(
     clientAddress: '',
     designatedWallet: '',
     chainId: '',
-    endpointId: triggerRequest.endpointId,
-    endpointName,
+    endpointId,
+    endpointName: triggerRequest.endpointName,
```
Siegrift, line 48: Github doesn't allow me to comment on non changed lines, but can you please fix a typo on line 61?

`An unkNown error occurred`

packages/node/src/handlers/test-api.ts
```diff
@@ -9,17 +9,25 @@ import { WORKER_CALL_API_TIMEOUT } from '../constants';
...
   const logOptions = logger.buildBaseOptions(config, { requestId: testCallId });
 
-  const triggerRequest = find(config.triggers.request, ['endpointName', endpointName]);
+  const triggerRequest = find(config.triggers.request, ['endpointId', endpointId]);
   if (!triggerRequest) {
-    return [new Error(`No such endpoint '${endpointName}'`), null];
+    return [new Error(`No such endpoint with ID '${endpointId}'`), null];
+  }
+
+  const endpoint = find(find(config.ois, ['title', triggerRequest.oisTitle])?.endpoints, [
```
Siegrift, line 25:
```suggestion
  const endpoint = find(config.ois, ['title', triggerRequest.oisTitle, "endpoints", "name", triggerRequest.endpointName])
```

doesn't work?

Siegrift (changes requested):
What about the README? Shouldn't it be updated as well? 

Or it's enough to update api3 docs and tell Warren?

amarthadan:
> What about the README? Shouldn't it be updated as well?
> 
> Or it's enough to update api3 docs and tell Warren?

I can mention the `dev:testApi` target in the README. I'll do that in the next PR with tests. Anything else should IMO go to the docs.

packages/node/src/handlers/test-api.ts
Siegrift: Sry, I was confused. It won't work. But you can make the whole thing more readble by:
```
const endpoints = find(config.ois, ['title', triggerRequest.oisTitle])?.endpoints
const endpoint = find(endpoint, ["name", triggerRequest.endpointName])
```

Siegrift (approved):

packages/node/src/handlers/test-api.ts
```diff
@@ -9,17 +9,27 @@ import { WORKER_CALL_API_TIMEOUT } from '../constants';
...
 
-  const triggerRequest = find(config.triggers.request, ['endpointName', endpointName]);
+  const triggerRequest = find(config.triggers.request, ['endpointId', endpointId]);
   if (!triggerRequest) {
-    return [new Error(`No such endpoint '${endpointName}'`), null];
+    return [new Error(`No such endpoint with ID '${endpointId}'`), null];
+  }
+
+  const endpoints = find(config.ois, ['title', triggerRequest.oisTitle])?.endpoints;
+  const endpoint = find(endpoints, ['name', triggerRequest.endpointName]);
```
andreogle, line 26: What happens if this `endpoints` is undefined? Maybe we need handle that error too before the second `find()`?

andreogle (changes requested):

packages/node/src/handlers/test-api.ts
amarthadan: We don't actually. Lodash's `find` is ok with `undefined` and it returns `undefined`, same as if the endpoint can't be found.

andreogle: oh nice 👍🏻

andreogle (approved):

PR #440, opened 2021-07-29
fix(#427):Admin package is not listening for the correct event

acenolaza:
Fixes regression bug.

Tests:
Ran e2e test suite.

Notes:
I didn't change the `fulfillWithdrawal` becuase in this particular case we are passing specific values as parameters to the filter.

PR #455, opened 2021-08-02
Fix admin CLI listening for incorrect event

Siegrift:
Basically the same changes as in https://github.com/api3dao/airnode/pull/440

but the base branch is current `master` and can be merged as of now. We might want to update the npm package as well.

Maybe sometime, we will be able to have better type awareness using typeechain https://github.com/ethereum-ts/TypeChain/issues/440


acenolaza (approved):
LGTM :shipit:

Issue #275, opened 2021-04-11
Consistent output of admin CLI


Siegrift:
Currently some of the commands use ad-hoc format, e.g.

```js
console.log(`Requester admin: ${requesterAdmin}`);
```

and some output the whole JSON:

```js
console.log(toJSON(parameters));
```

The output of the commands should be consistent and I'd say it's better to avoid ad-hoc format and just display the data. Reasons:
1) It's easier to test - no need to strip the ad-hoc prefix
2) Easier to process if someone uses the CLI in some pipeline - for the same reason as 1)
3) The output of the command has enough context for user to know what is returned - it should be clear even from the command name, but there is also description
4) We can wrap the response from blockchain as JSON in the SDK - e.g. return `{"requesterIndex": "2"}` from the CLI.

We could also introduce something like `--pretty` flag to pretty print json, but I think it's not necessary as users can just format it in code editor or online tool.

bbenligiray:
I know it's not really a bug-fix, but it's closer to that than an enhancement.

bbenligiray:
Tracked by https://api3dao.atlassian.net/browse/AN-151

PR #456, opened 2021-08-02
Fix admin package

acenolaza:
Main changes:
1. rename clients with requesters and requesters with sponsors
2. setAirnodeParameters was replaced by setAirnodeXpub ~~and serAirnodeAuthorizers~~
3. allow to send xpub as arg in derive-sponsor-wallet command

Tests:
1. ran e2e


packages/admin/test/e2e/cli.feature.ts
```diff
@@ -316,43 +291,51 @@ describe('CLI', () => {
...
+    const airnodeXpub = airnodeHdNode.neuter().extendedKey;
+
+    const setAirnodeXpubOut = execCommand(
+      'set-airnode-xpub',
+      ['--mnemonic', airnodeWallet.mnemonic.phrase],
+      ['--derivationPath', airnodeWallet.mnemonic.path],
+      ['--providerUrl', PROVIDER_URL],
+      ['--airnodeRrp', airnodeRrp.address]
+    );
+    expect(setAirnodeXpubOut).toMatch(new RegExp('Airnode xpub: \\w+'));
```
Siegrift, line 305:
```suggestion
    expect(setAirnodeXpubOut).toEqual(`Airnode xpub: ${airnodeXpub}`);
```

packages/admin/README.md
```diff
@@ -61,50 +61,50 @@ Please, refer to the implementation for more details.
 
 ## Requester commands
 
-### `derive-designated-wallet`
+### `derive-sponsor-wallet`
 
 Derives the address of the [wallet designated by an Airnode for a requester](https://github.com/api3dao/api3-docs/blob/master/request-response-protocol/designated-wallet.md).
```
Siegrift, line 66: This should probably be updated as well? (At least the filename should be renamed)

packages/admin/README.md
```diff
@@ -61,50 +61,50 @@ Please, refer to the implementation for more details.
...
+  --mnemonic "nature about salad..." \
+  --sponsor 0x9Ec6C4...
 ```
 
-### `endorse-client`
+### `endorse-requester`
 
-[Endorses](https://github.com/api3dao/api3-docs/blob/master/request-response-protocol/endorsement.md) a client contract so that its requests can be fulfilled by the requester's designated wallet.
-The account derived from the `mnemonic` you provide here has to belong to the requester admin (master wallet).
+[Endorses](https://github.com/api3dao/api3-docs/blob/master/request-response-protocol/endorsement.md) a requester contract so that its requests can be fulfilled by the sponsor's designated wallet.
```
Siegrift, line 77: These links regarding the request response protocol need to be updated as well.

packages/admin/src/cli.ts
```diff
@@ -20,27 +20,32 @@ const COMMON_COMMAND_ARGUMENTS = {
...
     demandOption: true,
-    describe: 'Requester address',
+    describe: 'Sponsor address',
   },
-  airnodeId: {
+  airnode: {
     type: 'string',
     demandOption: true,
-    describe: 'The onchain ID of the Airnode instance',
+    describe: 'Address of the Airnode wallet',
```
Siegrift, line 38: Is it airnode wallet? Isn't it an address where the airnode is deployed?

packages/admin/src/cli.ts
```diff
@@ -20,27 +20,32 @@ const COMMON_COMMAND_ARGUMENTS = {
...
   },
-  clientAddress: {
+  sponsorWallet: {
+    type: 'string',
+    demandOption: true,
+    describe: 'Address of the Sponsor wallet',
```
Siegrift, line 43: I see that some of the words are capitalized (Sponsor, Airnode) an some are not (requester). It should be somehow consistent I think.

packages/admin/README.md
```diff
@@ -8,7 +8,7 @@ Almost all commands require you to provide a `providerUrl` such as `https://rops
 The CLI connects to [AirnodeRrp.sol](https://github.com/api3dao/airnode/blob/master/packages/protocol/contracts/AirnodeRrp.sol) contract, which address is derived from the current chain.
 You can optionally specify the contract address yourself by providing optional `airnodeRrp` command argument with address of the deployed contract on your targeted chain.
 
-Commands that require `mnemonic` will make an on-chain transaction.
+Commands that require `mnemonic` will make an on-chain transaction (with the exception of `derive-sponsor-wallet`).
```
Siegrift, line 11: I think there are more exceptions, no? All view calls don't require mnemonic, no?

packages/admin/test/e2e/cli.feature.ts
```diff
@@ -88,118 +95,98 @@ describe('CLI', () => {
...
+      ['--sponsor', sponsor]
     );
 
     // Derive the wallet programatically
-    const designatedWallet = await deriveDesignatedWallet(airnodeWallet, requester);
+    const sponsorWallet = await deriveSponsorWallet(airnodeWallet, sponsor);
 
     // Check that they generate the same wallet address
-    expect(out).toBe(`Designated wallet address: ${designatedWallet.address}`);
+    expect(out).toBe(`Sponsor wallet address: ${sponsorWallet.address}`);
```
Siegrift, line 131: While I was doing the initial refactor of this package I wasn't sure whether to have these ad-hoc messages prepended or not, so I kept them.

WDYT? Should we remove them and only output the raw data? (We could say what the command returns in the README and CLI help).

packages/admin/src/cli.ts
Siegrift: It seems this is not a mistake... Can you explain how did we change using airnodeId and are using airnode wallet instead?

(Also, documentation in `WithdrawalUtils` for the airnode arguments could say `Airnode wallet address` instead of `Airnode address` to avoid confusion?)

Siegrift (changes requested):

packages/admin/src/cli.ts
acenolaza: Previously we hashed the airnode master wallet address in order to get the airnodeId. Now that we no longer have an airnodeAdmin and master wallet is the default admin/owner of the airnode then there's really no need for an ID and we've decided to just use the wallet address as the "airnode id".
More on this here: https://github.com/api3dao/airnode/pull/433

Regarding WithdrawalUtils I can agree on adding the wallet word but keep in mind that this contract is being extended by airnode contract and it is not meant to be deployed or used on it's own.

packages/admin/src/cli.ts
acenolaza: My bad. I'll make this naming consistent.

packages/admin/README.md
acenolaza: This is something I wasn't sure what to do. If I update the file name then the link will be broken until the docs are updated.

packages/admin/README.md
acenolaza: Same here. What should we link to? Should we just wait for the docs to be updated and then come back and fix these links?

packages/admin/test/e2e/cli.feature.ts
acenolaza: I think the SDK might be used by another application but CLI seems to me that it's meant to be used by a user. Therefore I like the messages we are currently showing becuase they give more info on the value being returned. With that said if we ever support localization and multiple languages on the admin package then it will be more work.

packages/admin/README.md
acenolaza: When I added that comment I saw only that one but let me double check. Actually after today's protocol-team meeting I need to refactor derive-sponsor-wallet again and it will no longer be an exception.

packages/admin/README.md
Siegrift: I am not sure how the docs are versioned, but the documentation in the docs should definitely update as well.
That said, it can be done afterwards. 

@bbenligiray @wkande what is the best approach for this?

Siegrift: We have https://docs.api3.org/next/ so it should be probably updated there.

I am not sure if you want to wait until it's done, but we should at least write a TODO: in comment to update it later (https://stackoverflow.com/questions/46734820/hidden-markdown-text-on-github/46734863)

packages/admin/src/cli.ts
Siegrift: I see, makes sense now.

packages/admin/test/e2e/cli.feature.ts
Siegrift: Ok.

In that case, write some message also in https://github.com/api3dao/airnode/pull/456/files#diff-27795c748e8300543cfa14348dc8f4789262138f8d88dcce430afbf765677a97R252

packages/admin/README.md
acenolaza: deprecated

packages/admin/test/e2e/cli.feature.ts
Siegrift: Actually, now I remember there are other reasons to avoid that "additional ad-hoc info" https://github.com/api3dao/airnode/issues/275

But it's OK to leave this as is for now.

Siegrift (approved):

acenolaza:
Removed the endorsement terminology from cli commands based on this [document](https://docs.google.com/document/d/1dWq7e0Kle8G16wscwwwz2SgpgiuoavBvGxdpCwMMhA4/edit?pli=1#heading=h.3jh2y6ljfa8p). 

I'm not really good at naming things so I'd like your opinion on the new names for the commands @bbenligiray @Siegrift

packages/admin/README.md
bbenligiray: I think the main problem here is that the docs and these READMEs are in separate repos. I'd rather keep the READMEs as light as possible and have these in the docs repo. Let's talk about it at the call next week.

packages/admin/README.md
bbenligiray: Remove the links for now maybe

bbenligiray (commented):
[Requester commands](https://github.com/api3dao/airnode/blob/681e76e6d75bd0f705feb700bd60ec2a1bd7ed36/packages/admin/README.md#requester-commands) -> Developer commands

start-sponsoring-requester -> sponsor-requester

stops-sponsoring-requester -> unsponsor-requester (sounds weird but very clear in meaning imo)

acenolaza:
> [Requester commands](https://github.com/api3dao/airnode/blob/681e76e6d75bd0f705feb700bd60ec2a1bd7ed36/packages/admin/README.md#requester-commands) -> Developer commands
> 
> start-sponsoring-requester -> sponsor-requester
> 
> stops-sponsoring-requester -> unsponsor-requester (sounds weird but very clear in meaning imo)

👍🏻 
I thought "unsponsor" was not a word in english and that's when I started getting creative with the command names 😆

PR #458, opened 2021-08-03
Tests for API test calls

amarthadan:


packages/node/src/handlers/test-api.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  const endpoints = endpoint ? [endpoint] : [];
+  return fixtures.buildConfig({ ois: [fixtures.buildOIS({ endpoints })] });
+}
+
+describe('testApi', () => {
+  it('returns an error if no endpoint trigger with given ID is found', async () => {
+    const nonexistentEndpointId = '0xeddc421714e1b46ef350e8ecf380bd0b38a40ce1a534e7ecdf4db7dbc931ffff';
+    const [err, res] = await testApi(fixtures.buildConfig(), nonexistentEndpointId, {});
+    expect(res).toBeNull();
+    expect(err).not.toBeNull();
```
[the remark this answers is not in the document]
Siegrift: I think it doesn't hurt and it carries a better intention than just `notNull` check.

packages/node/src/handlers/test-api.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    const unspecifiedEndpoint = fixtures.buildOIS().endpoints[0];
+    // eslint-disable-next-line functional/immutable-data
+    delete unspecifiedEndpoint.testable;
+
+    const [err, res] = await testApi(buildConfigWithEndpoint(unspecifiedEndpoint), ENDPOINT_ID, {});
+    expect(res).toBeNull();
+    expect(err).not.toBeNull();
+  });
+
+  it('returns an error if endpoint testability is off', async () => {
```
Siegrift, line 37:
```suggestion
  it('returns an error if endpoint testability is turned off', async () => {
```

packages/node/README.md
```diff
@@ -41,7 +43,15 @@ Airnode does not yet have an stable API for usage. However, you can run Airnode
 yarn run dev:invoke
 ```
 
-You must have a valid `config.json` file present in the `__dev__` folder. It is also recommended to use a `.env` file for handling secrets. See [Development](#Development) below for more details.
+### Testing API
+
+You can test the endpoints specified in your `config.json` by running the following command:
+
+```sh
+# -e Endpoint ID
```
Siegrift, line 51: I'd remove these aliases and instead used a full name...

Siegrift (approved):
:1st_place_medal: 

@amarthadan @andreogle WDYT about adding an e2e test like this into CI pipeline? It would spin up a lambda locally, call it, wait for response and asserts a certain form of response...
E.g. query current ETH/USD value and assert a single floating point number is returned.

I think it would bring up a huge benefit, because it would ensure a valid config.json (current one is not) and also describe the full airnode flow (and ensure we don't break it).

Currently it's kinda challenging to even perform the basic airnode flow, isn't it?

PR #463, opened 2021-08-05
Fixes operation package

acenolaza:
Main changes:
1. Renames clients with requesters and requesters with sponsors.
2. No longer sets authorizers on chain and just sets xpub (no needed but nice to have)
3. Updates requester contract reference (MockRrpRequesterFactory)

Tests:
1. ran all steps except for `invoke` since node package is still broken

### DIFF
--- a/packages/operation/README.md
+++ b/packages/operation/README.md
@@ -195,7 +195,7 @@ There are currently three types of requests that can be made. You can learn more
 
 **Withdrawals**
 
-`destination` - the ID of the sponsor to which the withdrawn funds should be sent. Can also be an address starting with `0x`.
+The withdrawn funds should be sent back to the address of the sponsor.
 
 #### Full Example
 
@@ -275,8 +275,7 @@ There are currently three types of requests that can be made. You can learn more
     {
       "sponsorId": "alice",
       "type": "withdrawal",
-      "airnode": "CurrencyConverterAirnode",
-      "destination": "alice"
+      "airnode": "CurrencyConverterAirnode"
     }
   ]
 }
--- a/packages/operation/README.md
+++ b/packages/operation/README.md
@@ -145,7 +145,7 @@ This is the index that will be used to select an account from the list of accoun
 
 `requesters` - a key/value object where the key represents the unique requester contract name and the value represents the requester options. All names defined correspond with actual contracts in the `contracts/folder`. See the [client documentation](https://github.com/api3dao/api3-docs/blob/master/request-response-protocol/requester.md) for more details.
 
-`requester.[name].endorsers` - a list of sponsors who have endorsed the requester. See the [endorsement documentation](https://github.com/api3dao/api3-docs/blob/master/request-response-protocol/endorsement.md) for more details.
+`requester.[name].sponsors` - a list of sponsors who have sponsored the requester. See the [sponsorship documentation](https://github.com/api3dao/api3-docs/blob/master/request-response-protocol/endorsement.md) for more details.
 
 ### 5. sponsors
 
@@ -230,7 +230,7 @@ There are currently three types of requests that can be made. You can learn more
     "public": "0x0000000000000000000000000000000000000000"
   },
   "requesters": {
-    "MockRrpRequesterFactory": { "endorsers": ["bob"] }
+    "MockRrpRequesterFactory": { "sponsors": ["bob"] }
   },
   "sponsors": [
     {
--- a/packages/operation/README.md
+++ b/packages/operation/README.md
@@ -1,3 +1,5 @@
+<!-- TODO: update links after api3-docs had been updated and use https://docs.api3.org/ url instead -->
+
 # @api3/operation
 
 > Development and testing utilities for Airnode
--- a/packages/operation/README.md
+++ b/packages/operation/README.md
@@ -41,7 +41,7 @@ yarn run dev:eth-node
 # OR to start as a background process
 yarn run dev:eth-node:background
 
-# Deploy Airnode RRP with Airnode parameters, templates, requesters etc. This creates a "deployment" file in a deployments/ folder.
+# Deploy Airnode RRP with Airnode xpub, templates, requesters etc. This creates a "deployment" file in a deployments/ folder.
 yarn run dev:eth-deploy
 
 # Make requests for Airnode to action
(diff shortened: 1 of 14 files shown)


packages/operation/README.md
```diff
@@ -127,45 +133,45 @@ Deployment can be configured by adjusting the `config/eth-dev-config.json` file.
 
 `templates.[name].oisTitle` - the title of an [OIS](https://github.com/api3dao/api3-docs/blob/master/airnode/ois.md). This is used to derive the [endpointId](https://github.com/api3dao/api3-docs/blob/master/request-response-protocol/endpoint.md) which should match an `endpointId` in a [config.json trigger](https://github.com/api3dao/api3-docs/blob/master/airnode/config-json.md)
 
-`templates.[name].parameters` - a list of parameters that be encoded directly using [airnode-abi](https://github.com/api3dao/airnode/tree/master/packages/airnode-abi)
+`templates.[name].parameters` - a list of parameters that will be encoded directly using [airnode-abi](https://github.com/api3dao/airnode/tree/master/packages/airnode-abi)
```
andreogle, line 136: 😅

packages/operation/README.md
```diff
@@ -127,45 +133,45 @@ Deployment can be configured by adjusting the `config/eth-dev-config.json` file.
...
-### 3. clients
+### 4. requesters
 
-`clients` - a key/value object where the key represents the unique client contract name and the value represents the client options. All names defined correspond with actual contracts in the `contracts/folder`. See the [client documentation](https://github.com/api3dao/api3-docs/blob/master/request-response-protocol/client.md) for more details.
+`requesters` - a key/value object where the key represents the unique requester contract name and the value represents the requester options. All names defined correspond with actual contracts in the `contracts/folder`. See the [client documentation](https://github.com/api3dao/api3-docs/blob/master/request-response-protocol/requester.md) for more details.
 
-`client.[name].endorsers` - a list of requesters who have endorsed the client. See the [endorsement documentation](https://github.com/api3dao/api3-docs/blob/master/request-response-protocol/endorsement.md) for more details.
+`requester.[name].endorsers` - a list of sponsors who have endorsed the requester. See the [endorsement documentation](https://github.com/api3dao/api3-docs/blob/master/request-response-protocol/endorsement.md) for more details.
```
andreogle, line 146: These docs should be updated to point to the https://docs.api3.org site at some point

packages/operation/src/evm/deploy/account-assignment.ts
```diff
@@ -55,8 +55,8 @@ export async function assignDesignatedWallets(state: State) {
       };
     });
 
-    requestersById[configRequester.id] = { ...requester, designatedWallets };
+    sponsorsById[configSponsor.id] = { ...sponsor, designatedWallets };
   }
 
-  return { ...state, requestersById };
+  return { ...state, requestersById: sponsorsById };
```
andreogle, line 61: Shouldn't this also be `sponsorsById`?
```suggestion
  return { ...state, sponsorsById };
```

andreogle: We should try set explicity types so that TS catches things like this

packages/operation/src/evm/utils.ts
```diff
@@ -6,29 +6,22 @@ export function deriveEndpointId(oisTitle: string, endpointName: string): string
...
 export function deriveExtendedPublicKey(mnemonic: string): string {
-  const wallet = ethers.Wallet.fromMnemonic(mnemonic);
-  const hdNode = ethers.utils.HDNode.fromMnemonic(wallet.mnemonic.phrase);
+  const hdNode = ethers.utils.HDNode.fromMnemonic(mnemonic);
   return hdNode.neuter().extendedKey;
 }
 
-export function deriveAirnodeId(masterWalletAddress: string): string {
-  return ethers.utils.keccak256(ethers.utils.defaultAbiCoder.encode(['address'], [masterWalletAddress]));
+export function deriveWalletFromPath(mnemonic: string, provider: ethers.providers.JsonRpcProvider, path?: string) {
```
andreogle, line 13: Shouldn't `path` be required here?

packages/operation/src/evm/utils.ts
```diff
@@ -6,29 +6,22 @@ export function deriveEndpointId(oisTitle: string, endpointName: string): string
...
 
-export function deriveWalletFromPath(mnemonic: string, path: string, provider: ethers.providers.JsonRpcProvider) {
-  const masterHdNode = ethers.utils.HDNode.fromMnemonic(mnemonic);
-  const designatorHdNode = masterHdNode.derivePath(path);
-  return new ethers.Wallet(designatorHdNode.privateKey, provider);
-}
-
-export function getDesignatedWallet(mnemonic: string, requester: string, provider: ethers.providers.JsonRpcProvider) {
-  return deriveWalletFromPath(mnemonic, `m/0/${addressToDerivationPath(requester)}`, provider);
+export function getDesignatedWallet(mnemonic: string, provider: ethers.providers.JsonRpcProvider, sponsor: string) {
```
andreogle, line 17: Maybe a bit clearer?

```suggestion
export function getDesignatedWallet(mnemonic: string, provider: ethers.providers.JsonRpcProvider, sponsorAddress: string) {
```

andreogle (changes requested):
🔥

packages/operation/src/evm/deploy/account-assignment.ts
acenolaza: Excelent catch! and I completely agree we should have explicit types.

acenolaza: Allthough in this particular explicit types might not have caught it because we did have a requestersById object

packages/operation/src/evm/utils.ts
acenolaza: My idea here was to call this function without a path when deriving the airnode wallet (will use the default path m/44/60/0/0/0) and call it with a path when deriving sponsor wallets.

Maybe this function should be called deriveWallet instead...

packages/operation/README.md
acenolaza: I'm still amazed my engrish caught this

packages/operation/README.md
acenolaza: I'll put a TODO then since docs still need to be updated

packages/operation/src/evm/utils.ts
andreogle: Either way works for me

andreogle (approved):

andreogle (approved):

Issue #462, opened 2021-08-04
Improve deployer package and its README


Siegrift:
I've verified that the new version of the deployer CLI works , but I faced some problems and the whole process was harder than it should.

Things to improve:
1. Deployer gave me error, but finished with 0 status code
2. The step `Make sure your AWS credentials are stored in the configuration file or exported as environment variables` wasn't easy for me because AWS documentation is kinda complex. We should link how to properly create the credentials and how to set them in config files or environment variables (Airnode starter has this explained well, but this README doesn not).
3. From the deployer error I had no idea what is wrong, but @amarthadan helped me to figure out that my AWS user didn't have any permissions. We should warn the user about this. (Will also be probably fixed by point 2)
4. The config.json.example uses localhost provider. It's enough for testing the deployer, but if the user wants to try the full flow, he has to modify it. I suggest having an example config which the user can **easily** modify to have a POC airnode which he can call. (This might be a stretch, because we also need protocol contracts to be deployed the target chain).

bbenligiray:
@Siegrift Can you check if the issues are resolved? For each one that is not, please create a separate issue.

Siegrift:
All points except 4) are resolved.

We discussed that airnode-starter should be a place that should work and have valid configuration (https://api3workspace.slack.com/archives/C01N6001YCE/p1628256549025600).

Let's close this. We can create docs/examples issues after we know what exactly we wanna do.

PR #467, opened 2021-08-06
Update deployer README - add AWS IAM setup video link

amarthadan:
Part of #462

I know this isn't much of a change but I looked through the Airnode starter docs and it feels like pretty much the same thing to me.

Also, I don't think it's our place to suggest some IAM setup for the users. Even the example from the video is just one way to do it. I mean, yes, we're basically providing a service they can run on AWS but it's still their infrastructure. I'd say that if you want to run something on AWS you should know how users and permissions work anyway.

But I'm open to suggestions.

### DIFF
--- a/packages/deployer/README.md
+++ b/packages/deployer/README.md
@@ -6,7 +6,7 @@
 
 ## Prerequisites
 * Install [Terraform v0.15.x](https://www.terraform.io/downloads.html) and make sure that the `terraform` binary is available in your `PATH`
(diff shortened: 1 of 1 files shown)


andreogle (approved):

Siegrift (approved):
> Also, I don't think it's our place to suggest some IAM setup for the users. Even the example from the video is just one way to do it. I mean, yes, we're basically providing a service they can run on AWS but it's still their infrastructure. I'd say that if you want to run something on AWS you should know how users and permissions work anyway.

I disagree. It wasted me 30 mins and I was hopeless just because I got this wrong. If it happened to me it can happen to others as well. If we are able to prevent these "silly" mistakes and save time and "motivation" of our users I see no reason why not include it.

PR #477, opened 2021-08-08
Add eslint-plugin-jest

andreogle:
### Changes

1. Adds [eslint-plugin-jest](https://www.npmjs.com/package/eslint-plugin-jest) to our base ESLint config and enables the recommended rules.
2. Fix the tests that don't pass the linting checks
3. Upgrade some ESLint related packages


packages/admin/test/e2e/cli.feature.ts
```diff
@@ -55,8 +55,6 @@ describe('CLI', () => {
   beforeAll(() => {
     jest.setTimeout(45_000);
 
-    expect(existsSync(`${CLI_EXECUTABLE}`)).toBe(true);
```
andreogle, line 58: Seems like `expect()` outside of an `it` or `test` block don't get executed: https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-standalone-expect.md#rule-details

packages/node/src/providers/actions.test.ts
```diff
@@ -164,15 +164,11 @@ describe('initialize', () => {
...
-    expect.assertions(1);
     const config = fixtures.buildConfig({ chains: [] });
     const workerOpts = fixtures.buildWorkerOptions();
-    // eslint-disable-next-line functional/no-try-statement
-    try {
-      await providers.initialize('abcdefg', config, workerOpts);
-    } catch (e) {
-      expect(e).toEqual(new Error('One or more chains must be defined in the provided config'));
-    }
+    await expect(providers.initialize('abcdefg', config, workerOpts)).rejects.toThrow(
```
andreogle, line 169: This syntax is a significant improvement imo. I wasn't aware of it before

Siegrift (approved):
Nice!

packages/node/src/providers/actions.test.ts
acenolaza: Yes, I like that pattern too. Also if you only care about the error message then you can only pass the error string to the `.toThrow()` function.

packages/admin/test/e2e/cli.feature.ts
acenolaza: Then if we want to keep this check we would need to replace it with an if statement followed by a throw, correct?

acenolaza: I just saw you put it in a separate `it` block which makes even more sense 🔥

acenolaza (approved):
LGTM :shipit:

PR #478, opened 2021-08-08
Use yargs in validator

### LINKED ISSUE #438: Use yargs for validator CLI
Switch from using npm to yarn and parse arguments with yargs

drgy:
Switch validator to use yarn instead of npm, version argument was removed from convertor, it will be added in a new format after template versions are sorted out.

closes #438


amarthadan:
I changed the base to `validator-autodocs` for easier review.

Siegrift (approved):
LGTM!

packages/validator/src/commands/convertCmd.ts
```diff
@@ -9,27 +10,52 @@ import { invalidConversionMessage } from '../utils/messages';
...
+    demandOption: 'to',
+    default: '',
+    string: true,
+  })
+  .option('to', {
+    description: 'Name of the target airnode specification format',
+    demandOption: 'from',
+    default: '',
+    string: true,
+  })
```
amarthadan, line 25: I think you can use [`choices`](https://github.com/yargs/yargs/blob/master/docs/api.md#choiceskey-choices) for these two options, can't you?

packages/validator/src/commands/convertCmd.ts
```diff
@@ -9,27 +10,52 @@ import { invalidConversionMessage } from '../utils/messages';
...
+  .option('template', {
+    description: 'Path to validator template file',
+    alias: 't',
+    default: '',
+    string: true,
+  })
```
amarthadan, line 31: If I understand it correctly you can choose to either provide a `template` or `from` and `to`, right? In that case you can use [`conflicts`](https://github.com/yargs/yargs/blob/master/docs/api.md#conflictsx-y) to enforce that.

packages/validator/src/commands/validateCmd.ts
```diff
@@ -1,23 +1,38 @@
...
+    default: '',
+    alias: 't',
+    string: true,
+  })
+  .option('specification', {
+    description: 'Path to specification file that will be validated',
+    default: '',
+    alias: ['specs', 's'],
+    string: true,
+  }).argv;
```
amarthadan, line 22: It's a bit weired that you provide a default as an empty string and than you're checking whether it is provided. You can use [`demandOption`](https://github.com/yargs/yargs/blob/master/docs/api.md#demandoptionkey-msg) flag to require some option and let yargs handle it.

packages/validator/src/commands/convertCmd.ts
amarthadan, line 25: You can use [`implies`](https://github.com/yargs/yargs/blob/master/docs/api.md#impliesx-y) to require an option A if an option B is provided. I think that would work well for `from` and `to`.

packages/validator/src/commands/convertCmd.ts
```diff
@@ -1,5 +1,6 @@
 import fs from 'fs';
```
amarthadan, line 1: Wouldn't it be better to use [`command`](https://github.com/yargs/yargs/blob/master/docs/api.md#commandmodule) and have the `convert` and `validate` CLIs as one CLI with two different commands (the same way as deployer's `deploy` and `remove` are modeled)?

packages/validator/src/commands/convertCmd.ts
```diff
@@ -9,27 +10,52 @@ import { invalidConversionMessage } from '../utils/messages';
...
   let res: Result = { valid: false, messages: [], output: {} };
+  args.from = args.from.toLowerCase();
+  args.to = args.to.toLowerCase();
 
-  if (!from || !to) {
+  if (!args['from'] || !args['to']) {
     res.messages.push(logger.error('Conversion source and target specification must be provided'));
-  } else if (from === 'oas' && to === 'ois') {
-    res = convert(specs, utils.getPath(oas2ois, messages, version));
+  } else if (args.from === 'oas' && args.to === 'ois') {
```
amarthadan, line 52: Why do you sometime access the field with `[]` and sometime with `.`? I don't see a reason to do that here.

packages/validator/src/commands/convertCmd.ts
```diff
@@ -41,9 +67,9 @@ if (process.env.npm_config_template) {
       res = tmp;
     }
   } else {
-    messages.push(invalidConversionMessage(from, to));
+    messages.push(invalidConversionMessage(args['from'] as unknown as string, args['to'] as unknown as string));
```
amarthadan, line 70: Why do you need to force those types? Both `args.from` and `args.to` are of type `string`.

amarthadan (changes requested):

packages/validator/src/commands/convertCmd.ts
```diff
@@ -9,27 +10,52 @@ import { invalidConversionMessage } from '../utils/messages';
...
+    default: '',
+    string: true,
+  })
+  .option('specification', {
+    description: 'Path to specification file that will be validated',
+    default: '',
+    alias: ['specs', 's'],
+    string: true,
+  })
+  .option('specs_only', {
```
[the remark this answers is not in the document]
drgy: Renamed in 77f3ae8d04c04d68691f3e0b7acbcaab70df1895

package.json
```diff
@@ -39,7 +39,7 @@
     "lint:eslint": "eslint . --ext .js,.ts",
     "lint:eslint:fix": "eslint . --ext .js,.ts --fix",
     "lint:solhint": "solhint ./**/*.sol",
-    "postinstall": "husky install",
+    "postinstall": "husky install && yarn run build",
```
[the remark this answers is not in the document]
drgy: Removed in 8d5fab39a91fd9e6533bdf0be1b6515817454331

packages/validator/package.json
```diff
@@ -23,6 +23,7 @@
   },
   "devDependencies": {
     "markdown-snippet-injector": "^0.2.4",
-    "ts-node": "^10.1.0"
+    "ts-node": "^10.1.0",
+    "yargs": "^17.1.0"
```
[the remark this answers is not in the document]
drgy: Moved in 7a933415b0e40a4d4f06a2f67a2c377ff20915e6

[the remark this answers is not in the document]
drgy: Added in 248d72dc17604f83fb62a6ae347139fca06e3480

packages/validator/src/commands/convertCmd.ts
drgy: No, because in that case `''` must be one of the options and yargs would display this option in hints even though `''` is not a valid conversion format.

packages/validator/src/commands/convertCmd.ts
drgy: Since `from` and `to` have default values set, using `conflicts` would cause to print error every time the command is ran with `template` argument

packages/validator/src/commands/convertCmd.ts
drgy: Updated in da523b323077b3a7c9a8b396bfd8640243d2c9d1

packages/validator/src/commands/convertCmd.ts
drgy: I had some issues with types, which I resolved, but forgot to update this line, updated in da523b323077b3a7c9a8b396bfd8640243d2c9d1

drgy:
@amarthadan I tried using `conflicts`, `choices`, `implies` etc. however it just didn't seem like it works well with typescript. If the arguments don't have set default value they might be possibly `undefined`, so checks for this must be added. Also if the default value is not set in `convert` command, arguments `from`, `to` and `template` must be all provided every time the command is ran, even when using `conflicts` or `implies`. Setting the default values also renders `conflicts`, `choices` and `implies` unusable as the arguments have always a value. That's why I left the old checks unchanged, but maybe you would be able to find a way out of this

packages/validator/src/commands/convertCmd.ts
drgy: That makes sense in deployer, however for users just running validations and conversions im trying to differentiate those 2 programs as much as possible so there is no confusion in what to expect from the command. It simply makes more sense to validate specifications with validator and convert them in convertor instead of making both validations and conversions in validator...

drgy:
@amarthadan looks good, I just didn't want to change much with the available conversions, as it will be reworked in the next PR, hopefully it will be possible to do with this structure.

amarthadan:
Stupid GitHub. I swear I wrote one more comment even with a screenshot when I was doing review but for some reason it's not here anymore :smile:

Currently, when you call `validate` and `convert` without any arguments the results have different error format and `validate` even exits with status code 0 (this was the case for `validate` as well before the latest changes).

What I would propose is to handle any CLI related errors (required options, file doesn't exist, etc.) before the actual validation/conversion and display it as a plain error message with non-zero status code. Leaving the object structure only as output of actual validation/conversion.

I'm ok with 0 status code for validation/conversion that contains error because the program itself was run correctly but for exmaple the JSON is not valid.

What do you think?

packages/validator/src/commands/validateCmd.ts
amarthadan: @drgy This is still not resolved.

drgy:
@amarthadan handling missing arguments with yargs is an elegant solution, however everything else shouldn't be checked with yargs. The reason is that CLI uses the same interface as packaged validator, which must include these checks. So I think it would be quite inefficient to first check everything with yargs, then pass it and check it again.

packages/validator/src/commands/validateCmd.ts
```diff
@@ -1,23 +1,39 @@
...
 if (template) {
   if (templates[template.toLowerCase() as keyof typeof templates]) {
     template = utils.getPath(templates[template.toLowerCase() as keyof typeof templates], messages, version);
   } else if (version) {
     messages.push(logger.warn('Version argument will be ignored when validating provided template file'));
   }
 
-  if (specs) {
-    const res = validate(specs, template);
+  if (args['specification']) {
```
amarthadan, line 35: Now that both `template` and `specification` are required you don't really need these checks.

amarthadan, line 35: You're again accessing `args` with both `.` and `[]` operator for no apparent reason.

amarthadan (changes requested):

amarthadan:
> @amarthadan handling missing arguments with yargs is an elegant solution, however everything else shouldn't be checked with yargs. The reason is that CLI uses the same interface as packaged validator, which must include these checks. So I think it would be quite inefficient to first check everything with yargs, then pass it and check it again.

I'm not saying it has to be checked with `yargs` nor am I saying that it should be checked twice. I'm just saying that we should differentiate between validation/conversion errors (returning specific JSON structure and 0 status code) and "runtime" errors (with non-zero status code and normal error message).

packages/validator/src/commands/validateCmd.ts
drgy: Removed in d2285cd9502fae3de31fcd2191e4c0239a98fa22

drgy: Yeah I missed that one somehow, fixed in ffaa558e18addc5796605e7ec9706ebaeb278beb

drgy:
@amarthadan I tried to return 0 code and standard validator message with explanation what failed. The reason behind this is, afaik validator will be mostly ran programmatically, for example a user edits config on the website and validates it, for the UI it might be easier to implement error messages display if all the errors are in singular format, rather than catching various errors. I think this is more for a longer discussion and definitely different PR though.

amarthadan (approved):

PR #498, opened 2021-08-11
Remove withdrawal destination

acenolaza:
AirnodeRrp caller will no longer be able to set a destination and funds from sponsored wallet will be sent to the sponsor (account that created the withdrawal request prior to airnode fulfilling it with sponsored wallet)

### DIFF
--- a/packages/protocol/README.md
+++ b/packages/protocol/README.md
@@ -211,5 +211,5 @@ The zero here is allocated for RRP, and the other branches will be used to deriv
 ## Withdrawal from the sponsor wallet
 
 Requesters may not want to use up all the ETH deposited in their sponsor wallet.
-Then, they can use `WithdrawalUtils` to request a withdrawal from the Airnode, which sends the entire balance of the sponsor wallet to the specified address.
+Then, they can use `WithdrawalUtils` to request a withdrawal from the Airnode, which sends the entire balance of the sponsor wallet to the sponsor address.
 Unlike template and full requests, there are no parameters that the Ethereum provider can tamper with, which is why withdrawal requests/fulfillments do not make similar hash checks.
--- a/packages/protocol/contracts/rrp/WithdrawalUtils.sol
+++ b/packages/protocol/contracts/rrp/WithdrawalUtils.sol
@@ -15,18 +15,16 @@ contract WithdrawalUtils is IWithdrawalUtils {
     mapping(bytes32 => bytes32) private withdrawalRequestIdToParameters;
 
     /// @notice Called by a sponsor to create a request for the Airnode to send
-    /// the funds kept in the respective sponsor wallet to the destination
+    /// the funds kept in the respective sponsor wallet to the sponsor
     /// @dev We do not need to use the withdrawal request parameters in the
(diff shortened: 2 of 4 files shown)


bbenligiray (approved):
The Solidity part looks good, leaving the tests for @Ashar2shahid to review.

We should also merge this to `beta-protocol` after the first audit report

Ashar2shahid (approved):
Had to go through the slack discussion to understand the solidity part in depth. 

> 1. A random msg.sender calls requestWithdrawal() with a sponsorWallet they don't own. The node derives the sponsor wallet of msg.sender, sees that it's not equal to sponsorWallet, ignores the withdrawal request.
> 2. A random msg.sender calls requestWithdrawal() with a sponsorWallet they don't own. The blockchain provider is malicious so spoofs the logs as if msg.sender is the actual owner of sponsorWallet. The node will attempt to fulfill the withdrawal request, but the malicious blockchain provider can only spoof the msg.sender in logs and not here (because it's on-chain data), and the withdrawal fulfillment will revert because the sponsor in this check will not match.

According to the above scenarios the Solidity and tests looks good, so i'll be approving these changes. I still have a question regarding why we are even having the airnode involved in the withdrawl process? wouldn't it easier and safer to do everything on-chain?

acenolaza:
> Had to go through the slack discussion to understand the solidity part in depth.
> 
> > 1. A random msg.sender calls requestWithdrawal() with a sponsorWallet they don't own. The node derives the sponsor wallet of msg.sender, sees that it's not equal to sponsorWallet, ignores the withdrawal request.
> > 2. A random msg.sender calls requestWithdrawal() with a sponsorWallet they don't own. The blockchain provider is malicious so spoofs the logs as if msg.sender is the actual owner of sponsorWallet. The node will attempt to fulfill the withdrawal request, but the malicious blockchain provider can only spoof the msg.sender in logs and not here (because it's on-chain data), and the withdrawal fulfillment will revert because the sponsor in this check will not match.
> 
> According to the above scenarios the Solidity and tests looks good, so i'll be approving these changes. I still have a question regarding why we are even having the airnode involved in the withdrawl process? wouldn't it easier and safer to do everything on-chain?

I'm guessing we don't even want to initiate the tx if we can know beforehand that it will be invalid. Also deriving the sponsor wallet  might consume too much gas 🤔

bbenligiray:
We need the Airnode to verify the sponsor wallet address off-chain because we can't derive it on-chain (it's computationally intensive to the degree that it's practically impossible, or at least that's what I understand). The same thing happens while making a request, the requester passes both the sponsor address and the sponsor wallet address because
1. We need to store the sponsor wallet address on-chain to prevent other addresses from fulfilling the respective request
2. The sponsor wallet address provided by the requester may not be right so we need the Airnode to verify it by deriving it again using the sponsor address
So you could have an Airnode fulfill a request with a wallet that is not actually a sponsor wallet of theirs if both the requester and the Airnode cooperated (so the requester uses an invalid sponsor wallet address and the Airnode ignores the fact that it's invalid). This is considered to be acceptable.

PR #504, opened 2021-08-13
Api3TokenLock Integration and Tests according to specifications

Ashar2shahid:
I have updated the tests to cover all blocking scenarios (thats what we are going for I presume instead of blacklisting).

Two things we should probably address: 
1. Contract Names
2. Directory Structure.


packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,305 @@
...
+        assert(IApi3Token(api3Token).transfer(msg.sender, amount));
+
+        emit Unlock(_airnode, _requesterAddress, msg.sender, amount);
+    }
+
+    /// @notice User calls this when the lock amount has been decreased and wants
+    /// to withdraw the redundantly locked tokens
+    /// @param _airnode The airnode address
+    /// @param _requesterAddress The requester address
+    function withdrawExcess(address _airnode, address _requesterAddress)
```
acenolaza, line 253: Remainder maybe? Again, I'm not good at naming things but excess sounds a litte bit weird to me 🙃

packages/protocol/contracts/rrp/authorizers/interfaces/IApi3TokenLock.sol
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+interface IApi3TokenLock {
+    event SetMinimumLockingTime(uint256 minimumLockingTime);
+
+    event SetLockAmount(uint256 lockAmount);
+
+    event BlockRequester(
```
acenolaza, line 9: Maybe events should be in past tense for consistency with other contracts: 

`event BlockedRequester(`

Same with Lock, Unlock, etc

acenolaza (changes requested):
I didn't go thru the tests just yet because I put some comments and asked some questions that I'd like to get answers first.

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,305 @@
...
+        minimumLockingTime = _minimumLockingTime;
+        lockAmount = _lockAmount;
+        api3RequesterRrpAuthorizer = _api3RequesterRrpAuthorizer;
+        api3Token = _api3Token;
+    }
+
+    /// @notice Called to get the rank of an admin for an adminned entity
+    /// @dev Overriden to use the bytes32(0) as the only adminnedId
+    ///      Overriden to use metaAdminned ranks
+    ///      Overriden to give Max Rank to the airnodeAdmin
```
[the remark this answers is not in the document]
acenolaza: One more thing, MAX_RANK is reserved for metaAdmin, shouldn't we use MAX_RANK - 1 or just return an admin uint256?

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,305 @@
...
+    /// @dev Address of Api3Token
+    address public api3Token;
+    /// @dev The Minimum locking time (in seconds)
+    uint256 public minimumLockingTime;
+    /// @dev Lock amount for each user
+    uint256 public lockAmount;
+
+    /// @dev Represents the locked amounts, periods for the sponsors
+    /// and the total whitelist counts of a airnode-requester pair
+    struct AirnodeRequester {
```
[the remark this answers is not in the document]
Ashar2shahid: this struct holds the tokens and lock amounts of the sponsors for an airnodeRequester pair, so maybe something like airnodeRequesterSponsor?

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,305 @@
...
+    string private constant ERROR_NOT_LOCKED = "No amount locked";
+    string private constant ERROR_LOCK_PERIOD_NOT_EXPIRED =
+        "Locking period not expired";
+    string private constant ERROR_LOCK_PERIOD_ZERO = "Zero lock period";
+    string private constant ERROR_REQUESTER_BLOCKED = "Requester Blocked";
+
+    /// @dev Address of api3RequesterRrpAuthorizer contract
+    address public api3RequesterRrpAuthorizer;
+    /// @dev Address of Api3Token
+    address public api3Token;
```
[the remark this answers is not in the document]
Ashar2shahid: fixing

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,305 @@
...
+    /// @param admin Admin address whose rank will be returned
+    /// @return Admin rank
+    function getRank(bytes32 adminnedId, address admin)
+        public
+        view
+        override(MetaAdminnable)
+        returns (uint256)
+    {
+        if (adminnedId == bytes32(abi.encode(admin))) return MAX_RANK;
+        return MetaAdminnable.getRank(bytes32(0), admin);
```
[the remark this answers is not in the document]
Ashar2shahid: comments are wrong, we are using the adminnedId to check if the caller is the airnodeAdmin, Ill have then written again.

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,305 @@
...
+        emit BlockRequester(_airnode, _requesterAddress, msg.sender);
+    }
+
+    /// @notice Locks API3 Tokens to gain access to Airnodes.
+    /// Airnode-requester pair gets authorized as long as there is at least one token lock for given pair.
+    /// @dev The amount to be locked is determined by a memory variable set by
+    /// the owners of the contract.
+    /// @param _airnode The airnode address
+    /// @param _requesterAddress The requester address
+    function lock(address _airnode, address _requesterAddress)
```
acenolaza, line 162: Question: What would happen if _airnode is address(0)? we'll be whitelisting the requester globally? Do we want that?

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,305 @@
...
+        mapping(address => uint256) sponsorLockAmount;
+        mapping(address => uint256) sponsorUnlockTime;
+        uint256 whitelistCount;
+    }
+
+    /// @dev Stores information about all token locks for airnode-requester pair
+    mapping(address => mapping(address => AirnodeRequester))
+        public airnodeToRequesterToTokenLocks;
+
+    /// @dev Stores information for blocked airndoe-requester pair
```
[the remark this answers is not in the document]
Ashar2shahid: oops, yeah agreed.

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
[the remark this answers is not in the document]
Ashar2shahid: yes the airnode master wallet is what I'm referring to, ill change the naming accordingly. Yes the airnode admin can block malicious users for their own airnode.

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,305 @@
...
+        require(_lockAmount != 0, ERROR_ZERO_AMOUNT);
+        lockAmount = _lockAmount;
+        emit SetLockAmount(_lockAmount);
+    }
+
+    /// @notice Called by a SuperAdmin or the airnodeAdmin to set the Block status of a airnode-requester pair
+    ///         also revokes access by setting whitelistExpiration to 0.
+    ///         Whitelist access is only revoked when the specified airnode address is specified
+    ///         otherwise specifying adress(0) would only result in the requester being unable
+    ///         to lock/unlock across all airnodes. This is an irreversible action.
```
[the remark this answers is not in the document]
Ashar2shahid: yes, its the same for a single airnode.

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
Ashar2shahid: > As adjectives the difference between remainder and excess is that remainder is remaining while excess is more than is normal, necessary or specified.

I think I'll keep excess.

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
Ashar2shahid: I've decided to add a zero address check for lock,unlock and withdrawExcess. On second thought they shouldnt be allowed to access the 0th address

packages/protocol/contracts/rrp/authorizers/interfaces/IApi3TokenLock.sol
Ashar2shahid: Done

packages/protocol/test/authorizers/Api3TokenLock.sol.js
```diff
@@ -0,0 +1,755 @@
...
+
+            expect(
+              await api3TokenLock.airnodeToRequesterToBlockStatus(airnodeAddress, roles.requester.address)
+            ).to.equal(true);
+
+            await expect(
+              api3TokenLock.connect(roles.sponsor).lock(airnodeAddress, roles.requester.address)
+            ).to.revertedWith(Errors.RequesterBlocked);
+
+            await api3TokenLock.connect(roles.sponsor).lock(roles.unknown.address, roles.requester.address);
```
acenolaza, line 297: A comment here explaning what should happen after calling the lock() function would be helpful.

packages/protocol/test/authorizers/Api3TokenLock.sol.js
```diff
@@ -0,0 +1,755 @@
...
+              api3TokenLock.connect(roles.superAdmin).blockRequester(airnodeAddress, roles.requester.address)
+            )
+              .to.emit(api3TokenLock, 'BlockedRequester')
+              .withArgs(airnodeAddress, roles.requester.address, roles.superAdmin.address);
+          });
+        });
+        context('requester is being blocked globally', async () => {
+          it('sets the block', async () => {
+            await api3TokenLock
+              .connect(roles.metaAdmin)
```
acenolaza, line 397: Shouldn't this be the superAdmin instead?

packages/protocol/test/authorizers/Api3TokenLock.sol.js
```diff
@@ -0,0 +1,755 @@
...
+              api3TokenLock.connect(roles.sponsor).lock(airnodeAddress, roles.requester.address)
+            ).to.revertedWith(Errors.RequesterBlocked);
+
+            await expect(
+              api3TokenLock.connect(roles.sponsor).lock(roles.unknown.address, roles.requester.address)
+            ).to.revertedWith(Errors.RequesterBlocked);
+          });
+
+          it('emits event', async () => {
+            await expect(api3TokenLock.connect(roles.metaAdmin).blockRequester(airnodeAddress, roles.requester.address))
```
acenolaza, line 417: Same here

packages/protocol/test/authorizers/Api3TokenLock.sol.js
```diff
@@ -0,0 +1,755 @@
...
+
+        it('emits event', async () => {
+          await expect(api3TokenLock.connect(roles.sponsor).lock(airnodeAddress, roles.requester.address))
+            .to.emit(api3TokenLock, 'Locked')
+            .withArgs(airnodeAddress, roles.requester.address, roles.sponsor.address, LOCK_AMOUNT);
+        });
+      });
+
+      context('caller has already locked', async () => {
+        it('is set for different sponser', async () => {
```
acenolaza, line 526: typo: sponser -> sponsor

packages/protocol/test/authorizers/Api3TokenLock.sol.js
```diff
@@ -0,0 +1,755 @@
...
+      });
+
+      context('already unlocked', async () => {
+        it('reverts', async () => {
+          utils.timeTravel(ONE_MINUTE_IN_SECONDS + 1000);
+          await api3TokenLock.connect(roles.sponsor).unlock(airnodeAddress, roles.requester.address);
+
+          await expect(
+            api3TokenLock.connect(roles.sponsor).unlock(airnodeAddress, roles.requester.address)
+          ).to.revertedWith('No amount locked');
```
acenolaza, line 654: Errors.ERROR_NOT_LOCKED?

packages/protocol/test/authorizers/Api3TokenLock.sol.js
```diff
@@ -0,0 +1,755 @@
...
+      });
+
+      context('requester is unlocking on address zero', async () => {
+        it('reverts', async () => {
+          await expect(
+            api3TokenLock.connect(roles.sponsor).unlock(hre.ethers.constants.AddressZero, roles.requester.address)
+          ).to.revertedWith(Errors.ZeroAddress);
+        });
+      });
+    });
```
acenolaza, line 676: Are we missing ERROR_LOCK_PERIOD_ZERO test?

packages/protocol/test/authorizers/Api3TokenLock.sol.js
```diff
@@ -0,0 +1,755 @@
...
+      context('caller is unknown', async () => {
+        it('reverts', async () => {
+          await expect(
+            api3TokenLock.connect(roles.unknown).blockRequester(airnodeAddress, roles.requester.address)
+          ).to.revertedWith(Errors.LowRank);
+        });
+      });
+    });
+
+    describe('lock', async () => {
```
acenolaza, line 489: Should we also add at least one happy path were we sponsor a single requester with multiple sponsors?

acenolaza (changes requested):
Tests are looking good 🥇 
Added a few minor comments and I think I've an issue where metaAdmin was used when it should have been superAdmin.
Also I think we should add tests to getRank() since we've changed the behavior by overriding.

packages/protocol/test/authorizers/Api3TokenLock.sol.js
```diff
@@ -0,0 +1,755 @@
...
+      admin: accounts[3],
+      superAdmin: accounts[4],
+      sponsor: accounts[5],
+      anotherSponsor: accounts[6],
+      requester: accounts[7],
+      anotherRequester: accounts[8],
+      unknown: accounts[9],
+    };
+
+    const generatedWallets = utils.generateRandomAirnodeWallet();
```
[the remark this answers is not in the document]
Ashar2shahid: yup your right, I refrenced another test and turned it into a 1-liner.

packages/protocol/test/authorizers/Api3TokenLock.sol.js
```diff
@@ -0,0 +1,755 @@
...
+          await expect(api3TokenLock.connect(roles.metaAdmin).setLockAmount(0)).to.revertedWith(Errors.ZeroAmount);
+        });
+      });
+    });
+
+    describe('blockRequester', async () => {
+      context('caller is metaAdmin', async () => {
+        context('requester is blocking on an airnode', async () => {
+          it('sets the block', async () => {
+            await api3TokenLock.connect(roles.metaAdmin).blockRequester(airnodeAddress, roles.requester.address);
```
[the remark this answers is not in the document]
Ashar2shahid: yes, I added checks in all block test cases accordingly.

packages/protocol/test/authorizers/Api3TokenLock.sol.js
Ashar2shahid: agreed, added the comments for better understanding

packages/protocol/test/authorizers/Api3TokenLock.sol.js
Ashar2shahid: oops, yes my bad. Fixed in the new commit

packages/protocol/test/authorizers/Api3TokenLock.sol.js
Ashar2shahid: resolved

packages/protocol/test/authorizers/Api3TokenLock.sol.js
Ashar2shahid: isn't [this](https://github.com/api3dao/airnode/blob/3137035fbabbb8967323cde9daec5f78e7b6b449/packages/protocol/test/authorizers/Api3TokenLock.sol.js#L612) what you're referring to?

packages/protocol/test/authorizers/Api3TokenLock.sol.js
Ashar2shahid: After going through the code, it isn't possible to ever reach the ERROR_LOCK_PERIOD_ZERO require statement. Consider the two cases:

> User locks tokens for himself which sets the lockamount and also sets the unlockTime which is > 0.
> User does not unlock tokens in which case the user has both the lockamount and unlockTime = 0.

so the revert condition is never reached because the lockamount require condition is always triggered first. One way to do this would be to AND them but I dont think its necessary and opted to remove it. What do you think?

packages/protocol/test/authorizers/Api3TokenLock.sol.js
acenolaza: You are right, my bad. I got confused because it was under 'caller has already locked' context and I was expecting that one to revert.

packages/protocol/test/authorizers/Api3TokenLock.sol.js
```diff
@@ -0,0 +1,842 @@
...
+            2
+          );
+
+          expect(
+            (await api3RequesterRrpAuthorizer.serviceIdToUserToWhitelistStatus(airnodeId, roles.requester.address))
+              .expirationTimestamp
+          ).to.equal(hre.ethers.BigNumber.from('0xffffffffffffffff'));
+        });
+
+        it('reverts for the same sponser', async () => {
```
acenolaza, line 644: typo sponser->sponsor

packages/protocol/test/authorizers/Api3TokenLockExternal.sol.js
```diff
@@ -0,0 +1,1046 @@
...
+          )
+            .to.emit(api3TokenLockExternal, 'Locked')
+            .withArgs(chainId, airnodeAddress, roles.requester.address, roles.sponsor.address, LOCK_AMOUNT)
+            .to.emit(api3TokenLockExternal, 'Authorized')
+            .withArgs(chainId, airnodeId, roles.requester.address, hre.ethers.BigNumber.from(2).pow(64).sub(1));
+        });
+      });
+
+      context('caller has already locked', async () => {
+        it('is set for different sponser', async () => {
```
acenolaza, line 703: typo sponser->sponsor

acenolaza (changes requested):

packages/protocol/test/authorizers/Api3TokenLock.sol.js
acenolaza: It might be redudant to check both values for 0 so I guess it's fine to remove that check

packages/protocol/contracts/rrp/authorizers/Api3TokenLock.sol
```diff
@@ -0,0 +1,304 @@
...
+    /// and the total whitelist counts of a airnode-requester pair
+    struct AirnodeRequester {
+        mapping(address => uint256) sponsorLockAmount;
+        mapping(address => uint256) sponsorUnlockTime;
+        uint256 whitelistCount;
+    }
+
+    /// @dev Stores information about all token locks for airnode-requester pair
+    mapping(address => mapping(address => AirnodeRequester))
+        public airnodeToRequesterToTokenLocks;
```
acenolaza, line 45: This mapping might not need to be public since we have getters for sponsorLockAmount and sponsorUnlockTime

Ashar2shahid:
Updated the contract names and interfaces and also the tests. If everything is good we can have it merged then work on iterating it later.

acenolaza (approved):
LGTM just one last thing before merging: 

shouldn't we put the `ITokenLockRrpAuthorizerAdmin` and `ITokenLockRrpAuthorizerAdminExternal` interfaces in a folder on the same level as the contracts (`rrp/authorizers/admin/interfaces`) instead of having them all mixed with authorizers interfaces?

Ashar2shahid:
> LGTM just one last thing before merging:
> 
> shouldn't we put the `ITokenLockRrpAuthorizerAdmin` and `ITokenLockRrpAuthorizerAdminExternal` interfaces in a folder on the same level as the contracts (`rrp/authorizers/admin/interfaces`) instead of having them all mixed with authorizers interfaces?

i actually did that initially but moved it back because I noticed the mock contracts were also this way. Ill move them to the new folder and merge it.

PR #511, opened 2021-08-18
Fixes protocol event names

acenolaza:
These changes will not guarantee that all tests and code related to protocol events will pass/work but they will get us one step closer to getting the node package to compile again and be able to run tests as well as start the app.

Changes in protocol event names:
- ClientFullRequestCreated -> MadeFullRequest
- ClientRequestCreated -> MadeTemplateRequest
- ClientRequestFailed -> FailedRequest
- ClientRequestFulfilled -> FulfilledRequest
- WithdrawalFulfilled -> FulfilledWithdrawal
- WithdrawalRequested -> RequestedWithdrawal

### DIFF
--- a/packages/node/src/evm/contracts/airnodeRrp.test.ts
+++ b/packages/node/src/evm/contracts/airnodeRrp.test.ts
@@ -42,15 +42,15 @@ describe('AirnodeRrp', () => {
     expect(events).toEqual([
       'AirnodeParametersSet',
       'ClientEndorsementStatusSet',
-      'ClientFullRequestCreated',
-      'ClientRequestCreated',
-      'ClientRequestFailed',
-      'ClientRequestFulfilled',
+      'MadeFullRequest',
+      'MadeTemplateRequest',
+      'FailedRequest',
+      'FulfilledRequest',
       'RequesterCreated',
       'RequesterUpdated',
       'TemplateCreated',
-      'WithdrawalFulfilled',
-      'WithdrawalRequested',
+      'FulfilledWithdrawal',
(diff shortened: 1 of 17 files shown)


packages/node/src/evm/contracts/airnodeRrp.test.ts
```diff
@@ -59,34 +59,34 @@ describe('AirnodeRrp', () => {
...
+      'FailedRequest',
+      'FulfilledRequest',
+      'FulfilledWithdrawal',
+      'RequestedWithdrawal',
     ]);
 
     // API calls
-    expect(airnodeRrpTopics.ClientRequestCreated).toEqual(
+    expect(airnodeRrpTopics.MadeTemplateRequest).toEqual(
       '0x8339fddbb81e588a9ed04dec82ee9ae6c7a185f44835adaaa2ace50ce3a14aaf'
```
andreogle, line 72: Haven't these values have changed too?

andreogle (approved):
Looks good, but I just have the one question about test values. I assume there are actually quite a few more tests that are broken? Will these be fixed later?

packages/node/src/evm/contracts/airnodeRrp.test.ts
acenolaza: Yes, probably but I was hoping to get the code in a state where I could run the tests and see what the new value would be. Do you think I should try to get this hash computed now and replace it?

andreogle: No need to if it's difficult to get. Can always fix this later

PR #512, opened 2021-08-18
Read mnemonic from configuration file (interpolated)

amarthadan:
My plan was to introduce reading of all values from `config.json` instead of `secrets.env` in one PR but moving just the mnemonic was much more complicated than I expected. I think (hope) that the rest of the values will go more smoothly.

Tricky part was that the `getMasterHDNode` method needed a mnemonic but didn't always have access to `config` structure (mainly the `refresh` method). I solved it so the config field is scrubed between different worker calls but is returned once we have the response (state from the worker).


packages/deployer/src/utils/files.ts
```diff
@@ -13,6 +18,28 @@ export function parseSecretsFile(secretsPath: string) {
...
+  mnemonic: string,
+  config: Config,
+  commandOutput: TerraformAirnodeOutput
+) {
+  const airnodeId = deriveAirnodeId(mnemonic);
+  const receipt = {
+    airnodeId,
+    airnodeIdShort: shortenAirnodeId(airnodeId),
+    // Keeping `chains` and `nodeSettings` fields from the configuration and removing mnemonic
+    config: omit(pick(config, ['chains', 'nodeSettings']), 'nodeSettings.airnodeWalletMnemonic'),
```
Siegrift, line 32: This should probably be done in a scalable way. We probably don't want any secret to be leaked out to `receipt.json`

Siegrift: I have some solutions in mind for this, but it deserves discussion

packages/node/src/config/index.ts
```diff
@@ -34,11 +34,11 @@ export function parseConfig(configPath: string, secrets: Record<string, string |
...
 }
 
-export function getMasterKeyMnemonic(): string {
-  const mnemonic = process.env.MASTER_KEY_MNEMONIC;
+export function getMasterKeyMnemonic(config: Config): string {
+  const mnemonic = config.nodeSettings.airnodeWalletMnemonic;
   // The node cannot function without a master mnemonic
   if (!mnemonic) {
-    throw new Error('Unable to find MASTER_KEY_MNEMONIC from the environment. Please ensure this is set first');
+    throw new Error('Unable to find Airnode wallet mnemonic in the configuration. Please ensure this is set first');
```
Siegrift, line 41: Is this possible? We are validating the config file so we expect mnemonic to be defined at this place.

packages/node/src/evm/workers.ts
```diff
@@ -4,58 +4,46 @@ import * as providerState from '../providers/state';
...
-    }
-    const log = logger.pend('ERROR', `Unable to process provider requests:${state.settings.name}`);
-    return [[log], null];
-  }
+  return spawn(state, workerOpts, 'initializeProvider', 'Unable to initialize provider');
+}
 
-  const refreshedState = providerState.refresh(res.data);
-  return [[], refreshedState];
+export async function spawnProviderRequestProcessor(
```
Siegrift, line 44: Nicely done!

packages/node/src/types.ts
```diff
@@ -260,8 +260,8 @@ export type EVMRequestCreatedLog = EVMTemplateRequestCreatedLog | EVMFullApiRequ
 
 export interface EVMRequestFulfilledLog extends EVMEventLogMetadata {
   readonly parsedLog:
-  | AirnodeLogDescription<AirnodeRrpLog<'ClientRequestFulfilled'>>
-  | AirnodeLogDescription<AirnodeRrpLog<'ClientRequestFailed'>>;
+    | AirnodeLogDescription<AirnodeRrpLog<'ClientRequestFulfilled'>>
```
Siegrift, line 263: How is this formatting change possbile? We should have prettier setup

packages/node/src/types.ts
```diff
@@ -372,6 +372,7 @@ export type NodeCloudProvider = 'local' | 'aws';
 
 export interface NodeSettings {
   readonly airnodeIdShort?: string;
+  readonly airnodeWalletMnemonic?: string;
```
Siegrift, line 375: Shouldn't these be required?

Siegrift (approved):
:1st_place_medal: Nicely done, just a couple minor comments.

packages/node/test/setup/init/set-env-vars.ts
```diff
@@ -1,4 +1,3 @@
-process.env.MASTER_KEY_MNEMONIC = 'achieve climb couple wait accident symbol spy blouse reduce foil echo label';
```
andreogle, line 1: Can the rest of the env variables here be removed too or not yet?

andreogle (approved):

packages/node/test/setup/init/set-env-vars.ts
amarthadan: Not yet, I'm still working on moving the rest of the secrets. I'll clean it in the process.

packages/deployer/src/utils/files.ts
amarthadan: > Why do we _need_ the `config` in the receipt?

We need `stage`, `cloudProvider` and `region` in order to remove the deployment from `receipt.json`. Do you think it would be better to provide just those?

Also, I've just realized that currently the provider URL is also in the `receipt.json` and probably shouldn't be. Maybe we should rethink the `receipt.json` altogether.

What else (except those fields I mentioned) should be in the `receipt.json`? Maybe instead removing what is not needed we should just pick what is.

amarthadan: > This should probably be done in a scalable way. We probably don't want any secret to be leaked out to `receipt.json`
> I have some solutions in mind for this, but it deserves discussion

I agree and I'm happy to hear your ideas but it looks like we may rework the `receipt.json` in general so I would wait till we decide what we actually want to do about it.

packages/node/src/config/index.ts
amarthadan: We're not validating the config file yet. I think we can remove it once the validator is integrated.

packages/node/src/types.ts
amarthadan: TBH I have no idea :smile: My IDE will format it wrongly for some reason (by wrongly I mean the state in which it was before) but pre-commit hook formats it correctly hence the change. I don't know how it could have been commited there without a check :man_shrugging:

packages/node/src/types.ts
amarthadan: I don't actually think the mnemonic should be required because you can run the deployer without it and it will be created for you. Yes, you have to add it there afterwards but still.

And I don't really get the `airnodeIdShort` there. As far as I can tell it's never read from the `config.json`. It's not even in the new documentation. I think that should have been removed some time ago but was forgotten.

packages/deployer/src/utils/files.ts
andreogle: I would say it's better to only include the fields we need like `stage`, `cloudProvider` and `region` rather than the entire config. It will be easy to add sensitive fields to the config and forget to remove them from the receipt.

Do we need the the provider URL in the receipt? 😄

amarthadan: I moved the discussion to [slack](https://api3workspace.slack.com/archives/C02AYRX8D89/p1629448851051700), this is something that should be definitelly resolved in a separate PR.

packages/node/src/types.ts
Siegrift: I see your point. I still think it should be strictly defined in node. I would define a subtype in deployer `type DeployerNodeSettings = Partial<NodeSettings>` and use that there instead.

Regarding `airnodeIdShort` there is a discussion, but it still appears to be used? https://api3workspace.slack.com/archives/C027Y2FAV0S/p1629381176035600

amarthadan: > I see your point. I still think it should be strictly defined in node. I would define a subtype in deployer type DeployerNodeSettings = Partial<NodeSettings> and use that there instead.

Yep, makes sense, will do that.

> Regarding airnodeIdShort there is a discussion, but it still appears to be used? https://api3workspace.slack.com/archives/C027Y2FAV0S/p1629381176035600

I'm not saying we should get rid of the concept of `airnodeIdShort`. I'm just saying that it's never read from the configuration. It's always derived from `airnodeId` which is derived from the mnemonic. And it will be the same with the `airnodeAddress` and its short version.

PR #513, opened 2021-08-18
Validator versions update

### LINKED ISSUE #437: Change version argument in validator CLI
Version argument should ber removed from `validate` and `convert` command, version should be optionally passed with specification name

`yarn run validate config@0.1.0 [specsPath]`
`yarn run convert oas@3.0.0 ois@0.1.0 [specsPath]`

drgy:
Separating `pre-alpha` and `0.1.0` versions, `pre-alpha` should correctly work with specifications that are currently in use, including validation and conversions. Version `0.1.0` is not up to date and will be updated after it is more clear how the `config.json` will look in `0.1.0`.

Updated versioning in `convert` command to match `validate` command:
```
yarn run convert --from="OIS@pre-alpha" --to="config@pre-alpha" --specs="exampleSpecs/ois.specs.json"
```

The way conversion templates were handled previously was not scalable and didnt make much sense. All conversion templates were moved to `templates/conversion/` and based on the file names it is determined what conversions are available.

Validation templates are still in folders in `templates/` grouped by version, each version has all the necessary templates, which brings up question, when nothing changes in some versions like in `ois.json` right now, should the whole template from `pre-alpha` be copy pasted into `0.1.0`, or should just the `ois.json` in `0.1.0` point to `ois.json` in `pre-alpha` via nested templates? 
closes #437


packages/validator/src/commands/utils.ts
```diff
@@ -49,3 +66,68 @@ export function getPath(template: string, messages: Log[], version = ''): string
...
+      toLatest = !toLatest || (toLatest < version && version.match(/^[0-9\.]+$/)) ? version : toLatest;
+    }
+
+    if (!toLatest) {
+      messages.push({ level: 'error', message: `Unknown conversion from ${from} to ${to}` });
+      return '';
+    }
+
+    toVersion = toLatest;
+  }
```
amarthadan, line 121: This is something I can't decide if it's a correct/desired behavior or not.

When you don't pass a version to `from` and `to` arguments it's expected to choose the latest one. But, what is the latest one? Imagine that we support these conversions:

`apple@1.0.0` -> `orange@1.0.0`
`apple@1.0.0` -> `blueberry@1.0.0`
`apple@1.0.0` -> `banana@1.0.0`
`apple@2.0.0` -> `orange@2.0.0`
`apple@2.0.0` -> `blueberry@2.0.0`

When moving to version `2.0.0` we've droped the conversion from `apple` to `banana`. So currently the latest version of the "product" is `2.0.0`. But if you call `convert --from apple --to banana` it will currently still run but using version `1.0.0` because that's the latest that can do the conversion. Even though there's already a newer version of `apple` available, it just doesn't support conversion to `banana`.

I'd say this is a bit confusing for the user. I, as a user, would expect the message saying that the latest version of `apple` can't be converted to `banana`. And if I'm ok with the older version saying so explicitly.

WDYT?

packages/validator/src/commands/convertCmd.ts
```diff
@@ -1,33 +1,18 @@
...
-
-const supportedConversions = [
-  ['oas', 'ois'],
-  ['oas', 'config'],
-  ['ois', 'config'],
-];
+import { convert } from '../convertor';
+import { Log } from '../types';
 
 const args = yargs(hideBin(process.argv))
```
amarthadan, line 7: It would be nice if we could add something like `--list-conversions` listing all the available conversions (including versions). That's just something nice to have for the future.

packages/validator/src/commands/utils.ts
```diff
@@ -49,3 +66,68 @@ export function getPath(template: string, messages: Log[], version = ''): string
...
+    for (const version in conversions[from]) {
+      if (!conversions[from][version][to]) {
+        continue;
+      }
+
+      fromLatest = !fromLatest || (fromLatest < version && version.match(/^[0-9\.]+$/)) ? version : fromLatest;
+    }
+
+    if (!fromLatest) {
+      messages.push({ level: 'error', message: `Unknown conversion from ${from} to ${to}` });
```
amarthadan, line 96: These error messages have pretty much the same text, please extract them to a common place.

packages/validator/src/commands/utils.ts
```diff
@@ -49,3 +66,68 @@ export function getPath(template: string, messages: Log[], version = ''): string
...
+  if (
+    !fs.existsSync(
+      path.resolve(__dirname, `../../templates/conversion/${from}@${fromVersion}->${to}@${toVersion}.json`)
+    )
+  ) {
+    messages.push({ level: 'error', message: `Unknown conversion from ${from}@${fromVersion} to ${to}@${toVersion}` });
+    return '';
+  }
+
+  return path.resolve(__dirname, `../../templates/conversion/${from}@${fromVersion}->${to}@${toVersion}.json`);
```
amarthadan, line 132: This path is long and complicated, it would be better to extract it into one common constant.

amarthadan (changes requested):

amarthadan:
> which brings up question, when nothing changes in some versions like in ois.json right now, should the whole template from pre-alpha be copy pasted into 0.1.0, or should just the ois.json in 0.1.0 point to ois.json in pre-alpha via nested templates?

Good question. What can we do with the nested templates? Can we nest the `pre-alpha` version and "override" some fields if necessary or would we have to copy & paste and modify the `pre-alpha` if we need to make some changes in the future?

packages/validator/src/commands/utils.ts
```diff
@@ -49,3 +66,68 @@ export function getPath(template: string, messages: Log[], version = ''): string
...
+ * @param from - name of source specification format
+ * @param to - name of target specification format
+ * @param messages - array into which warnings/errors will pushed
+ * @param fromVersion (optional) - version of source specification format, if not specified, latest version of the template is returned
+ * @param toVersion (optional) - version of target specification format, if not specified, latest version of the template is returned
+ */
+export function getConversionPath(from: string, to: string, messages: Log[], fromVersion = '', toVersion = ''): string {
+  if (!conversions[from]) {
+    messages.push({ level: 'error', message: `Unknown conversion from ${from} to ${to}` });
+    return '';
```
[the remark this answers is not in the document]
drgy: Changed in 269b0db5482f61f62fdb56b809b958fa53df90fd

packages/validator/src/commands/utils.ts
drgy: Now I understood, in f8ab06605f0b989f68ec3d055cc6bb368ba553d0 it will issue a warning if the latest version does not have the conversion.

packages/validator/src/commands/utils.ts
```diff
@@ -92,21 +93,36 @@ export function getConversionPath(
   }
 
   if (!fromVersion) {
-    let fromLatest;
+    let fromLatest, latestOverall;
+    const versionRegex = /^[0-9\.]+$/;
 
     for (const version in conversions[from]) {
-      if (!conversions[from][version][to]) {
+      latestOverall =
```
amarthadan, line 100: You can compute `latestOverall` with the `sort` function on the `conversions[from]` keys, I'd say that would be more readable.

amarthadan (approved):
It would be nice if every variable you define with `let` had its type so you can immediately see what you're dealing with.

That's the last thing and you can merge it after.

PR #514, opened 2021-08-19
Removes setting airnode parameters on airnode initialization, removes airnodeAdmin and replaces airnodeId with airnodeAddress

acenolaza:
This  PR is for fixing compilation errors in node package and correspond to these issues:

1. [Remove on-chain record creation on Airnode initialization](https://api3dao.atlassian.net/browse/AN-84)
2. [Remove airnodeAdmin](https://api3dao.atlassian.net/browse/AN-85)
3. [Use address of  default BIP 44 wallet instead of airnodeId](https://api3dao.atlassian.net/browse/AN-86)

I left a TODO in [here](https://github.com/api3dao/airnode/blob/fb1be6ab4e1952e359ed4858de5340346f07980c/packages/node/src/evm/templates/template-fetching.ts#L55) because:
1. @bbenligiray I think we've forgot to add to the [What changed](https://docs.google.com/document/d/1dWq7e0Kle8G16wscwwwz2SgpgiuoavBvGxdpCwMMhA4/edit?pli=1#heading=h.z82ek4xxcuz3) document that `getTemplate()` was removed from the protocol and we now only have a `getTemplates()` available. This might require a ticket in Jira but not sure if another ticket for documentation will be required.
2. @andreogle since we no longer have a getter for single template maybe that code in `template-fetching `can be removed.


packages/node/src/evm/handlers/initialize-provider.ts
```diff
@@ -51,32 +50,10 @@ export async function initializeProvider(
   const state1 = state.refresh(initialState);
 
   // =================================================================
-  // STEP 2: Get current block number, and verify or set Airnode parameters
+  // STEP 2: Get current block number
```
andreogle, line 53: Nice one fixing the comment. It's easy to forget

packages/node/src/evm/initialization.ts
```diff
@@ -1,221 +0,0 @@
-import { ethers } from 'ethers';
```
andreogle, line 1: Removing this is a significant improvement to the code 🎉

packages/node/src/evm/templates/template-fetching.ts
```diff
@@ -53,6 +52,7 @@ async function fetchTemplateGroup(
   if (err || !rawTemplates) {
     const groupLog = logger.pend('ERROR', 'Failed to fetch API call templates', err);
 
+    // TODO: is this still needed now that getTemplate() was removed from protocol?
```
andreogle, line 55: I'm not quite sure why `getTemplate()` was removed from the protocol, but if it doesn't exist anymore and will throw an error if called, then we should definitely remove it

andreogle (approved):
👍🏻

packages/node/src/evm/templates/template-fetching.ts
bbenligiray: It's now called [`templates()`](https://github.com/api3dao/airnode/blob/3dfd88a627e058fb42883258ad755c1828d500d8/packages/protocol/contracts/rrp/interfaces/ITemplateUtils.sol#L27) because we're using the default getter

packages/node/src/evm/handlers/initialize-provider.ts
```diff
@@ -51,32 +50,10 @@ export async function initializeProvider(
...
-  }
-
-  // If the Airnode parameters do not yet exist onchain then we can't start processing anything.
-  // This is to be expected for new Airnode deployments and is not an error case
-  if (airnodeParametersData.xpub === '') {
-    return state1;
-  }
-
-  const state2 = state.update(state1, { currentBlock: airnodeParametersData.blockNumber });
+  const currentBlock = await state1.provider.getBlockNumber();
```
bbenligiray, line 55: Apparently this is compatible with more chains (optimistic L2 stuff) than the way we did it before so this is an unexpected improvement

packages/node/src/evm/handlers/initialize-provider.ts
acenolaza: These comments have been extremely helpful to me when trying to follow the coordinator-worker flow and understand what's actually happening. Well done!

packages/node/src/evm/initialization.ts
acenolaza: I have to admit that it was super scary to remove all of these and had to read that code quite a few times before actually decide on getting rid of it but this will make the setup process of an airnode much more simple than it is today.

packages/node/src/evm/templates/template-fetching.ts
acenolaza: Right, we can use the default getter now or we can use the [getTemplates()](https://github.com/api3dao/airnode/blob/fb1be6ab4e1952e359ed4858de5340346f07980c/packages/protocol/contracts/rrp/TemplateUtils.sol#L53) function if we need to pass a list of templateIds as argument.

The thing here is that I think the fallback mechanism now makes no sense since we would be calling the same function. Previously if getTemplates() call failed then we would call getTemplate(). Now we only have getTemplates(). So I think I'll just remove the fallback code and keep the error.

packages/node/src/evm/handlers/initialize-provider.ts
acenolaza: wohooo! good point and I hadn't realized that 🎉

packages/node/src/evm/templates/template-fetching.ts
andreogle: Yeah I think that's fine then. The `getTemplates()` function is already retried once if it fails anyway

acenolaza: I've created a separate [issue](https://api3dao.atlassian.net/browse/AN-118) for doing this refactor so I can merge this PR.

bbenligiray: I don't get the problem here. Previously we fell back from `getTemplates()` to `getTemplate()`. Now we'll fall back from `getTemplates()` to `templates()`. There was only supposed to be a naming change.

andreogle: Ok I misunderstood, sorry. Let's just switch to `templates()` instead then

PR #515, opened 2021-08-20
Read HTTP gateway & heartbeat settings from configuration file

amarthadan:
Followup for #512

I'm aware that the `receipt.json` still contains some secrets after the deployment but I'll deal with that according to [our discussion](https://api3workspace.slack.com/archives/C02AYRX8D89/p1629461128051800?thread_ts=1629448851.051700&cid=C02AYRX8D89) separately.


andreogle (approved):
🎉

PR #517, opened 2021-08-20
AN-105: Removes destination address param when fulfilling withdrawal …

acenolaza:
…requests

Fixes compilation errors due to protocol no longer taking the destination address as param for withdrawal requests. Now funds will be sent back to the sponsor. 

@andreogle I found this call to [fixtures.operation.buildWithdrawal()](https://github.com/api3dao/airnode/blob/ecd78ef7ff504a9ce6e5c52da0afa7eaf6b577f5/packages/node/test/e2e/withdrawals.feature.ts#L13) and I'm not sure why we didn't use fixtures.requests.buildWithdrawal() instead. It's also kind of weird to have this dependency on the operation package isn't it?

### DIFF
--- a/packages/node/test/e2e/withdrawals.feature.ts
+++ b/packages/node/test/e2e/withdrawals.feature.ts
@@ -10,7 +10,7 @@ it('processes withdrawals only once', async () => {
 
   const provider = e2e.buildProvider();
 
-  const requests = [fixtures.operation.buildWithdrawal({ requesterId: 'alice', destination: 'alice' })];
+  const requests = [fixtures.operation.buildWithdrawal({ requesterId: 'alice' })];
 
   const deployerIndex = e2e.getDeployerIndex(__filename);
   const deployConfig = fixtures.operation.buildDeployConfig({ deployerIndex, requests });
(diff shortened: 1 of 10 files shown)


packages/node/test/e2e/withdrawals.feature.ts
```diff
@@ -10,7 +10,7 @@ it('processes withdrawals only once', async () => {
 
   const provider = e2e.buildProvider();
 
-  const requests = [fixtures.operation.buildWithdrawal({ requesterId: 'alice', destination: 'alice' })];
+  const requests = [fixtures.operation.buildWithdrawal({ requesterId: 'alice' })];
```
andreogle, line 13: This is different to `fixtures.requests.buildWithdrawal()`. Here it builds a request in the format intended for the `operation` package here:

https://github.com/api3dao/airnode/blob/78ee8da7339fc0f4b111ce72230f523a90053abf/packages/operation/src/config/evm-dev-config.json#L75-L80

The withdrawal isn't a valid request object in the `node` package. But I think it would actually be better to expose these fixture functions (currently in `node/test/fixtures/operation/`) from the `operation` package, rather than the `node` package.

andreogle (approved):

packages/node/test/e2e/withdrawals.feature.ts
acenolaza: doh! you are right. The expected objects have different attributes.

I'll create a separate issue for moving everything under `node/test/fixtures/operation/ `to the `operation` package.

PR #518, opened 2021-08-20
AN-90: Fixes call to AirnodeRrp.checkAuthorizationStatus()

acenolaza:
AirnodeRrp.checkAuthorizationStatus() no longer needs designatedWallet being passed but it does need the authorizers address array to be passed as argument. I'm using the authorizers list defined in config.json under chain config.

### DIFF
--- a/packages/node/src/evm/authorization/authorization-fetching.test.ts
+++ b/packages/node/src/evm/authorization/authorization-fetching.test.ts
@@ -15,12 +15,13 @@ import { RequestStatus } from '../../types';
 import { AirnodeRrp } from '../contracts';
 
 describe('fetch (authorizations)', () => {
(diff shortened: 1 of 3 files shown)


packages/node/src/evm/authorization/authorization-fetching.test.ts
```diff
@@ -15,12 +15,13 @@ import { RequestStatus } from '../../types';
 import { AirnodeRrp } from '../contracts';
 
 describe('fetch (authorizations)', () => {
-  let mutableFetchOptions: any;
+  let mutableFetchOptions: authorization.FetchOptions;
```
andreogle, line 18: Nice one

andreogle (approved):

PR #523, opened 2021-08-24
AN-92: Replace client with requester

acenolaza:
last compilation error fixed! 🎉 

* renamed `ClientRequest` type to [`Request`](https://github.com/api3dao/airnode/blob/2c9a4236b3fc2d8041e05edb9fd2e0f1bfdc9911/packages/node/src/types.ts#L52) because RequesterRequest seems a little bit redundant 😆 
* renamed var `clientAddress` to [`requesterAddress`](https://github.com/api3dao/airnode/blob/2c9a4236b3fc2d8041e05edb9fd2e0f1bfdc9911/packages/node/src/types.ts#L69) and not just `requester` because I used the same pattern when renaming `airnodeIdAddress` to `airnodeAddress`.  (I mention this because in AirnodeRrp.sol `airnodeAddress` is `airnode` and `requesterAddress` is `requester`)

PR #524, opened 2021-08-24
AN-94: Replace requester with sponsor

acenolaza:
3 compilation error down, 1 more to go 💪🏻 

* Used var name [`sponsorAddress`](https://github.com/api3dao/airnode/blob/96a4464270a388e2c37d32aea43bf2c1d471e4f6/packages/node/src/types.ts#L90) instead of just sponsor because I used the same pattern when renaming airnodeIdAddress to airnodeAddress. (I mention this because in AirnodeRrp.sol airnodeAddress is airnode and sponsorAddress is sponsor)
* Now [`Withdrawal`](https://github.com/api3dao/airnode/blob/96a4464270a388e2c37d32aea43bf2c1d471e4f6/packages/node/src/types.ts#L88) interface seems to share `airnodeAddress` with [`ApiCall`](https://github.com/api3dao/airnode/blob/96a4464270a388e2c37d32aea43bf2c1d471e4f6/packages/node/src/types.ts#L66) interface and `sponsorAddress` is already present in generic type [`ClientRequest<T>`](ClientRequest) so I was wondering if `Withdrawal` interface could be removed or merged. If so then I'll work on this on a separate branch.
* Replaced derivation path calculation function that splits the sponsor address into 31bits chunks.

### DIFF
--- a/packages/node/src/coordinator/calls/aggregation.test.ts
+++ b/packages/node/src/coordinator/calls/aggregation.test.ts
@@ -24,7 +24,7 @@ describe('aggregate (API calls)', () => {
     const res = aggregation.aggregate(fixtures.buildConfig(), apiCalls);
     expect(res).toEqual({
       apiCallId: {
-        requesterIndex: '3',
+        sponsorAddress: '3', //TODO: fix value
         airnodeAddress: 'airnodeAddress',
         clientAddress: 'clientAddress',
         sponsorWallet: 'sponsorWallet',
@@ -47,7 +47,7 @@ describe('aggregate (API calls)', () => {
     const res = aggregation.aggregate(fixtures.buildConfig(), apiCalls);
     expect(res).toEqual({
       apiCallId: {
-        requesterIndex: '3',
+        sponsorAddress: '3', //TODO: fix value
         airnodeAddress: 'airnodeAddress',
         clientAddress: 'clientAddress',
         sponsorWallet: 'sponsorWallet',
--- a/packages/node/src/evm/handlers/process-transactions.test.ts
+++ b/packages/node/src/evm/handlers/process-transactions.test.ts
@@ -38,14 +38,17 @@ describe('processTransactions', () => {
       hash: '0xad33fe94de7294c6ab461325828276185dff6fed92c54b15ac039c6160d2bac3',
     });
 
-    const apiCall = fixtures.requests.buildSubmittableApiCall({ requesterIndex: '4' });
-    const withdrawal = fixtures.requests.buildWithdrawal({ requesterIndex: '5' });
+    const apiCall = fixtures.requests.buildSubmittableApiCall({ sponsorAddress: '4' }); //TODO: fix value
+    const withdrawal = fixtures.requests.buildWithdrawal({ sponsorAddress: '5' }); //TODO: fix value
     const requests: GroupedRequests = {
       apiCalls: [apiCall],
       withdrawals: [withdrawal],
     };
-    const transactionCountsByRequesterIndex = { 4: 79, 5: 212 };
-    const state = fixtures.buildEVMProviderState({ requests, transactionCountsByRequesterIndex });
+    const transactionCountsBySponsorAddress = { 4: 79, 5: 212 }; //TODO: fix value
+    const state = fixtures.buildEVMProviderState({
+      requests,
+      transactionCountsBySponsorAddress,
+    });
 
     const res = await processTransactions(state);
     expect(res.requests.apiCalls[0]).toEqual({ ...apiCall, nonce: 79 });
@@ -57,7 +60,7 @@ describe('processTransactions', () => {
     expect(fulfillWithdrawalMock).toHaveBeenCalledWith(
       withdrawal.id,
       withdrawal.airnodeAddress,
-      withdrawal.requesterIndex,
+      withdrawal.sponsorAddress,
       {
         gasPrice,
         gasLimit: ethers.BigNumber.from(70_000),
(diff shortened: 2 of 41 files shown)


packages/node/src/coordinator/calls/aggregation.test.ts
```diff
@@ -24,7 +24,7 @@ describe('aggregate (API calls)', () => {
     const res = aggregation.aggregate(fixtures.buildConfig(), apiCalls);
     expect(res).toEqual({
       apiCallId: {
-        requesterIndex: '3',
+        sponsorAddress: '3', //TODO: fix value
```
andreogle, line 27: I don't really like leaving all of these TODOs in the codebase. What's preventing us from fixing this now?

packages/node/src/evm/handlers/process-transactions.test.ts
```diff
@@ -38,14 +38,17 @@ describe('processTransactions', () => {
...
-    const withdrawal = fixtures.requests.buildWithdrawal({ requesterIndex: '5' });
+    const apiCall = fixtures.requests.buildSubmittableApiCall({ sponsorAddress: '4' }); //TODO: fix value
+    const withdrawal = fixtures.requests.buildWithdrawal({ sponsorAddress: '5' }); //TODO: fix value
     const requests: GroupedRequests = {
       apiCalls: [apiCall],
       withdrawals: [withdrawal],
     };
-    const transactionCountsByRequesterIndex = { 4: 79, 5: 212 };
-    const state = fixtures.buildEVMProviderState({ requests, transactionCountsByRequesterIndex });
+    const transactionCountsBySponsorAddress = { 4: 79, 5: 212 }; //TODO: fix value
```
andreogle, line 47: Why the TODO here?

packages/node/src/evm/handlers/process-transactions.test.ts
```diff
@@ -85,13 +88,16 @@ describe('processTransactions', () => {
...
 
-    const apiCall = fixtures.requests.buildSubmittableApiCall({ requesterIndex: '4' });
+    const apiCall = fixtures.requests.buildSubmittableApiCall({ sponsorAddress: '4' }); //TODO: fix value
     const requests: GroupedRequests = {
       apiCalls: [apiCall],
       withdrawals: [],
     };
-    const transactionCountsByRequesterIndex = { 4: 79 };
-    const state = fixtures.buildEVMProviderState({ requests, transactionCountsByRequesterIndex });
+    const transactionCountsBySponsorAddress = { 4: 79 }; //TODO: fix value
```
andreogle, line 96: and here

packages/node/src/evm/transaction-counts.test.ts
```diff
@@ -14,16 +14,16 @@ import { ethers } from 'ethers';
...
+  it('calls getTransactionCount once for each unique sponsor', async () => {
     getTransactionCountMock.mockResolvedValueOnce(5);
     const options = {
       currentBlock: 10716084,
       masterHDNode: wallet.getMasterHDNode(),
       provider: new ethers.providers.JsonRpcProvider(),
     };
     const indices = ['1', '1'];
-    const [logs, res] = await transactions.fetchByRequesterIndex(indices, options);
+    const [logs, res] = await transactions.fetchBySponsor(indices, options);
```
andreogle, line 26: The `indices` variable should be updated to be `sponsorAddresses` and ideally use actual addresses for clarity. Applies to all tests here

packages/node/src/evm/wallet.ts
```diff
@@ -1,16 +1,30 @@
...
-function getPathFromIndex(index: number | string) {
-  return `m/0/${index}`;
-}
+/**
+ * HD wallets allow us to create multiple accounts from a single mnemonic.
+ * Each sponsor creates a designated wallet for each provider to use
+ * in order for them to be able to respond to the requests their requesters make.
+ *
+ * By convention derivation paths start with a master index
+ * followed by child indexes that can be any integer up to 2^31.
```
andreogle, line 10:
```suggestion
 * followed by child indices that can be any integer up to 2^31.
```

Thanks for adding the explanation here

andreogle (changes requested):

packages/node/src/coordinator/calls/aggregation.test.ts
acenolaza: I know and I feel the same way but after merging this PR and this other one https://github.com/api3dao/airnode/pull/523 I'll be to compile and run the test. My next task is actually fixing all broken tests in the node package and I plan to remove all TODOs as my next task.

packages/node/src/evm/handlers/process-transactions.test.ts
acenolaza: I basically wasn't sure what sponsor address to use and where to get it so my main focus was to get the codebase to compile again first.

packages/node/src/evm/transaction-counts.test.ts
acenolaza: Good catch 🎉

packages/node/src/evm/handlers/process-transactions.test.ts
andreogle: Ah right. I forgot that this is by sponsor address now so the keys need to change

packages/node/src/coordinator/calls/aggregation.test.ts
andreogle: Ok as long as they won't be left in the code for long 👍🏻

acenolaza: I promise 😇

Issue #424, opened 2021-07-21
Integrate Pocket Network


crisog:
**Rationale**
API3 data feeds are secured by Airnodes. Airnodes are connected to the blockchain to read and write data into it to operate successfully. As of now, most of the Airnodes running are using Infura (a centralized infrastructure provider to connect to Ethereum).

Given how centralized systems work when Infura has its downtimes, this affects the service of Airnodes. Pocket can decentralize this aspect of the Airnode stack, leading to more reliable service and fewer network bottlenecks. [API3 and Pocket did a token swap in February](https://medium.com/api3/announcing-the-api3-partnership-and-token-swap-with-pocket-network-f45420790b6c) to kick off this partnership, and now Pocket Network wants to contribute code as well to make the integration seamless for Airnodes.

**What is Pocket Network?**
Pocket Network is a decentralized blockchain API built for Web3 apps, relaying data to and from any blockchain through a network of thousands of nodes. The Pocket Network protocol validates all relayed data and proportionally rewards the participating nodes with POKT.

**Why Pocket Network?**
Instead of deploying to a single service provider, paying absurd fees for random outages and leaked data, you could use Pocket’s trustless API protocol and route your requests to thousands of independent full nodes. Available for all major blockchain networks.

You now don't rely on Infura but on a decentralized network of people providing their services to access any major blockchain with a huge variety of locations, capacity & availability.

**Benefits:**
- Easy management & monitoring via a dashboard
- No single point of failure
- Decentralized
- Up to 1M requests free plan
- Incentivizing more full nodes across different chains

**How would Pocket be integrated?**
For the [Airnode repository](https://github.com/api3dao/airnode),

1. Replace all [Infura references](https://github.com/api3dao/airnode/blob/pre-alpha/packages/deployer/src/config-data/config.json.example#L24) with Pocket's. 
2. Suggest Airnode operators sign up for a Pocket instead of Infura.
3. Use a Pocket endpoint in case there's no provider in the configuration, [see here](https://github.com/api3dao/airnode/blob/pre-alpha/packages/deployer/src/evm/evm.ts#L21) (it doesn't default to anything now).

[Sign up for a free Application ID on Pocket Network.](https://mainnet.dashboard.pokt.network/#/)

andreogle:
Hey! Thanks for this and really sorry for the slow reply. I like what Pocket Network is doing and I support suggesting that users register with Pocket as an option for getting a provider URL.

[...]

PR #525, opened 2021-08-24
docs: update provider urls with Pocket

crisog:
Related to #424.

Pocket doesn't support Ropsten, I see there are a few tests using Ropsten (I guess because there are contracts deployed on it already).

Do you guys have also the same contracts deployed to Rinkeby so we can update those Infura Ropsten references with Pocket Ethereum Rinkeby?

As discussed in the issue, I'll also submit PRs to [admin package](https://github.com/api3dao/airnode/tree/master/packages/admin) and [api3-docs](https://github.com/api3dao/api3-docs).

### DIFF
--- a/packages/admin/README.md
+++ b/packages/admin/README.md
@@ -3,7 +3,7 @@
 
 > A package/CLI tool to interact with the Airnode contracts across chains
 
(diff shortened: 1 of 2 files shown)


andreogle (approved):
I'm not sure if the contracts have been deployed to Rinkeby yet. @bbenligiray would be able to answer that. Otherwise, I'm happy with this change 👍🏻

PR #529, opened 2021-08-26
Beta protocol cherry pick to master

Siegrift:
Yesterday I've tried rebasing the whole thing on master and it was challenging, but doable (took me like ~3 hours)... That doesn't mean that it's rebased correctly though.

I want to make it as easier as possible to review the changes and compare the rebase differences, so I've decided to redo it in a different way. I will start cherry picking the commits introduced in `beta-protocol-revision` branch and will gradually ask @acenolaza  for review. Once I finish the whole rebase I'd like all of you @bbenligiray @acenolaza and @Ashar2shahid to approve and review it.

To simplify the process (both for cherry picking and reviewing) I've created a small utility: https://gist.github.com/Siegrift/427129b166b8179b8408a1f23dab57e1

### For reviewers

While cherry picking I will always commit the original cherry picked commit (even if it adds no changes) to make it easier reviewable. This means you can review the PR commit by commit and you will see basically the same commits that have been commited to `beta-protocol-revision`. The only ugly thing is that merge commits might make it difficult.

There is another PR https://github.com/api3dao/airnode/pull/532 which addresses the issues I failed to resolve while fixing the conflicts. Before this PR is merged, the fixes PR should be included first.

#### Tip

Use the script I created (referenced above). You can download it anywhere and you can when you run: `node ../rebase-utils/git-utils.js diff` it will show you a diff of the original commit and the one created by cherry pick. You can also use `node ../rebase-utils/git-utils.js diff <number_of_commits_to_skip>` if you want to check only some of the later commits.

Also, this way should be quite adaptive to changes both in `master` and `beta-protocol-revision` :)


andreogle (commented):
Thanks for the utility script - I did give it a try. I've tried to review as thoroughly as I could but it's still very difficult to ensure everything is present. From what I can see, everything looks good

packages/admin/test/fixtures/authorizers.json
```diff
@@ -1,3 +1,4 @@
 [
-  "0x0000000000000000000000000000000000000000"
+    "0x0000000000000000000000000000000000000000",
```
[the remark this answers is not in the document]
acenolaza: Good catch! I imagine this file is not even used nor needed anymore 🤔

packages/node/src/evm/authorization/authorization-fetching.test.ts
```diff
@@ -15,12 +15,13 @@ import { RequestStatus } from '../../types';
...
 describe('fetch (authorizations)', () => {
-  let mutableFetchOptions: any;
+  let mutableFetchOptions: authorization.FetchOptions;
 
   beforeEach(() => {
     mutableFetchOptions = {
-      address: '0xD5659F26A72A8D718d1955C42B3AE418edB001e0',
-      airnodeId: '0xf5ad700af68118777f79fd1d1c8568f7377d4ae9e9ccce5970fe63bc7a1c1d6d',
+      authorizers: ['0x711c93B32c0D28a5d18feD87434cce11C3e5699B', '0x9E0e23766b0ed0C492804872c5164E9187fB56f5'],
+      airnodeAddress: '0xf5ad700af68118777f79fd1d1c8568f7377d4ae9e9ccce5970fe63bc7a1c1d6d',
```
acenolaza, line 23: Here I should have used an actual address instead of keeping the same bytes32 id. This is a mistake made by me and I'll fix it on another branch.

packages/node/src/evm/fulfillments/api-calls.ts
```diff
@@ -53,11 +46,11 @@ async function testFulfill(
     `Attempting to fulfill API call with status code:${statusCode.toString()} for Request:${request.id}...`
   );
 
-  const operation = () =>
+  const operation = (): Promise<StaticResponse> =>
     airnodeRrp.callStatic.fulfill(
       request.id,
-      // TODO: make sure airnodeId is not null
-      request.airnodeId!,
+      // TODO: make sure airnodeAddress is not null
```
acenolaza, line 52: I wonder if we still need to check what the TODO says 🤔

packages/protocol/test/AirnodeParameterStore.js
```diff
@@ -1,10 +1,10 @@
 /* globals context ethers */
```
acenolaza, line 1: This file should have been deleted because the contract doesn't exist anymore

acenolaza (approved):
A few minor comments but looks good to me 👍🏻

packages/admin/test/fixtures/authorizers.json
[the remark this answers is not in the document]
Siegrift: It's not. I'll remove it and apply it on https://github.com/api3dao/airnode/pull/532

packages/node/src/evm/authorization/authorization-fetching.test.ts
Siegrift: Ok. I'll resolve this conversation. You can apply the fix after this one is merged.

packages/node/src/evm/fulfillments/api-calls.ts
Siegrift: I think at that point it must be defined, but TS type is still `string | null`. We could maybe rethink the typings to avoid `!`, but that doesn't have a high prio.

packages/protocol/test/AirnodeParameterStore.js
Siegrift: Already removed in https://github.com/api3dao/airnode/pull/532

acenolaza (approved):
All good with latests changes 💪🏻

PR #541, opened 2021-09-01
AN-130: Skip authorization checks if authorizers is empty array

acenolaza:
Fix: 
Skips calling AirnodeRrp.checkAuthorizationStatuses when authorizers array is empty and returns true for all pending requests.

Operation package commands should be back to working condition.

@Siegrift Please checkout the branch and run the e2e locally. For some weird reason requester-fulfill.feature tests started to pass automagically. 

We can also work on a separate branch for fixing e2e since this is actually for fixing public authorizers.

### DIFF
--- a/packages/node/src/evm/authorization/authorization-fetching.ts
+++ b/packages/node/src/evm/authorization/authorization-fetching.ts
@@ -109,6 +109,14 @@ export async function fetch(
     return [[], {}];
   }
 
+  // If there are no authorizer contracts then endpoint is public
+  if (isEmpty(fetchOptions.authorizers)) {
+    const authorizationByRequestIds = pendingApiCalls.map((pendingApiCall) => ({
+      [pendingApiCall.id]: true,
+    }));
+    return [[], Object.assign({}, ...authorizationByRequestIds) as AuthorizationByRequestId];
+  }
+
   // Request groups of 10 at a time
   const groupedPairs = chunk(pendingApiCalls, CONVENIENCE_BATCH_SIZE);
 
--- a/packages/node/__dev__/config.json.example
+++ b/packages/node/__dev__/config.json.example
@@ -2,9 +2,7 @@
   {
     "chains": [
       {
-        "authorizers": [
-          "0x0000000000000000000000000000000000000000"
-        ],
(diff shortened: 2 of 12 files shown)


packages/node/src/evm/authorization/authorization-fetching.ts
```diff
@@ -109,6 +109,14 @@ export async function fetch(
     return [[], {}];
   }
 
+  // If there are no authorizer contracts then endpoint is public
+  if (isEmpty(fetchOptions.authorizers)) {
```
andreogle, line 113: Nice and simple 👍🏻

andreogle (approved):

packages/node/src/evm/authorization/authorization-fetching.ts
Siegrift: Actually a nice trick to convey the same information is to create a very specific function/variable for that:
```js
const isEndpointPublic = isEmpty(fetchOptions.authorizers)
if (isEndpointPublic) ...
```

Ofc, this is perfectly fine - but I've read this tip from a smartass from some book :) :D

Siegrift (approved):
:1st_place_medal: I've tried running unit tests locally (because they failed on CI) but they pass locally for me so I retried them on CI...

The e2e tests still fail for me locally and on CI as well. I'll try to debug them as part https://api3dao.atlassian.net/browse/AN-177 :)

packages/node/src/evm/authorization/authorization-fetching.ts
acenolaza: I really like that idea but I think I like consistency more tbh 😄

acenolaza:
>  I've tried running unit tests locally (because they failed on CI) but they pass locally for me so I retried them on CI...
> 
> The e2e tests still fail for me locally and on CI as well. I'll try to debug them as part https://api3dao.atlassian.net/browse/AN-177 :)

There is also a "Check documentation" also failing. I'm going to look at it later.

Thanks @Siegrift for taking care of the e2e! I really appreciate your help.
Also thank you @andreogle for the patience and the attention to detail when reviewing PRs.

Issue #507, opened 2021-08-17
Requester cannot specify arbitrary off-chain computation on the API response


bbenligiray:
Say the API returns this as the response
```json
{
  "data": {
    "fieldA": 123,
    "fieldB": 456
  },
  "version": "7.8.9"
}
```
Currently, we only allow the requester to specify `_type`, `_path` and `_times` reserved parameters, which can be used to read a single field of the JSON object and typecast it as a 32 bytes-long type.

Ideally, the requester would be able to specify arbitrary off-chain computation such as:
- Return `true` if `data.fieldA > 100`, return `false` otherwise
- ABI-encode `data.fieldA` and `data.fieldB` to be decoded by the requester contract (because it needs both fields)

Issue #484, opened 2021-08-09
Requester can't specify gasLimit


bbenligiray:
Currently, `gasLimit` is hardcoded (https://github.com/api3dao/airnode/issues/464). The requester may want to override this, and we don't have any reason not to allow them to.

Tracked by https://api3dao.atlassian.net/browse/AN-81

bbenligiray:
Closed to be tracked by https://api3dao.atlassian.net/jira/software/projects/AN/boards/18/backlog?selectedIssue=AN-81

bbenligiray:
Here's one case where this would have helped https://github.com/api3dao/airnode/pull/542#issuecomment-918155800

bbenligiray:
At least for now, we will only allow gas price/limit parameters to be configurable by the Airnode operator in their configuration file

PR #542, opened 2021-09-01
Parse API response to BigNumber

An-dz:
## Description

Large numbers get destroyed because JavaScript only works on IEEE-754 float numbers.

While #93 did some changes to support large integers, there's more to be done because when the number reaches the _[adapter](/api3dao/airnode/tree/master/packages/adapter)_ the number has already been destroyed way before at the _[node](/api3dao/airnode/tree/master/packages/node)_.

As can be seen on the following transactions on ropsten, the rightmost 60 bits are always zero, and comparing with what the API did respond only the first 53 bits were correct and the rightmost 60 bits were not all zero.

[0x59112a449054cfdd7bc324c54df4eba26e296c84888c40d690c6fe7d388a4227](https://ropsten.etherscan.io/tx/0x59112a449054cfdd7bc324c54df4eba26e296c84888c40d690c6fe7d388a4227#eventlog)
[0x1a1ce21657b0ca34e0e11662dae5296034bde994a7159726ef32996d00348555](https://ropsten.etherscan.io/tx/0x1a1ce21657b0ca34e0e11662dae5296034bde994a7159726ef32996d00348555#eventlog)
[0x2dd15eaef0605720f8b0cbd3bb409cd18bb085469c177fcbe4d8f180a9605404](https://ropsten.etherscan.io/tx/0x2dd15eaef0605720f8b0cbd3bb409cd18bb085469c177fcbe4d8f180a9605404#eventlog)
[0x8bd5d221648bd50baa247926a41454af2c594c394501e75ee9cdf68013e9e63f](https://ropsten.etherscan.io/tx/0x8bd5d221648bd50baa247926a41454af2c594c394501e75ee9cdf68013e9e63f#eventlog)
[0x8f26294acab7d427dd857651b06ea2d96dd9b3ea316496a88f96beccc9151341](https://ropsten.etherscan.io/tx/0x8f26294acab7d427dd857651b06ea2d96dd9b3ea316496a88f96beccc9151341#eventlog)
[0x776aeef840e60f44edb70bb25cb394eb3591894b7ff2a5d524fd01ad12edabac](https://ropsten.etherscan.io/tx/0x776aeef840e60f44edb70bb25cb394eb3591894b7ff2a5d524fd01ad12edabac#eventlog)
[0x9143cbb5502f459d5e82a59887de9e8bc3fa087233c8d7d71a321a89908c9e9d](https://ropsten.etherscan.io/tx/0x9143cbb5502f459d5e82a59887de9e8bc3fa087233c8d7d71a321a89908c9e9d#eventlog)
[0xa8fea1369604f8918b491c8a2f3bcd78028060b37d1851442eecaa93537dc42c](https://ropsten.etherscan.io/tx/0xa8fea1369604f8918b491c8a2f3bcd78028060b37d1851442eecaa93537dc42c#eventlog)

The problem lies in this part of the code:

https://github.com/api3dao/airnode/blob/9577aeb7a8f6aea3ca2571eb923ce59acff16b8e/packages/node/src/workers/cloud-platforms/aws.ts#L22

`JSON.parse` will destroy any number too large for JS, you need a custom JSON parser that handles large numbers.

## Version

Commit 9577aeb7a8f6aea3ca2571eb923ce59acff16b8e

## Present Behaviour

Any number too larger for JS to handle gets broken.

## Expected Behaviour

The app should handle any integer that is 256 bits or less.

## Steps to reproduce

Add the tests that were included in this PR to the master branch and see it fail.

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -28,10 +28,12 @@
     "@api3/operation": "^0.1.0",
     "@api3/protocol": "^0.1.0",
     "@lifeomic/attempt": "^3.0.0",
+    "@types/json-bigint": "^1.0.1",
     "aws-sdk": "^2.753.0",
     "date-fns": "^2.16.1",
     "dotenv": "^10.0.0",
     "ethers": "^5.4.4",
+    "json-bigint": "^1.0.0",
     "lodash": "^4.17.21"
   },
   "devDependencies": {
--- a/packages/node/src/workers/cloud-platforms/aws.test.ts
+++ b/packages/node/src/workers/cloud-platforms/aws.test.ts
@@ -1,3 +1,5 @@
+import { BigNumber } from 'bignumber.js';
+
 const invokeMock = jest.fn();
 jest.mock('aws-sdk', () => ({
   Lambda: jest.fn().mockImplementation(() => ({
@@ -25,6 +27,54 @@ describe('spawn', () => {
     };
     const res = await aws.spawn(parameters);
     expect(res).toEqual({ value: 7777 });
+    expect(invoke).toHaveBeenCalledTimes(1);
+    expect(invoke).toHaveBeenCalledWith(
+      {
+        FunctionName: 'airnode-19255a4-test-some-function',
+        Payload: JSON.stringify({ from: 'ETH', to: 'USD' }),
+      },
+      expect.any(Function)
+    );
+  });
+
+  it('should not kill a large number', async () => {
+    const lambda = new AWS.Lambda();
+    const invoke = lambda.invoke as jest.Mock;
+    invoke.mockImplementationOnce((params, callback) =>
+      callback(null, { Payload: '{"body":"{\\\"value\\\":115792089237316195423570985008687907853269984665640564039457584007913129639935}"}' })
+    );
+    const workerOpts = fixtures.buildWorkerOptions({ cloudProvider: 'aws' });
+    const parameters = {
+      ...workerOpts,
+      functionName: 'some-function' as WorkerFunctionName,
+      payload: { from: 'ETH', to: 'USD' },
+    };
+    const res = await aws.spawn(parameters);
+    expect(res).toEqual({ value: new BigNumber('115792089237316195423570985008687907853269984665640564039457584007913129639935') });
+    expect(invoke).toHaveBeenCalledTimes(1);
+    expect(invoke).toHaveBeenCalledWith(
+      {
+        FunctionName: 'airnode-19255a4-test-some-function',
+        Payload: JSON.stringify({ from: 'ETH', to: 'USD' }),
+      },
+      expect.any(Function)
+    );
+  });
+
+  it('should not kill a large float number', async () => {
+    const lambda = new AWS.Lambda();
+    const invoke = lambda.invoke as jest.Mock;
+    invoke.mockImplementationOnce((params, callback) =>
+      callback(null, { Payload: '{"body":"{\\\"value\\\":1157920892373161954235709850086879078532699.84665640564039457584007913129639935}"}' })
+    );
+    const workerOpts = fixtures.buildWorkerOptions({ cloudProvider: 'aws' });
+    const parameters = {
+      ...workerOpts,
+      functionName: 'some-function' as WorkerFunctionName,
+      payload: { from: 'ETH', to: 'USD' },
+    };
+    const res = await aws.spawn(parameters);
+    expect(res).toEqual({ value: new BigNumber('1157920892373161954235709850086879078532699.84665640564039457584007913129639935') });
     expect(invoke).toHaveBeenCalledTimes(1);
     expect(invoke).toHaveBeenCalledWith(
       {
--- a/packages/node/src/workers/cloud-platforms/aws.ts
+++ b/packages/node/src/workers/cloud-platforms/aws.ts
@@ -1,4 +1,5 @@
 import AWS from 'aws-sdk';
+import JSON from 'json-bigint';
 import { WorkerParameters, WorkerResponse } from '../../types';
 
 export function spawn(params: WorkerParameters): Promise<WorkerResponse> {
--- a/yarn.lock
+++ b/yarn.lock
@@ -2738,6 +2738,11 @@
     jest-diff "^26.0.0"
     pretty-format "^26.0.0"
 
+"@types/json-bigint@^1.0.1":
+  version "1.0.1"
+  resolved "https://registry.yarnpkg.com/@types/json-bigint/-/json-bigint-1.0.1.tgz#201062a6990119a8cc18023cfe1fed12fc2fc8a7"
+  integrity sha512-zpchZLNsNuzJHi6v64UBoFWAvQlPhch7XAi36FkH6tL1bbbmimIF+cS7vwkzY4u5RaSWMoflQfu+TshMPPw8uw==
+
 "@types/json-schema@*", "@types/json-schema@^7.0.6", "@types/json-schema@^7.0.7":
   version "7.0.7"
   resolved "https://registry.yarnpkg.com/@types/json-schema/-/json-schema-7.0.7.tgz#98a993516c859eb0d5c4c8f098317a9ea68db9ad"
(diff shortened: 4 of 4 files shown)


packages/node/package.json
```diff
@@ -28,10 +28,12 @@
     "@api3/operation": "^0.1.0",
     "@api3/protocol": "^0.1.0",
     "@lifeomic/attempt": "^3.0.0",
+    "@types/json-bigint": "^1.0.1",
```
andreogle, line 31: Can we move this to `devDependencies`?

packages/node/src/workers/cloud-platforms/aws.ts
```diff
@@ -1,4 +1,5 @@
 import AWS from 'aws-sdk';
+import JSON from 'json-bigint';
```
andreogle, line 2: I would prefer to not override the default global `JSON`. Can we rather call this something else? e.g. 
```suggestion
import JSONBigInt from 'json-bigint';
```

packages/node/src/workers/cloud-platforms/aws.test.ts
```diff
@@ -1,3 +1,5 @@
+import { BigNumber } from 'bignumber.js';
```
andreogle, line 1: I think we should be explicit about this dependency in the `node` package's package.json. Can we add `bignumber.js` there too please?

bbenligiray:
My assumption was that the API will return big numbers as strings. Referring to [here](https://datatracker.ietf.org/doc/html/rfc7159#section-6), this seems to be recommended but there are no strict rules.
To me, not acting on this (to avoid the added dependency and complexity) is a valid option because there is no limit to the extent of out of the ordinary JSON formats that we can support and I don't see this solution playing well with what we're planning to build to address https://github.com/api3dao/airnode/issues/507 at all.

Siegrift:
I agree with Burak that this is a more generic problem and should be handled as part of #507 instead.

An-dz:
> My assumption was that the API will return big numbers as strings. Referring to [here](https://datatracker.ietf.org/doc/html/rfc7159#section-6), this seems to be recommended but there are no strict rules.

The JSON specification does not set a limit because it depends on where it would be used. Since most APIs have a use on frontend applications in JS it's common to use strings for large numbers. But it's an implementation choice. Since we developed an endpoint specific for EVM we went with a 256 bit number.

> To me, not acting on this (to avoid the added dependency and complexity) is a valid option because there is no limit to the extent of out of the ordinary JSON formats that we can support and I don't see this solution playing well with what we're planning to build to address #507 at all.

I'm guessing #507 will be done on the adapter or somewhere else outside the node, because at some point it must be cast to big number to do comparisons and other computations correctly. Because `18014398509481984 > 18014398509481983` is `false` and that's only 0.018 $ETH.

An-dz:
I changed our response to be a string but now Airnode fails.

TX: [0xeebf30cd0f215b4be06d84ad3f63dca7b6e83cd1e429140adb6d96f19b68f60a](https://ropsten.etherscan.io/tx/0xeebf30cd0f215b4be06d84ad3f63dca7b6e83cd1e429140adb6d96f19b68f60a)

_callAPI_, _startCoordinator_ and _initializeProvider_ shows nothing on log, the API did get the request and responded correctly.

The only relevant log is from _processProviderRequests_ that has this:
```json
{
    "timestamp": "2021-09-02 15:01:47.945",
    "level": "INFO",
    "message": "Submitting API call fail for Request:0xca6d813ccdc041ffffd229674ebd9b0652cce289a34c98ae01bb1dfb56a2201e...",
    "coordinatorId": "bdfd74d5f23421e2",
    "providerName": "ropsten",
    "chainType": "evm",
    "chainId": "3"
}
```

Siegrift:
Hi, @An-dz Can you please provide more details? Is it based on airnode starter? Can you share your config.json?

I am not sure why the transaction fails, but it's because of `onlyCorrectFulfillmentParameters` which checks whether you are fulfilling with the same parameters as when the request was created. However this check doesn't care about the encoded data at all - so I find it really strange why it fails for you.

An-dz:
Hi @Siegrift 

It's based on api3dao/airnode-starter@b521d9d77dc3c4d3f6b27adf674adca3a3fba05f, it's using the airnode-deployer pre-alpha `c378a59487c8` docker image.

The difference between the half-working transactions and the failing ones is only the API responding with a string instead of a number. Both contract and aws airnode were not changed.

Here's the transaction that initiated the request:
[0x267eaaa40b5d8f08655adb35fbf10856d2d127d555da9e668d7fb3117339a23b](https://ropsten.etherscan.io/tx/0x267eaaa40b5d8f08655adb35fbf10856d2d127d555da9e668d7fb3117339a23b#eventlog)

I can't see why it would fail here, all 5 parameters (requestId, providerId, designatedWallet, fulfillAddress, fulfillFunctionId) are the same from when it was requested.

But another problem I see in the failed/reverted transaction is that it called the `fail` function and not `fullfill`, so our Airnode did fail to execute correctly.

Here's out config file: [config.json.txt](https://github.com/api3dao/airnode/files/7116162/config.json.txt)

Siegrift:
@An-dz Does your API still return strings? I've tried calling https://beta.heimdall.land/v2/coins/btc but all the numbers in the response were (small) numbers.

I tried reproducing your problem with my own tiny API, which returns large number. There are two API routes:
- https://airnode-mocked-api-netlify.netlify.app/.netlify/functions/big-number returns the number as `string`
- https://airnode-mocked-api-netlify.netlify.app/.netlify/functions/big-number-raw returns it as regular `number`

I've deployed Airnode that is able to call these two endpoints and both of them work as expected for me. If I call the endpoint that returns regular `number` I Iose precision (exactly how you reported in this issue). If I use the `string` version, I get the correct API response back without error.

I've pushed the changes to: https://github.com/Siegrift/airnode-starter/commit/cf8c9056d93ac7c22a780fd282e8777d9a73cd5e in case you are interested.

However, I still don't know why does it fail for you. The revert reason (`Incorrect fulfillment parameters`) seems to be unrelated to the shape of the API response :/ 

I'd like to investigate further. Can you tell me the endpoint that you are calling (along with it's parameters)? I'd like to reproduce the **exactly** same error as you have.

An-dz:
@Siegrift Sorry, forgot to explain that, only one endpoint responds with big numbers, that's `getMultipleSocialScores`. It's still responding with a string, you can test it out on our beta docs page: https://beta.heimdall.land/v2/docs#/social%20score/symbols_score_uint256_v2_coins_scores_post

The contract is requesting with `period = "30d"` and `symbols = "btc,eth,matic,link,uni,sushi,aave,chz,theta,rsr,grt,enj,ocean,kacy"`

Both are being requested as a string, even though `period` could go with _bytes32_. Here's the part of the code in the contract that encodes the parameters:
```solidity
_parametersHeimdall = abi.encode(
    bytes32("1SS"),
    bytes32("period"), "30d",
    bytes32("symbols"), symbols
);
```
Though I don't think that's the problem, as the API is receiving the request and is responding correctly.

Siegrift:
Sorry for responding so late, I had some personal errands throughout the day :/

I have still no idea why Airnode fails for you. I've updated my `config.json`, `ExampleClient.sol`, `parameters.js` and `make-request.js` script to call your API and used the same encoding as you and it works for me.

Maybe there are some differences between mine Airnode and yours I am not aware of. You can check out my configuration in:
- https://github.com/api3dao/airnode-starter/commit/8d30aedec0336a31cd401c709bdb5f3de9f4d2e2

NOTE: The output of `npm run make-request` returned `-57896044152017083202966606769401135416575027563102747117842488555085535247157` which is right, because the API returns number larger than int256 and solidity overflows.

An-dz:
@Siegrift
I also changed my airnode starter script to make the request and it indeed worked. I guess the problem is in the contract. I'll investigate the contract to see what could be the problem.

And the API doesn't return numbers larger than int256, it just answers them as **u**int256 but the first bit is always one so they end up negative when they reach the contract.

An-dz:
I pinpointed the problem to here:

https://github.com/api3dao/airnode/blob/19c9dce9ce281f3735b8377425169891b7ec05c2/packages/node/src/evm/fulfillments/api-calls.ts#L56-L71

My requests are calling `submitFail` but `testFulfill` doesn't call the error log. So I'm guessing that static call to `fullfill` in `testFullfill` is reverting. Unfortunately there's no log for that so I'm trying to debug that, unfortunately my AWS skills are not that good.

I've tested the contract directly with the data that should be sent to my contract and no errors at all, so I'm puzzled if something in the Airnode contract would revert, but again, nothing indicates that.

I'm now trying to build a new docker to put a new airnode online to better debug this.

Siegrift:
Can I help you? Can you share your contract and the parameters you are calling it with?

An-dz:
Thanks for the offer, but I found out what is the problem. Because the code that receives the API3 call also makes calculations with the response the Gas is going above the limit of 500_000 gwei. I'll have to first save the response and move the bulk of the calculations to a second call we'll have to do. There're quite some optimisations we can do too.

So problem solved, my only request would be to add a log in Airnode for the case when the `callStatic.fulfill` reverts, and maybe add a explanation field in the Airnode contract `fail` function too so users can have an idea of why it failed.

bbenligiray:
@An-dz we will implement a `_gasLimit` reserved parameter that the requester can set (https://github.com/api3dao/airnode/issues/484) but this is a bit low on the priority list. Thank you for reporting back.

PR #543, opened 2021-09-01
AN-98: Replace endorsing with sponsoring

acenolaza:
Everything has been taken care in regards to endorser->sponsor renaming. Documentation update is the only thing left here.

All I was able to find was a comment and a link to https://github.com/api3dao/api3-docs/blob/master/request-response-protocol/endorsement.md. I remember that something like "READMEs should not have a dependency to docs" therefore I removed all the links in the operation package README.

### DIFF
--- a/packages/node/test/setup/e2e/deployment.ts
+++ b/packages/node/test/setup/e2e/deployment.ts
@@ -21,7 +21,7 @@ export async function deployAirnodeRrp(config: operation.Config): Promise<operat
   // Set Airnode xpub
   const state11 = await operation.setAirnodeXpub(state10);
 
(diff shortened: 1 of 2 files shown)


andreogle (approved):

Siegrift:
> I remember that something like "READMEs should not have a dependency to docs" therefore I removed all the links in the operation package README.

I am not sure about such strict rule (but my memory is not very sharp). Our focus is to make the READMEs developer oriented, so by definition there shouldn't be a reason to link to docs. (Apart from saying "For customer facing documentation see <API3 docs link>").

Siegrift (approved):

PR #547, opened 2021-09-03
Fixes relay metadata parameters

acenolaza:
### Bug:
Relay metadata parameters are not included in http requests even after properly configuring the airnode

### Fix:
When implementing this feature I made the wrong assumption that adding the parameters to the reservedParameters list in [call-api.ts](https://github.com/api3dao/airnode/blob/1c94313f636da4c7f4e08baf6327fc24cb46c8df/packages/node/src/handlers/call-api.ts#L83) was enough and that they were going to be sent on http requests. This was not properly tested and the metadata parameters where ignored and never included.

 So I've created a new method on the adapters package to include them and tested both GET and POST requests using the operation package tools. Confirmed that the values reached the api by console logging the querystring when request was a GET and body when request was a POST.
 
Also added type checking for the V1 relay metadata parameters by adding a new interface to types.ts.

One thing I didn't like too much was extracting the relay metadata parameters by searching all keys that started with "_adapter_" on the parameters array. I would have prefer to use the new `RelayMetadataV1` interface instead of the string but I wasn't able to do it since interfaces do not exists on runtime. Another alternative was have those keys hardcoded in a string[] but I didn't like that duplication either.

**This fix also needs to be applied to pre-alpha since it's blocking Cam with an integration.**

### DIFF
--- a/packages/adapter/src/request-building/parameters.ts
+++ b/packages/adapter/src/request-building/parameters.ts
@@ -76,13 +76,13 @@ function buildUserParameters(options: CachedBuildRequestOptions): BuilderParamet
   }, initalParameters());
 }
 
-function buildRelayMetadataParameters(options: CachedBuildRequestOptions): BuilderParameters {
-  const { parameters } = options;
+function buildMetadataParameters(options: CachedBuildRequestOptions): BuilderParameters {
+  const { metadataParameters } = options;
 
-  const parameterKeys = Object.keys(parameters).filter((key) => key.startsWith('_airnode_'));
+  const parameterKeys = Object.keys(metadataParameters);
 
   return parameterKeys.reduce((acc, key) => {
-    return appendParameter(acc, 'query', key, parameters[key]);
+    return appendParameter(acc, 'query', key, metadataParameters[key]);
   }, initalParameters());
 }
 
@@ -90,7 +90,7 @@ export function buildParameters(options: CachedBuildRequestOptions): RequestPara
   const auth = authentication.buildParameters(options);
   const fixed = buildFixedParameters(options);
   const user = buildUserParameters(options);
-  const relayMetadata = buildRelayMetadataParameters(options);
+  const metadata = buildMetadataParameters(options);
 
   const cookie = cookies.buildHeader({ ...user.cookies, ...fixed.cookies, ...auth.cookies });
 
@@ -98,7 +98,7 @@ export function buildParameters(options: CachedBuildRequestOptions): RequestPara
   // overwrite fixed and authentication parameters
   return {
     paths: { ...user.paths, ...fixed.paths },
-    query: { ...user.query, ...fixed.query, ...auth.query, ...relayMetadata.query },
+    query: { ...user.query, ...fixed.query, ...auth.query, ...metadata.query },
     headers: { ...user.headers, ...fixed.headers, ...auth.headers, ...cookie },
   };
 }
--- a/packages/adapter/src/request-building/parameters.ts
+++ b/packages/adapter/src/request-building/parameters.ts
@@ -76,10 +76,21 @@ function buildUserParameters(options: CachedBuildRequestOptions): BuilderParamet
   }, initalParameters());
 }
 
+function buildRelayMetadataParameters(options: CachedBuildRequestOptions): BuilderParameters {
+  const { parameters } = options;
+
+  const parameterKeys = Object.keys(parameters).filter((key) => key.startsWith('_airnode_'));
+
+  return parameterKeys.reduce((acc, key) => {
+    return appendParameter(acc, 'query', key, parameters[key]);
+  }, initalParameters());
+}
+
 export function buildParameters(options: CachedBuildRequestOptions): RequestParameters {
   const auth = authentication.buildParameters(options);
   const fixed = buildFixedParameters(options);
   const user = buildUserParameters(options);
+  const relayMetadata = buildRelayMetadataParameters(options);
 
   const cookie = cookies.buildHeader({ ...user.cookies, ...fixed.cookies, ...auth.cookies });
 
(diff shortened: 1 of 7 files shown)


packages/adapter/src/request-building/parameters.ts
```diff
@@ -76,18 +76,29 @@ function buildUserParameters(options: CachedBuildRequestOptions): BuilderParamet
   }, initalParameters());
 }
 
+function buildRelayMetadataParameters(options: CachedBuildRequestOptions): BuilderParameters {
+  const { parameters } = options;
+
+  const parameterKeys = Object.keys(parameters).filter((key) => key.startsWith('_airnode_'));
```
andreogle, line 82: I don't think this is ideal...

What about rather adding a new option to the `BuildRequestOptions`, specifically for parameters that bypass the OIS checks and always get included?

acenolaza: I thought about that as well but it seemed like too big of a change for some reason. I thought of adding something like this:

```
readonly metadata: {
        readonly [key: string]: string;
    }
```
    
to the BuildRequestOptions interface instead of using the same `parameters` object. Is that what you're suggesting or did I get it wrong?

andreogle: Yeah that's what I was thinking. It's not really ideal either but it kinda feel better than just filtering on the parameter name including `_airnode_`. That feels like it might have unintended consequences at some point

acenolaza: Done. I like this much better since I agree that searching by startsWith('_adapter_') might lead to unexpected issues.

I'm still not sure if double checking the metadata parameters is required in the adapter package. Meaning looking for the relay metadata reserved parameter value and checking the metataParameters props based on the specified version. It seems unnecessary but I'd like your opinion.

acenolaza:
Operation package dummy API output: 

```
Server is running at http://localhost:5000
::ffff:127.0.0.1 - - [06/Sep/2021:14:33:25 +0000] "GET /convert?from=ETH&to=USD&access_key=%3Center+your+API+key%3E&_airnode_airnode_id=0x19255a4ec31e89cea54d1f125db7536e874ab4a96b4d4f6438668b6bb10a6adb&_airnode_client_address=0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512&_airnode_designated_wallet=0xa46c4b41d72Ada9D14157b28A8a2Db97560fFF12&_airnode_endpoint_id=0xeddc421714e1b46ef350e8ecf380bd0b38a40ce1a534e7ecdf4db7dbc9319353&_airnode_requester_index=2&_airnode_request_id=0xe79be653cd8ab2d03ff2996078d84fc72f59233f61b2798e11d3ea2a5ae9b812&_airnode_chain_id=31337&_airnode_chain_type=evm&_airnode_airnode_rrp=0x5FbDB2315678afecb367f032d93F642f64180aa3 HTTP/1.1" 200 38 "-" "axios/0.21.1"
::ffff:127.0.0.1 - - [06/Sep/2021:14:33:25 +0000] "GET /convert?from=ETH&to=USD&access_key=%3Center+your+API+key%3E&_airnode_airnode_id=0x19255a4ec31e89cea54d1f125db7536e874ab4a96b4d4f6438668b6bb10a6adb&_airnode_client_address=0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512&_airnode_designated_wallet=0xa46c4b41d72Ada9D14157b28A8a2Db97560fFF12&_airnode_endpoint_id=0xeddc421714e1b46ef350e8ecf380bd0b38a40ce1a534e7ecdf4db7dbc9319353&_airnode_requester_index=2&_airnode_request_id=0x750370320cdd7b8f07733896947db8047ff9309f1076a6359f348d7fc71ab1d0&_airnode_chain_id=31337&_airnode_chain_type=evm&_airnode_airnode_rrp=0x5FbDB2315678afecb367f032d93F642f64180aa3 HTTP/1.1" 200 38 "-" "axios/0.21.1"
```

packages/node/src/handlers/call-api.ts
```diff
@@ -36,8 +26,12 @@ function addMetadataParameters(
         _airnode_chain_type: chain.type,
         _airnode_airnode_rrp: chain.contracts.AirnodeRrp,
       };
+      return {
```
andreogle, line 29: No need to spread here

```suggestion
      return metadataParametersV1;
```

packages/node/src/handlers/call-api.ts
```diff
@@ -49,14 +43,15 @@ function buildOptions(
   apiCredentials: adapter.ApiCredentials[]
 ): adapter.BuildRequestOptions {
   // Include airnode metadata based on _relay_metadata version number
-  const parametersWithMetadata = addMetadataParameters(chain, aggregatedApiCall, reservedParameters);
+  const metadataParameters: adapter.Parameters = addMetadataParameters(chain, aggregatedApiCall, reservedParameters);
```
andreogle, line 46: Nit: I would change this function name to `buildMetadataParameters()` since it's not really doing any "adding"

andreogle (approved):
Looks good to me :shipit:

packages/node/src/handlers/call-api.ts
acenolaza: I think I do because I said that the `addMetadataParameters` method returns `adapter.Parameters` object and `metadataParametersV1` is of type `adapter.RelayMetadataV1`.

Maybe I overcomplicated things with this type checking but I was trying to make it more future proof by adding `adapter.RelayMetadataV1`. In the future we might have `adapter.RelayMetadataV2` interface etc and I wanted to have some sort of type checking to maker sure the properties we are adding are the correct ones.

acenolaza: I could maybe make `adapter.RelayMetadataV1` extend `adapter.Parameters` and the compiler would tell if a property is missing but it won't complain if we are adding more props to the object.

acenolaza: I'll merge to unblock an API integration but I could still improve this afterwards.

PR #549, opened 2021-09-07
Certik RBS-6/7 Replace the sponsorship check with an additional whitelist

bbenligiray:
I didn't implement template-specific whitelisting. If that is needed, it can be implemented in a proxy contract and the sponsor can permit that to request updates instead.

Tests are not implemented/fixed

### DIFF
--- a/packages/protocol/contracts/rrp/requesters/RrpBeaconServer.sol
+++ b/packages/protocol/contracts/rrp/requesters/RrpBeaconServer.sol
@@ -32,6 +32,10 @@ contract RrpBeaconServer is
     int256 private constant MIN_INT224 = type(int224).min;
     uint256 private constant MAX_UINT32 = type(uint32).max;
 
+    mapping(address => mapping(address => bool))
+        public
+        override sponsorToUpdateRequesterToPermissonStatus;
+
     mapping(bytes32 => Beacon) private templateIdToBeacon;
     mapping(bytes32 => bytes32) private requestIdToTemplateId;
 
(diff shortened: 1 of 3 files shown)


Ashar2shahid:
I'm working on the tests for this, I'll push it on this branch because the implementation is more or less the same

acenolaza:
This change is pretty straightforward if I'm not mistaken. We just need to replicate sponsorship flow from AirnodeRrp in here.

packages/protocol/contracts/rrp/requesters/interfaces/IRrpBeaconServer.sol
```diff
@@ -42,4 +53,9 @@ interface IRrpBeaconServer is IMetaAdminnable, IWhitelister {
         external
         view
         returns (int224 value, uint32 timestamp);
+
+    function sponsorToUpdateRequesterToPermissonStatus(
```
Ashar2shahid, line 57: Just curious but is it standard practice to explicitly state public mappings/structs in the interface? Shouldn't you be able to query them regardless? I've seen this done in the TemplateUtils as well.

packages/protocol/contracts/rrp/requesters/RrpBeaconServer.sol
```diff
@@ -42,12 +46,35 @@ contract RrpBeaconServer is
...
+    /// @notice Called by the sponsor to set the update request permission
+    /// status of an account
+    /// @param updateRequester Update requester address
+    /// @param updatePermissionStatus Update permission status of the update
+    /// requester
+    function setUpdatePermissionStatus(
+        address updateRequester,
+        bool updatePermissionStatus
+    ) external override {
+        require(updateRequester != address(0), "updateRequester address zero");
```
Ashar2shahid, line 58: I have added this check just to be sure.

Ashar2shahid (approved):
I've updated the tests and added an extra check. Can have a look

bbenligiray:
Agreed and closed.

PR #550, opened 2021-09-07
Endpoint-level whitelisting

bbenligiray:


packages/protocol/contracts/rrp/authorizers/RequesterRrpAuthorizer.sol
```diff
@@ -23,10 +70,19 @@ abstract contract RequesterRrpAuthorizer is
...
         address requester
     ) external view override returns (bool) {
-        return userIsWhitelisted(bytes32(abi.encode(airnode)), requester);
+        return
+            userIsWhitelisted(deriveAdminnedId(airnode, endpointId), requester);
+    }
+
+    function deriveAdminnedId(address airnode, bytes32 endpointId)
+        internal
+        view
```
acenolaza, line 83: Isn't this a pure function since it's not reading from state?

bbenligiray: It is

bbenligiray:
Closing this because the approach doesn't work

PR #552, opened 2021-09-07
AN86: deployer airnodeid to airnodeaddress

acenolaza:
As part of fixing [this](https://api3dao.atlassian.net/browse/AN-86) issue I forgot to update `deployer` package with all required changes from protocol refactor.

Thanks @amarthadan for helping me troubleshoot the issue regarding AWS not accepting uppercase letter in S3 bucket and Lamda function names.

There's also a clean up regarding not needing to pre-fund airnode wallet anymore since AirnodeRrp.setAirnodeParameters() doesn't need to be called after deployment (in fact, this method doesn't even exist anymore https://api3dao.atlassian.net/browse/AN-84)

Test:
ran deploy and remove from local


packages/deployer/src/utils/evm.ts
```diff
@@ -36,10 +29,12 @@ export function generateMnemonic() {
...
-    throw new Error('airnodeId is not a valid hex string');
+export function shortenAirnodeAddress(airnodeAddress: string) {
+  logger.debug('Shortening Airnode Address');
+  try {
+    ethers.utils.getAddress(airnodeAddress);
+  } catch {
+    throw new Error('airnodeAddress is not a valid hex string');
   }
-  return airnodeId.substring(2, 9);
+  return airnodeAddress.substring(2, 9);
```
Siegrift, line 39: @amarthadan You mentioned that this needs to be lowercased. 
But maybe I misunderstood?

Siegrift (approved):
👍 LGTM

andreogle (approved):
I haven't tested actually deploying this, but it looks good from what I can see 👍🏻  Thanks for also cleaning up the functions that are no longer used as a part of this PR

packages/deployer/src/utils/evm.ts
acenolaza: That is correct. I'm just calling toLowerCase() where I'm calling [this](https://github.com/api3dao/airnode/blob/fa68fc7fe787ca4235d386d17f04a67ea641cfe3/packages/deployer/src/cli/commands.ts#L49) function because I thought that making lower case is AWS specific and this  shortenAirnodeAddress() might be called for other purposes in the future.

amarthadan: I'd still move it here instead. Even if it is AWS specific issue I would like to have the lowercased version everywhere, not just there.

amarthadan (approved):

packages/deployer/src/utils/evm.ts
acenolaza: Ok I'll move it here and I'll apply the same logic in node package and also add it in [here](https://github.com/api3dao/airnode/blob/e0acbf5f3c8fc263e09e32ec818940a7886f3bdb/packages/node/src/evm/wallet.ts#L45).

PR #555, opened 2021-09-08
AN-176: Certik ARO-02 | requestId should include airnode

acenolaza:
### Fix:
Adds `airnode` (address) to `requestId` when calculating the hash and updates `node` packing to also use `airnodeAddress` while doing api request call verification.

### Test:
Ran tests
Ran operation package commands


acenolaza:
Converted this PR to draft because I need to fix  a unit test that requires regenerating the events emitted by AirnodeRrp (this will cause changes in other unit tests probably)

andreogle (approved):
🔥

PR #556, opened 2021-09-09
Endpoint whitelisting + refactor

bbenligiray:
Contracts and tests are complete, I'll review the comments and the readme while you guys are reviewing


packages/protocol/contracts/adminnable/SelfAdminnable.sol
```diff
@@ -0,0 +1,70 @@
...
+        uint256 newRank
+    )
+        external
+        override
+        onlyWithRank(
+            adminned,
+            max(adminnedToAdminToRank[adminned][targetAdmin], newRank) + 1
+        )
+    {
+        adminnedToAdminToRank[adminned][targetAdmin] = newRank;
```
acenolaza, line 45: Should we be preventing from setting `address(0)` on either level of the mapping?

packages/protocol/contracts/adminnable/SelfAdminnable.sol
```diff
@@ -0,0 +1,70 @@
...
+    /// @notice Called by an admin to decrease its rank for the adminned
+    /// address
+    /// @param adminned Adminned address
+    /// @param newRank Rank to be set
+    function decreaseSelfRank(address adminned, uint256 newRank)
+        external
+        override
+        onlyWithRank(adminned, newRank + 1)
+    {
+        adminnedToAdminToRank[adminned][msg.sender] = newRank;
```
acenolaza, line 58: Same here and now that I think about it in Adminned.sol too.

packages/protocol/contracts/rrp/authorizers/DaoRequesterRrpAuthorizer.sol
```diff
@@ -0,0 +1,135 @@
...
+/// @title Authorizer contract that whitelists requesters where the API3 DAO is
+/// the meta-admin
+/// @dev Unlike the MetaAdminnable contract that it inherits, this contract
+/// does not support the independent adminning of different entities. In other
+/// words, when an address is assigned to be an Admin/SuperAdmin, it becomes
+/// the Admin/SuperAdmin of all entities being adminned. To achieve this, the
+/// interface that allows the caller to specify an `adminnedId` is disabled,
+/// and an alternative one is implemented where `adminnedId` is forced to be
+/// `bytes32(0)`.
+contract DaoRequesterRrpAuthorizer is Adminnable, RequesterRrpAuthorizer {
```
acenolaza, line 16: It seems we are missing an interface similar to https://github.com/api3dao/airnode/blob/beta-protocol-endpoint-whitelisting/packages/protocol/contracts/rrp/authorizers/interfaces/IAirnodeRequesterRrpAuthorizer.sol#L0-L1

acenolaza (commented):
I wonder if there's a way to avoid duplicated code between `AirnodeRequesterRrpAuthorizer` and `DaoRequesterRrpAuthorizer` 🤔 

Since the logic of all these functions seems to be the same on both contracts: `extendWhitelistExpiration`, `setWhitelistExpiration` and `setWhitelistStatusPastExpiration` maybe they can go in RequesterRrpAuthorizer.sol. Then we could override them in both authorizer contracts just to set the proper `onlyWithRank` modifier and then call the parent implementation.

packages/protocol/contracts/adminnable/Whitelister.sol
```diff
@@ -1,115 +1,45 @@
...
-contract Whitelister is RankedAdminnable, IWhitelister {
-    enum AdminRank {
-        Unauthorized,
-        Admin,
-        SuperAdmin
-    }
-
+/// @title Contract where users are whitelisted for specific services (could be
+/// Airnodes, dAPIs, beacons, etc.) until an expiration time or indefinitely
+contract Whitelister {
```
Ashar2shahid, line 6: I don't see the point of this file anymore. I think we should probably move all of the functionality to RequesterRrpAuthorizer and rename that to something more apt.

packages/protocol/contracts/adminnable/Adminnable.sol
```diff
@@ -0,0 +1,73 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+import "./interfaces/IAdminnable.sol";
+
+/// @title Contract that implements multiple levels of admins
+contract Adminnable is IAdminnable {
```
Ashar2shahid, line 7: Should this be an abstract contracts with a non implemented getRank() function? I can see getRank being a useful method to have

packages/protocol/contracts/adminnable/SelfAdminnable.sol
```diff
@@ -0,0 +1,70 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+import "./interfaces/ISelfAdminnable.sol";
+
+/// @title Contract that implements multiple levels of admins for multiple
+/// "adminned" addresses independently
+contract SelfAdminnable is ISelfAdminnable {
```
Ashar2shahid, line 8: Same here,  should this be an abstract contracts with a non implemented getRank() function?

packages/protocol/contracts/rrp/authorizers/RequesterRrpAuthorizer.sol
```diff
@@ -5,28 +5,66 @@ import "../../adminnable/Whitelister.sol";
...
+    /// service ID, and we do this by calculating their hash.
+    /// @param airnode Airnode address
+    /// @param endpointId Endpoint ID
+    /// @return serviceId Service ID
+    function deriveServiceId(address airnode, bytes32 endpointId)
+        internal
+        pure
+        returns (bytes32 serviceId)
+    {
+        return keccak256(abi.encodePacked(airnode, endpointId));
```
Ashar2shahid, line 68: this is more of a question but if we use encode we can unpack the contents later. Wouldn't that be useful instead of just hashing it?

Ashar2shahid (commented):
Didn't go through the tests yet. Left some of my thoughts. Also agree with Santiago on the code duplication between Adminnable and SelfAdminnable and if there is a way to avoid it apart from what we did previously.

packages/protocol/contracts/adminnable/Whitelister.sol
acenolaza: I think `RrpBeaconServer` is using it, isn't it? 🤔

Ashar2shahid: Ah right, then this is fine. The only issue I have with the whitelister is that the setting functions are separated and duplicated across contracts as you've mentioned before. The old abstract way was a cleaner approach, atleast to me.

packages/protocol/contracts/adminnable/SelfAdminnable.sol
bbenligiray: Admin ranks for zero `adminned` will all be zero so those calls will revert anyway. I added checks for zero `targetAdmin` https://github.com/api3dao/airnode/pull/556/commits/adee69da44fc5f93d0bfa97165221bdad02ededf

packages/protocol/contracts/rrp/authorizers/DaoRequesterRrpAuthorizer.sol
bbenligiray: Good catch, added https://github.com/api3dao/airnode/pull/556/commits/6317f579409338dd003b7962b11fd276d9730b0c

packages/protocol/contracts/adminnable/Whitelister.sol
bbenligiray: The problem with the old approach were mainly the events. We want to emit `airnode`+`endpointId` and not `serviceId` for the authorizers and this is pretty much the only way of doing that.

packages/protocol/contracts/adminnable/Adminnable.sol
bbenligiray: `adminToRank()` is exposed as `public` now so that's no longer needed. The previous implementation depended on not exposing that and overriding `getRank()` to implement difference in functionality between the contracts, that's no longer the case.

packages/protocol/contracts/rrp/authorizers/RequesterRrpAuthorizer.sol
bbenligiray: We can't unpack the contents of the encoding because it's being hashed. It doesn't matter if `abi.encode` or `abi.encodePacked` is used here, and in these cases I use `encodePacked`, see https://github.com/api3dao/airnode/blob/f09794864e3daf58a4bb41dbc6e35d1d1c0c6dcf/packages/protocol/contracts/rrp/AirnodeRrp.sol#L130 (I think this changed with beta)

bbenligiray:
> Since the logic of all these functions seems to be the same on both contracts: extendWhitelistExpiration, setWhitelistExpiration and setWhitelistStatusPastExpiration maybe they can go in RequesterRrpAuthorizer.sol. Then we could override them in both authorizer contracts just to set the proper onlyWithRank modifier and then call the parent implementation.

I did this in https://github.com/api3dao/airnode/pull/556/commits/779a6d873bc844f4cea1b0dfe3fd8045e147a469 and like the outcome

packages/protocol/contracts/rrp/authorizers/RequesterRrpAuthorizer.sol
Ashar2shahid: My question is why are we hashing it, is it not possible to just encode and return the result?

bbenligiray: Because we need an identifier of `bytes32` type. `abi.encode` returns a `bytes` type that is longer than 32 bytes.

Ashar2shahid: I see thanks.

PR #557, opened 2021-09-11
PSP draft

bbenligiray:
Just some POCs

### DIFF
--- a/packages/protocol/contracts/psp/AirnodePsp.sol
+++ b/packages/protocol/contracts/psp/AirnodePsp.sol
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+contract AirnodePsp {
+    mapping(address => address) public airnodeToAuthorizer;
+    mapping(address => mapping(address => bool))
+        public sponsorToSubscriberToSponsorshipStatus;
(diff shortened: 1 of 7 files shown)


packages/protocol/contracts/psp/AirnodePsp.sol
```diff
@@ -0,0 +1,25 @@
...
+pragma solidity 0.8.6;
+
+contract AirnodePsp {
+    mapping(address => address) public airnodeToAuthorizer;
+    mapping(address => mapping(address => bool))
+        public sponsorToSubscriberToSponsorshipStatus;
+    mapping(address => uint256) public subscriberToSubscriptionCountPlusOne;
+
+    function setAuthorizer(address authorizer) external {
+        airnodeToAuthorizer[msg.sender] = authorizer;
```
acenolaza, line 11: Is msg.sender the airnode operator? does this mean that an airnode has authorizers by operator instead of global authorizers like in RRP?

bbenligiray: Here it's the `airnodeAddress` calling this function for simplicity. We can also implement some adminship functionality where `airnodeAddress` can authorize an admin address to call this function, but that's not needed yet.

bbenligiray:
Closing this in favor of the new one

PR #558, opened 2021-09-12
Small validator updates

drgy:
- Move conversion templates from `templates/conversion` to `conversions`
- Fix tests
- Update readme example commands
- Update build


packages/validator/package.json
```diff
@@ -26,6 +28,7 @@
   },
   "devDependencies": {
```
amarthadan, line 29: You're missing `rimraf` as a dev dependency.

packages/validator/tsconfig.build.json
```diff
@@ -2,7 +2,7 @@
   "extends": "../../tsconfig.build.json",
   "compilerOptions": {
     "baseUrl": "./",
-    "outDir": "./dist"
+    "outDir": "./dist/src"
```
amarthadan, line 5: I'm a bit worried that if you do this the imports will be different for local imports and imports from the package version (because the packaged version will have all the code under `src` and not directly, basically one level deeper).

But feel free to leave it as it is for now, I'll be going through all the packages anyway when I'll be sorting packaging.

amarthadan (changes requested):

packages/validator/src/commands/utils.ts
```diff
@@ -5,13 +5,23 @@ import { Log } from '../types';
...
+    if (!a.match(/^[0-9\.]+$/)) {
+      if (!b.match(/^[0-9\.]+$/)) {
+        return 0;
+      }
+
+      return 1;
+    }
+
+    return b.localeCompare(a, undefined, { numeric: true });
+  });
```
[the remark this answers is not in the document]
drgy: Thanks for that, I was missing one exit condition, fixed in 35d6fbee7bbe79fec124cfe6f8664f65e8dcbf42

packages/validator/package.json
drgy: Added in ff66de5eef05f1fb88e6ea103cb1817d27844f3d

packages/validator/tsconfig.build.json
drgy: Im not sure I follow this, the source of validator reads all of the templates relatively, which means if a file is located in `src/` it would look in `../templates/`. I thought package would consist only of the contents of `dist/`, and thats why the build was moved to `dist/src/` and templates are copied to `dist/templates/` so that package can access templates the exact same way.

packages/validator/src/commands/utils.ts
```diff
@@ -5,13 +5,27 @@ import { Log } from '../types';
...
 export const templateVersions = fs
   .readdirSync(validatorTemplatesPath, { withFileTypes: true })
-  .filter((dirent) => dirent.isDirectory() && dirent.name !== 'conversion')
+  .filter((dirent) => dirent.isDirectory())
   .map((dirent) => dirent.name)
-  .sort((a: any, b: string) => b.localeCompare(a, undefined, { numeric: true }));
+  .sort((a: string, b: string) => {
+    if (!a.match(/^[0-9\.]+$/)) {
+      if (!b.match(/^[0-9\.]+$/)) {
+        return 0;
```
amarthadan, line 17: But this means that each version that is not in a semantic versioning format (something like just text, e.g. `alpha`) is evaluated as the same. I think we should sort those alphabetically as well.

amarthadan (changes requested):

packages/validator/tsconfig.build.json
amarthadan: I wasn't talking about imports within the validator but about some other package importing validator. But in the meantime I was looking more into it as I'm doing the node dockerization and it looks like it shouldn't be a problem after all.

drgy: Yeah I wasnt talking about imports as well, I found that reading the templates might be a problem when I was working on the integration with deployer. I updated validator so that it always resolves paths relative to the current file, so even if it's imported and ran from completely different directory it will be reading the correct directory. I'm not 100% sure if this will work as package, but it is javascript after all so hopefully it wont need any more adjustments, but we can fix this if it becomes a problem in packaging.

packages/validator/src/commands/utils.ts
drgy: Yes this is intentional, as it was discussed versioning will be primarily numerical from now on, validator will be focused on the numerical versions so the latest version will be `0.1.0` until the next numerical version will come out. It will just throw all the alphabetical versions at the very end, so that they are not considered, sorting them alphabetically does not make much sense as for example `beta` would be ranked higher than `pre-alpha`, to correctly sort versions we would have to import a library, which I believe is a bit overkill and focusing only on numerical versions should be sufficient.

amarthadan: I understand and I agree that we will focus on the semantic versioning from now on. However, it makes sense to provide a deterministic ordering of all versions (even though migth be wrong as you mentioned). Listing a directory is one of those things that can be very much OS dependent so the order may not be guaranteed. Forcing the ordering in such a case is just overall less error prone.

amarthadan (approved):

PR #559, opened 2021-09-13
Refactor Request<T> type

acenolaza:
My instinct tells me that we should have a base `Request` inteface with shared properties and then extend this interface in other interfaces like `ApiCall` which will have it's own set of properties. Now I really like what was done by mixing interfaces, types and generics. The problem is that I've now realized that `Withdrawal` interface no longer have any property that is not shared with `ApiCall` and specifying a type as `Request<{}>` looks kind of weird to me.

### DIFF
--- a/packages/node/src/evm/templates/template-application.test.ts
+++ b/packages/node/src/evm/templates/template-application.test.ts
@@ -12,7 +12,7 @@ describe('mergeApiCallsWithTemplates', () => {
 
   it('merges the template into the API call', () => {
     const apiCall = fixtures.requests.buildApiCall({
-      airnodeAddress: 'templateAirnode-0',
+      airnodeAddress: '0xA30CA71Ba54E83127214D3271aEA8F5D6bD4Dace',
       endpointId: null,
       parameters: {},
       templateId: 'templateId-0',
@@ -20,7 +20,7 @@ describe('mergeApiCallsWithTemplates', () => {
 
     const templatesById: { readonly [id: string]: ApiCallTemplate } = {
       'templateId-0': {
-        airnodeAddress: 'templateAirnode-0',
+        airnodeAddress: '0xA30CA71Ba54E83127214D3271aEA8F5D6bD4Dace',
         encodedParameters:
           '0x315375000000000000000000000000000000000000000000000000000000000066726f6d0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000a0616d6f756e74000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003e800000000000000000000000000000000000000000000000000000000000000034554480000000000000000000000000000000000000000000000000000000000',
         endpointId: 'templateEndpointId-0',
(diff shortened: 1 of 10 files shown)


packages/node/src/types.ts
```diff
@@ -85,14 +85,9 @@ export interface ApiCallTemplate {
...
 
-export interface Withdrawal {
-  readonly airnodeAddress: string;
-  readonly sponsorAddress: string;
-}
-
 export interface GroupedRequests {
   readonly apiCalls: Request<ApiCall>[];
-  readonly withdrawals: Request<Withdrawal>[];
+  readonly withdrawals: Request<{}>[];
```
andreogle, line 90: I'm not quite sure I understand this. Why not just make separate `Withdrawal` interface instead? Is this in anticipation of refactoring later?

andreogle (commented):
I don't like the `Request<{}>` either.

Can I defer to @Siegrift for this? He is the one added the TODO comment to the request types and has a much better understanding of TypeScript than me 😅

packages/node/src/types.ts
acenolaza: The only reason that I found for not doing that is because I got a linting error complaining about an empty interface.

acenolaza:
> I don't like the `Request<{}>` either.
> 
> Can I defer to @Siegrift for this? He is the one added the TODO comment to the request types and has a much better understanding of TypeScript than me

That's perfectly fine. I'll add Emo to the reviewers list. Thanks!

Siegrift:
Yeah, @acenolaza suggestion makes sense. Basically have:
```ts
interface BaseRequest {
...
}

// and then
type Withdrawal = BaseRequest
interface ApiCall extends BaseRequest {...}
```

packages/node/src/types.ts
acenolaza: Even though I like [this](https://github.com/api3dao/airnode/pull/559#issuecomment-921163176) idea the changes I had to make were way too many in comparison with the benefits of that approach so I ended up adding an empty `Withdrawal` interface with a `es-lint-disable-next-line`

Siegrift: :/ Mhh, why did it cause so many changes? 

I think is simpler if it is done like in https://github.com/api3dao/airnode/pull/559#issuecomment-921163176

packages/node/src/evm/templates/template-application.test.ts
```diff
@@ -12,7 +12,7 @@ describe('mergeApiCallsWithTemplates', () => {
 
   it('merges the template into the API call', () => {
     const apiCall = fixtures.requests.buildApiCall({
-      airnodeAddress: null,
+      airnodeAddress: 'templateAirnode-0',
```
Siegrift, line 15: seems like a strange address to me

Siegrift (approved):
:+1: Alltogether the changes LGTM, but I am a bit sad that the request couldn't be refactored the suggested way.

packages/node/src/types.ts
acenolaza: Maybe I did something wrong but these types are used all over in the node package. When I finished I had over 25 files modified and there were still some compilation errors so at this point I realized that the change was not worth the trouble.

packages/node/src/evm/templates/template-application.test.ts
acenolaza: Yeah I thought the same but I was trying to be consistent with other values in the same file. We don't really care in this tests if it's a valid address. All we care about is that it is not null but I'll change it to an address just because it makes more sense.

acenolaza:
>  Alltogether the changes LGTM, but I am a bit sad that the request couldn't be refactored the suggested way.

Right, me too at first but then I realized that the main benefit of the refactor would be to not have the `eslint-disable-next-line` on the empty interface which in the future might even go away. 
Also this `export interface Withdrawal {}` started looking no so different than this `type WithdrawalRequest = Request`
And one last thing I realized is that by loosing the generics I was forced to do type casting on functions that for instance returned an object of type `Request`
I will keep thinking about this and try to refactor it again in the next releases.

PR #562, opened 2021-09-14
A3P-23: Replace airnode with sponsor when generating requestId hash

acenolaza:
A [decision](https://api3workspace.slack.com/archives/C027Y2FAV0S/p1631374192018100) was made to ignore ARO-02 (add `airnodeAddress` to `requestId` hash) in order to be consistent with upcoming pub-sub protocol when generating a new hash for `requestId` but we will be adding the `sponsorAddress` to this hash.


bbenligiray (approved):
Looks good, merging

PR #563, opened 2021-09-15
An176 airnode to sponsor in requestid

acenolaza:
Having the `airnode` address in the `requestId` will cause an extra verification in node package code which is unnecessary and redundant since we are already checking `airnode` address on-chain when fulfilling the request so a decision was made to not add it and instead add the `sponsor` address to the `requestId` hash to mitigate a possible attack described [here](https://api3workspace.slack.com/archives/C027Y2FAV0S/p1631374683018300).


andreogle (approved):

Siegrift (approved):
LGTM

PR #566, opened 2021-09-16
Docker container for Airnode containerized version

amarthadan:
Issues [AN-196](https://api3dao.atlassian.net/browse/AN-196) and [AN-197](https://api3dao.atlassian.net/browse/AN-197)

It took a while but I'm quite happy with the result. Main changes are that you can now build the Docker containers (both Airnode and Deployer) without the need for packages being published. I'm also using one common artifact image so the build doesn't have to happen multiple times.

I added a convenient Yarn targets for container building so all you have to do is just run `yarn docker`.

Please, try building the images and let me know if something's not working for you.

I highly recommend using the new [Docker BuildKit](https://docs.docker.com/develop/develop-images/build_enhancements/), it's much faster.


package.json
```diff
@@ -36,6 +36,10 @@
     "dev:invoke": "(cd packages/node && yarn run dev:invoke)",
     "dev:list": "(cd packages/operation && yarn run dev:list)",
     "dev:stop": "(cd packages/operation && yarn run dev:stop)",
+    "docker": "yarn docker:artifacts && yarn docker:deployer && yarn docker:node",
+    "docker:artifacts": "docker rmi -f api3/artifacts && docker build -t api3/artifacts:latest -f docker/Dockerfile .",
```
Siegrift, line 40: Nit: Maybe use full args instead of shorthand's?

docker/README.md
```diff
@@ -0,0 +1,35 @@
+This is a Docker container for building all the packages in the monorepo and providing their build artifacts. Other package specific containers can use this one to obtain files necessary for their run without any additional dependencies.
```
Siegrift, line 1: any for their run without **building or installing** any additional dependencies.

docker/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+# Build
```
Siegrift, line 3: Maybe add a single hashtag title and use ## as subheaders for the other sections?

docker/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Local
+You can build the container image from your current workspace by providing a build type `local` (this is the default)
+```bash
+docker build --build-arg build=local -f docker/Dockerfile -t api3/artifacts:latest .
+# or with no build argument
+```
+
+## Git
+You can build the container image from the Git repository and you can also provide a branch from which the build should be done
```
Siegrift, line 16: you can also provide a branch **commit or tag**

docker/README.md
```diff
@@ -0,0 +1,35 @@
...
+docker build --build-arg build=git --build-arg branch=my-branch -f docker/Dockerfile -t api3/artifacts:latest .
+# the default branch is master
+```
+
+# Usage
+Artifacts Docker image is not suitable for a direct usage. It's a building step for other containers.
+
+Once built the image contains 3 directories with artifacts that can be used for building other images:
+* `/build` - contains copy of the monorepo with all the packages build, including all the dev dependencies
+* `/packages` - contains directories representing unpack packages of all the packages in the monorepo
```
Siegrift, line 27: what are those "unpack packages"?

packages/node/docker/README.md
```diff
@@ -0,0 +1,42 @@
...
+└── config
+    ├── config.json
+    └── secrets.env
+$ docker run -v $(pwd)/config:/app/config ...
+```
+
+# Usage
+Example directory structure and commands for running the Airnode Docker container. The below commands are run from the depicted directory.
+
+> If you are using Windows, use CMD (and not PowerShell), replace `\` with `^` and `$(pwd)` with `%cd%`.
```
Siegrift, line 26: Why can't I use PowerShell?

Siegrift (commented):
Code and documentation looks great! I will try this out tomorrow morning.

packages/node/bin/invoke.ts
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env node
```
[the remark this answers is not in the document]
amarthadan: It will and no I'm not using it with `ts-node`. At least not for the container. Yes, we use it like that if you want to run it localy but in production (or in container in our case) you want to run the transpiled version with `node`.

package.json
```diff
@@ -40,7 +40,8 @@
     "lint:eslint": "eslint . --ext .js,.ts",
     "lint:eslint:fix": "eslint . --ext .js,.ts --fix",
     "lint:solhint": "solhint ./**/*.sol",
-    "postinstall": "husky install && yarn run build",
```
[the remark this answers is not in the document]
Ashar2shahid: yup I agree adding it to the root readme

packages/node/docker/README.md
amarthadan: Don't actually know. I don't have a Windows machine, I was just relying on a wisdom of the person who did the dockerization befor me and added this line :smile:

packages/node/package.json
```diff
@@ -2,17 +2,21 @@
...
   "version": "0.1.0",
   "license": "MIT",
-  "main": "dist/index",
-  "types": "dist/index",
+  "bin": {
+    "invoke": "./dist/bin/invoke.js",
+    "test-api": "./dist/bin/test-api.js"
+  },
+  "main": "dist/src/index",
+  "types": "dist/src/index",
```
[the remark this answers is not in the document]
amarthadan: I'll look into it as part of [AN-195](https://api3dao.atlassian.net/browse/AN-195)

package.json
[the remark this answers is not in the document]
amarthadan: When you run `yarn bootsrtap` the build is done twice that's why I removed it.
But yes, I will add that to the README.

packages/node/docker/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+    yarn run pack
+
+# Run image
+FROM environment
+
+ENV airnodeBin="/usr/local/bin/${name}" \
+    cronjob="/etc/cron.d/${name}"
+
+LABEL application=${name} \
+    description="Arinode"
```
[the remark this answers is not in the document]
amarthadan: Yeah, I was thinking about that. I wanted to keep it consistent with other multiline commands and indent it 4 spaces. But I guess aligning it with the command keyword would be better.

packages/node/docker/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+
+COPY --from=deps ${airnodeDir}/node_modules ${airnodeDir}/node_modules
+COPY --from=build ${airnodeDir}/packages/*/*.tgz ${airnodeDir}/node_modules/@api3/
+COPY --from=build ${airnodeDir}/packages/node/dist ${airnodeDir}/
+COPY packages/node/docker/airnode-crontab ${cronjob}
+
+    # Install Tini to pass signals correctly
```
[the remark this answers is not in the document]
amarthadan: This is actually on purpose. I wanted the comment to be command running, not to the `RUN` command itself. Same as the comments below.

packages/node/docker/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+# Production dependencies
+FROM preparation AS deps
+
+RUN yarn global add lerna && \
+    lerna bootstrap -- --production --no-optional --ignore-scripts
+
+# Build
+FROM preparation AS build
+
+RUN yarn bootstrap && \
```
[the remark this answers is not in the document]
amarthadan: I know, it's a bit confusing :slightly_smiling_face:

Imagine that each section from `FROM` till the next `FROM` (or till the end of file) is a separate image. You can name them with `AS` and then copy between them with `COPY --from=<name>`. If you want to learn more you can read [the documentation on multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/).

What I'm doing is that I'm using one container with all the dependencies to do the build to get the transpiled JS code and another one for installing just the runtime dependencies and I'm mixing the results (that's a simplyfied TL;DR version :smile: )

packages/node/docker/README.md
Siegrift: I could possibly try this on windows if there is need for this (I have dual boot) - but I don't have docker installed there so it would be better if someone working on windows already would check this.

packages/node/.gitignore
```diff
@@ -9,5 +9,3 @@ coverage/
 # Config files
 secrets.env
 config.json
-**/secrets.env
```
Siegrift: Maybe this could be fixed for all .gitignores in airnode monorepo? (in a separate PR?)

packages/deployer/docker/Dockerfile
```diff
@@ -6,14 +6,14 @@ ENV name="airnode-deployer" \
...
-    description="Arinode Deployer CLI"
+      description="Arinode Deployer CLI"
 
 COPY --from=api3/artifacts /dependencies ${appDir}/node_modules
 COPY --from=api3/artifacts /packages ${appDir}/node_modules/@api3/
 COPY --from=api3/artifacts /build/packages/deployer/dist ${appDir}/
 COPY packages/deployer/docker/entrypoint.sh /entrypoint.sh
 
-# Install external dependencies
+    # Install external dependencies
```
Siegrift, line 16: I find this formatting very strange tbh :D

Siegrift:
I've tried the node container and it worked for me (although I only tested with incorrect config.json).
The airnode started correctly and I saw correct logs after running  `docker logs <id> -f`

Siegrift (approved):
:1st_place_medal: LGTM. Although I would mention how to get the logs as well in the README.

I'll test this in depth while doing https://api3dao.atlassian.net/browse/AN-169

packages/adapter/package.json
```diff
@@ -9,8 +9,9 @@
   ],
   "scripts": {
     "build": "yarn run clean && yarn run compile",
-    "clean": "rimraf -rf ./dist",
+    "clean": "rimraf -rf ./dist *.tgz",
```
Siegrift, line 12: why not clean the whole dist?

Siegrift (approved):

packages/adapter/package.json
amarthadan: I am cleaning the whole `dist`. The `*.tgz` part is for cleaning the packages made with `yarn pack`.

Siegrift: I would argue that yestedray it looked like `dist/*.tgz` but I probably should just buy glasses instead...

PR #567, opened 2021-09-16
AN-208: Fix node e2e tests in beta-protocol-revision after merging be…

acenolaza:
…ta-protocol changes

Some node e2e tests are expecting a fixed set of events to be emitted and since the protocol is now emitting an extra `SetSponsorshiptStatus` event then they break. I refactored those tests to only expect only on the relevant events for the specific test.


Siegrift (approved):
:+1:

PR #568, opened 2021-09-17
Merge beta protocol

Siegrift:
https://api3dao.atlassian.net/browse/AN-173

@acenolaza helped me debug some problems - the biggest one is that I forgot to apply https://github.com/api3dao/airnode/pull/563

I've applied his commits as well, but it's hard to review :/ 
Also, some commits seem to be duplicated.

Shall I squash merge this PR?


packages/node/test/e2e/requester-fulfill.feature.ts
```diff
@@ -57,6 +45,7 @@ it('should call fail function on AirnodeRrp contract and emit FailedRequest if r
 
   const preInvokelogNames = await fetchAllLogNames(provider, deployment.contracts.AirnodeRrp);
   expect(preInvokelogNames).toEqual([
+    'SetSponsorshipStatus', // RrpRequester constructor
```
acenolaza, line 48: doh! in this [PR](https://github.com/api3dao/airnode/pull/567) I missed to replace this fixed array with `expect.arrayContaining(postInvokeExpectedLogs)` 🤦🏻‍♂️ 

Since I added the missing `SetSponsorshipStatus` event the test is passing anyways but we should change it to be consistent with the rest of the code I think.

packages/node/test/e2e/requester-fulfill.feature.ts
```diff
@@ -68,6 +57,7 @@ it('should call fail function on AirnodeRrp contract and emit FailedRequest if r
 
   const postInvokeLogs = await fetchAllLogs(provider, deployment.contracts.AirnodeRrp);
   expect(postInvokeLogs.map(({ name }) => name)).toEqual([
+    'SetSponsorshipStatus', // RrpRequester constructor
```
acenolaza, line 60: Same here, this should be changed with `expect.arrayContaining(postInvokeExpectedLogs)`

packages/node/test/setup/e2e/testing.ts
```diff
@@ -4,7 +4,7 @@ import { buildChainConfig, buildProvider, getDeployerIndex } from './utils';
 import { deployAirnodeRrp, makeRequests } from './deployment';
 import { buildConfig, operation } from '../../fixtures';
 
-export const increaseTestTimeout = (timeoutMs = 45_000) => jest.setTimeout(timeoutMs);
+export const increaseTestTimeout = (timeoutMs = 60_000) => jest.setTimeout(timeoutMs);
```
acenolaza, line 7: Did I commit this by accident or this is something we want to update?

acenolaza (approved):
Changes I've made look good but I just realized I missed to update something in 1 e2e test. Should we do that in this branch or should I update it in the source PR which is still open https://github.com/api3dao/airnode/pull/567?

packages/node/test/e2e/requester-fulfill.feature.ts
Siegrift: Good point. I noticed this while fixing the test but then I forgot about it.

I will create a new PR for this.

packages/node/test/setup/e2e/testing.ts
Siegrift: Explicit choice by me: https://github.com/api3dao/airnode/commit/3ba5ae7e842d98c3bbd46485d296cbc6643a6be6

I am not sure if my machine is just slow, but it might be because of the recent changes? (The fact is it doesn't pass locally until I increase the timeout)

packages/node/test/e2e/requester-fulfill.feature.ts
acenolaza: I have already changed it in the other branch if you like I'll push them now

acenolaza: https://github.com/api3dao/airnode/pull/567/commits/cd501b2a963de1f7c74d0238e5b59fae14c4b6a7

PR #569, opened 2021-09-18
[WIP,BLOCKED]: Call fulfill with empty bytes string if API error code is not zero

Siegrift:
https://api3dao.atlassian.net/jira/software/projects/AN/boards/18?selectedIssue=AN-143

### DIFF
--- a/packages/node/src/evm/fulfillments/api-calls.ts
+++ b/packages/node/src/evm/fulfillments/api-calls.ts
@@ -7,6 +7,23 @@ import { DEFAULT_RETRY_TIMEOUT_MS } from '../../constants';
 import { ApiCall, Request, LogsErrorData, RequestErrorCode, RequestStatus, TransactionOptions } from '../../types';
 import { AirnodeRrp } from '../contracts';
 
+/**
+ * When the API fails to respond, we still need to provide a response data. We decided to return "0x", because decoding
+ * this value will cause error for consumers (causing transaction revert). This is good, because people often forget to
+ * check the error code on the airnode response.
+ *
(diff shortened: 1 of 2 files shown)


packages/node/src/evm/fulfillments/api-calls.ts
```diff
@@ -7,6 +7,23 @@ import { DEFAULT_RETRY_TIMEOUT_MS } from '../../constants';
...
+ * check the error code on the airnode response.
+ *
+ * See: https://github.com/api3dao/airnode/issues/416
+ */
+export const API_ERROR_RESPONSE_VALUE = '0x';
+
+const getFulfillmentValue = (request: Request<ApiCall>) => {
+  const statusCode = requests.getErrorCode(request);
+  // NOTE: Allow empty string as responseValue
+  if (statusCode !== 0 || request.responseValue === undefined) return API_ERROR_RESPONSE_VALUE;
```
andreogle, line 22: Do we maybe want to check for `null` as well here? Can use `isNil` from lodash
 
```suggestion
  if (statusCode !== 0 || isNil(request.responseValue)) return API_ERROR_RESPONSE_VALUE;
```

andreogle (approved):

packages/node/src/evm/fulfillments/api-calls.ts
Siegrift: It actually can't be `null` (at least that's what TS is telling me :) )

bbenligiray:
This became obsolete with https://github.com/api3dao/airnode/pull/577, closing

PR #574, opened 2021-09-22
Update packaging for all packages

amarthadan:
Issue [AN-195](https://api3dao.atlassian.net/browse/AN-195)

Took a bit longer than I expected but I tested:
* packaging of all packages
* publishing to registry (even though local private one)
* including all packages as dependencies (both code and types)
* running all packages that has binaries from installed packages
* building and testing once again Docker images after the changes


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "api3",
-  "version": "0.0.1",
+  "version": "0.1.0",
```
andreogle, line 3: Is there any way we could manage the version from a single place like `lerna.json`? Right now, there are loads of places with 0.1.0 as the version and it's going to be easy to miss updating one

andreogle (approved):

packages/deployer/src/infrastructure/index.ts
```diff
@@ -99,7 +99,7 @@ async function deploy(
   await runCommand(command, options);
 
   const httpGatewayApiKeyVar = httpGatewayApiKey ? `-var='api_key=${httpGatewayApiKey}'` : '';
-  command = `terraform apply -var="aws_region=${region}" -var="airnode_id_short=${airnodeAddressShort}" -var="stage=${stage}" -var="configuration_file=${path.resolve(
+  command = `terraform apply -var="aws_region=${region}" -var="airnode_address_short=${airnodeAddressShort}" -var="stage=${stage}" -var="configuration_file=${path.resolve(
```
acenolaza, line 102: 😮 it seems we missed this while merging `beta-protocol-revision`

package.json
amarthadan: I don't think there is but it's not really a problem. The publishing will be done with Lerna and it will take care of updating all the versions for us :slightly_smiling_face:

packages/deployer/src/infrastructure/index.ts
Siegrift: Yeah, but this might be just this file was not touched in the rebase (maybe it was renamed or whatever) and I didn't have merge conflict with it. 

Anyway, these were the only changes that I needed to do to make airnode starter work - so I am still confident about the rebase :)

Siegrift (approved):
:1st_place_medal: LGTM.

I didn't test manually, because I intend to test this heavily while working on airnode starter.

I'll report any issues and we can fix them separately :)

packages/admin/package.json
```diff
@@ -2,7 +2,9 @@
   "name": "@api3/admin",
   "license": "MIT",
   "version": "0.1.0",
-  "bin": "./dist/bin/admin.js",
+  "bin": {
+    "api3-admin": "./dist/bin/admin.js"
```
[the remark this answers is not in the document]
andreogle: I like this idea

PR #576, opened 2021-09-24
Merge beta-protocol branch to master

acenolaza:
Rebased `beta-protocol` on top of master and this branch is ahead by 7 commits.

I'm assuming that Certik will no longer be using this branch or we will create a new one out of master if they do another follow up review.


bbenligiray (approved):
Thanks :rocket:

PR #577, opened 2021-09-26
Replace status codes with error messages

bbenligiray:


packages/protocol/test/rrp/AirnodeRrp.sol.js
```diff
@@ -442,32 +457,59 @@ describe('fulfill', function () {
...
+              .callStatic.fulfill(
+                requestId,
+                airnodeAddress,
+                fulfillData,
+                rrpRequester.address,
+                rrpRequester.interface.getSighash('fulfillAlwaysReverts'),
+                { gasLimit: 500000 }
+              );
+            expect(staticCallResult.callSuccess).to.equal(false);
+            expect(utils.decodeRevertString(staticCallResult.callData)).to.equal('Always reverts');
```
bbenligiray, line 479: This is how we retrieve the revert string to use in `errorMessage` for the `fail()` call

packages/protocol/test/rrp/requesters/RrpRequester.sol.js
```diff
@@ -67,12 +67,13 @@ describe('onlyAirnodeRrp', function () {
...
       const requestId = hre.ethers.utils.keccak256(
         hre.ethers.utils.solidityPack(
-          ['uint256', 'uint256', 'address', 'bytes32', 'bytes'],
+          ['uint256', 'uint256', 'address', 'bytes32', 'address', 'bytes'],
           [
             (await airnodeRrp.requesterToRequestCountPlusOne(rrpRequester.address)).sub(1),
             (await hre.ethers.provider.getNetwork()).chainId,
             rrpRequester.address,
             templateId,
+            roles.sponsor.address,
```
bbenligiray, line 76: This was skipped

packages/protocol/test/utils.js
```diff
@@ -48,4 +48,14 @@ module.exports = {
     const currentBlock = await provider.getBlock(currentBlockNumber);
     return currentBlock.timestamp;
   },
+  decodeRevertString: (callData) => {
+    // Refer to https://ethereum.stackexchange.com/a/83577
+    // Assuming `callData` is a hex string
+    if (callData.length < 2 + 68 * 2) {
```
acenolaza: This seems a little bit different than what was checked in the code from the link above. @bbenligiray you mind explaining what 2 + 68 * 2 means here?

packages/protocol/test/utils.js
```diff
@@ -48,4 +48,14 @@ module.exports = {
...
+      // callData is exactly "0x" when there is no revert string but let's do what everybody else does
+      return 'No revert string';
+    }
+    // We probably don't want the signature either
+    return ethers.utils.defaultAbiCoder.decode(['string'], `0x${callData.substring(2 + 4 * 2)}`)[0];
```
acenolaza: Here I know we are trimming out the sighash but what do these values represent 2 + 4 * 2 when used as a starting point in the substring call?

acenolaza (approved):
LGTM

I just asked some questions for me to better understand the function that decodes the revert string since I'll be probably using that same code in the node package.

packages/protocol/contracts/rrp/AirnodeRrp.sol
```diff
@@ -316,8 +317,25 @@ contract AirnodeRrp is
         )
     {
         delete requestIdToFulfillmentParameters[requestId];
-        // Failure is recorded so that it can be checked externally
-        requestWithIdHasFailed[requestId] = true;
-        emit FailedRequest(airnode, requestId);
+        emit FailedRequest(airnode, requestId, errorMessage);
```
Ashar2shahid, line 320: The only problem with doing that could be that the errorMessage is string and data is of type bytes

Ashar2shahid (approved):
Everything Looks good to me, just some comments

packages/protocol/test/utils.js
bbenligiray: I think the idea in the link is that if the call reverted with a string, `callData` will be the signature of the function that reverted (not 100% on this but it doesn't matter) and the revert string. The signature is 4 bytes, the string is minimum 64 bytes. So `callData` should be at least 68 bytes. I'm multiplying by 2 here because we're operating on a hex string (where 2 hex characters is a byte) and the added 2 is the `0x`.

This is outdated though, check out the new implementation.

packages/protocol/test/utils.js
bbenligiray: Correct

packages/protocol/README.md
```diff
@@ -3,11 +3,6 @@
 
 > The contracts that implement the Airnode protocols
 
-***This documents the Beta version of the protocol.
-We have also published and documented the pre-alpha version widely.
-The pre-alpha and the Beta versions are very different in implementation and the terminology they use is contradictory.
-If you are referring to any outside source, make sure that it is not referring to the pre-alpha version, or at least interpret it accordingly.***
-
```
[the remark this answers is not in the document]
bbenligiray: We need to remove it at some point before release. It may be a bit premature though.

packages/protocol/contracts/rrp/AirnodeRrp.sol
```diff
@@ -280,31 +276,36 @@ contract AirnodeRrp is
...
+        if (callSuccess) {
+            emit FulfilledRequest(airnode, requestId, data);
+        } else {
+            // We do not bubble up the revert string from `callData`
+            emit FailedRequest(
+                airnode,
+                requestId,
+                "Fulfillment failed unexpectedly"
+            );
+        }
```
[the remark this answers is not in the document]
bbenligiray: Semantically they are different "events" though (one does the callback and the other doesn't) so it makes sense for them to be different

bbenligiray:
Updated PR description

bbenligiray:
@acenolaza I think you can continue working on this branch, what do you think?

packages/protocol/README.md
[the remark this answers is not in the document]
bbenligiray: Restored this

acenolaza:
> @acenolaza I think you can continue working on this branch, what do you think?

Sounds good. I have already started working on this branch since yesterday in the evening. Today I plan on updating the tests to see if everything works as expected.

packages/node/src/evm/contracts/airnodeRrp.ts
```diff
@@ -8,8 +8,8 @@ const MadeFullRequest = ethers.utils.id(
   'MadeFullRequest(address,bytes32,uint256,uint256,address,bytes32,address,address,address,bytes4,bytes)'
 );
 
-const FulfilledRequest = ethers.utils.id('FulfilledRequest(address,bytes32,uint256,bytes)');
-const FailedRequest = ethers.utils.id('FailedRequest(address,bytes32)');
+const FulfilledRequest = ethers.utils.id('FulfilledRequest(address,bytes32,bytes)');
```
andreogle, line 11: Not in scope here, but I wonder if these could be generated from importing the contract ABI. i.e. something like:

```ts
// Not sure what the exact ABI import looks like
// Assuming events have the structure of "Name(...types)"
const eventTopics = airnodeRrpAbi.events.map((e) => {
  const name = e.name.split('(')[0];
  const topic = ethers.utils.id(e);
  return { name, topic };
});

export function getTopics() {
  // Build the same topic object
  return eventTopics.reduce((acc, event) => {
    return { ...acc, [event.name]: event.topic };
  }, {});
}
```

Then there wouldn't be any need to maintain the topics here

acenolaza: That's a really good idea but do you still see this as something needed after we deploy the contracts to mainnet? 
For newer versions of the protocol maybe 🤔

acenolaza:
@amarthadan do you mind reviewing [this](https://github.com/api3dao/airnode/pull/577/commits/7fb62ef0044790b82e9184b5ae559bf6d724e627) commit? The idea is that Airnode will not check for on-chain params anymore so I did some clean up in the deployer package. Also the Airnode xpub must be for this path: `m/44'/60'/0'`

packages/node/src/evm/contracts/airnodeRrp.ts
andreogle: Probably not and it might actually be an over-optimization anyway. I just see these getting updated all the time, but they should probably settle in later versions. Ignore me 😁

amarthadan:
> @amarthadan do you mind reviewing [this](https://github.com/api3dao/airnode/pull/577/commits/7fb62ef0044790b82e9184b5ae559bf6d724e627) commit? The idea is that Airnode will not check for on-chain params anymore so I did some clean up in the deployer package. Also the Airnode xpub must be for this path: `m/44'/60'/0'`

LGTM :+1: Make sure to update the `airnodeXpub` change in the [`receipt.json` documentation](https://docs.api3.org/airnode/next/reference/deployment-files/receipt-json.html)

packages/protocol/contracts/rrp/AirnodeRrp.sol
```diff
@@ -264,60 +223,92 @@ contract AirnodeRrp is
...
         );
-        require(callSuccess, "Fulfillment failed");
+        if (callSuccess) {
+            emit FulfilledRequest(airnode, requestId, data);
+        } else {
+            // We do not bubble up the revert string from `callData`
+            emit FailedRequest(
+                airnode,
+                requestId,
+                "Fulfillment failed unexpectedly"
```
Siegrift, line 260: Shouldn't this be more descriptive? Like `Fulfillment callback failed unexpectedly.`?

packages/admin/README.md
```diff
@@ -155,30 +154,6 @@ npx @api3/airnode-admin check-withdrawal-request \
 
 ## Airnode commands
 
-### `set-airnode-xpub`
```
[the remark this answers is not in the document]
bbenligiray: 1. Majority of the API providers won't use it, it's too expensive.
2. It's not possible to verify that the announced xpub is correct on-chain. The value is going to have to be verified off-chain so may as well announce it off-chain along with `airnode` (like in https://qrng.anu.edu.au/contact/how-to-use-qrng-on-web3/)
3. We can always curate a separate on-chain registry, there is no need for it to be on this contract
4. In theory, an Airnode can utilize multiple HD wallets with different xpubs with a recent update so this is semantically misleading too

packages/protocol/contracts/rrp/AirnodeRrp.sol
Siegrift: Probably. My reasoning was to stress that the failure is in the requester callback and not our implementation.

bbenligiray: Actually it should probably say `"Fulfillment reverted unexpectedly"`, then it would be obvious that the failure happened in the fulfillment transaction. I'm not a fan of using "callback" because that is not used in the protocol context to refer to fulfillments.

packages/node/src/evm/fulfillments/api-calls.ts
```diff
@@ -113,22 +118,39 @@ async function testAndSubmitFulfill(
...
   options: TransactionOptions
 ): Promise<LogsErrorData<SubmitResponse>> {
+  const errorMessage = requests.getErrorMessage(request);
+  if (errorMessage) {
+    const [submitLogs, submitErr, submitData] = await submitFail(airnodeRrp, request, errorMessage, options);
+    return [submitLogs, submitErr, submitData];
+  }
+
+  // Compute signature once.
+  // Not computing this beforehand because it might be a very computational expensive operation
```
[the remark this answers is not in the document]
acenolaza: Haven't really checked but AFAIK lambdas run in a very small instance on AWS. The main idea was to prevent any issues with multiple requests being fulfilled simultaneously. Do you think we should pre-compute this signature for all requests during provider initialization no matter if the fulfill function is going to be called?

packages/node/src/evm/fulfillments/api-calls.ts
```diff
@@ -4,15 +4,32 @@ import { go } from '../../utils/promise-utils';
...
+type ApiCallWithSignature = ApiCall & { readonly signature: string };
+
+async function signResponseMessage(request: Request<ApiCall>, masterHDNode: ethers.utils.HDNode) {
+  const airnodeWallet = ethers.Wallet.fromMnemonic(masterHDNode.mnemonic!.phrase);
+  return await airnodeWallet.signMessage(
+    ethers.utils.arrayify(
+      ethers.utils.keccak256(
+        ethers.utils.solidityPack(
+          ['bytes32', 'bytes'],
+          [request.id, request.responseValue || ethers.constants.HashZero]
```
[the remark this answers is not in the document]
acenolaza: Right, if we reach this far then it means that `responseValue` is not `undefined`, `null` or empty. I guess this is here just because `responseValue` is set as optional in `ApiCall` [interface](https://github.com/api3dao/airnode/blob/e2e6009c8a257566a05119b34eaa2bf38ef5b9ac/packages/node/src/types.ts#L92) and typescript will complain if we call `fulfill` with a possibly `undefined` arg.

It would be cool if we could enforce some sort of XOR between `responseValue` and `errorMessage` since one of those properties must always have a value.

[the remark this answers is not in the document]
Siegrift: That is not easy to type. You essentially need to have two versions of APIResponse:
1) ApiSuccess
2) ApiFail

both extend the ApiBase and and in code you differentiate between the two.. Not sure, but it's probably out of scope to type it like this (and also not worth it)...

That said, if the value is defined for sure, just use `!` as that conveys this meaning...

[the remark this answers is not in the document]
Siegrift: But only the api call lambda, which is probably fine. I think if we fulfilled the transaction with invalid data the consquences could be bigger.

acenolaza:
@bbenligiray rebased on top of master and I think this PR should be ready to be merged

PR #578, opened 2021-09-27
Remove typechain hacks and web3 artifact exports

Siegrift:
I received an email that TypeChain fixed the issues which needed the hacks (they now export the events).

https://api3dao.atlassian.net/browse/AN-156


packages/protocol/src/index.ts
```diff
@@ -27,13 +25,19 @@ const AirnodeRequesterRrpAuthorizerAddresses: { [chainId: number]: string } = {
...
 };
-// TODO:
 const authorizers = {
   AirnodeRequesterRrpAuthorizerFactory,
 };
 
-// NOTE: For now, we only want to expose AirnodeRrp contract (and it's mock)
 export { AirnodeRrpAddresses, AirnodeRequesterRrpAuthorizerAddresses, AirnodeRrpFactory, mocks, authorizers };
-// Export the artifacts for API consumers using web3
-export { AirnodeRrpArtifact, MockRrpRequesterArtifact };
```
Siegrift, line 38: We do not need to export the artifacts, because the ABI is available on the generated contracts from TypeChain...

I don't think end users will depend on protocol package - so I do not think we need to export these anymore.

packages/node/src/evm/requests/event-logs.ts
```diff
@@ -28,12 +26,12 @@ interface GroupedLogs {
   readonly withdrawals: (EVMFulfilledWithdrawalLog | EVMRequestedWithdrawalLog)[];
 }
 
-export function parseAirnodeRrpLog<T extends keyof AirnodeRrpFilters>(
-  log: ethers.providers.Log
-): AirnodeLogDescription<AirnodeRrpLog<T>> {
-  const airnodeRrpInterface = new ethers.utils.Interface(AirnodeRrpArtifact.abi);
+// NOTE: The generic parameter have a better default value (unknown instead of any) but doing so would make the tests
```
andreogle, line 29:
```suggestion
// NOTE: The generic parameter has a better default value (unknown instead of any) but doing so would make the tests
```

andreogle (approved):
Nice one

packages/node/src/evm/requests/event-logs.ts
Siegrift: It should be `could have`

PR #579, opened 2021-09-28
AN-229: Rename Admin CLI command derive-sponsor-wallet-address

acenolaza:
More details: https://api3dao.atlassian.net/browse/AN-229

### DIFF
--- a/packages/admin/README.md
+++ b/packages/admin/README.md
@@ -61,12 +61,12 @@ Please, refer to the implementation for more details.
 
 ## Developer commands
 
-### `derive-sponsor-wallet-address`
+### `derive-sponsor-wallet`
 
-Derives the address of the wallet designated by an Airnode for a requester. You need to specify the xpub but if it is not provided then this command will try to fetch it from the AirnodeRrp contract.
+Derives a sponsor wallet designated by an Airnode for a sponsor and returns the address of the wallet. You need to specify the xpub but if it is not provided then this command will try to fetch it from the AirnodeRrp contract.
 
 ```sh
-npx @api3/airnode-admin derive-sponsor-wallet-address \
+npx @api3/airnode-admin derive-sponsor-wallet \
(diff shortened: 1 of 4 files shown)


packages/admin/README.md
```diff
@@ -61,12 +61,12 @@ Please, refer to the implementation for more details.
 
 ## Developer commands
 
-### `derive-sponsor-wallet-address`
+### `derive-sponsor-wallet`
 
-Derives the address of the wallet designated by an Airnode for a requester. You need to specify the xpub but if it is not provided then this command will try to fetch it from the AirnodeRrp contract.
+Derives a sponsorWallet designated by an Airnode for a sponsor and returns the address of the wallet. You need to specify the xpub but if it is not provided then this command will try to fetch it from the AirnodeRrp contract.
```
Siegrift, line 66: I am not a fan of `sponsorWallet` in the README description. Why not just `sponsor wallet`?

Siegrift (approved):

Siegrift (commented):
On the second thought, I am not sure about this change - I think it's more consistent with other commands with the `address` suffix.

The issues says:
```
The command `derive-sponsor-wallet` creates a sponsorWallet and returns its public address.
```
but that is confusing since this command doesn't **create** the wallet, it just derives the wallet address (public key)

acenolaza:
> On the second thought, I am not sure about this change - I think it's more consistent with other commands with the `address` suffix.
> 
> The issues says:
> 
> ```
> The command `derive-sponsor-wallet` creates a sponsorWallet and returns its public address.
> ```
> 
> but that is confusing since this command doesn't **create** the wallet, it just derives the wallet address (public key)

@bbenligiray would be the best person to decide between the 2 options at this point.

andreogle (approved):

bbenligiray:
Yeah I agree that it should go `derive-sponsor-wallet-address` because it's a bit misleading currently

PR #580, opened 2021-09-28
Error message node

acenolaza:
Airnode should omit the statusCode when calling fulfill() (happy path)
Airnode should call fail() instead of fulfill() with an error string when api call request has non-zero error code 
Airnode should decode the callData and use the revert string to build the error message and call fail() with it If the static-call’ed fulfill() returned false


bbenligiray:
Reviewers can also refer to https://github.com/api3dao/airnode/pull/577#issue-1007411484

packages/node/src/evm/fulfillments/api-calls.ts
```diff
@@ -151,17 +153,25 @@ async function testAndSubmitFulfill(
 async function submitFail(
   airnodeRrp: AirnodeRrp,
   request: Request<ApiCall>,
-  options: TransactionOptions
+  options: TransactionOptions,
+  errorMessage: string
```
andreogle, line 157: Nit: Maybe put options as the last argument?

andreogle (approved):

PR #581, opened 2021-09-30
Add executable flag to CLIs

Siegrift:
https://api3dao.atlassian.net/browse/AN-230

Cause of the issue: https://github.com/microsoft/TypeScript/issues/37583

I think we have only 4 CLI tools as of now - unfortunately we will have to think about this issue when adding new CLI tools.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   },
   "scripts": {
     "bootstrap": "yarn install && lerna bootstrap",
-    "build": "lerna run build --stream",
+    "build": "lerna run build --stream && ts-node scripts/make-cli-executable.ts",
     "build:adapter": "(cd packages/adapter && yarn run build)",
     "build:admin": "(cd packages/admin && yarn run build)",
     "build:airnode-abi": "(cd packages/airnode-abi && yarn run build)",
--- a/scripts/make-cli-executable.ts
+++ b/scripts/make-cli-executable.ts
@@ -1,0 +1,8 @@
+import { chmodSync } from 'fs';
+
+// See: https://github.com/microsoft/TypeScript/issues/37583
+const addExecuteRightsMode = '755'; // default is 664
+chmodSync('packages/admin/dist/bin/admin.js', addExecuteRightsMode);
+chmodSync('packages/deployer/dist/bin/deployer.js', addExecuteRightsMode);
+chmodSync('packages/validator/dist/bin/convertor.js', addExecuteRightsMode);
+chmodSync('packages/validator/dist/bin/validator.js', addExecuteRightsMode);


packages/deployer/package.json
```diff
@@ -10,12 +10,13 @@
...
   ],
   "scripts": {
-    "build": "yarn clean && yarn compile && yarn webpack && yarn copy:webpack && yarn copy:terraform",
+    "build": "yarn clean && yarn compile && yarn make-cli-executable && yarn webpack && yarn copy:webpack && yarn copy:terraform",
     "clean": "rimraf -rf ./dist *.tgz",
     "cli": "ts-node bin/deployer.ts",
     "compile": "tsc -p tsconfig.build.json",
     "copy:terraform": "copyfiles terraform/**/**/*.tf terraform/**/**/*.tpl  dist/",
     "copy:webpack": "copyfiles .webpack/**/**/*.js dist/",
+    "make-cli-executable": "chmod +x dist/bin/deployer.js",
```
andreogle, line 19: Do we want/need this to work on Windows? I'm not sure if building is possible now on Windows though, but this will almost certainly break it (?)

amarthadan: Yes, this will most definitely break build on Windows :sweat_smile: Maybe we can use Node's [`fs.chmod`](https://nodejs.org/api/fs.html#fs_fs_chmod_path_mode_callback) instead. I guess they have that covered.

Siegrift: Yeah, good point. 

But maybe we could drop the windows support and recommend people to use WSL2?

dev-scripts/make-cli-executable.ts
```diff
@@ -0,0 +1,8 @@
+import { chmodSync } from 'fs';
```
amarthadan, line 1: I'd call the directory just `scripts` instead of `dev-scripts` but that's just nitpicking :slightly_smiling_face:

dev-scripts/make-cli-executable.ts
```diff
@@ -0,0 +1,8 @@
...
+
+// See: https://github.com/microsoft/TypeScript/issues/37583
```
amarthadan, line 3: The more commonly used permissions for executable files are `755` (the default is 644, not 664). The executable flag is added for others, group, and owner.

amarthadan (changes requested):
I'm thinking whether it's better to do it like this or to let each package handle its binaries. I know, the code would be duplicated but it's just like three lines. WDYT?

dev-scripts/make-cli-executable.ts
Siegrift: Yeah, I used 755 before as well - but I decided to respect the defaults. Do we need 755 for some reason?

amarthadan: But the default is `644`. Which means everybody can at least read the file. It makes sense that in case of a binary everybody would be able to run it as well. You can't know in what setup it's going to be used. Being able to execute the file for a group might be needed. I don't really see a reason not to add the execution flag for group and others as well.

Siegrift (commented):
> I'm thinking whether it's better to do it like this or to let each package handle its binaries. I know, the code would be duplicated but it's just like three lines. WDYT?

I think this is favorable for multiple reasons:
1) It's all in one place, when new CLIs are added/removed only this must be changed
2) I need a JS file to use the `chmodSync` and I don't want to recreate it in each package
3) I can add a comment explaining why the file is needed.

amarthadan (approved):

PR #582, opened 2021-10-01
Fix AWS credentials env variable name

amarthadan:
Fixes Airnode removal from AWS issue.


Siegrift (approved):
Btw. the docs for this https://docs.api3.org/airnode/next/grp-providers/guides/build-an-airnode/deploying-airnode.html#creating-cloud-credentials need to be updated as well.

PR #596, opened 2021-10-01
Fix testing lambda deployment

amarthadan:
As I was testing Airnode starter I realized that when I deploy Airnode without the HTTP gateway, the lambda that is called by the gateway is deployed anyway, even though nothing will ever call it. This fixes that.


Siegrift (approved):
:+1:

PR #597, opened 2021-10-04
Fix long arguments casing for CLIs 

amarthadan:
Issue [AN-228](https://api3dao.atlassian.net/browse/AN-228)

### DIFF
--- a/packages/deployer/package.json
+++ b/packages/deployer/package.json
@@ -29,7 +29,7 @@
     "dotenv": "^10.0.0",
     "ethers": "^5.4.5",
     "lodash": "^4.17.21",
-    "ora": "^6.0.1",
+    "ora": "^5.4.1",
     "yargs": "^17.0.1"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


packages/deployer/package.json
```diff
@@ -29,7 +29,7 @@
     "dotenv": "^10.0.0",
     "ethers": "^5.4.5",
     "lodash": "^4.17.21",
-    "ora": "^6.0.1",
+    "ora": "^5.4.1",
```
andreogle, line 32: Why the downgrade here? I updated this recently with dependabot https://github.com/api3dao/airnode/pull/594/files

Siegrift: @amarthadan mentioned that it broke the deployer package (but this could have been in a seprate PR to avoid confusion).

Siegrift (approved):

packages/deployer/package.json
andreogle: Ah sorry about that. I wasn't aware

amarthadan: @andreogle I mentioned it in the commit message :slightly_smiling_face: `ora` is now supporting only EMS modules and our current TS configuration doesn't transpile to that. So, I figured it's better to downgrade the not that important dependency instead of changing our TS configuration :smile:

PR #603, opened 2021-10-08
Use prettify for markdown files

Siegrift:


andreogle:
Nice - I wasn't aware that Prettier supported markdown 👍🏻

PR #607, opened 2021-10-09
Update airnode client image docs

Siegrift:
https://api3dao.atlassian.net/browse/AN-268

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "docker": "yarn docker:artifacts && yarn docker:deployer && yarn docker:node",
     "docker:artifacts": "docker rmi --force api3/artifacts && docker build --tag api3/artifacts:latest --file docker/Dockerfile .",
     "docker:deployer": "docker build --tag api3/airnode-deployer:latest --file packages/deployer/docker/Dockerfile .",
-    "docker:node": "docker build --tag api3/airnode:latest --file packages/node/docker/Dockerfile .",
+    "docker:node": "docker build --tag api3/airnode-client:latest --file packages/node/docker/Dockerfile .",
     "lint": "yarn run lint:eslint && yarn run lint:solhint",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "docker": "yarn docker:artifacts && yarn docker:deployer && yarn docker:node",
     "docker:artifacts": "docker rmi --force api3/artifacts && docker build --tag api3/artifacts:latest --file docker/Dockerfile .",
     "docker:deployer": "docker build --tag api3/deployer:latest --file packages/deployer/docker/Dockerfile .",
-    "docker:node": "docker build --tag api3/airnode:latest --file packages/node/docker/Dockerfile .",
+    "docker:node": "docker build --tag api3/airnode-client:latest --file packages/node/docker/Dockerfile .",
```
Siegrift, line 46: Because the image is called like this in docker hub https://hub.docker.com/r/api3/airnode-client

@bbenligiray @amarthadan do we want to use the same name as in pre-alpha for the dockers (same question for deployer).

packages/node/docker/README.md
```diff
@@ -1,73 +1,19 @@
...
 ## Build
 
 In order to build Airnode Docker image you need to build the [artifacts image first](../../../docker/README.md). Once
 you've done that, you can build the Docker image by running following command from the root directory:
 
 ```bash
-docker build -f packages/node/docker/Dockerfile -t api3/airnode:latest .
-```
-
-## Configuration
```
Siegrift, line 15: I'll move all of this to the api3 docs

Siegrift: https://github.com/api3dao/api3-docs/pull/292

package.json
bbenligiray: I personally like `api3/airnode-deployer` and `api3/airnode-client` as image names/have been using them

PR #609, opened 2021-10-09
Add e2e tests to examples package 

Siegrift:
https://api3dao.atlassian.net/browse/AN-170
https://api3dao.atlassian.net/browse/AN-171


packages/examples/scripts/aws-test/coingecko-aws-test.ts
```diff
@@ -0,0 +1,68 @@
...
+      integration: 'coingecko',
+      airnodeType: 'aws',
+      network: 'rinkeby',
+      mnemonic: secrets.MNEMONIC,
+      providerUrl: secrets.RINKEBY_PROVIDER_URL,
+    },
+    null,
+    2
+  );
+  writeFileSync(join(__dirname, '../../integration-info.json'), content);
```
acenolaza, line 30: Is there a cleanup step at the end of running the tests to remove this file since it seems it might contain sensitive information?

packages/examples/scripts/aws-test/coingecko-aws-test.ts
```diff
@@ -0,0 +1,68 @@
...
+};
+
+const createAwsSecrets = () => {
+  const content = formatSecrets([
+    `# Generated by ${__filename}`,
+    `AWS_ACCESS_KEY_ID=${secrets.AWS_ACCESS_KEY_ID}`,
+    `AWS_SECRET_ACCESS_KEY=${secrets.AWS_SECRET_ACCESS_KEY}`,
+    `AWS_SESSION_TOKEN=${secrets.AWS_SESSION_TOKEN}`,
+  ]);
+  writeFileSync(join(__dirname, '../../aws.env'), content);
```
acenolaza, line 40: Same here

packages/examples/scripts/aws-test/coingecko-aws-test.ts
acenolaza: Yes, I guess that should be enough 💯 
I'm now just wondering if these files are really needed 🤔 Wouldn't it be better to have a helper ts file to read these values from evn vars in all other scripts instead? Maybe I'm missing something that forces us to have them in json files

acenolaza: Thank you so much for your detailed explanation.

acenolaza (approved):
🥇

packages/admin/README.md
```diff
@@ -63,10 +63,18 @@ The SDK will also provide TS typings out of the box. Please, refer to the implem
 
 ## Developer commands
 
+### `derive-airnode-xpub`
```
Siegrift, line 66: @bbenligiray I've added this as it seems useful. Agree?

acenolaza: 💯 This seems really useful to me. I think we should add at least 1 e2e test for this new command

packages/admin/src/cli.ts
```diff
@@ -48,6 +48,11 @@ const COMMON_COMMAND_ARGUMENTS = {
       describe: 'Derivation path to be used for deriving the wallet account',
     },
   },
+  airnodeMnemonic: {
```
acenolaza, line 51: You could use `mnemonicCommands` instead of adding a new one since `derivation-path` is optional.

packages/admin/src/implementation.ts
```diff
@@ -9,7 +9,7 @@ const assertAllParamsAreReturned = (params: object, ethersParams: any[]) => {
 };
 
 const verifyAirnodeXpub = (airnodeXpub: string, airnodeAddress: string): ethers.utils.HDNode => {
-  // The xpub is exptected to be from the hardened path m/44'/60'/0'
+  // The xpub is expected to be from the hardened path m/44'/60'/0'
```
acenolaza, line 12: 🤦🏻‍♂️ 😆

packages/admin/src/cli.ts
acenolaza: I'm just trying to re-use the same object and also make it simpler to the end user. I mean, if I'm used to adding `--mnemonic` arg to all other commands that require a mnemonic it might be simpler to keep the same name instead of having a single command that requires a `--airnode-mnemonic`.

acenolaza: I'm not sure TBH because I don't think this tool has been used enough to determine what is simpler for the user but I do see we are not being consistent if we make a "special" mnemonic arg just for this command. For instance: `sponsor-requester`, `unsponsor-requester` and `request-withdrawal` should only be called with the sponsor wallet mnemonic. Does that mean we need to use a `--sponsor-mnemonic` arg instead?
I feel like the arg name should be just `--mnemonic` and if that mnemonic should be for a specific wallet then those are details that must be added in the docs for the user to know, don't you think?

Siegrift: This is a good point. I forgot that we use this for other commands in a similar way.

> I feel like the arg name should be just --mnemonic and if that mnemonic should be for a specific wallet then those are details that must be added in the docs for the user to know, don't you think?

I think user should not need to go to docs for this and it should be clear even from the CLI. Especially, since user can easily confuse these concepts when first trying airnode... I think we should bring this up at the planning call.

For now, I am going to use the mnemonic for consistency.

packages/admin/README.md
```diff
@@ -63,10 +63,18 @@ The SDK will also provide TS typings out of the box. Please, refer to the implem
...
+
+Derives the Airnode extended public key (xpub). Many of the CLI commands require the xpub as a parameter.
```
acenolaza, line 68: I'm not sure if "many" is right here since only a single command actually requires it (`derive-sponsor-wallet-address`)

packages/admin/README.md
```diff
@@ -63,10 +63,18 @@ The SDK will also provide TS typings out of the box. Please, refer to the implem
...
+
+```sh
+npx @api3/airnode-admin derive-airnode-xpub --airnode-mnemonic "nature about salad..."
```
acenolaza, line 71:
```suggestion
npx @api3/airnode-admin derive-airnode-xpub --mnemonic "nature about salad..."
```

packages/admin/test/e2e/cli.feature.ts
```diff
@@ -110,14 +110,25 @@ describe('CLI', () => {
...
 
+  describe('derive-airnode-xpub', () => {
+    it('derives airnode xpub', () => {
+      const airnodeMnemonic = airnodeWallet.mnemonic.phrase;
+
+      // Derive the xpub programatically
+      const airnodeXpub = admin.deriveAirnodeXpub(airnodeMnemonic);
+
+      // Derive the wallet using CLI
+      const out = execCommand('derive-airnode-xpub', ['--airnode-mnemonic', airnodeMnemonic]);
```
acenolaza, line 121: I wonder how this is passing 🤔 

```suggestion
      const out = execCommand('derive-airnode-xpub', ['--mnemonic', airnodeMnemonic]);
```

PR #610, opened 2021-10-10
PSP draft v2

bbenligiray:


packages/protocol/contracts/psp/AirnodeSponsorPspAllocator.sol
```diff
@@ -0,0 +1,104 @@
...
+
+    // At the start of each invocation, the node calls this where `limit` is the number of
+    // slots specified for this allocator in the config file
+    function getActiveSubscriptions(address airnode, uint256 limit)
+        external
+        view
+        returns (bytes32[] memory subscriptionIds,
+        address[] memory sponsors)
+    {
+        // Did a similar thing in the DAO ConvenienceUtils.sol
```
bbenligiray, line 83: I suppose this is pointless, we can just return all the first `limit` items and let the node filter them

packages/protocol/contracts/psp/ExampleConditionAndFulfillContract.sol
```diff
@@ -0,0 +1,54 @@
...
+        // Check if you know about the subscription first
+        // require(subscriptionId == ...)
+        // You can also just check if the template ID/parameters is correct
+        // (bytes32 templateId, , , , , bytes parameters) = airnodePsp.subscriptions(subscriptionId);
+        // require(templateId == ...);
+        // require(parameters.length == 0); // No additional parameters
+
+        // Do nothing if the condition doesn't hold.
+        // Although this check is optional, it's highly recommended.
+        require(condition(data), "Condition not met");
```
bbenligiray, line 38: If `condition()` was in a separate contract, this would have been an external call. I'm a bit conflicted about this.

packages/protocol/contracts/psp/DaoPspAllocator.sol
```diff
@@ -0,0 +1,65 @@
...
+
+    function setSlot(
+        address airnode,
+        uint256 slotIndex,
+        bytes32 subscriptionId,
+        address sponsor
+    ) external {
+        // Check if caller is DAO admin
+        // require(msg.sender == ...)
+        airnodeToSlotIndexToSlot[airnode][slotIndex] = Slot({
```
acenolaza, line 25: When trying to assign a slot to a new `subscriptionId`, how will we know if the `slotIndex` is not already being used? It feels to me like we are missing a `numberOfSlotsUsed` state var counter that could also be used in `getActiveSubscriptions` by replacing the local var `count`.

packages/protocol/contracts/psp/AirnodeSponsorPspAllocator.sol
```diff
@@ -0,0 +1,104 @@
...
+        bytes32 subscriptionId;
+        address sponsor;
+    }
+
+    IAirnodeRrp public airnodeRrp;
+    AirnodePsp public airnodePsp;
+    mapping(address => mapping(address => uint256))
+        public airnodeToSponsorToNoOfTotalSlots;
+    mapping(address => mapping(address => uint256))
+        public airnodeToSponsorToNoOfUsedSlots;
```
acenolaza, line 19: I just posted a message in the DaoPspAllocator about having something like this in there because I reviewed it before this one 🤭

packages/protocol/contracts/psp/AirnodeSponsorPspAllocator.sol
```diff
@@ -0,0 +1,104 @@
...
+        address airnode,
+        address sponsor,
+        uint256 noOfTotalSlots
+    ) external {
+        // Check if the caller is an Airnode admin, i.e., this contract will inherit SelfAdminnable
+        // require(msg.sender == ...)
+        airnodeToSponsorToNoOfTotalSlots[airnode][sponsor] = noOfTotalSlots;
+    }
+
+    // Need to come up with a way to prevent people from constantly switching between slots to grief
```
acenolaza, line 38: You mean like checking if mapping already has a subscriptionId set for airnode/sponsor pair in different slots?

packages/protocol/contracts/psp/AirnodeSponsorPspAllocator.sol
```diff
@@ -0,0 +1,104 @@
...
+        (address airnode, , ) = airnodeRrp.templates(templateId);
+        require(
+            airnodeToSponsorToNoOfTotalSlots[airnode][msg.sender] >
+                airnodeToSponsorToNoOfUsedSlots[airnode][msg.sender],
+            "Sponsor has no available slots"
+        );
+        address revokedSponsor = airnodeToSlotIndexToSlot[airnode][slotIndex]
+            .sponsor;
+        if (revokedSponsor != address(0)) {
+            require(
```
acenolaza, line 52: I'm having a hard time understanding in which case this will be true

packages/protocol/contracts/psp/AirnodeSponsorPspAllocator.sol
bbenligiray: I mean say I'm a regular user that wants to subscribe. When I send a transaction that specifies a slot, someone that also has a slot available makes a transaction with a high gas price, frontrunning to get my slot. My tx gets reverted and I see that the slot I attempted to get is now unavailable so I make another tx for another slot. The attacker quickly unsubscribes and subscribes again for the next slot I attempted to get, and this keeps happening.

packages/protocol/contracts/psp/DaoPspAllocator.sol
bbenligiray: This `DaoPspAllocator` is just a very heavy-handed implementation that allows the admin to kick subscriptions out at will

packages/protocol/contracts/psp/AirnodePsp.sol
```diff
@@ -0,0 +1,109 @@
...
+    // This is because botched fulfillments are not a threat (read more at `fulfill()` comments).
+    // We do not specify `sponsor` because it is returned by the allocator along with the respective
+    // `subscriptionId`. This implies that subscriptions can be served by anyone with
+    // signed data.
+    // The struct below could do without the `parameters` parameter (and we would force the
+    // user to put everything in the template `parameters`). However, RRP is designed around the
+    // reuse of templates and Airnode ABI is not very efficient in encoding parameters, which
+    // is good enough reason to allow PSP to reuse more generic templates with additional
+    // subscription-specific parameters.
+    struct Subscription {
```
Siegrift, line 25: Could the subscription be ownable and support updates? 

An use case is that the API slightly changes it's parameters and you want to only alter the `parameters` (to save some eth)

packages/protocol/contracts/psp/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+The main difficulty in providing a subscription service is that an API/Airnode can serve a limited number of them at a time (compared to this, RRP is much more scalable). Then, the resource that we must allow the Airnode to manage are these subscription "slots". This slot allocation needs to happen on-chain, programmatically. To achieve this, we implement "allocators", contracts similar to "authorizers" that implement arbitrary business logic that the Airnode operator can choose to opt-in to by adding them in their config file. Note that a requester both needs to have a slot allocated by an allocator, and be authorized by an authorizer to receive service.
```
Siegrift, line 7: 1) What if there are multiple subscription for the same `subscriptionId` with different `sponsor`?

I guess it will send the response transaction twice, right?

packages/protocol/contracts/psp/README.md
```diff
@@ -0,0 +1,23 @@
...
+- `airnode`: `airnode`
+- `endpointId`: `endpointId`
+- `sponsor`: `sponsor`
+- `requester`: `fulfillAddress`
+Alternatively, we can refer to the subscriber as the requester (instead of the sponsor), and implement an additional sponsorship layer on top.
+
+For each chain, the config file should specify a list of allocators with the following fields:
+- Allocator chain + provider(s)
+- Allocator address
+- Maximum number of slots
```
Siegrift, line 20: So this is global for the allocator? Shouldn't this be in the contract?

packages/protocol/contracts/psp/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+The main difficulty in providing a subscription service is that an API/Airnode can serve a limited number of them at a time (compared to this, RRP is much more scalable). Then, the resource that we must allow the Airnode to manage are these subscription "slots". This slot allocation needs to happen on-chain, programmatically. To achieve this, we implement "allocators", contracts similar to "authorizers" that implement arbitrary business logic that the Airnode operator can choose to opt-in to by adding them in their config file. Note that a requester both needs to have a slot allocated by an allocator, and be authorized by an authorizer to receive service.
+
+The authorizer parameters will be filled as such:
+- `requestId`: `subscriptionId`
+- `airnode`: `airnode`
+- `endpointId`: `endpointId`
+- `sponsor`: `sponsor`
+- `requester`: `fulfillAddress`
+Alternatively, we can refer to the subscriber as the requester (instead of the sponsor), and implement an additional sponsorship layer on top.
```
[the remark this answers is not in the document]
Siegrift: Yeah, I find it strange that the entity calling `subscribe` must be the sponsor, but I'd just rename the contract functions:
- Instead of `subscribe` call it `sponsorSubscription` 
- and `unsponsorSubscription` instead of `unsubscribe`

packages/protocol/contracts/psp/AirnodeSponsorPspAllocator.sol
```diff
@@ -0,0 +1,104 @@
...
+            msg.sender == airnodeToSlotIndexToSlot[airnode][slotIndex].sponsor,
+            "Caller does not own slot"
+        );
+        airnodeToSponsorToNoOfUsedSlots[airnode][msg.sender]--;
+        delete airnodeToSlotIndexToSlot[airnode][slotIndex];
+    }
+
+    // At the start of each invocation, the node calls this where `limit` is the number of
+    // slots specified for this allocator in the config file
+    function getActiveSubscriptions(address airnode, uint256 limit)
```
Siegrift, line 77: Shouldn't this be paginated? (just in case)?

packages/protocol/contracts/psp/AirnodeSponsorPspAllocator.sol
Siegrift: Any advantage of not doing it in the contract function directly? 

I can imagine an implementation of Allocator where slots will have some automated "time to live" (e.g. block number or timestamp) and this function can filter those "outdated" slots... So I like this function as a part of an "AllocatorInterface"

packages/protocol/contracts/psp/AirnodePsp.sol
```diff
@@ -0,0 +1,109 @@
...
+        bytes32 templateId;
+        address conditionAddress;
```
Siegrift, line 27: I like that condition and fulfill address can be different - I think that is very important

packages/protocol/contracts/psp/AirnodePsp.sol
```diff
@@ -0,0 +1,109 @@
...
+    function createSubscription(
+        bytes32 templateId,
+        address conditionAddress,
+        bytes4 conditionFunctionId,
+        address fulfillAddress,
+        bytes4 fulfillFunctionId,
+        bytes calldata parameters
+    ) external returns (bytes32 subscriptionId) {
+        // Subscription IDs do not need to be unique. Airnode will not serve the same
+        // sbuscription ID multiple times
```
Siegrift, line 53: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker :D

packages/protocol/contracts/psp/AirnodePsp.sol
```diff
@@ -0,0 +1,109 @@
...
+    ) external returns (bool callSuccess, bytes memory callData) {
+        Subscription storage subscription = subscriptions[subscriptionId];
+        // Note that we are not verifying that msg.sender is a pre-determined `sponsorWallet`.
+        // This because one cannot "botch" (e.g., by specifying a low gas limit) a PSP fulfillment
+        // like an RRP fulfillment (because previous failed fulfillments don't prevent future
+        // successful ones).
+        (address airnode, , ) = airnodeRrp.templates(subscription.templateId);
+        // It is a concern for someone to have the API provider sign a response, to be used much
+        // later (for example, would be very problematic for asset price data feeds). Therefore,
+        // we add the timestamp to the hash to be signed.
```
Siegrift, line 91: The idea is that a requester will compare the `timestamp` with the `now` in solidity? and decide depending on that?

I wonder whether this achieves anything the API provider can sign it with a future timestamp and there is no way for requester to check if the timestamp is indeed valid... Maybe you could ask the API provider to specify the hash and the block number of the latest mined block - since that can be at least verified by the requester.

packages/protocol/contracts/psp/AirnodePsp.sol
```diff
@@ -0,0 +1,109 @@
...
+        // like an RRP fulfillment (because previous failed fulfillments don't prevent future
+        // successful ones).
+        (address airnode, , ) = airnodeRrp.templates(subscription.templateId);
+        // It is a concern for someone to have the API provider sign a response, to be used much
+        // later (for example, would be very problematic for asset price data feeds). Therefore,
+        // we add the timestamp to the hash to be signed. 
+        require(
+            (
+                keccak256(
+                    abi.encodePacked(subscriptionId, timestamp, data)
```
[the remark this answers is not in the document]
Siegrift: I don't think I follow. Do you mean that a requester can save a timestamp when the request was made and then later once the Airnode fulfill transaction is received it will reject, because the timestamp on the Airnode response transaction is much larger? (a lot of time has passed)

Because I don't see a problem with this requester behaviour.

packages/protocol/contracts/psp/AirnodeSponsorPspAllocator.sol
```diff
@@ -0,0 +1,104 @@
...
+        public airnodeToSponsorToNoOfUsedSlots;
+    mapping(address => mapping(uint256 => Slot))
+        public airnodeToSlotIndexToSlot;
+
+    constructor(address airnodeRrp_, address airnodePsp_) {
+        airnodeRrp = IAirnodeRrp(airnodeRrp_);
+        airnodePsp = AirnodePsp(airnodePsp_);
+    }
+
+    function setTotalNoOfSlots(
```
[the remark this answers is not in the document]
Siegrift: What is the use case for this?

packages/protocol/contracts/psp/AirnodePsp.sol
bbenligiray: Word is hard

packages/protocol/contracts/psp/AirnodePsp.sol
```diff
@@ -0,0 +1,109 @@
...
+            parameters: parameters
+        });
+    }
+
+    // We are having the node fulfill through this function because we don't want the requester
+    // contract to have to import ECDSA.sol
+    // The node doesn't attempt to fulfill if `callSuccess` is false, logs the revert string internally.
+    function fulfill(
+        bytes32 subscriptionId,
+        uint256 timestamp,
```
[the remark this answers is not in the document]
bbenligiray: The fulfillment target can already fetch `block.timestamp` itself so I don't see why we would send it

packages/protocol/contracts/psp/AirnodePsp.sol
```diff
@@ -0,0 +1,109 @@
...
+        bytes calldata data,
+        bytes calldata signature
+    ) external returns (bool callSuccess, bytes memory callData) {
+        Subscription storage subscription = subscriptions[subscriptionId];
+        // Note that we are not verifying that msg.sender is a pre-determined `sponsorWallet`.
+        // This because one cannot "botch" (e.g., by specifying a low gas limit) a PSP fulfillment
+        // like an RRP fulfillment (because previous failed fulfillments don't prevent future
+        // successful ones).
+        (address airnode, , ) = airnodeRrp.templates(subscription.templateId);
```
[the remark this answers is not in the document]
bbenligiray: I see. I think of RRP as the main protocol and PSP as something that most people won't need to concern themselves with. One good thing about keeping templates in RRP is that it won't need to make an external call to check the `airnode` in a template for example, which will save a bit of gas, but this is probably not very relevant. The main downside of a separate TemplateStore.sol imo is that we'll have to handle one additional protocol contract address per chain.

packages/protocol/contracts/psp/AirnodePsp.sol
bbenligiray: > API provider can sign it with a future timestamp
Good point, though I was mostly thinking about transactions stuck in the mempool more than a malicious API provider, they can straight out misreport anyway

> specify the hash and the block number of the latest mined block
You can only get the hash of one of the 256 latest blocks in contract, so this would be pretty limiting (and probably not future-proof). Also we would have to make more blockchain provider calls by the node. So this is creating more problems than it solves imo (mostly because it doesn't protect against all kinds of attacks that an API provider performs).

packages/protocol/contracts/psp/AirnodeSponsorPspAllocator.sol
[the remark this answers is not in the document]
bbenligiray: You allocate N slots to a sponsor (because they paid you or you signed an agreement), so they can make N consecutive subscriptions but not any more (because slots are a finite resource)

packages/protocol/contracts/psp/AirnodeSponsorPspAllocator.sol
bbenligiray: Yeah, good point

packages/protocol/contracts/psp/AirnodeSponsorPspAllocator.sol
bbenligiray: Doing computation in a static call is more expensive than doing the same computation at the node (the blockchain provider will charge you for the compute units and/or the calls will be responded to with more delay)

packages/protocol/contracts/psp/ExampleConditionAndFulfillContract.sol
bbenligiray: > What if API provider is malicious and responds `true`

This call is a static call done to the blockchain provider so they would be the one being malicious (maybe you meant this anyway)

> Airnode would then fulfill the transaction (draining sponsor funds) even though the requester didn't want this data.

Yes, but blockchain providers have other ways of fucking over the Airnode, such as reporting an unnecessarily high gas price

> Shouldn't this be called in the fulfill of AirnodePsp?

The main reason it's not there is that that would have been a second external untrusted low level call, and such calls need to be the last thing you do to avoid a reentrancy attack, but that's not a solution when you have two such calls (and we probably don't want a reentrancy guard just for this). So you can just leave it to the target function to check the condition (and implement the reentrancy guard if `condition()` is not trusted, in this case it is), it will be just as secure.

packages/protocol/contracts/psp/README.md
Siegrift: But which sponsor should pay for the transaction?

packages/protocol/contracts/psp/AirnodeSponsorPspAllocator.sol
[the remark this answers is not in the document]
Siegrift: Actually I misread this function. This sets the number of slots for a particular sponsor (I thought it's global).

Anyway, shouldn't the contract support setting a global maximum number of slots? Since, this number might impact the Airnode throughtput and I also see a business value here (provably saying that a given allocator only serves at most X subscribers)

packages/protocol/contracts/psp/README.md
Siegrift: Actually, do I understand this field correctly? This is the number that will be used as `limit` when calling `getActiveSubscriptions`.

I mean the limit could be optional in the contract (default value -1) and settable through admin CLI by the API provider? But this is a bit unfortunate that you have to make TX for this... Also, a big pain is to keep this in sync with what's in config json.

Can't you keep this `limit` number in sync in the contract instead? (update it when the maximum changes). Some allocator updates will be more costly then others, but I don't think it's a big deal.

Also, I wonder how will the allocator interface look like. Will it have only the `getActiveSubscriptions`? I guess that is all the Airnode needs... As mentioned above, I'd like if the `limit` (maximum number of slots of sponsor) was managed by the contract. I'd then imagine the following interface:

```solidity
interface IAllocator {
    function getTotalNumberOfActiveSlots() external view returns (uint256);

    function getMaxSlotNumberOfAirnode(address airnode) external view returns (uint256);

    function getNumberOfActiveSlotsOfAirnode(address airnode) external view returns (uint256);

    function getActiveSlots(address airnode)
        external
        view
        returns (bytes32[] memory subscriptionIds, address[] memory sponsors);

    // in case the number of active slots is large - pagination is required???
    // the node can decide if this is necessary based on "getMaxSlotNumberOfAirnode"
    function getActiveSlotsPage(address airnode, uint256 from, uint256 to)
        external
        view
        returns (bytes32[] memory subscriptionIds, address[] memory sponsors);
}
```

packages/protocol/contracts/psp/AirnodeSponsorPspAllocator.sol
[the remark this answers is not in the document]
bbenligiray: A global maximum number of slots can only be set at the node side (in the config file) because the allocator contracts don't know about each other. When the config file uses one allocator contract for 5 slots and another for 10, it implicitly has a limit of 15 on the global number of slots.

packages/protocol/contracts/psp/AirnodeSponsorPspAllocator.sol
```diff
@@ -0,0 +1,104 @@
...
+        external
+        view
+        returns (bytes32[] memory subscriptionIds,
+        address[] memory sponsors)
+    {
+        // Did a similar thing in the DAO ConvenienceUtils.sol
+        uint256 count = 0;
+        mapping(uint256 => Slot)
+            storage slotIndexToSlot = airnodeToSlotIndexToSlot[airnode];
+        for (uint256 ind = 0; ind < limit; ind++) {
```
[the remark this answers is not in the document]
bbenligiray: The idea was that I want the function to check the first `limit` slots and return the valid ones. It is non-standard/confusing so I'll update it.

Siegrift:
@bbenligiray what about these PSP drafts branches? Shall we close them? (and keep only v4)

bbenligiray:
Leave them pls, I still need them but I'll clean them up soon

bbenligiray:

PR #614, opened 2021-10-12
Include admin unit tests in CI, simplify configuration

Siegrift:
Since, we have unlimited minutes we can just run the e2e tests and the unit tests at the same time...

And since e2e tests take way longer than unit tests, we can just run the unit tests sequentially... (And it actually might run in parallel anyway, since it uses `lerna --stream` )


.github/workflows/build-test.yml
```diff
@@ -80,141 +80,12 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: '14.17.1'
-      - run: yarn run test:adapter
-      - name: Slack Notification
```
andreogle, line 84: Can we keep a slack notification for the unit tests please?

andreogle (approved):

packages/validator/test/nested/template.json
```diff
@@ -6,4 +6,4 @@
       "__message": "Error in template nested in __prefix"
     }
   }
-}
+}
```
Siegrift, line 9: Why is this change happening? :/

PR #615, opened 2021-10-12
Fix formatting issues

Siegrift:


Siegrift:
This one is not needed anymore since the README will be moved to docs.

PR #620, opened 2021-10-12
Add new values to "_types" parameter

Siegrift:
https://api3dao.atlassian.net/browse/AN-255

Adds support for new values to `_type` reserved parameter ([docs](https://docs.api3.org/airnode/v0.2/reference/specifications/reserved-parameters.html#type)). The new supported values are: `address`, `bytes`, `string` and arrays (also nested and fixed) for any of the "base" types already supported.


packages/adapter/src/response-processing/encoding.ts
```diff
@@ -1,42 +1,54 @@
...
-export function convertStringToBytes32(value: string) {
-  const trimmedValue = trimValue(value);
   const bytes32String = ethers.utils.formatBytes32String(trimmedValue);
   return ethers.utils.defaultAbiCoder.encode(['bytes32'], [bytes32String]);
 }
 
-export function convertBoolToBytes32(value: boolean) {
-  return ethers.utils.defaultAbiCoder.encode(['bool'], [value]);
+export function convertAddress(value: string) {
+  return ethers.utils.defaultAbiCoder.encode(['address'], [value]);
```
acenolaza, line 26: Should we check if `value` is actually a valid `address`? https://docs.ethers.io/v5/api/utils/address/#utils-isAddress

Or maybe this should be done in a specific `castAddress()` function in [casting.ts](https://github.com/api3dao/airnode/blob/b83870b816f7ad7c36747844b80d23387976c6ec/packages/adapter/src/response-processing/casting.ts) instead of using `castStringLike()`.

Siegrift: Yeah, that is probably a good idea.

packages/adapter/src/response-processing/casting.ts
```diff
@@ -53,25 +54,56 @@ function castBoolean(value: unknown): boolean {
...
 }
 
-function castBytes32(value: any): string {
+function castStringLike(value: any, target: 'string' | 'bytes32'): string {
+  // TODO: Should we throw here? Isn't String(JSON.stringify(value)) better?
   // Objects convert to "[object Object]" which isn't very useful
   if (isArray(value) || isPlainObject(value)) {
-    throw new Error(`Unable to convert: '${JSON.stringify(value)}' to bytes32`);
+    throw new Error(`Unable to convert: '${JSON.stringify(value)}' to ${target}`);
   }
```
acenolaza, line 62: This now looks like a good candidate for a separate function since we are duplicating these lines in 3 different places now.

packages/airnode-adapter/src/response-processing/encoding.test.ts
```diff
@@ -161,8 +139,12 @@ describe('Encoding array values', () => {
   });
 
   it('throws on invalid encoding type', () => {
-    expect(() => encodeValue('', 'unit256')).toThrow(`Unknown type: unit256`);
-    expect(() => encodeValue('', 'int256[[]]')).toThrow(`Unknown type: int256[[]]`);
-    expect(() => encodeValue('', '[int256]')).toThrow(`Unknown type: [int256]`);
+    expect(() => encodeValue('', 'unit256')).toThrow(`invalid type (argument=\"type\", value=\"unit256\"`);
+    expect(() => encodeValue('', 'int256[[]]')).toThrow(
+      `unexpected character at position 7 (argument=\"param\", value=\"int256[[]]\"`
```
Siegrift, line 144: I have reverted back to letting ethers throw the errors. It's not as descriptive as before, but we don't have to check types validity anymore while encoding.

LMK if you'd prefer the previous approach.

packages/airnode-adapter/src/response-processing/encoding.ts
```diff
@@ -1,66 +1,8 @@
...
-function convertArray(value: ValueType, type: ResponseType): string {
-  return ethers.utils.defaultAbiCoder.encode([type], [value]);
-}
-
 export function encodeValue(value: ValueType, type: ResponseType): string {
-  const parsedArrayType = parseArrayType(type);
-  if (parsedArrayType) {
-    return convertArray(value, type);
-  }
+  if (type === 'bytes32') return ethers.utils.formatBytes32String(value as string);
```
Siegrift, line 5: I have mixed feelings about this... There are two stages - casting the value and encoding. The logic works like this: 
- if the type is `bytes32`, the casting will trim the value to at most 31 characters, encoding is done by `formatBytes32String`
- if the type is `string`, casting will just stringify and encoding is done by `defaultAbiCoder` (line below)
- if the type is `bytes` casting will verify that the string is a valid hex string (e.g. `0x7468697320697320616e206578616d706c6520737472696e672074686174206973206120626974206c6f6e676572`) and again encode using `defaultAbiCoder`

Result:
- The handling is very different for `bytes32` and `bytes` (former expects string, latter "encoded" hex string)
- If I implemented the `bytes` to expect string and encode it the same way - I'd be essentially doing the same as what `string` does

WDYT should be the correct approach?

Siegrift: @andreogle @bbenligiray

packages/airnode-adapter/src/response-processing/array-type.ts
```diff
@@ -0,0 +1,54 @@
...
+export interface ParsedArrayType {
+  readonly baseType: BaseResponseType;
+  readonly dimensions: number;
+}
+
+export function parseArrayType(type: ResponseType): ParsedArrayType | null {
+  if (!type.includes('[')) return null;
+
+  const typeMatch = type.match(/^([^\[]+)(.*)$/);
+  if (!typeMatch || !baseResponseTypes.includes(typeMatch[1] as any)) return null;
```
acenolaza, line 17: Maybe we should lowercase the input just in case:
```suggestion
  if (!typeMatch || !baseResponseTypes.includes(typeMatch[1].toLocaleLowerCase() as any)) return null;
```

packages/airnode-adapter/src/response-processing/encoding.ts
bbenligiray: In the previous version, there was only one way to return a string from the API response (encoded in `bytes32`), which is why we did it this way. This is confusing for the user but can be worked around https://api3workspace.slack.com/archives/C02AYRX8D89/p1633444784102900

Here, we definitely want to treat `bytes` as raw bytes and not an encoded string (because there is no other way to return raw bytes from the API and we want that). I guess one question is do we still treat `bytes32` as an encoded string (which is not obvious) or should we also treat `bytes32` as `bytes32` and tell people to use `string` if they want to return a string.

packages/airnode-adapter/src/response-processing/array-type.ts
```diff
@@ -0,0 +1,54 @@
...
+  const baseType = typeMatch[1] as BaseResponseType;
+
+  // eslint-disable-next-line functional/no-let
+  let dimensionsString = typeMatch[2];
+  // eslint-disable-next-line functional/no-let
+  let dimensions = 0;
+  // eslint-disable-next-line functional/no-loop-statement
```
andreogle, line 24: The ESLint functional plugin is bothering me more and more. The only part of it that I *really* like is the enforced immutability (with opt-out when needed)

packages/airnode-adapter/src/response-processing/encoding.test.ts
andreogle: I'm fine with this

packages/airnode-adapter/src/response-processing/encoding.ts
andreogle: Responded in Slack, but this:

> should we also treat bytes32 as bytes32 (so break backwards compatibility for it to make more sense) and tell people to use string if they want to return a string (which will actually be more expensive than bytes32).

seems like the correct thing to do to me?

Siegrift: I had also an idea bout introducing "made up" `string32` which would encode the same way as `bytes32` currently.

Siegrift: I am resolving this thread -let's continue the discussion in slack

packages/airnode-adapter/src/response-processing/array-type.ts
Siegrift: Tell me something about it... I think in this function I wrote more eslint disable lines than logic. I've bumped the priority of https://api3dao.atlassian.net/browse/AN-190

packages/airnode-adapter/src/response-processing/array-type.ts
Siegrift: do we want that? I mean `UiNt256` should throw an error

acenolaza: Correct me if I'm wrong but that value is coming from the config.json, right? Since the API provider might not be so familiar with solidity types he could make a simple mistake and put something like "String", "uInt256" or "Uint256" causing the request to fail.

Troubleshooting the Airnode requests is not a trivial task at the moment (maybe with the RRP Explorer will be easier) so  I'm keen to take a more self-healing approach when possible.

Since this is more of a personal preference you can take it or leave as is.

Siegrift: This should be the responsibility of validator to reject such config.json

But you raised a very valid point - currently we only check whether the name of the reserved parameters is correct (https://github.com/api3dao/airnode/blob/feature/gcp-state/packages/airnode-validator/templates/0.3/endpoints.json#L41). I've created https://api3dao.atlassian.net/browse/AN-340 for this

packages/airnode-adapter/src/response-processing/casting.ts
```diff
@@ -53,26 +55,71 @@ function castBoolean(value: unknown): boolean {
...
       return castBoolean(value);
-
     case 'bytes32':
       return castBytes32(value);
+    case 'string':
+      return castString(value);
+    case 'address':
+      return castAddress(value);
+    case 'bytes':
+      return castHexString(value);
```
acenolaza, line 119: Just out of curiosity, how will the requester decode a bytes data?

Will something like the following give us the expected result if we send the hexlified string?:
```
byte decodedData = abi.decode(data, (byte));
```

acenolaza: 🤔 that was my suspicion...maybe because the hex string starts with `"0x..."`? What I'm not sure about is if the hex string is the valid encoding for bytes.

Siegrift: I mean that `byte` trigger compile error for me.

> What I'm not sure about is if the hex string is the valid encoding for bytes solidity type.

It does - I mean the e2e test is passing correctly. Or maybe I am misunderstanding something

packages/airnode-adapter/src/response-processing/encoding.test.ts
Siegrift: Actually, I've changed my mind - I think it will be better to handle it ourselves. It's a bit more code, but the ethers error is not very descriptive and it might be a pain to debug that when looking at AWS logs.

packages/airnode-adapter/src/response-processing/casting.ts
acenolaza: > I mean that `byte` trigger compile error for me.

It should be:
```
bytes decodedData = abi.decode(data, (bytes)); 
```
> It does - I mean the e2e test is passing correctly. Or maybe I am misunderstanding something

I'm just trying to make sure that `data` can be successfully decoded in the requester contract when type is `bytes`

Siegrift: Yeah, but it does :D 

Or am I still misunderstanding?

acenolaza (approved):

PR #621, opened 2021-10-12
AN-180: Adds verify-airnode-xpub command to admin-cli

acenolaza:
Task details: https://api3dao.atlassian.net/browse/AN-180

### DIFF
--- a/packages/admin/README.md
+++ b/packages/admin/README.md
@@ -71,6 +71,17 @@ Derives the Airnode extended public key (xpub).
 npx @api3/airnode-admin derive-airnode-xpub --mnemonic "nature about salad..."
 ```
 
+### `verify-airnode-xpub`
+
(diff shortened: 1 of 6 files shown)


packages/admin/README.md
```diff
@@ -71,6 +71,17 @@ Derives the Airnode extended public key (xpub).
...
 ```
 
+### `verify-airnode-xpub`
+
+Verifies that the xpub belongs to the HDNode with the path `m/44'/60'/0'` of the Airnode wallet. This command checks
+that the Airnode address can be derived with provided Airnode xpub and default derivation path `m/44'/60'/0'/0/0`.
+
+```sh
+npx @api3/airnode-admin verify-airnode-xpub \
+  --airnode-xpub xpub6CUGRUo... \
```
Siegrift, line 81: I guess we could now say that "Many commands require airnode-xpub as parameter" :D

PR #622, opened 2021-10-13
AN-132 Update validator templates for 1.0.0

drgy:


bbenligiray:
Don't validate any rules that are not in the docs / If a rule needs to be validated add it in the docs. I remember that we only allowed alphanumeric characters and `-` for some fields such as `stage`.

Everything needs to be modeled after our own docs, which are modeled after OAS. If our docs are underdefined, they need to be extended instead of referring to an external source.

PR #623, opened 2021-10-13
AN-159 Validator integration

amarthadan:
Issue [AN-159](https://api3dao.atlassian.net/browse/AN-159)

I had to include the validator's templates in the exported lambdas so that the validator will work there. Not that happy about the solution but not sure what else I can do about it.

I also "simplified" the template for `config.json` just not to break the build. https://github.com/api3dao/airnode/pull/622 will fix that.

I also added a template for `receipt.json` but again in a simplified version. @drgy can you please come up with the correct one according to the [documentation](https://docs.api3.org/airnode/next/reference/deployment-files/receipt-json.html)? Everything except `heartbeatId` and `httpGatewayUrl` is mandatory.

Lastly, I'm not passing any template version when requesting the validation which means that the validator will validate against the latest one. That's ok for now but I'll open a discussion about this on Slack as I think there's a bigger issue here.


packages/deployer/webpack.config.js
```diff
@@ -16,4 +17,12 @@ module.exports = {
     mainFields: ['main'],
   },
   target: 'node',
+  plugins: [
+    new CopyPlugin({
+      patterns: [
+        { from: '../validator/dist/templates', to: 'templates' },
```
Siegrift, line 23: Why don't w have to do this for the JS files copied using the `copy:webpack`?

amarthadan: There's just one JS file and that's the `index.js`, the result of the Webpack run. Or maybe I don't understand the question.

amarthadan: Ok, I'll try to break it down.

When you run `yarn build` in the `deployer` directory these targets are called:
* `compile` - TS transpilation outputting into the `dist` directory.
* `webpack` - Starting from `dist/src/handlers/aws/index.js` bundles everything that is imported/required into `.webpack`. This is the code for Lambdas. Now, with the validator integration, in order for this bundle to run we have to include validator templates. Since they are not imported in the validator code they need to be included separately. That's where the plugin comes in. It copies the template and convertor files needed for validation next to the bundle (in `.webpack`).
* `copy:webpack` - Copies the `.webpack` directory into `dist` directory so that when packaged node's lambda function is also included.
* `copy:terraform` - Copies the Terraform recipes into `dist` directory so that when packaged the deployment can be run from them.

Does it make sense now?

Siegrift: Thanks, I was missing the part that the validator files are not required in the code and that's why they are left out of by webpack...

Maybe add link to your reply above the plugin (for future q)

Siegrift (approved):
:+1: LGTM

PR #625, opened 2021-10-15
Add safe navigator to heartbeat and httpGateway config.json objects.

aquarat:
The docs originally listed these two objects as optional config options. If they're omitted both writeReceiptFile and other code fails cryptically.

Additionally, the mnemonic validator doesn't differentiate between an invalid mnemonic and an undefined mnemonic - something I encountered when I tried to bring in an old env file which lacked that env entirely.

I have tested these changes.


packages/deployer/src/cli/commands.ts
```diff
@@ -21,13 +21,12 @@ export async function deploy(configFile: string, secretsFile: string, receiptFil
...
   if (!validateMnemonic(mnemonic)) {
-    logger.fail('AIRNODE_WALLET_MNEMONIC in your secrets.env file is not valid');
-    throw new Error('Invalid mnemonic');
+    throw new Error('AIRNODE_WALLET_MNEMONIC invalid or not found');
   }
 
   const httpGateway = config.nodeSettings.httpGateway;
   let httpGatewayApiKey: string | undefined = undefined;
-  if (httpGateway.enabled) {
+  if (httpGateway?.enabled) {
```
Siegrift, line 29: This is not correct. The HTTP gateway and heartbeat are optional features, but you need to explicitely turn them off and this property should be present. We should update the docs if this is not clear.

The schema check for this should be handled in https://api3dao.atlassian.net/browse/AN-284

Siegrift: That's because you are looking at `pre-alpha` docs. It's correct in https://docs.api3.org/airnode/next/grp-providers/tutorial/config-json.html

aquarat:
Changes not strictly necessary due to other issues.

PR #626, opened 2021-10-17
Fix Deployer version retrieval

amarthadan:
Deployer version wasn't displayed correctly when run from the installed package or container. This fixes it.


packages/deployer/src/cli/index.ts
```diff
@@ -7,6 +7,7 @@ import omitBy from 'lodash/omitBy';
 import isEmpty from 'lodash/isEmpty';
 import { hideBin } from 'yargs/helpers';
 import { deploy, removeWithReceipt, remove } from './commands';
+import { version } from '../../package.json';
 import * as logger from '../utils/logger';
 
 // TODO: Get nodeVersion from the package
```
Siegrift, line 13: Can't you resolve this one the similarly? (check the node version from dependencies)

amarthadan: You can't get that from the deployer's `package.json` as the version there doesn't have to necessarily match the actually used version. I would have to go to the `node_modules` directory, find `node`, and read its `package.json` to see what version was installed. This may be complicated in the monorepo where the `node_modules` are symlinked.

What I will probably do in this regard is to export a `version` function from `node` that would return the Airnode's version.

amarthadan: > Will this fn read package.json or we will need to update this everytime we release?

No, I want to read it from `package.json` the same way as here.

PR #633, opened 2021-10-26
AirnodeTokenLock with AirnodeFeeRegistry

Ashar2shahid:
This Draft is based on the [specifications ](https://docs.google.com/document/d/1OlSLsjIBZfqliGt7deY2oCX3q8itnfoTMnkJte4Ko88/edit?usp=sharing) in this document. Few Things to note:

- Still incomplete due to the New Access Control Scheme still being worked on.
- Haven't updated the tests because of the above, the new tests used the old adminnable scheme but I removed all traces of adminnable and the tests as well.
- The Prices in the AirnodeFeeRegistry are multiplied with the decimals places, I'm not too sure about this
- AirnodeFeeRegistry has different mappings for all the different default prices, maybe we can use a single mapping i.e price[chainId][airnode][endpoint] and access the defaults when any one of them is 0 ? Would this be a better approach?


packages/protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
```diff
@@ -0,0 +1,433 @@
...
+
+    /// @dev Address of AirnodeFeeRegistry
+    address public immutable airnodeFeeRegistry;
+
+    /// @dev The price of API3 in terms of USD
+    /// @notice The price is specified upto 6 decimal places
+    uint256 public api3PriceInUsd;
+
+    /// @dev A coefficient used to calculate the amount of tokens to be locked
+    uint8 public multiplierCoefficient;
```
acenolaza, line 50: Since `multiplierCoefficient` is type `uint8`, shouldn't we move above `api3PriceInUsd` to try pack it in a single memory slot with `airnodeFeeRegistry` (20bytes)?

Ashar2shahid: I've changed it to be a uint256, mainly because it didn't make sense to have it be uint8 integer. We would want precision in the coefficient when setting it, before it could only be set without precision. Now it can be set as a number with 18 digit decimals which also makes the calculation of the tokenlock cleaner, all the numbers are in their respective decimals and we don't have to multiply/divide by anything.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
```diff
@@ -0,0 +1,231 @@
...
+        public defaultChainAirnodePrice;
+
+    /// @dev The default price of an endpoint on an airnode
+    mapping(address => uint256) public defaultAirnodePrice;
+
+    /// @dev The default price of an endpoint for a chain
+    mapping(uint256 => uint256) public defaultChainPrice;
+
+    /// @dev The global default price of an endpoint
+    uint256 public defaultPrice;
```
acenolaza, line 34: How do you feel about replacing all these variables for a single mapping using a data structures like the this below:

```
    struct EndpointPrice {
        uint256 price;
    }

    struct AirnodePrice {
        uint256 defaultPrice;
        uint256 price;
        mapping(bytes32 => EndpointPrice) endpointPrices;
    }

    struct ChainPrice {
        uint256 defaultPrice;
        uint256 price;
        mapping(address => AirnodePrice) airnodePrices;
    }

    mapping(uint256 => ChainPrice) public chainIdToAirnodeToEndpointToPrice;
``` 

The `getEndpointPrice()` function would be something like this:
```
    function getEndpointPrice(
        uint256 _chainId,
        address _airnode,
        bytes32 _endpointId
    ) external view returns (uint256) {
        if (
            chainIdToAirnodeToEndpointToPrice[_chainId]
                .airnodePrices[_airnode]
                .endpointPrices[_endpointId]
                .price != 0
        )
            // returns the endpoint price for an airnode in a chain
            return
                chainIdToAirnodeToEndpointToPrice[_chainId]
                    .airnodePrices[_airnode]
                    .endpointPrices[_endpointId]
                    .price;

        if (
            airnodeEndpointFlag[_airnode] &&
            chainIdToAirnodeToEndpointToPrice[0]
                .airnodePrices[_airnode]
                .endpointPrices[_endpointId]
                .price !=
            0
        )
            // returns the endpoint price for an airnode across all chains
            return
                chainIdToAirnodeToEndpointToPrice[0]
                    .airnodePrices[_airnode]
                    .endpointPrices[_endpointId]
                    .price;
        // returns the default price for an airnode in a chain
        else
            return
                chainIdToAirnodeToEndpointToPrice[_chainId]
                    .airnodePrices[_airnode]
                    .defaultPrice;

        if (
            chainIdToAirnodeToEndpointToPrice[0]
                .airnodePrices[_airnode]
                .defaultPrice != 0
        )
            // returns the default price for an airnode across all chains
            return
                chainIdToAirnodeToEndpointToPrice[0]
                    .airnodePrices[_airnode]
                    .defaultPrice;
        if (chainIdToAirnodeToEndpointToPrice[_chainId].defaultPrice != 0)
            // returns the default price for a chain
            return chainIdToAirnodeToEndpointToPrice[_chainId].defaultPrice;

        // returns the default price for all airnodes across all chains
        return
            chainIdToAirnodeToEndpointToPrice[0]
                .airnodePrices[address(0)]
                .endpointPrices[bytes32(0)]
                .price;
    }
```

Ashar2shahid: This was something I thought of doing initially, this could be made even simpler if we treat the [0] indexes as the default price so something like this

```
       function getEndpointPrice(
        uint256 _chainId,
        address _airnode,
        bytes32 _endpointId
    ) external view returns(uint256) {
        if (
            chainIdToAirnodeToEndpointToPrice[_chainId][_airnode][_endpointId]
            != 0
        )
            // returns the endpoint price for an airnode in a chain
            return
                chainIdToAirnodeToEndpointToPrice[_chainId][_airnode][_endpointId]
        if (
            airnodeEndpointFlag[_airnode] &&
            chainIdToAirnodeToEndpointToPrice[0][_airnode][_endpointId] != 0
        )
            // returns the endpoint price for an airnode across all chains
            return
                chainIdToAirnodeToEndpointToPrice[0][_airnode][_endpointId]
            // returns the default price for an airnode in a chain
        else
            return
                chainIdToAirnodeToEndpointToPrice[_chainId][_airnode][0];

        if (
            chainIdToAirnodeToEndpointToPrice[0][_airnode][0] != 0
        )
            // returns the default price for an airnode across all chains
            return
                chainIdToAirnodeToEndpointToPrice[0][_airnode][0];
        if (chainIdToAirnodeToEndpointToPrice[_chainId][0][0] != 0)
            // returns the default price for a chain
            return 
                chainIdToAirnodeToEndpointToPrice[_chainId][0][0];

        // returns the default price for all airnodes across all chains
        return
            chainIdToAirnodeToEndpointToPrice[0][0][0]
    }

```

I agree I will probably change it to a variation of the above, one thing to keep in mind is that endpointId can be 0(correct me if I am wrong on this) so thats why I was hesitant to go with this approach, but your approach makes it easier, I would just return a global default price similar to now and keep everything else the same.

Ashar2shahid: So this was the what I ended up with 

```
struct EndpointPrice {
        uint256 price;
    }

    struct AirnodePrice {
        uint256 defaultPrice;
        mapping(bytes32 => EndpointPrice) endpointPrices;
    }

    struct ChainPrice {
        uint256 defaultPrice;
        mapping(address => AirnodePrice) airnodePrices;
    }

    /// @dev A mapping structure used to store all the prices for all the different chain-airnode-endpoint pairs.
    /// Default prices are also stored for the respective chains,airnodes. 
    mapping(uint256 => ChainPrice) public chainIdToAirnodeToEndpointToPrice;

    /// @dev The global default price of an endpoint
    uint256 public defaultPrice;
    
    function getEndpointPrice(
        uint256 _chainId,
        address _airnode,
        bytes32 _endpointId
    ) external view override returns (uint256) {
        if (
            chainIdToAirnodeToEndpointToPrice[_chainId]
                .airnodePrices[_airnode]
                .endpointPrices[_endpointId]
                .price != 0
        )
            // returns the endpoint price for an airnode in a chain
            return
                chainIdToAirnodeToEndpointToPrice[_chainId]
                .airnodePrices[_airnode]
                .endpointPrices[_endpointId]
                .price;

        if (airnodeEndpointFlag[_airnode]) {

            if (
                chainIdToAirnodeToEndpointToPrice[0]
                    .airnodePrices[_airnode]
                    .endpointPrices[_endpointId]
                    .price != 0
            )
                // returns the endpoint price for an airnode across all chains
                return
                    chainIdToAirnodeToEndpointToPrice[0]
                    .airnodePrices[_airnode]
                    .endpointPrices[_endpointId]
                    .price;

            
            if (
                chainIdToAirnodeToEndpointToPrice[_chainId]
                    .airnodePrices[_airnode]
                    .defaultPrice != 0
            )
                // returns the default price of an endpoint on an airnode on a chain
                return
                    chainIdToAirnodeToEndpointToPrice[_chainId]
                        .airnodePrices[_airnode]
                        .defaultPrice;
        
        } else {

            if (
                chainIdToAirnodeToEndpointToPrice[_chainId]
                    .airnodePrices[_airnode]
                    .defaultPrice != 0
            )
                // returns the default price of an endpoint on an airnode on a chain
                return
                    chainIdToAirnodeToEndpointToPrice[_chainId]
                        .airnodePrices[_airnode]
                        .defaultPrice;
            if (
                chainIdToAirnodeToEndpointToPrice[0]
                    .airnodePrices[_airnode]
                    .endpointPrices[_endpointId]
                    .price != 0
            )
                // returns the endpoint price for an airnode across all chains
                return
                    chainIdToAirnodeToEndpointToPrice[_chainId]
                        .airnodePrices[_airnode]
                        .endpointPrices[_endpointId]
                        .price;
        }

        if (
            chainIdToAirnodeToEndpointToPrice[0]
                .airnodePrices[_airnode]
                .defaultPrice != 0
        )
         // returns the default price for an airnode across all chains
            return 
                chainIdToAirnodeToEndpointToPrice[0]
                    .airnodePrices[_airnode]
                    .defaultPrice;
                    
        if (chainIdToAirnodeToEndpointToPrice[_chainId].defaultPrice != 0)
        // returns the default price for a chain
            return chainIdToAirnodeToEndpointToPrice[_chainId].defaultPrice;
        // returns the global default price
        return defaultPrice;
    }
```

Which definitely looks a bit cleaner but there is a problem with this approach in that the mapping inside the structs don't get initialized with [the normal getters](https://docs.soliditylang.org/en/v0.5.7/contracts.html#visibility-and-getters), this means we would have to write our own getters to access the mappings for us to be able to test the values inside. Just felt like a longer workaround than the old way. One other way is how I described above with 0 indexes being treated as default prices

acenolaza: Makes sense, thanks for trying it out anyways 👏🏻 

I think it's fine to go back to the old way since my suggestion didn't really improve the number of txs needed to set this whole structure of fees and storage savings might not be impacted that much. Also, I didn't consider the endpointId possibly being 0.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
```diff
@@ -0,0 +1,231 @@
...
+        require(
+            hasAirnodeFlagAndPriceSetterRole(msg.sender),
+            "Not airnode flag and price setter"
+        );
+        require(_airnode != address(0), "Address is zero");
+        airnodeEndpointFlag[_airnode] = status;
+        emit SetAirnodeEndpointFlag(_airnode, status, msg.sender);
+    }
+
+    /// @dev Called by a global default price setter to set the default price in USD
```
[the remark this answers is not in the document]
Ashar2shahid: yes, I get confused between when to use dev and notice but its probably better to use notice [here](https://docs.soliditylang.org/en/develop/natspec-format.html)

packages/airnode-validator/test/nested/template.json
```diff
@@ -6,4 +6,4 @@
       "__message": "Error in template nested in __prefix"
     }
   }
-}
\ No newline at end of file
+}
```
Ashar2shahid, line 9: Not sure how this showed up

Ashar2shahid:
Ready for review, Added tests checking everything I could think of.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistryRolesWithManager.sol
```diff
@@ -0,0 +1,162 @@
...
+    AccessControlClient,
+    IAirnodeFeeRegistryRolesWithManager
+{
+    // There are three roles implemented in this contract:
+    // Root
+    // └── (1) Admin (can grant and revoke the roles below)
+    //     ├── (2) Global Default Price setter
+    //     └── (3) Airnode Flag and Price setter
+    // Their IDs are derived from the descriptions below. Refer to
+    // AccessControlRegistry for more information.
```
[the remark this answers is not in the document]
Ashar2shahid: The reason why I didn't go for an owner was approach because it is highly likely that multiple addresses are going to be price setters i.e we don't want only 1 address to be able to control the price so we need to be able to have multiple owners which we could have done via a proxy to an only owner AirnodeFeeRegistry but then thats still a workaround. We might not need two roles but we definently need 1 role that can be assigned to muliple people.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
```diff
@@ -0,0 +1,234 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+import "./AirnodeFeeRegistryRolesWithManager.sol";
+import "./interfaces/IAirnodeFeeRegistry.sol";
+
+/// @title Contract for storing the price of a airnode endpoints
```
acenolaza, line 7:
```suggestion
/// @title Contract for storing the price of an airnode endpoints
```

packages/airnode-protocol/contracts/rrp/authorizers/admin/interfaces/IAirnodeFeeRegistry.sol
```diff
@@ -0,0 +1,79 @@
...
+        address airnodeFlagAndPriceSetter
+    );
+
+    event SetChainAirnodeEndpointPrice(
+        uint256 chainId,
+        address airnode,
+        bytes32 endpointId,
+        uint256 price,
+        address airnodeFlagAndPriceSetter
+    );
```
acenolaza, line 45: At least `chaindId`, `airnode` and `endpointId` should be indexed in order to have them available in the "topics" for filtering.
https://docs.soliditylang.org/en/v0.8.9/contracts.html#events

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
```diff
@@ -0,0 +1,234 @@
...
+
+    /// @notice The global default price of an endpoint
+    uint256 public defaultPrice;
+
+    /// @notice A flag that indicates which price to default to in
+    /// the getAirnodeEndpointFee function
+    mapping(address => bool) public airnodeEndpointFlag;
+
+    /// @notice Prices will have upto 6 decimal places
+    uint8 public immutable decimals = 6;
```
acenolaza, line 41: Why is this state var immutable if we are not initializing it in the constructor? Are we planning to do that or should it just be a constant?

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
```diff
@@ -0,0 +1,231 @@
...
+        address _manager
+    )
+        AirnodeFeeRegistryRolesWithManager(
+            _accessControlRegistry,
+            _adminRoleDescription,
+            _manager
+        )
+    {}
+
+    function setAirnodeEndpointFlag(address _airnode, bool status)
```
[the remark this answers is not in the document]
acenolaza:
```suggestion
    function setAirnodeEndpointFlag(address _airnode, bool _status)
```
natspec should also be updated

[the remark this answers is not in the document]
acenolaza:
```suggestion
    function setAirnodeEndpointFlag(address _airnode, bool _status)
```
natspec should also be updated

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
```diff
@@ -0,0 +1,234 @@
...
+            hasGlobalDefaultPriceSetterRoleOrIsManager(msg.sender),
+            "Not global default price setter"
+        );
+        require(_chainId != 0, "ChainId is zero");
+        require(_price != 0, "Price is zero");
+        defaultChainPrice[_chainId] = _price * 10**decimals;
+        emit SetDefaultChainPrice(_chainId, _price, msg.sender);
+    }
+
+    /// @notice Called by an airnode flag and price setter to set the default price on a chain in USD
```
acenolaza, line 103:
```suggestion
    /// @notice Called by an airnode flag and price setter to set the default price on an airnode in USD
```

acenolaza (commented):
@Ashar2shahid please note that branch is failing to build due to contracts importing different other contracts with different pragma version.

packages/airnode-protocol/contracts/rrp/authorizers/admin/interfaces/IAirnodeFeeRegistry.sol
```diff
@@ -0,0 +1,79 @@
...
+
```
acenolaza, line 46: We are missing all the view fuctions for the public state vars here and we should add override to all them after adding these functions.

Ashar2shahid: I've added them but do we normally implement view functions for state vars aren't they alreadt declared when declaring a public variable?

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
Ashar2shahid: changed a bunch of comments, can have a look again

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
Ashar2shahid: It was going to be initialized in the constructor but its better to have it be a constant because we don't want it to change in the future

packages/airnode-protocol/contracts/rrp/authorizers/admin/interfaces/IAirnodeFeeRegistry.sol
Ashar2shahid: Agreed

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
Ashar2shahid: changed a bunch of comments can have a look again

packages/airnode-protocol/contracts/rrp/authorizers/admin/interfaces/IAirnodeFeeRegistry.sol
Ashar2shahid: Also have indexed key events in the  AirnodeTokenLock which I think will be needed when calculating provider payments off chain.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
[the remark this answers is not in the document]
Ashar2shahid: fixed

packages/airnode-protocol/contracts/rrp/authorizers/admin/interfaces/IAirnodeFeeRegistry.sol
acenolaza: It's something I started seeing being done by Burak in the Access Control Registry implementation.

acenolaza: I guess we should do the same for IAirnodeTokenLock.sol

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
```diff
@@ -0,0 +1,495 @@
...
+    }
+
+    /// @notice Called by the airnode to set the opt status for itself
+    /// @param _airnode The airnode address
+    /// @param _status The Opted status for the airnode
+    function setSelfOptOutStatus(address _airnode, bool _status)
+        external
+        override
+    {
+        require(msg.sender == _airnode, ERROR_NOT_AIRNODE);
```
acenolaza, line 215: This check is a little bit weird isn't it? 

Maybe what I would do instead is fetch the `airnodeOptInStatus` by `msg.sender` and check that is `true` since that is a requirement.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
```diff
@@ -0,0 +1,495 @@
...
+        );
+    }
+
+    /// @notice Called by a block requester role to set the Block status of a airnode-requester pair.
+    /// A blocked requester can have its Locks be withdrawn into the address specified
+    /// by the blockWithdrawDestination setter. This is an irreversible action.
+    /// @dev A requester blocked on address(0) is blocked globally
+    /// @param _airnode The airnode address
+    /// @param _requesterAddress requester address
+    function blockRequester(address _airnode, address _requesterAddress)
```
acenolaza, line 393: Isn't it weird that blocking a requester doesn't take the whitelisting out? It only prevents the unstake but the requester would still be able to make requests until withdrawBlocked is called.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
```diff
@@ -0,0 +1,495 @@
...
+    /// @notice Returns the current amount a locker has to
+    /// lock for a given chainId-airnode-endpointId
+    /// @param _chainId The id of the chain
+    /// @param _airnode The airnode address
+    /// @param _endpointId The endpointId
+    function getLockAmount(
+        uint256 _chainId,
+        address _airnode,
+        bytes32 _endpointId
+    ) public view override returns (uint256 _lockAmount) {
```
acenolaza, line 472: This question I'm not sure about but do we need underscore for the return args as well?

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
```diff
@@ -0,0 +1,495 @@
...
+    mapping(uint256 => mapping(address => mapping(bytes32 => mapping(address => TokenLocks))))
+        public tokenLocks;
+
+    /// @notice Stores information for blocked airnode-requester pair
+    ///      Stores true if blocked, false otherwise
+    mapping(address => mapping(address => bool))
+        public airnodeToRequesterToBlockStatus;
+
+    /// @notice The Address to which blocked tokens will be withdrawn to
+    address public blockWithdrawDestination;
```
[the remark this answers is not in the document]
Ashar2shahid: Yes, we are not expecting the address to change very often. One of the requirments of withdrawBlocked was that anyone is able to call it which is why we can't have withdrawBlocked be able to specify the address.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
```diff
@@ -0,0 +1,495 @@
...
+    /// @param _requesterAddress The requester address
+    modifier isNotBlocked(address _airnode, address _requesterAddress) {
+        require(
+            !airnodeToRequesterToBlockStatus[address(0)][_requesterAddress],
+            ERROR_REQUESTER_BLOCKED
+        );
+        require(
+            !airnodeToRequesterToBlockStatus[_airnode][_requesterAddress],
+            ERROR_REQUESTER_BLOCKED
+        );
```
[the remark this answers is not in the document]
Ashar2shahid: Yes this makes sense but its probably an || instead of &&

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
Ashar2shahid: both of these statuses are mutually exclusive, if an airnode is not opted in by an admin , we still want to give the airnode the ability to opt out incase they are opted in without consent.  Only the airnode needs to be able to set this status so the require makes sense or is there anything else I am missing?

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
Ashar2shahid: unfourtunately to take the whitelisting out we need the endpointId and blocking on specific endpoints seem redundant. Which is why we had to do it through withdrawBlocked.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
Ashar2shahid: Hm probably not, Ill change it

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
acenolaza: > if an airnode is not opted in by an admin

because of that I made the assumption that `isOptedIn()` will return `false` therefore it's like this opted out value is only meaningful when and admin has previously opted in the airnode. Maybe I'm the one missing something tho 😄

packages/airnode-protocol/contracts/rrp/authorizers/admin/interfaces/IAirnodeFeeRegistry.sol
Ashar2shahid: Ill have to get back on this

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
Ashar2shahid: haha, its more like this opted out value is a failsafe for the airnode to not mistakenly become part of the subscription model. 

Another use case is if they are already opted in, then they can choose to opt out temporarily and then join back in when they want.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
```diff
@@ -0,0 +1,240 @@
...
+        external
+        override
+    {
+        require(
+            hasGlobalDefaultPriceSetterRoleOrIsManager(msg.sender),
+            "Not global default price setter"
+        );
+        require(_chainId != 0, "ChainId is zero");
+        require(_price != 0, "Price is zero");
+        defaultChainPrice[_chainId] = _price * 10**DECIMALS;
```
bbenligiray, line 104: - `_price` will already have been multiplied by `10**DECIMALS`, don't do it again
- There are no `chainId` and `_price` parameters in the scope so you can use them, no need for the leading underscores.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
```diff
@@ -0,0 +1,240 @@
...
+
+    /// @notice The default price of an endpoint for a chain
+    mapping(uint256 => uint256) public override defaultChainPrice;
+
+    /// @notice The global default price of an endpoint
+    uint256 public override defaultPrice;
+
+    /// @notice A flag that indicates which price to default to in
+    /// the getAirnodeEndpointFee function
+    mapping(address => bool) public override airnodeEndpointFlag;
```
bbenligiray, line 40: Call this `prioritizeEndpointPriceOverChainPrice`

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
```diff
@@ -0,0 +1,240 @@
...
+        public
+        override airnodeToEndpointToPrice;
+
+    /// @notice The default price of an endpoint for an airnode on a specific chain
+    mapping(uint256 => mapping(address => uint256))
+        public
+        override defaultChainAirnodePrice;
+
+    /// @notice The default price of an endpoint on an airnode
+    mapping(address => uint256) public override defaultAirnodePrice;
```
bbenligiray, line 30: `airnodeToPrice`

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
```diff
@@ -0,0 +1,240 @@
...
+
+    /// @notice The USD price of an endpoint for an airnode across all chains
+    mapping(address => mapping(bytes32 => uint256))
+        public
+        override airnodeToEndpointToPrice;
+
+    /// @notice The default price of an endpoint for an airnode on a specific chain
+    mapping(uint256 => mapping(address => uint256))
+        public
+        override defaultChainAirnodePrice;
```
bbenligiray, line 27: `chainToAirnodeToPrice`

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLockRolesWithManager.sol
```diff
@@ -0,0 +1,337 @@
...
+contract AirnodeTokenLockRolesWithManager is
+    RoleDeriver,
+    AccessControlClient,
+    IAirnodeTokenLockRolesWithManager
+{
+    // There are Six roles implemented in this contract:
+    // Root
+    // └── (1) Admin (can grant and revoke the roles below)
+    //     ├── (2) Oracle address setter
+    //     ├── (3) coefficient and registry setter
```
bbenligiray, line 20: I can imagine a case where we can expect an address to need to be able to update the coefficient but not the fee registry addresss, so I think this should be broken down for consistency

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLockRolesWithManager.sol
```diff
@@ -0,0 +1,337 @@
...
+        requesterAuthorizerWithManagerSetterRole = _deriveRequesterAuthorizerWithManagerSetterRole(
+            _manager
+        );
+    }
+
+    /// @notice Derives the admin role for the specific manager address
+    /// @param _manager Manager address
+    /// @return _adminRole Admin role
+    function _deriveAdminRole(address _manager)
+        internal
```
bbenligiray, line 128: These also should be `private`

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
```diff
@@ -0,0 +1,492 @@
...
+    /// @notice Stores information for blocked airnode-requester pair
+    ///      Stores true if blocked, false otherwise
+    mapping(address => mapping(address => bool))
+        public airnodeToRequesterToBlockStatus;
+
+    /// @notice The Address to which blocked tokens will be withdrawn to
+    address public blockWithdrawDestination;
+
+    /// @notice mapping used to store all the oracle addresses
+    mapping(address => bool) public isOracle;
```
bbenligiray, line 76: This should be implemented as a role adminned by `oracleAddressSetterRole`

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
```diff
@@ -0,0 +1,492 @@
...
+            tokenLock.lockerToLockAmount[msg.sender] == 0,
+            ERROR_ALREADY_LOCKED
+        );
+
+        uint256 lockAmount = getLockAmount(_chainId, _airnode, _endpointId);
+
+        require(
+            IERC20(api3Token).balanceOf(msg.sender) >= lockAmount,
+            ERROR_INSUFFICIENT_AMOUNT
+        );
```
bbenligiray, line 295: This is redundant

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
```diff
@@ -0,0 +1,492 @@
...
+                );
+        }
+
+        assert(
+            IERC20(api3Token).transferFrom(
+                msg.sender,
+                address(this),
+                lockAmount
+            )
+        );
```
bbenligiray, line 317: We can `require()` with an error message here assuming one can use another ERC20 implementation instead of API3 where `transferFrom` returns `false` instead of reverting

packages/protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
```diff
@@ -0,0 +1,381 @@
...
+            IDaoRequesterRrpAuthorizer(
+                chainIdToDaoRequesterRrpAuthorizer[_chainId]
+            ).setWhitelistExpiration(
+                    _airnode,
+                    _endpointId,
+                    _requesterAddress,
+                    type(uint64).max
+                );
+        }
+
```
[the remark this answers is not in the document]
bbenligiray: Nah, the frontend is supposed to do the allowance check and not let the user make the transaction

packages/airnode-protocol/contracts/rrp/authorizers/admin/interfaces/IAirnodeFeeRegistry.sol
bbenligiray: If you import an interface to another contract and the public state variable is not declared as a view function in the interface, you won't be able to read them

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
bbenligiray: This should be implemented like
```
function setSelfOptOutStatus(bool _status)
        external
        override
    {
        airnodeSelfOptOutStatus[msg.sender] = _status;
        emit SetSelfOptOutStatus(msg.sender _status);
    }
```

bbenligiray:
The logic looks good apart from one place where a role needs to be broken into two. I think some variable names need to be changed and a lot more/repeated documentation is needed in some places. So I'd like to see this merged as soon as possible (ideally today) so that I can polish it. Also it feels like this is a bit too complicated to only go by with the in-code documentation while auditing, so I'd like to see it being documented in the README, including the price logic diagram.

Ashar2shahid:
>  So I'd like to see this merged as soon as possible (ideally today) so that I can polish it. Also it feels like this is a bit too complicated to only go by with the in-code documentation while auditing, so I'd like to see it being documented in the README, including the price logic diagram.

Sounds good, I'll try to get these out later today

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
Ashar2shahid: isn't that what's happening [here ](https://github.com/api3dao/airnode/blob/2584aa6fb3fb2ec419801e5d68fafb309952b39a/packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol#L158-L166)?

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
Ashar2shahid: api3Token is immutable so its not possible that some other token is being used unless the deployer put in the wrong address

acenolaza (approved):

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeRequesterAuthorizerRegistryClient.sol
```diff
@@ -0,0 +1,21 @@
...
+pragma solidity 0.8.9;
+
+import "./interfaces/IAirnodeRequesterAuthorizerRegistryClient.sol";
+
+contract AirnodeRequesterAuthorizerRegistryClient is
+    IAirnodeRequesterAuthorizerRegistryClient
+{
+    /// @notice Address of the AirnodeRequesterAuthorizerRegistry contract that keeps the
+    /// RequesterAuthorizerWithManager addresses for different chains.
+    address public immutable override airnodeRequesterAuthorizerRegistry;
```
acenolaza, line 11: Should this really be immutable? I'm wondering if it would ever make sense to set a new  registry contract address in the future via setter function

acenolaza: Also, we do have a setter for AirnodeFeeRegistry in AirnodeTokenLock

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
```diff
@@ -0,0 +1,232 @@
...
+    uint256 public override defaultPrice;
+
+    /// @notice A flag that indicates which price to default to in
+    /// the getAirnodeEndpointFee function
+    mapping(address => bool)
+        public
+        override prioritizeEndpointPriceOverChainPrice;
+
+    /// @notice Prices will have upto 18 decimal places
+    uint8 public constant DECIMALS = 18;
```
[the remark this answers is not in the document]
Ashar2shahid: just to give context, doesn't hurt :D

packages/airnode-protocol/README.md
```diff
@@ -125,6 +125,16 @@ The contracts are under the `contracts/` directory.
     recent response for each template is stored and can be accessed by the whitelisted users where the whitelists are
     managed by a single account.
 
+  `rrp/authorizers/admin`: Houses the token locking and fee registry contacts
+
+  - `admin/AirnodeTokenLock.sol`: A contract that Requesters will use to get authorized by the RequesterAuthorizer by locking in API3 tokens
```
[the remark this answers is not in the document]
Ashar2shahid: Requesters get authorized , sponsor/locker locks token for a requester

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenLock.sol
bbenligiray: "api3Token" may not be API3 but a different ERC20

PR #640, opened 2021-10-29
Bump contract dependencies

bbenligiray:
I figured it would be a good idea to bump the OpenZeppelin contracts package version after Ashar pointed out that AccessControl.sol, but that seems to have happened in 4.4.0 which is not released yet. So this PR shouldn't really change anything.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -22,7 +22,7 @@
     {
       "files": "*.sol",
       "options": {
-        "compiler": "0.8.6",
+        "compiler": "0.8.9",
         "printWidth": 80,
         "tabWidth": 4,
         "useTabs": false,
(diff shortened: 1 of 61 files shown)


bbenligiray:
> Are we planning to replace _setupRole() with _grantRole() after 4.4 is released? if so, shouldn't we replace it now?

I don't think so because I don't want to redeploy contracts only to use 4.4 instead of 4.3. We can't replace it now because in 4.3 _grantRole() is still private and _setupRole() is its internal version.

bbenligiray:
I'm force-merging to be able to release, refer to this to see that there is not much difference
https://github.com/OpenZeppelin/openzeppelin-contracts/compare/release-v4.2...release-v4.3

PR #656, opened 2021-11-02
AirnodeTokenPayment

Ashar2shahid:
Initial Draft. Access control and tests needs to be implemented but the core logic is here.


packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeStablecoinPayment.sol
```diff
@@ -0,0 +1,290 @@
...
+    /// @notice Locks API3 Tokens to gain access to Airnodes.
+    /// @dev chainId-airnode-endpoint-requester pair gets authorized as long as there is
+    /// at least one token lock for given pair.
+    /// @notice The amount to be locked is determined by the fee set in the AirnodeFeeRegistry Contract
+    /// @param _stablecoin The address of the stablecoin token contract
+    /// @param _chainId The id of the chain
+    /// @param _airnode The airnode address
+    /// @param _endpointId The endpointId
+    /// @param _requesterAddress The address of the requester for which tokens are being locked
+    function makePayment(
```
Ashar2shahid, line 183: Running into Stack is too deep here, any ideas on how to get around this?

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeStablecoinPayment.sol
```diff
@@ -0,0 +1,290 @@
...
+        uint64 _days
+    ) public view override returns (uint256 paymentAmount) {
+        uint256 endpointFee = IAirnodeFeeRegistry(airnodeFeeRegistry)
+            .getEndpointPrice(_chainId, _airnode, _endpointId);
+        uint8 endpointFeeDecimals = IAirnodeFeeRegistry(airnodeFeeRegistry)
+            .decimals();
+        uint16 endpointInterval = IAirnodeFeeRegistry(airnodeFeeRegistry)
+            .interval();
+        uint8 erc20Decimals = IERC20Extended(_stablecoin).decimals();
+        paymentAmount =
```
Ashar2shahid, line 285: I did it this way in case we end up changing the interval / endpointfee decimals of the airnodeFeeRegistry contract (if we deploy a new one) this will not be affected. wdyt

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeStablecoinPayment.sol
```diff
@@ -0,0 +1,290 @@
...
+    /// @notice Called by the airnode to set the opt status for itself
+    /// @param _airnode The airnode address
+    /// @param _status The Opted status for the airnode
+    function setSelfOptInStatus(address _airnode, bool _status)
+        external
+        override
+    {
+        require(msg.sender == _airnode, ERROR_NOT_AIRNODE);
+        airnodeSelfOptInStatus[_airnode] = _status;
+        if (_status && airnodePaymentAddress[_airnode] == address(0)) {
```
Ashar2shahid, line 133: first time airnode opts in and hasn't set the payment address. The airnode itself is the payment address

packages/airnode-protocol/contracts/rrp/authorizers/admin/interfaces/IERC20Extended.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+interface IERC20Extended is IERC20 {
```
Ashar2shahid, line 4: Standard ERC20 doesn't let you access the decimals :/

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeStablecoinPayment.sol
```diff
@@ -0,0 +1,289 @@
...
+    string private constant ERROR_NOT_AIRNODE = "Not airnode";
+    string private constant ERROR_ZERO_AMOUNT = "Zero amount";
+    string private constant ERROR_INSUFFICIENT_AMOUNT = "Insufficient amount";
+    string private constant ERROR_AIRNODE_NOT_OPTED_IN = "Airnode not opted in";
+
+    /// @notice Address of AirnodeFeeRegistry
+    address public airnodeFeeRegistry;
+
+    /// @notice The maximum whitelisting duration in seconds
+    uint64 public maximumWhitelistingDuration;
```
acenolaza, line 22: I can't seem to find where this is being set 🤔

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeStablecoinPayment.sol
```diff
@@ -0,0 +1,289 @@
...
+    mapping(uint256 => address) public chainIdToRequesterAuthorizerWithManager;
+
+    constructor(address _airnodeFeeRegistry) {
+        require(_airnodeFeeRegistry != address(0), ERROR_ZERO_ADDRESS);
+        airnodeFeeRegistry = _airnodeFeeRegistry;
+    }
+
+    /// @notice Reverts if the airnode is not opted in
+    /// @param _airnode The airnode Address
+    modifier isOptedIn(address _airnode) {
```
acenolaza, line 49: Nit: we might not need this modifier since we are using it in a single function

acenolaza: same goes for `isSupportedERC20`

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeStablecoinPayment.sol
```diff
@@ -0,0 +1,289 @@
...
+    /// transfer for a given chainId-airnode-endpointId
+    /// @param _chainId The id of the chain
+    /// @param _airnode The airnode address
+    /// @param _endpointId The endpointId
+    function getPaymentAmount(
+        address _stablecoin,
+        uint256 _chainId,
+        address _airnode,
+        bytes32 _endpointId,
+        uint64 _days
```
acenolaza, line 276: Since this is a public function shouldn't we also add checks like the following one here:

```
        require(
            _days * 1 days <= maximumWhitelistingDuration,
            "Exceed maximum whitelisting"
        );
```

acenolaza (commented):
* all the roles check are missing in the setters
* public state vars getters are missing in the interface

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeStablecoinPayment.sol
```diff
@@ -0,0 +1,289 @@
...
+
+    /// @notice Mapping to store the default payment address for an airnode
+    mapping(address => address) public airnodePaymentAddress;
+
+    /// @notice Mapping to store the default supported stablecoins
+    mapping(uint256 => mapping(address => bool)) public defaultSupportedERC20;
+
+    /// @notice Mapping to store the supported stablecoins for an airnode
+    mapping(uint256 => mapping(address => mapping(address => bool)))
+        public airnodeSupportedERC20;
```
acenolaza, line 32: I'm now confused about why we need to have chainId in these to mappings if we are only going to support payments in mainnet stablecoins 🤔 

I mean if we are always going to make payments to an address in mainnet (airnodePaymentAddress) why do we need to check by `chainId` in the `isSupportedERC20()` modifier?

Ashar2shahid: Ah yes, we went over this yesterday. Ill change this

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeStablecoinPayment.sol
Ashar2shahid: Hm maybe, Its more to get the price more than anything, so I don't think that check is really necessary

packages/airnode-protocol/contracts/rrp/authorizers/admin/interfaces/IAirnodeStablecoinPayment.sol
```diff
@@ -0,0 +1,87 @@
...
+    ) external;
+
+    function setAirnodeSupportedERC20(
+        uint256 chainId,
+        address airnode,
+        address stablecoin,
+        bool status
+    ) external;
+
+    function setSelfOptInStatus(address _airnode, bool _status) external;
```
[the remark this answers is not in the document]
Ashar2shahid: No, these were typos. I wanted to get the draft out to get feedback on the logic before I pushed the full PR and fixed the comments/typos. Especially because I couldn't get the stack too deep error resolved.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeStablecoinPayment.sol
Ashar2shahid: yeah there is a setter function for this with its own access control role in the next commit

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeStablecoinPayment.sol
```diff
@@ -0,0 +1,289 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+import "../../../authorizers/interfaces/IRequesterAuthorizerWithManager.sol";
+import "./interfaces/IAirnodeFeeRegistry.sol";
+import "./interfaces/IERC20Extended.sol";
+import "./interfaces/IAirnodeStablecoinPayment.sol";
+
+/// @title The contract used to lock API3 Tokens in order to gain access to Airnodes
```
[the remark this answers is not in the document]
Ashar2shahid: make payments in stablecoin tokens to providers directly in order to gain access to Airnodes? I was confused on what wording to use

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeStablecoinPayment.sol
```diff
@@ -0,0 +1,289 @@
...
+            _stablecoin,
+            _status,
+            msg.sender
+        );
+    }
+
+    /// @notice Called by the airnode to set the opt status for itself
+    /// @param _airnode The airnode address
+    /// @param _status The Opted status for the airnode
+    function setSelfOptInStatus(address _airnode, bool _status)
```
bbenligiray, line 127: I don't like that stablecoin payments require a transaction from the node operator

bbenligiray (commented):
This contract has a bunch of functions that need to be adminned (adding token addresses, authorizer addresses, etc.). I wonder this should also be "managed" similar to the TokenLock contract.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistryClient.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+import "./interfaces/IAirnodeFeeRegistryClient.sol";
+
+contract AirnodeFeeRegistryClient is IAirnodeFeeRegistryClient {
+    /// @notice Address of the AirnodeFeeRegistry contract that keeps the price
+    /// of airnode endpoints across different chains
+    address public override airnodeFeeRegistry;
```
acenolaza, line 9: Not immutable because AirnodeTokenLock.sol has a setter

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
```diff
@@ -44,7 +44,11 @@ contract AirnodeFeeRegistry is
         override prioritizeEndpointPriceOverChainPrice;
 
     /// @notice Prices will have upto 18 decimal places
-    uint8 public constant DECIMALS = 18;
+    uint8 public constant override DECIMALS = 18;
+
+    /// @notice Prices are represented in seconds for 30 day intervals
+    /// @dev interval will be counted in days
+    uint24 public constant override INTERVAL = 30 days;
```
acenolaza, line 51: I'm thinking that maybe we may need to make this value configurable per Airnode and default to this constant if none is specified.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
```diff
@@ -0,0 +1,282 @@
...
+        );
+
+        uint256 amount = getPaymentAmount(
+            _chainId,
+            _airnode,
+            _endpointId,
+            expirationTimestamp > uint64(block.timestamp)
+                ? (uint64(block.timestamp) + _whitelistDuration) -
+                    expirationTimestamp
+                : _whitelistDuration
```
acenolaza, line 199: We should only charge for the delta between the existing expiration and current whitelist duration.

acenolaza: I'm now thinking that this should go inside the `getPaymentAmount()` function 🤔

acenolaza:
@Ashar2shahid this should be ready for review now

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeFeeRegistry.sol
Ashar2shahid: Yeah I think that would probably be better considering we are having everything else configurable.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
```diff
@@ -0,0 +1,286 @@
...
+                    _requesterAddress
+                );
+        require(
+            indefiniteWhitelistCount == 0,
+            "Already whitelisted indefinitely"
+        );
+        require(
+            uint64(block.timestamp) + _whitelistDuration > expirationTimestamp,
+            "Already whitelisted"
+        );
```
Ashar2shahid, line 194: So just to be clear the way it is currently we require `now + whitelistduration > expirationTimestamp` i.e  if the user whitelisted for 3 months, we then attempt to whitelist for 1 month it would revert. The way it should work is if the user is whitelisted for 3 months and we attempt to whitelist for 1 month it would extend the whitelisting by 1 month to 4 months. Obviously we can extend only till the maximum limit after which it would revert.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
Ashar2shahid: if we go by what I mentioned above we won't need to do this and just have to calculate for `_whitelistDuration`

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
```diff
@@ -0,0 +1,286 @@
...
+
+        emit MadePayment(
+            _chainId,
+            _airnode,
+            _endpointId,
+            _requesterAddress,
+            msg.sender,
+            getAirnodeToPaymentDestination(_airnode),
+            amount,
+            IERC20Metadata(paymentTokenAddress).symbol(),
```
Ashar2shahid, line 230: I think address is smaller than symbol so might be better to use that?

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
```diff
@@ -0,0 +1,286 @@
...
+    /// @notice The address of the ERC20 token that is used to pay for the
+    /// Airnode whitelisting
+    address public immutable paymentTokenAddress;
+
+    /// @notice The price of the ERC20 token in USD
+    /// @dev The price is used to calculate the amount of tokens that are
+    /// required to be paid and it is defaulted to 1 for stable coin tokens
+    /// but if the token has a different price then an oracle must call
+    /// setPaymentTokenPrice() in order to keep the price up to date
+    uint256 public paymentTokenPrice = 1;
```
Ashar2shahid, line 37: this will be 1 * 10**18

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
```diff
@@ -0,0 +1,286 @@
...
+        uint256 _chainId,
+        address _airnode,
+        bytes32 _endpointId,
+        uint64 _whitelistDuration
+    ) public view override returns (uint256 amount) {
+        uint256 feeInUsd = IAirnodeFeeRegistry(airnodeFeeRegistry)
+            .getEndpointPrice(_chainId, _airnode, _endpointId);
+        uint24 feeInterval = IAirnodeFeeRegistry(airnodeFeeRegistry).INTERVAL();
+        amount =
+            (feeInUsd * _whitelistDuration) /
```
Ashar2shahid, line 254: Im thinking about this more and it seems like we have to multiply this with decimal places of erc20 because otherwise the payment amount isn't right
```
feeInUsd = 100 * 10**18
whitelistduration = 100
paymentTokenPrice = 1 * 10**18
feeInterval = 1000

amount = (100 * 10**18) * 100 / ((1 * 10**18) * 1000) = 10 

```

but it should give 10 * (decimal places of erc20)

important to keep the decimal places because different erc20s have different decimal places.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPaymentRolesWithManager.sol
```diff
@@ -0,0 +1,218 @@
...
+contract AirnodeTokenPaymentRolesWithManager is
+    RoleDeriver,
+    AccessControlClient,
+    IAirnodeTokenPaymentRolesWithManager
+{
+    // There are 3 roles implemented in this contract:
+    // Root
+    // └── (1) Admin (can grant and revoke the roles below)
+    //     ├── (2) Payment token price setter (Oracle)
+    //     ├── (3) Airnode to minimum whitelist duration setter
```
Ashar2shahid, line 20: can probably change this to whitelist duration setter for both minimum and maximum if we do end up setting both

packages/airnode-protocol/contracts/rrp/authorizers/admin/interfaces/IERC20Extended.sol
Ashar2shahid, line 4: I thought this was not needed anymore because of the openzepplin meta contract?

Ashar2shahid (commented):
Some changes have to be made, after which I will review the tests.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    /// @notice Called by an Airnode to minimum whitelist duration setter to
+    /// set the minimum whitelisting duration for an Airnode
+    /// @param _minimumWhitelistDuration Minimum whitelist duration in seconds
+    function setAirnodeToMinimumWhitelistDuration(
+        uint64 _minimumWhitelistDuration
+    ) external override {
+        require(
+            hasAirnodeToMinimumWhitelistDurationSetterRoleOrIsManager(
```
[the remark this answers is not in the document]
acenolaza: The thing is that we stopped using constants for error messages a while ago so in order to be consistent I guess I should just make it shorter.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
```diff
@@ -0,0 +1,286 @@
...
+        uint64 _whitelistDuration
+    ) external override {
+        require(_chainId != 0, "Invalid chainId");
+        require(_airnode != address(0), "Invalid Airnode address");
+        require(_requesterAddress != address(0), "Invalid requester address");
+        require(
+            _whitelistDuration != 0 &&
+                _whitelistDuration <= DEFAULT_MAXIMUM_WHITELIST_DURATION &&
+                _whitelistDuration >= getMinimumWhitelistDuration(_airnode),
+            "Invalid whitelist duration"
```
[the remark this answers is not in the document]
acenolaza: At first I thought the same but I then I thought of the use case and realized that an Airnode operator would probably only want to set the minimum (instead of charging by day they might want to charge by month minimum, etc). I could add a setter to the maximum too but that would require a new role that might never be used.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
```diff
@@ -0,0 +1,286 @@
...
+        );
+
+        address requesterAuthorizerWithManager = IAirnodeRequesterAuthorizerRegistry(
+                airnodeRequesterAuthorizerRegistry
+            ).chainIdToRequesterAuthorizerWithManager(_chainId);
+        require(
```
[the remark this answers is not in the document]
acenolaza: The `chainIdToRequesterAuthorizerWithManager()` function is an auto-generated function because the mapping is set to public. What we could do is create a convenience function to get the authorizer and check the require and then call it from both AirnodeTokenLock and AirnodeTokenPayment.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
acenolaza: Good catch 🥇

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
acenolaza: True. I'll fix it by fetching the decimals from the payment token like we were doing on the initial draft of this contract.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPaymentRolesWithManager.sol
acenolaza: This might actually be a good idea to avoid creating a new role 👍🏻 if we do add the setter.

packages/airnode-protocol/contracts/rrp/authorizers/admin/interfaces/IERC20Extended.sol
acenolaza: I probably deleted the implementation and forgot to delete the interface. Good catch.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
acenolaza: The think here is that in the event I think it might be more useful to have the symbol instead. Don't you agree? Should I add both maybe?

acenolaza:
@Ashar2shahid you should now be able continue reviewing this PR now

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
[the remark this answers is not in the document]
Ashar2shahid: Hmm yeah maybe that would be better

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPaymentRolesWithManager.sol
```diff
@@ -0,0 +1,216 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+import "../../../access-control-registry/RoleDeriver.sol";
+import "../../../access-control-registry/AccessControlClient.sol";
+import "../../../access-control-registry/interfaces/IAccessControlRegistry.sol";
+import "./interfaces/IAirnodeTokenPaymentRolesWithManager.sol";
+
+/// @title Contract that implements generic AccessControlRegistry roles for
+/// the AirnodePaymentLock contract
```
Ashar2shahid, line 10: AirnodeTokenPayment*

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
Ashar2shahid: this seem alright, I guess I have just one small caveat with this approach, right now we assume that the fee we receive form the AirnodeFeeRegistry contract is implicitly going to be 18 places. If for some reason we plan to deploy a new contract with different decimal places , it will break this logic so for that reason what we can do is multiply by 1e18 to cancel the paymentToken price decimals and divide by the decimals from the feeRegistry contract .

Ashar2shahid: Maybe this is not even needed tbh...

Ashar2shahid (commented):
Everything looks good to me just added some small comments. I made a pass of the tests and they seem alright, will do one more pass tomorrow but this is approved from my side (can't approve my own pull request 😅  ) .

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
acenolaza: Right, but since in the last protocol we'd decided that everything will be expressed in 18 decimals I made that assumption that fee and price would always be in 18 decimal places. Also, now that I think about it I don't even need to use the token decimals in the formula 🤔 I could just have a multiplier coefficient and multiply the numerator by that coefficient in order to make sure I get 18 decimals back from the integer division.

Ashar2shahid: I don't think I am completely following but essentially the payment will be made in using the erc20 decimals, so if an erc20 has 6 decimals and we end up using 18 decimal places then we will be specifying the wrong amount by a large magnitude.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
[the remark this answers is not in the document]
acenolaza: I've changed my mind and I think it's better to keep the error handling on the caller or the `chainIdToRequesterAuthorizerWithManager` function for now.

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
acenolaza: I think I now get what you mean 👍🏻

acenolaza:
> Everything looks good to me just added some small comments. I made a pass of the tests and they seem alright, will do one more pass tomorrow but this is approved from my side (can't approve my own pull request sweat_smile ) .
> 
> One more thing is @acenolaza you should probably also update the README similar to the AirnodeTokenLock changes.

Updated the README and added a new test for paying with a contract that only has 6 decimals (Mock USDC stablecoin contract)

I'm feeling pretty confident that this could be merge tomorrow if you agree @Ashar2shahid

packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
acenolaza: > Right, but since in the last protocol we'd decided that everything will be expressed in 18 decimals I made that assumption that fee and price would always be in 18 decimal places. Also, now that I think about it I don't even need to use the token decimals in the formula thinking I could just have a multiplier coefficient and multiply the numerator by that coefficient in order to make sure I get 18 decimals back from the integer division.

Implemented it [here](https://github.com/api3dao/airnode/pull/656/commits/2ea56225ee6cbdccbb1ce7ad86c290a27c1c3a6b)

acenolaza (approved):
Approving my own PR 🤭

Ashar2shahid:
Went through the tests again and everything looks good. Shouldn't be a problem to merge this now

PR #658, opened 2021-11-02
Add mnemonic generation

Siegrift:
https://api3dao.atlassian.net/browse/AN-277


package.json
```diff
@@ -42,6 +42,7 @@
     "dev:list": "(cd packages/airnode-operation && yarn run dev:list)",
     "dev:stop": "(cd packages/airnode-operation && yarn run dev:stop)",
     "docker": "yarn docker:artifacts && yarn docker:deployer && yarn docker:node",
+    "docker:admin": "docker build --tag api3/airnode-admin:latest --file packages/airnode-admin/docker/Dockerfile .",
```
andreogle, line 45: Minor thing, but maybe `docker:build:admin` is a bit more descriptive as a script name? (And the same would apply to the other `docker build` scripts. What do you think?

packages/airnode-admin/docker/README.md
```diff
@@ -0,0 +1,16 @@
...
+This documentation focuses on the admin CLI docker image, not the admin CLI itself. If you want to learn more about the
+admin CLI, please read [its documentation](../README.md).
+
+## Build
+
+In order to build Airnode Docker image you need to build the [artifacts image first](../../../docker/README.md). Once
+you've done that, you can build the Docker image by running following command from the root directory:
+
+```bash
+docker build -f packages/airnode-admin/docker/Dockerfile -t api3/airnode-admin:latest .
```
andreogle, line 15: Can't the user just use the `docker:admin` (or `docker:build:admin`) script instead?

packages/airnode-admin/src/cli.ts
```diff
@@ -443,6 +443,14 @@ yargs
       console.log(`Is requester whitelisted: ${isRequesterWhitelisted}`);
     }
   )
+  .command(
+    'generate-mnemonic',
+    'Generates a random mnemonic. Uses "ethers.Wallet.createRandom" under the hood.',
+    async () => {
+      const mnemonic = await admin.generateMnemonic();
+      console.log(`Generated mnemonic: ${mnemonic}`);
```
andreogle, line 451: Do we want to add some warnings here about writing it down somewhere safe?

andreogle (approved):
Very minor comments, otherwise looks good 💯

package.json
andreogle: Makes it easier to add other Docker commands in the future too. e.g. `docker run ...`

Siegrift: Agree.

packages/airnode-admin/docker/README.md
Siegrift: Yeah, they could - I kept this in sync with other READMEs, but I'll change that now.

packages/airnode-admin/src/cli.ts
Siegrift: I do not think it's necessary. The user have an intention to generate the mnemonic for. He should know where to write the mnemonic and how to store it (or write it down).

package.json
```diff
@@ -41,10 +41,11 @@
     "dev:invoke": "(cd packages/airnode-node && yarn run dev:invoke)",
     "dev:list": "(cd packages/airnode-operation && yarn run dev:list)",
     "dev:stop": "(cd packages/airnode-operation && yarn run dev:stop)",
-    "docker": "yarn docker:artifacts && yarn docker:deployer && yarn docker:node",
-    "docker:artifacts": "docker rmi --force api3/airnode-artifacts && docker build --tag api3/airnode-artifacts:latest --file docker/Dockerfile .",
-    "docker:deployer": "docker build --tag api3/airnode-deployer:latest --file packages/airnode-deployer/docker/Dockerfile .",
-    "docker:node": "docker build --tag api3/airnode-client:latest --file packages/airnode-node/docker/Dockerfile .",
+    "docker": "yarn docker:build:artifacts && yarn docker:build:deployer && yarn docker:build:node && yarn docker:build:admin",
+    "docker:build:admin": "docker build --tag api3/airnode-admin:latest --file packages/airnode-admin/docker/Dockerfile .",
+    "docker:build:artifacts": "docker rmi --force api3/airnode-artifacts && docker build --tag api3/airnode-artifacts:latest --file docker/Dockerfile .",
```
andreogle, line 46: This one isn't building 😉 
```suggestion
    "docker:rm:artifacts": "docker rmi --force api3/airnode-artifacts && docker build --tag api3/airnode-artifacts:latest --file docker/Dockerfile .",
```

andreogle (approved):

package.json
andreogle: Nevermind, I just saw the second part 🤦🏻‍♂️

Siegrift: It is.

I mean it's also removing the old container, but it also does the fresh build

amarthadan (changes requested):
Can you also please change the binary name for deployer and node containers as we've discussed on [slack](https://api3workspace.slack.com/archives/D01TXKGUCUV/p1635855374025100)?

packages/airnode-admin/docker/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+ADMIN_BINARY="/usr/local/bin/airnode-admin"
+USER_ID="${USER_ID:-0}"
+GROUP_ID="${GROUP_ID:-0}"
+
+su-exec ${USER_ID}:${GROUP_ID} ${ADMIN_BINARY} "$@"
```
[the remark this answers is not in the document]
Siegrift: Yeah, good idea.

packages/airnode-admin/docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+WORKDIR ${appDir}
+
+COPY --from=api3/airnode-artifacts /dependencies/common ${appDir}/node_modules
+COPY --from=api3/airnode-artifacts /dependencies/airnode-admin ${appDir}/node_modules
+COPY --from=api3/airnode-artifacts /packages ${appDir}/node_modules/@api3/
+COPY --from=api3/airnode-artifacts /build/packages/airnode-admin/dist ${appDir}/
+COPY packages/airnode-admin/docker/entrypoint.sh /entrypoint.sh
+
+RUN apk add --update --no-cache su-exec
```
amarthadan, line 18: You don't need the `su-exec` package anymore.

packages/airnode-admin/docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# Make the binary available within PATH
+RUN ln -s ${appDir}/bin/admin.js ${appBin} && chmod +x ${appBin}
+# Create Airnode user
+RUN adduser -h ${appDir} -s /bin/false -S -D -H ${name}
```
amarthadan, line 22: You have to switch to that user with the `USER` directive in order to run the CLI under that user.

packages/airnode-admin/docker/entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+ADMIN_BINARY="/usr/local/bin/airnode-admin"
+
+node ${ADMIN_BINARY} "$@"
```
amarthadan, line 5: Do we even need this file? You can call the binary directly from `ENTRYPOINT` and arguments are passed automatically by Docker.

packages/airnode-validator/package.json
```diff
@@ -6,8 +6,8 @@
   "main": "./dist/src/index.js",
   "types": "./dist/src/index",
   "bin": {
-    "api3-validator": "./dist/bin/validator.js",
-    "api3-convertor": "./dist/bin/convertor.js"
+    "validator": "./dist/bin/validator.js",
+    "convertor": "./dist/bin/convertor.js"
```
amarthadan, line 10:
```suggestion
    "airnode-validator": "./dist/bin/validator.js",
    "airnode-convertor": "./dist/bin/convertor.js"
```

amarthadan (dismissed):

packages/airnode-validator/package.json
Siegrift: Do we need this? This is only visible to the user if running with `npx`. You invoke this like this:

```sh
npx -p @api3/airnode-validator -c validator
```

so I find the prefix for this unnecessary.

packages/airnode-admin/docker/Dockerfile
Siegrift: Actually, do we need that user at all? Does docker run with sudo by default?

amarthadan: The Docker service is usually run by root and base images like `node` here usually let you have root access so you can do whatever. But for the specific application image it's recommended to run the application under some non-root user.

packages/airnode-validator/package.json
amarthadan: That's not true. Based on this the package is packed and binaries are created. So this is to avoid having a command in the system that's called just `validator`.

Siegrift: Yeah, you're right. I forgot about the global install

PR #660, opened 2021-11-02
Fix formatting

Siegrift:


.husky/pre-commit
```diff
@@ -1,3 +1,3 @@
 yarn --cwd packages/airnode-validator/ run docs
-yarn prettier --write "./**/*.md" --loglevel silent
+yarn prettier --write "./packages/airnode-validator/**/*.md" --loglevel silent
```
amarthadan, line 2: Why do we write anything during the pre-commit hook? Pre-commit hook should make you aware that there is a problem and not allow you to commit unless you solve it. It shouldn't silently change the content of files. That should always be a conscious decision of a developer.

packages/airnode-admin/test/e2e/cli.feature.ts
```diff
@@ -48,9 +48,9 @@ describe('CLI', () => {
     if (DEBUG_COMMANDS) console.log(`Executing command: ${formattedCommand}`);
     try {
       return execSync(`node ${CLI_EXECUTABLE} ${formattedCommand}`).toString().trim();
-    } catch (e: any) {
+    } catch (e) {
       // rethrow the output of the CLI
-      throw new Error(e.stdout.toString().trim());
+      throw new Error((e as any).stdout.toString().trim());
```
amarthadan, line 53: Why is this change needed? If you don't state it explicitly the variable is implicitly of type `any`. There's also no need for type assertion.

amarthadan (changes requested):
Can you also add exceptions for:
* `<some path>/.terraform/<some path>`
* `config.json` anywhere
* `receipt.json` anywhere

Adding screenshot so you know what I mean
Screenshot_20211103_090757

Also, if possible, can you include `config.json.example` as JSON to be checked as well? I guess it's currently ignored based on the extension.

packages/airnode-admin/test/e2e/cli.feature.ts
Siegrift: Not sure why is this included - I think I was debugging some stuff and forget to revert this.

.husky/pre-commit
Siegrift: Because validator docs are genereated (and written on disk) on pre commit hook, but thoes files are not correctly formatted.

But we could make it a pre-push hook, wdyt?

Siegrift:
Regarding the screenshot. I think you forgot to run `yarn` as locally the formatting passes for me

.husky/pre-commit
andreogle: I kinda like making the prettier check into something more conscious for the developer instead of doing it automatically. I think we need to remove the `yarn run pretty-quick` below though?

I also generally prefer pre-commit to pre-push but that's just my personal preference

.prettierrc
```diff
@@ -6,12 +6,6 @@
   "trailingComma": "es5",
   "useTabs": false,
   "overrides": [
-    {
```
andreogle, line 9: Why remove this? Is it inferred by prettier now?

.prettierignore
```diff
@@ -12,3 +12,4 @@
 **/.webpack/**
 
 packages/airnode-validator/test/fixtures
+packages/airnode-protocol/src/contracts
```
andreogle, line 15: Why add this to be ignored?

andreogle (approved):

.prettierrc
Siegrift: Yes. It's not necessary, TS is default for `.ts` and `.tsx`

.prettierignore
Siegrift: Because those files are generated.

.husky/pre-commit
Siegrift: I am confused. What do you suggest?

andreogle: Sorry I think I may have misunderstood myself. This should probably be kept as is if the files are generated during pre-commit. Ignore me 😬

PR #661, opened 2021-11-03
AN-324 Ignore patch version

drgy:
Patch version will be ignored when validating, also in `validate` and `convert` commands.

### DIFF
--- a/packages/airnode-validator/package.json
+++ b/packages/airnode-validator/package.json
@@ -28,6 +28,7 @@
   "devDependencies": {
     "@types/yargs": "^17.0.2",
     "copyfiles": "^2.4.1",
+    "lodash": "^4.17.21",
     "markdown-snippet-injector": "^0.2.4",
     "rimraf": "^3.0.2",
     "ts-node": "^10.1.0",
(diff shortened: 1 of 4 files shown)


packages/airnode-validator/src/commands/utils.ts
```diff
@@ -66,8 +66,9 @@ function getLatestPath(template: string): string | null {
  */
 export function getPath(template: string, messages: Log[], version = ''): string | null {
   if (version) {
-    if (fs.existsSync(path.resolve(validatorTemplatesPath, version, template))) {
-      return path.resolve(validatorTemplatesPath, version, template);
+    const parsedVersion = version.replace(/[0-9]+$/, '0');
```
amarthadan, line 69: I think it makes more sense to rename the template directories to omit the patch version and ignore it instead of arbitrarily adding 0. I think this would be quite confusing for the users.

packages/airnode-validator/package.json
```diff
@@ -28,6 +28,7 @@
   "devDependencies": {
     "@types/yargs": "^17.0.2",
     "copyfiles": "^2.4.1",
+    "lodash": "^4.17.21",
```
amarthadan, line 31: Why is this here?

amarthadan (changes requested):

packages/airnode-validator/src/commands/utils.ts
drgy: removed in ddd8229

packages/airnode-validator/package.json
drgy: that's my bad commiting, it belongs to another branch, removed in 9f1ef8c

amarthadan (approved):

PR #662, opened 2021-11-03
AN-133 Add tests to check if the validator is up to date

drgy:
Added tests and updated fixtures to pass the tests,
@amarthadan updated fixtures are passing tests, but please check the updates I made to them wouldn't cause any problems somewhere else.


packages/airnode-validator/test/validator.test.ts
```diff
@@ -1,10 +1,19 @@
...
 const tests = fs.readdirSync(path.resolve(__dirname, 'validatorTests'));
 const validOutput = { valid: true, messages: [] };
 
+// Regular expression that does not match anything, ensuring no escaping or interpolation happens
+// https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L199
+const NO_MATCH_REGEXP = /($^)/;
+// Regular expression matching ES template literal delimiter (${}) with escaping
+// https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L175
+const ES_MATCH_REGEXP = /\$\{([^\\}]*(?:\\.[^\\}]*)*)\}/g;
+
```
amarthadan, line 16: Copy & paste from the `airnode-node` package is really not an ideal solution because if the implementation from the package changes the tests will still pass.

I would suggest refactoring a bit (if you must) the `parseConfig` function and importing the result instead.

amarthadan (changes requested):
I don't really see a reason to change those `secrets.example.env`s. You can mock the values within the test, you don't have to read them from the file. I'm ok with you uniting the variable names in config and secrets files to make it easier to mock. You can also come up with multiple sets of mock values (one for local, one for AWS, etc.) and run them against all the configs.

packages/airnode-validator/test/validator.test.ts
drgy: I can simply use the `parseConfig` function from `airnode-node`, but current jest config is causing trouble importing it to tests in `airnode-validator`, im not sure how can I import it, perhaps the config might have to change...

amarthadan: I'm ok with that if it's needed.

Siegrift:
@drgy what is the status of this?

drgy:
@Siegrift It was decided that validator should do interpolation on its own, since these tests require interpolation it has to wait for #672

drgy:
@Siegrift @amarthadan Currently each `config.json` is tested with secrets that are defined [here](https://github.com/api3dao/airnode/blob/1fc7443b8f5b32fe56d9e49a934984039f089a37/packages/airnode-validator/test/validator.test.ts#L8), but each of them has a `secrets.env` in it's directory. Would it be possible to fill the `secrests.env` with some example values, interpolate with `config.json` and test like that?

Im asking this because there was a new variable added into one of the `secrets.env`, this variable then needed to be added to the variable in the validator tests. If each `secrets.env` would have some example values, it can be freely edited without any changes needed in the validator tests.

Siegrift:
For examples, there is always a `secrets.example.env` which you can populate and use for testing

amarthadan:
To be honest I'm not a big fan of populating any `secrets.env` with dummy values. I'm more ok with possibly having to update the tests when an example is added.

packages/airnode-validator/test/validator.test.ts
```diff
@@ -1,10 +1,24 @@
...
 const validOutput = { valid: true, messages: [] };
 
+const secretsEnv = {
+  PROVIDER_URL: 'http://127.0.0.1:8545',
+  AIRNODE_RRP_ADDRESS: '0x0000000000000000000000000000000000000000',
+  AIRNODE_WALLET_MNEMONIC: 'achieve climb couple wait accident symbol spy blouse reduce foil echo label',
+  SS_CURRENCY_CONVERTER_API_KEY: '<enter your API key>',
+  CMC_PRO_API_KEY: '<enter your API key>',
+  HTTP_GATEWAY_API_KEY: 'apikey_example_apikey_example_apikey_example',
+  HEARTBEAT_API_KEY: '',
```
amarthadan, line 15: I would suggest using UUID as example values for API keys.

packages/airnode-validator/test/validator.test.ts
```diff
@@ -1,10 +1,24 @@
...
+  HEARTBEAT_ID: '',
+  HEARTBEAT_URL: '',
+  CHAIN_ID: '',
```
amarthadan, line 18: Why are these values ok empty? Shouldn't validator complain about each of them being empty? Also, about their format? `HEARTBEAT_URL` should be a valid URL and `CHAIN_ID` should be a number, shouldn't it?

amarthadan (changes requested):

packages/airnode-validator/test/validator.test.ts
drgy: Templates are based on docs, it was not specified that `CHAIN_ID` should be a number, but I can change it if it should.

All urls are using this regex: `^(([^:\/?#]+):)?(\/\/([^\/?#]*))?([^?#]*)$`, which basically allows anything. There was whole discussion regarding this, it was decided it should not be limiting and this one was chosen...

packages/airnode-validator/test/validator.test.ts
drgy: Updated in d83d5fe

Siegrift:
FYI: I've populated the examples in the https://github.com/api3dao/airnode/pull/705 and they are now generated (and there are tests verifying they are up to date).

So this PR can be updated to validate on those generated files (config.example.json and secrets.example.env)

packages/airnode-validator/test/validator.test.ts
Siegrift: There is a TODO in node and we can refactor it when the times comes. I'd like this to happen after 0.4

packages/airnode-validator/test/validator.test.ts
Siegrift: The `chainId` is indeed typed as string, so this is correct (I don't see it a big deal, since I wonder how big of a convention this is and whether there exist any non number like providers).

Siegrift (approved):
LGTM :D

amarthadan (approved):

PR #664, opened 2021-11-05
Update nodeVersion in config.json files

dcroote:


amarthadan (commented):
We are already at `0.2.2` so this won't help much because before releasing `0.2.3` we would have to bump them again. Maybe bumping them directly to `0.2.3` and doing the release would be better. I think @bbenligiray is planning to do that anyway.

bbenligiray:
I think it would be ideal if we used [this](https://github.com/api3dao/airnode/blob/master/package.json#L4) as the node version and have a script at the root that does these updates.

amarthadan:
> I think it would be ideal if we used [this](https://github.com/api3dao/airnode/blob/master/package.json#L4) as the node version and have a script at the root that does these updates.

The problem is that that specific version number won't ever be bumped by the changeset release mechanism (@Siegrift correct me if I'm wrong).

But I agree that we should automate this. I'll create a task and we can discuss the priority on the next planning call.

bbenligiray:
> The problem is that that specific version number won't ever be bumped by the changeset release mechanism

I know, but leaving it at 0.1.0 doesn't feel right either so we'll also need to address that (maybe the script pulls the version number from the packages and updates this too)

dcroote:
I'm certainly in favor of automation, but in the meantime I don't see harm in merging this, especially since the docs pull the `.example` config files that currently contain `0.1.0`

wkande:
In the mean time I have disabled the GitHub action in the docs repo that create a PR when these config file change in the Airnode repo. The files in the docs have been manually changed and will re-sync when the GitHub action is enabled again.

bbenligiray:
Pls merge

amarthadan (approved):

PR #665, opened 2021-11-09
Psp draft v3

bbenligiray:
Just some small updates/comments, it looks like AirnodePsp.sol doesn't need much more changes. I wanted to do some allocator work but I'll have to import AccessControlRegistry so I'll work on that on another branch.


bbenligiray:

PR #666, opened 2021-11-09
PSP draft v4

bbenligiray:
This mirrors the authorizer patterns and pushes the complexity to its client contracts. One potential downside is that assume the node reserved 5 slots to an allocator, which has two client contracts (or two multisigs, whatever). One of these is allowed to use up all 5 slots, leaving none for the other. The only way to reverse this is to take away its slot setter role. So the slot setters are expected to "play nice", limit the number of slots they set, etc. Someone trying to set a slot and failing is preferable over someone losing service, so this is alright imo.

### DIFF
--- a/packages/airnode-protocol/contracts/psp/AirnodePsp.sol
+++ b/packages/airnode-protocol/contracts/psp/AirnodePsp.sol
@@ -9,6 +9,7 @@ contract AirnodePsp {
 
     struct Subscription {
         bytes32 templateId;
+        address sponsor;
         address conditionAddress;
         bytes4 conditionFunctionId;
         address fulfillAddress;
@@ -26,6 +27,7 @@ contract AirnodePsp {
 
     function createSubscription(
         bytes32 templateId,
+        address sponsor,
         address conditionAddress,
         bytes4 conditionFunctionId,
         address fulfillAddress,
@@ -35,6 +37,7 @@ contract AirnodePsp {
         subscriptionId = keccak256(
             abi.encode(
                 templateId,
+                sponsor,
                 conditionAddress,
                 conditionFunctionId,
                 fulfillAddress,
@@ -44,6 +47,7 @@ contract AirnodePsp {
         );
         subscriptions[subscriptionId] = Subscription({
             templateId: templateId,
+            sponsor: sponsor,
             conditionAddress: conditionAddress,
             conditionFunctionId: conditionFunctionId,
             fulfillAddress: fulfillAddress,
--- a/packages/airnode-protocol/contracts/psp/AirnodePsp.sol
+++ b/packages/airnode-protocol/contracts/psp/AirnodePsp.sol
@@ -1,0 +1,77 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "../rrp/interfaces/IAirnodeRrp.sol";
+
+contract AirnodePsp {
+    using ECDSA for bytes32;
+
+    struct Subscription {
+        bytes32 templateId;
+        address conditionAddress;
+        bytes4 conditionFunctionId;
+        address fulfillAddress;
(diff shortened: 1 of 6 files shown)


packages/airnode-protocol/contracts/psp/AirnodePsp.sol
```diff
@@ -0,0 +1,81 @@
...
+
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "../rrp/interfaces/IAirnodeRrp.sol";
+
+contract AirnodePsp {
+    using ECDSA for bytes32;
+
+    struct Subscription {
+        bytes32 templateId;
+        address sponsor;
```
bbenligiray, line 12: Added this after the call because we need a way to match subscription IDs to sponsors (for example see https://github.com/api3dao/airnode/pull/666/files#diff-45d4b511538c5f4fe3ce1dda8891270fe0a6eb927a72603ecceea7d16a7cdbeaR322)

packages/airnode-protocol/contracts/psp/BeaconServer.sol
```diff
@@ -0,0 +1,435 @@
...
+        (int256 decodedData, uint256 timestamp) = abi.decode(
+            data,
+            (int256, uint256)
+        );
+        require(
+            decodedData >= type(int224).min && decodedData <= type(int224).max,
+            "Value typecasting error"
+        );
+        require(timestamp <= type(uint32).max, "Timestamp typecasting error");
+        require(block.timestamp - 1 hours < timestamp, "Fulfillment stale");
```
acenolaza, line 263: We probably want the `1 hour` to be configurable, don't we?

acenolaza: Also, do we really want to revert here or should we just emit an event and return to ignore the fulfillment tx. 

I guess the question is: does Airnode care about this require condition when fulfilling a request?

packages/airnode-protocol/contracts/psp/BeaconServer.sol
```diff
@@ -0,0 +1,435 @@
...
+            templateId,
+            requestId,
+            int224(decodedData),
+            uint32(timestamp)
+        );*/
+    }
+
+    function setUpdatePercentageThreshold(
+        bytes32 subscriptionId,
+        uint256 updatePercentageThreshold
```
acenolaza, line 315: How is this percentage represented? Something like 5% is equal to 5 * 10 ** 18?

packages/airnode-protocol/contracts/psp/BeaconServer.sol
bbenligiray: I think 1 hour is reasonable (confirmation time under ideal conditions is <5 mins so this should be a proportionate number) and it's fine to hardcode it. Otherwise, it's not clear who should be able to configure it. We can initialize it at the constructor but that means 1 additional storage read for fulfillments = +2000 gas so I'm not sure if that's worth it.

Requiring these conditions here means `callSuccess` [here](https://github.com/api3dao/airnode/blob/master/packages/airnode-protocol/contracts/rrp/AirnodeRrp.sol#L253) will return `false` and AirnodeRrp will emit `FailedRequest` instead of `FulfilledRequest`. It feels like that's the correct thing to do.

packages/airnode-protocol/contracts/psp/BeaconServer.sol
bbenligiray: 10**18 means 100%, everything else relates linearly

bbenligiray (commented):
@acenolaza I commented on the PR when I made it but I realize that I didn't finalize the review :man_facepalming:

packages/airnode-protocol/contracts/psp/BeaconServer.sol
bbenligiray: https://github.com/api3dao/airnode/pull/666#discussion_r746364027

bbenligiray:

PR #668, opened 2021-11-12
AN-333: Fix endpoint ID derivation implementation

acenolaza:
Endpoint id derivation in admin-cli was out of date.

https://api3dao.atlassian.net/browse/AN-333


andreogle (approved):

Siegrift (approved):

Siegrift:
Optionally you might want to update the `endpointId`s in the examples package, but since that's just a convention it's not really necessary I guess.

Issue #516, opened 2021-08-20
Requester can't specify multiple fields to be returned


bbenligiray:
Currently, the `_type`, `_path` and `_times` parameters expect a single value each, which will specify a single field to be returned from the API response. It's very common for the requester to need multiple fields to be returned, and these parameters should support the specification of multiple values accordingly.

See https://github.com/api3dao/airnode/issues/507 for a more general version.

Tracked by https://api3dao.atlassian.net/browse/AN-125

PR #669, opened 2021-11-13
Support multiple values for Airnode responses

Siegrift:
https://api3dao.atlassian.net/browse/AN-125

Docs 1: https://github.com/api3dao/api3-docs/pull/408
Docs 2: https://github.com/api3dao/api3-docs/pull/417

Closes https://github.com/api3dao/airnode/issues/516

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,11 +3,6 @@ ARG build=local
 # Environment
 FROM node:14.17-alpine3.14 AS environment
 
-# Download solidity compiler as per: https://github.com/nomiclabs/hardhat/issues/1280#issuecomment-949822371
-# The binary filename is taken from: https://github.com/ethereum/solc-bin/tree/gh-pages/linux-amd64
-RUN mkdir -p /root/.cache/hardhat-nodejs/compilers/linux-amd64 \
-  &&  wget -O /root/.cache/hardhat-nodejs/compilers/linux-amd64/solc-linux-amd64-v0.8.9+commit.e5eed63a https://binaries.soliditylang.org/linux-amd64/solc-linux-amd64-v0.8.9+commit.e5eed63a
-
 ENV buildDir="/build" \
     dependenciesDir="/dependencies" \
     packagesDir="/packages"
@@ -34,6 +29,11 @@ RUN git clone ${repository} ${buildDir} -b ${branch}
 
 # Production dependencies
 FROM source${build} AS deps
+
+# Download solidity compiler as per: https://github.com/nomiclabs/hardhat/issues/1280#issuecomment-949822371
+# The binary filename is taken from: https://github.com/ethereum/solc-bin/tree/gh-pages/linux-amd64
+RUN mkdir -p /root/.cache/hardhat-nodejs/compilers/linux-amd64 \
+  &&  wget -O /root/.cache/hardhat-nodejs/compilers/linux-amd64/solc-linux-amd64-v0.8.9+commit.e5eed63a https://binaries.soliditylang.org/linux-amd64/solc-linux-amd64-v0.8.9+commit.e5eed63a
 
 RUN yarn global add lerna && \
     lerna bootstrap -- --production --no-optional --ignore-scripts && \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,6 +3,11 @@ ARG build=local
 # Environment
 FROM node:14.17-alpine3.14 AS environment
 
+# Download solidity compiler as per: https://github.com/nomiclabs/hardhat/issues/1280#issuecomment-949822371
+# The binary filename is taken from: https://github.com/ethereum/solc-bin/tree/gh-pages/linux-amd64
+RUN mkdir -p /root/.cache/hardhat-nodejs/compilers/linux-amd64 \
+  &&  wget -O /root/.cache/hardhat-nodejs/compilers/linux-amd64/solc-linux-amd64-v0.8.9+commit.e5eed63a https://binaries.soliditylang.org/linux-amd64/solc-linux-amd64-v0.8.9+commit.e5eed63a
+
 ENV buildDir="/build" \
     dependenciesDir="/dependencies" \
     packagesDir="/packages"
--- a/packages/airnode-adapter/e2e/index.ts
+++ b/packages/airnode-adapter/e2e/index.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
 import { ethers } from 'hardhat';
-import { extractAndEncodeResponse } from '../src';
+import { extractAndEncodeResponse, ReservedParameters } from '../src';
 import type { Contract } from 'ethers';
 
 // Chai is able to assert that "expect(BigNumber).to.equal(string)" but fails to assert
@@ -69,6 +69,13 @@ const apiResponse = {
   },
 } as const;
 
+function extractAndEncodeSingle(reservedParams: ReservedParameters) {
+  const encoded = extractAndEncodeResponse(apiResponse, reservedParams);
+  if (Array.isArray(encoded)) expect.fail();
+
+  return (encoded as any)[0].encodedValue;
+}
+
 describe('Extraction, encoding and simple on chain decoding', () => {
   // eslint-disable-next-line functional/no-let
   let testDecoder: Contract;
@@ -85,7 +92,7 @@ describe('Extraction, encoding and simple on chain decoding', () => {
 
       expect(
         await testDecoder[methodName](
-          extractAndEncodeResponse(apiResponse, {
+          extractAndEncodeSingle({
             _type: type,
             _path: 'big.decimal',
           }).encodedValue
@@ -94,7 +101,7 @@ describe('Extraction, encoding and simple on chain decoding', () => {
 
       expect(
         await testDecoder[methodName](
-          extractAndEncodeResponse(apiResponse, {
+          extractAndEncodeSingle({
             _type: type,
             _path: 'decimal',
           }).encodedValue
@@ -103,7 +110,7 @@ describe('Extraction, encoding and simple on chain decoding', () => {
 
       expect(
         await testDecoder[methodName](
-          extractAndEncodeResponse(apiResponse, {
+          extractAndEncodeSingle({
             _type: type,
             _path: 'big.float',
             _times: '1000000',
@@ -113,7 +120,7 @@ describe('Extraction, encoding and simple on chain decoding', () => {
 
       expect(
         await testDecoder[methodName](
-          extractAndEncodeResponse(apiResponse, {
+          extractAndEncodeSingle({
             _type: type,
             _path: 'float',
             _times: '1000000',
@@ -126,34 +133,28 @@ describe('Extraction, encoding and simple on chain decoding', () => {
   it('floors the number after multiplying it', async () => {
     expect(
       await testDecoder.decodeSignedInt256(
-        extractAndEncodeResponse(apiResponse, { _type: 'int256', _times: '100', _path: 'float' }).encodedValue
+        extractAndEncodeSingle({ _type: 'int256', _times: '100', _path: 'float' }).encodedValue
       )
     ).to.equal(1234567);
   });
 
   it('floors the number without using "_times" parameter', async () => {
     expect(
-      await testDecoder.decodeSignedInt256(
-        extractAndEncodeResponse(apiResponse, { _type: 'int256', _path: 'float' }).encodedValue
-      )
+      await testDecoder.decodeSignedInt256(extractAndEncodeSingle({ _type: 'int256', _path: 'float' }).encodedValue)
     ).to.equal(12345);
   });
 
   it('decodes bool encoded by the adapter package', async () => {
     expect(
-      await testDecoder.decodeBool(
-        extractAndEncodeResponse(apiResponse, { _type: 'bool', _path: 'boolTrue' }).encodedValue
-      )
+      await testDecoder.decodeBool(extractAndEncodeSingle({ _type: 'bool', _path: 'boolTrue' }).encodedValue)
     ).to.equal(true);
     expect(
-      await testDecoder.decodeBool(
-        extractAndEncodeResponse(apiResponse, { _type: 'bool', _path: 'strFalse' }).encodedValue
-      )
+      await testDecoder.decodeBool(extractAndEncodeSingle({ _type: 'bool', _path: 'strFalse' }).encodedValue)
     ).to.equal(false);
   });
 
   it('decodes bytes32 encoded by the adapter package', async () => {
-    const encodedBytes = extractAndEncodeResponse(apiResponse, { _type: 'bytes32', _path: 'bytes' }).encodedValue;
+    const encodedBytes = extractAndEncodeSingle({ _type: 'bytes32', _path: 'bytes' }).encodedValue;
 
     const fromContract = await testDecoder.decodeBytes32(encodedBytes);
     const decoded = ethers.utils.parseBytes32String(fromContract);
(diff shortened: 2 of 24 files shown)


packages/airnode-adapter/e2e/index.ts
```diff
@@ -235,27 +217,31 @@ describe('Extraction, encoding and simple on chain decoding', () => {
...
+    const encodedBytes = extractAndEncode({ _type: 'string32', _path: 'string' });
 
     const fromContract = await testDecoder.decodeString32(encodedBytes);
     const decoded = parseBytes32String(arrayify(fromContract)).toString();
     expect(decoded).to.equal(apiResponse.string);
   });
 
+  it('decodes multiple params encoded by the adapter package', async () => {
+    const decoded = await testDecoder.decodeMultipleParameters(
+      extractAndEncode({ _type: 'string,uint256,address', _path: 'big.string,float,address', _times: ',10000,' })
```
Siegrift, line 229: Note that I am not using `null` to specify that a value should not be multiplied, but rather empty string.

This is because `_path` is optional as well - and that returns the whole API response. Also, `null` in path would be ambiguous (do you want key or the whole response?). Empty string solves this.

packages/airnode-adapter/src/response-processing/extraction.test.ts
```diff
@@ -66,11 +64,95 @@ describe('extractAndEncodeValue', () => {
...
     expect(res).toEqual({
-      rawValue: '750.51',
-      value: '75051',
+      rawValue: data,
+      values: ['75051'],
       encodedValue: '0x000000000000000000000000000000000000000000000000000000000001252b',
     });
   });
+
+  it('accepts "" (empty string) for _times parameter', () => {
```
Siegrift, line 75: Since we allow using empty string when specifying multiple parameters It makes sense to support this for single values as well (it's more consistent and easier to document).

packages/airnode-adapter/src/response-processing/extraction.test.ts
```diff
@@ -66,11 +64,95 @@ describe('extractAndEncodeValue', () => {
...
+    });
+  });
+});
+
+describe('extract and encode multiple values', () => {
+  it('works for basic request', () => {
+    const data = { a: { b: [{ c: 1 }, { d: '750.51' }] } };
+    const parameters: ReservedParameters = { _path: 'a.b.1.d,a.b.0.c', _type: 'int256,bool', _times: '100,' };
+    const res = extractAndEncodeResponse(data, parameters);
+    expect(res).toEqual({
```
Siegrift, line 103: I needed to change the signature of the encoding function. Airnode mostly cares about the `encodedValue` which meaning didn't change.

However, I've changed testing gateway output. @andreogle this maybe impacts chainAPI

packages/airnode-adapter/src/response-processing/extraction.test.ts
```diff
@@ -66,11 +64,95 @@ describe('extractAndEncodeValue', () => {
...
+    expect(splitReservedParameters({ _type: 'uint256' })).toEqual([{ _type: 'uint256' }]);
+    expect(splitReservedParameters({ _type: 'uint256,string', _path: 'key,anotherKey' })).toEqual([
+      { _type: 'uint256', _path: 'key' },
+      { _type: 'string', _path: 'anotherKey' },
+    ]);
+    expect(splitReservedParameters({ _type: 'uint256, invalid' })).toEqual([
+      { _type: 'uint256' },
+      { _type: ' invalid' },
+    ]);
+    expect(splitReservedParameters({ _type: 'uint256,,,invalid' })).toEqual([
```
Siegrift, line 133: FYI: This type is ofc invalid, but I want to double check that the function works correctly.

packages/airnode-adapter/src/response-processing/extraction.test.ts
```diff
@@ -66,11 +64,95 @@ describe('extractAndEncodeValue', () => {
...
+      { _type: 'uint256', _path: 'usd' },
+      { _type: 'string', _path: '' },
+      { _type: 'bytes', _path: '' },
+    ]);
+    expect(splitReservedParameters({ _type: 'uint256', _times: '100' })).toEqual([{ _type: 'uint256', _times: '100' }]);
+  });
+
+  it('throws when there are different number of splits', () => {
+    expect(() => splitReservedParameters({ _type: 'uint256', _path: 'usd,eur' })).toThrow(
+      `Unexpected number of parsed reserved parameters. Number of "_types" parameters = 1, but "_path" has only 2`
```
Siegrift, line 149: This should help catch some (hard to debug) issues if the API provider forgets to escape things.

packages/airnode-adapter/src/types.ts
```diff
@@ -49,18 +50,21 @@ export interface Config {
...
 // Use might pass a complex type (e.g. int256[3][]) which we cannot type
 export type ResponseType = string;
 
 export interface ReservedParameters {
-  readonly _path?: string;
-  readonly _times?: string | BigNumber;
-  readonly _type: ResponseType;
-  readonly _relay_metadata?: string;
+  _path?: string;
+  _times?: string;
```
Siegrift, line 59: I am not sure why this function accepted `BigNumber` - I think the only valid value is `string` (also there were no issues when I removed BigNumber from the type).

packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -29,9 +30,30 @@ export async function initializeProvider(event: any) {
...
+  });
+
+  return pick(event, requiredKeys) as handlers.CallApiPayload;
+}
+
+export async function callApi(event: unknown) {
+  // TODO: Verify that logOptions are present
+  const { logOptions } = event as any;
+
+  const [error, callApiPayload] = promiseUtils.goSync(() => parseCallApiPayload(event));
```
Siegrift, line 50: @andreogle @amarthadan LMK WDYT about this...

All of these AWS handlers are "black magic" boxes to me. I find it dangerous to expect that the `event` will contain the expected properties (especially after we do a bunch of refactors). I think we should check for them explicitly...

Also, I still don't understand completely what is contained in the `event` object (especially how `config` ges there in this example).

packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -61,5 +83,6 @@ export async function testApi(event: any) {
     return { statusCode: 400, body: JSON.stringify({ error: err.toString() }) };
   }
 
-  return { statusCode: 200, body: JSON.stringify(result) };
+  // NOTE: We do not want the user to see {"value": <actual_value>}, but the actual value itself
+  return { statusCode: 200, body: JSON.stringify(result!.value) };
```
Siegrift, line 87: @amarthadan I've changed the output of testing gateway. I think this is not a big deal (major change) though since this output is only for testing

packages/airnode-node/src/adapters/http/worker.ts
```diff
@@ -9,17 +9,21 @@ import {
...
   workerOpts: WorkerOptions,
-  encodeResponse = true
+  apiCallOptions?: ApiCallOptions
 ): Promise<LogsData<ApiCallResponse | null>> {
   const options = {
     ...workerOpts,
     functionName: 'callApi' as WorkerFunctionName,
-    payload: { aggregatedApiCall, logOptions, encodeResponse },
+    // TODO: The payload should have the same type for all handlers (local and AWS). Currently it does not.
+    // AWS handlers (in deployer) expects the `config` property to be present in the payload.
```
Siegrift, line 25: WDYT? Did I miss something?

packages/airnode-adapter/src/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const solidityBaseTypes = ['uint256', 'int256', 'bool', 'bytes32', 'address', 'bytes', 'string'] as const;
+export const artificialTypes = ['string32'] as const;
+export const baseResponseTypes = [...solidityBaseTypes, ...artificialTypes] as const;
+
+export const MULTIPLE_PARAMETERS_DELIMETER = ',';
+export const PATH_DELIMETER = ',';
```
bbenligiray, line 6: Should this be `,`? I couldn't find where it's used.

packages/airnode-adapter/e2e/index.ts
bbenligiray: Does this also apply for `_path`, i.e. is `big.string,` valid as a `_path` value?

packages/airnode-adapter/src/types.ts
bbenligiray: I couldn't recall but this is probably important

packages/airnode-adapter/src/constants.ts
Siegrift: > Should this be .? 

It should.

>  I couldn't find where it's used.

Nowhere - I forgot to use it... But I've fixed it in https://github.com/api3dao/airnode/pull/673/files#diff-4940026c48ea44c3e3afc561a628cc4a8b15cc39341356e9eba334e3c268dd29R6

packages/airnode-adapter/src/types.ts
Siegrift: @andreogle do you remember?

packages/airnode-adapter/e2e/index.ts
Siegrift: Yes.

packages/airnode-adapter/src/types.ts
andreogle: I can't quite remember either, sorry. Should be fine to remove it then

packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -1,5 +1,6 @@
 import * as path from 'path';
 import { config, handlers, logger, promiseUtils, providerState, WorkerResponse } from '@api3/airnode-node';
+import { pick } from 'lodash';
```
andreogle, line 3:
```suggestion
import pick from 'lodash/pick';
```

packages/airnode-deployer/src/handlers/aws/index.ts
andreogle: Yeah I agree. I like this.

packages/airnode-node/src/adapters/http/worker.ts
andreogle: I'm not sure I understand, sorry. The AWS handlers in the deployer don't look like they expect the `config` object to be present in the payload.

https://github.com/api3dao/airnode/blob/b152f7e3426aa13e3a12e9ec616036642ecdcba3/packages/airnode-deployer/src/handlers/aws/index.ts#L18
https://github.com/api3dao/airnode/blob/b152f7e3426aa13e3a12e9ec616036642ecdcba3/packages/airnode-deployer/src/handlers/aws/index.ts#L41
https://github.com/api3dao/airnode/blob/b152f7e3426aa13e3a12e9ec616036642ecdcba3/packages/airnode-deployer/src/handlers/aws/index.ts#L5

packages/airnode-deployer/src/handlers/aws/index.ts
amarthadan: I agree that checking the content of the `event` object is a good idea but wouldn't it be better to do it on the `airnode-node` side instead of `airnode-deployer`? In that case code would be only at one place even for different cloud providers.

amarthadan: > Also, I still don't understand completely what is contained in the event object (especially how config ges there in this example).

I'm not sure what you mean. The config is parsed from `config.json` (line 5).

packages/airnode-node/src/handlers/call-api.ts
```diff
@@ -55,17 +54,19 @@ function buildOptions(
   };
 }
 
-function createPrintableValue(value: adapter.ValueType): string {
-  if (Array.isArray(value)) return JSON.stringify(value);
```
amarthadan, line 59: This was definitely done this way for a reason but I don't remember why. Are you sure that stringifying everything is ok?

packages/airnode-node/src/handlers/call-api.ts
```diff
@@ -55,17 +54,19 @@ function buildOptions(
...
+export interface ApiCallOptions {
+  readonly forTestingGateway?: boolean;
+}
 
-  return BigNumber.isBigNumber(value) ? adapter.bigNumberToString(value) : value.toString();
+export interface CallApiPayload {
+  readonly config: Config;
+  readonly aggregatedApiCall: AggregatedApiCall;
+  readonly apiCallOptions: ApiCallOptions;
```
amarthadan, line 64: Wouldn't be better to make this optional?

packages/airnode-node/src/adapters/http/worker.ts
Siegrift: Yeah, thanks. I was confused because actually the I've broken things with this refactor - I was not passing the `config` correctly and I was confused why is not passed anywhere...

Anyway, I'll separate the refactor part out of this PR and try to do in in a separate branch (and PR).

packages/airnode-deployer/src/handlers/aws/index.ts
Siegrift: This is my bad - I wrongly refactored the code and then it made no sense to me... I have removed the refactor parts and will improve that in a separate PR

packages/airnode-node/src/handlers/call-api.ts
Siegrift: I've changed the response to be a JSON, but I've confirmed that it works (I deployed Airnode and used the testing gateway). I will document it's output as part of https://api3dao.atlassian.net/browse/AN-355

packages/airnode-node/src/handlers/call-api.ts
Siegrift: By not making it optional I can enforce the AWS `callApi` handler to pass this value - so I'd say this is better.

Siegrift:
> I agree that checking the content of the event object is a good idea but wouldn't it be better to do it on the airnode-node side instead of airnode-deployer? In that case code would be only at one place even for different cloud providers.

Maybe we can enforce it on the type level only - now I see that mixing this refactor was a premature mistake by me (sorry) so let's postpone this conversation.

packages/airnode-node/src/handlers/call-api.ts
Siegrift: Btw. the response can now be a really big JSON, because the `rawValue` will be whatever the API response is - but I think it's a worthy information to have and users can just copy the JSON to any online format tool to better visualize what's in there.

And I plan to add all of that to the docs tomorrow.

packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -61,5 +61,6 @@ export async function testApi(event: any) {
     return { statusCode: 400, body: JSON.stringify({ error: err.toString() }) };
   }
 
-  return { statusCode: 200, body: JSON.stringify(result) };
+  // NOTE: We do not want the user to see {"value": <actual_value>}, but the actual value itself
+  return { statusCode: 200, body: result!.value };
```
Siegrift, line 65: I've remove `toString` because the output was unusable. Now it's proper JSON and that makes things like formatting a bit easier.

Siegrift: https://github.com/api3dao/api3-docs/pull/417

docker/Dockerfile
```diff
@@ -3,6 +3,11 @@ ARG build=local
 # Environment
 FROM node:14.17-alpine3.14 AS environment
 
+# Download solidity compiler as per: https://github.com/nomiclabs/hardhat/issues/1280#issuecomment-949822371
+# The binary filename is taken from: https://github.com/ethereum/solc-bin/tree/gh-pages/linux-amd64
+RUN mkdir -p /root/.cache/hardhat-nodejs/compilers/linux-amd64 \
+  &&  wget -O /root/.cache/hardhat-nodejs/compilers/linux-amd64/solc-linux-amd64-v0.8.9+commit.e5eed63a https://binaries.soliditylang.org/linux-amd64/solc-linux-amd64-v0.8.9+commit.e5eed63a
```
amarthadan, line 9: Why are you running this in the `environment` phase? All other steps inherit from this one, is it needed everywhere? I think the `preparation` phase would be sufficient, wouldn't it?

amarthadan (approved):

docker/Dockerfile
Siegrift: Yeah, it only needs to be downloaded before `yarn install` - thanks for noticing

PR #670, opened 2021-11-14
Add retryOn429 for CI doc link checking

dcroote:
The Continuous Build workflow fails somewhat frequently because of 429 "Too Many Requests" errors returned when checking many github.com links. This PR adds a retry with delay.

### DIFF
--- a/.github/workflows/mlc_config.json
+++ b/.github/workflows/mlc_config.json
@@ -1,0 +1,5 @@
+{
+  "retryOn429": true,
+  "retryCount": 3,
+  "fallbackRetryDelay": "10s"
+}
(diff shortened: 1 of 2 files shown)


.github/workflows/mlc_config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "retryOn429": true,
+  "retryCount": 3,
+  "fallbackRetryDelay": "45s"
```
Siegrift, line 4: 45 s = 45 seconds? Seems like too much.

Maybe something like 5s would be enough?

dcroote: Yes "45s" = 45 seconds. Github rate limits quite aggressively, and [most open source packages using this Action](https://sourcegraph.com/search?q=context:global+%22fallbackRetryDelay%22:+lang:json&patternType=literal) use "30s", but we can decrease it to 10s and increase it if we continue receiving 429 errors.

Siegrift: Let's see if its enough :)

Siegrift (approved):

PR #671, opened 2021-11-14
Improve admin CLI generate-mnemonic command output

Siegrift:
https://api3dao.atlassian.net/browse/AN-341

image

### DIFF
--- a/packages/airnode-admin/src/cli.ts
+++ b/packages/airnode-admin/src/cli.ts
@@ -448,7 +448,13 @@ yargs
     'Generates a random mnemonic. Uses "ethers.Wallet.createRandom" under the hood.',
     async () => {
       const mnemonic = await admin.generateMnemonic();
-      console.log(`Generated mnemonic: ${mnemonic}`);
+      const lines = [
+        'This mnemonic is created locally on your machine using "ethers.Wallet.createRandom" under the hood.',
+        'Make sure to back it up securely, e.g., by writing it down on a piece of paper:',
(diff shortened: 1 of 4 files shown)


Siegrift:
I don't think this change needs a changeset.

packages/airnode-admin/src/cli.ts
```diff
@@ -448,7 +448,13 @@ yargs
     'Generates a random mnemonic. Uses "ethers.Wallet.createRandom" under the hood.',
     async () => {
       const mnemonic = await admin.generateMnemonic();
-      console.log(`Generated mnemonic: ${mnemonic}`);
+      const lines = [
+        'This mnemonic is created locally on your machine using "ethers.Wallet.createRandom" under the hood.',
+        'Make sure to back it up securely, e.g., by writing it down on a piece of paper:',
+        '',
+        mnemonic,
```
andreogle, line 455: We could maybe make some of the outputs a bit nicer using chalk? Not in the scope of this PR, but something to consider. What do you think?

```suggestion
        chalk.red(mnemonic),
```

andreogle (approved):

packages/airnode-admin/src/cli.ts
amarthadan: I wouldn't recommend that for the data that should be copied as it can copy the coloring markers from the terminal (e.g. `\033[0;31m`) and therefore broke the pasted output.

PR #672, opened 2021-11-14
AN-337 - Add interpolation to validator

drgy:


packages/airnode-validator/README.md
```diff
@@ -31,6 +31,13 @@ Custom templates can be used, by providing path to the validator template file i
 yarn run validate --template="templates/0.2/config.json" --specs="exampleSpecs/config.specs.json"
 ```
 
+Specification files can be interpolated with .env files, interpolation will inject values of variables from the .env
+file into the specification where variables are in the following format: `${VARIABLE_NAME}`
+
+```sh
+yarn run validate --template="config" --specs="exampleSpecs/secrets.config.json" --interpolate="exampleSpecs/secrets.env"
```
Siegrift, line 38: I think this makes a bit more sense

```suggestion
yarn run validate --template="config" --specs="exampleSpecs/secrets.config.json" --secrets="exampleSpecs/secrets.env"
```

packages/airnode-validator/README.md
```diff
@@ -66,7 +73,8 @@ validator templates with conversion actions.
 
 ## Usage
 
-Convertor works the same way as validator and can be invoked with the `convert` command, for example:
+Convertor works the same way as validator (including interpolation) and can be invoked with the `convert` command, for
+example:
 
 ```sh
 yarn run convert --template="conversions/oas@3.0------ois@0.2.json" --specs="exampleSpecs/OAS.specs.json"
```
Siegrift, line 80: Can you change this command as well?

I think the validator should always depend on both `secrets.env` and `config.json` and throw an error if one is not provbided.

Siegrift (changes requested):
Could you also write some basic tests for the interpolation?

packages/airnode-validator/README.md
amarthadan: I think that we agreed on the call that if the secrets file is not provided the interpolation is simply skipped.

packages/airnode-validator/exampleSpecs/secrets.env
```diff
@@ -0,0 +1,4 @@
+gateway_key=012345678901234567890123456789
+airnode_rrp=0x534DfA0d242e7fDCFD096C2B1a3D70F172008778
+cloud_provider=aws
+node_version=0.2.0
```
amarthadan, line 4: Can you name the variables as we usually do, all uppercase?

packages/airnode-validator/src/commands/utils.ts
```diff
@@ -45,6 +47,13 @@ conversionTemplates.forEach((file) => {
   conversions[fromName][fromVersion][toName].push(toVersion);
 });
 
+// Regular expression that does not match anything, ensuring no escaping or interpolation happens
+// https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L199
+const NO_MATCH_REGEXP = /($^)/;
+// Regular expression matching ES template literal delimiter (${}) with escaping
+// https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L175
+const ES_MATCH_REGEXP = /\$\{([^\\}]*(?:\\.[^\\}]*)*)\}/g;
```
amarthadan, line 55: As I mentioned before, I don't like that you just copy&pasted the regular expressions and code from the `airnode-node` package. Feel free to refactor the code there so you can import just the interpolation itself or move the implementation here and import it in the `airnode-node` package but the code should be only in one place.

packages/airnode-validator/src/convertor.ts
```diff
@@ -1,66 +1,75 @@
...
   }
 
   let template, specs;
+  const messages: Log[] = [];
 
-  try {
-    template = fs.readFileSync(path.resolve(templatePath), 'utf-8');
-  } catch (e) {
-    return { valid: false, messages: [logger.error(`Unable to read file ${path.resolve(templatePath)}`)], output: {} };
+  if ((template = utils.readJson(templatePath, messages)) === undefined) {
```
amarthadan, line 25: You usually don't need to check specifically for `undefined` unless your function can return multiple falsy values (`undefined`, `null`, `false`) and you need to distinguish between them.
Doing `if (!functionCall())` is usually enough.

amarthadan, line 25: Also, not a big fan of assignments within an `if` statement. I find it unnecessarily confusing at first glance.

packages/airnode-validator/README.md
Siegrift: I don't recall. 

Anyway, is that a good idea? This can cause problems for the user with no real benefit.

amarthadan (changes requested):

packages/airnode-validator/README.md
amarthadan: Not all validated files contain sensitive information and don't need an interpolation (e.g. `receipt.json`). You would have to provide an empty file just to make the flow work. That doesn't make much sense to me. And I don't really see problems it would cause, can you elaborate?

packages/airnode-validator/README.md
drgy: Changed in 65f4f38

packages/airnode-validator/README.md
drgy: This sounds like not very intuitive behavior, it will be very confusing and cause many problems. From what I've understood the longterm goal is to limit the work with config files needed to run airnode to as little as possible, especially these kind of checks should run automatically in the background without user knowing. There is already github action set up in one of the repositories that's validating the OIS files in the repository. Having to provide valid `secrets.env` file in order to validate the OIS just does not make much sense, and it's much easier to just put one extra argument when the interpolation is needed rather than creating an empty file every single time when it's not.

packages/airnode-validator/exampleSpecs/secrets.env
drgy: Renamed in 7c90b7f

packages/airnode-validator/src/commands/utils.ts
drgy: Removed in 538ee33

packages/airnode-validator/src/convertor.ts
drgy: Changed in 2cd2f5f

drgy: I think this comes down to personal preference. If it would be used in more complex condition I would agree with you, but I believe this condition is simple enough and it is easy to understand, moving it out of the condition would make it unnecessarily long, so I would prefer leaving it as is.

drgy:
I added [a test](https://github.com/api3dao/airnode/blob/2cd2f5fe3486d16ed186c2627481abf65b4010d5/packages/airnode-validator/src/validator.test.ts#L360), I think it might be just enough, as the interpolation is not something complex. If you have any idea for other tests that would be useful let me know.

amarthadan (approved):
Why is there an empty file `packages/airnode-validator/exampleSpecs/ois.secrets.env`?

Other than that LGTM.

packages/airnode-validator/README.md
Siegrift: > Not all validated files contain sensitive information and don't need an interpolation

Good point. But I was mainly talking about validating `config.json` since interpolations is not supported (and doesn't make sense) for `receipt.json`. I'd like to have `secrets.env` required when validting `config.json` - because it;'s easy to forget to supply the file and accept invalid configuration. 

Another option is to issue a warning in this case (when validating config.json without secrets)

drgy: Having a special behavior just for `config.json` brings another layer to the confusion imho. From the perspective of the user I still think its going to be much easier to figure out that I forgot to provide `.env` file when its needed (just by looking at the values that are in the error messages), rather then accidentally validating with `.env` file that I didn't even know Im validating with.

Siegrift: I disagree actually - `config.json` on its own doesn't make much sense. We should always be talking a about the pair (config.json + secrets.env) as this form "the airnode configuration". Also, most people will be using the secrets as well because the providerUrl is a secret and you need that for every Airnode deployment.

That said, I don't wanna block this PR because of this - so feel free to merge and we can implement the check for secrets separately.

Siegrift (approved):

packages/airnode-validator/README.md
bbenligiray: I agree with Jakub in this in general (interpolation should be optional and explicit)

PR #674, opened 2021-11-15
Implement timestamp value for _type parameter

Siegrift:
https://api3dao.atlassian.net/browse/AN-344

docs: https://github.com/api3dao/api3-docs/pull/413

### DIFF
--- a/packages/airnode-adapter/src/response-processing/extraction.ts
+++ b/packages/airnode-adapter/src/response-processing/extraction.ts
@@ -97,11 +97,21 @@ export function splitReservedParameters(parameters: ReservedParameters): Reserve
 }
 
 function extractSingleResponse(data: unknown, parameters: ReservedParameters) {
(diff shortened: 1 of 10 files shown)


packages/airnode-adapter/src/response-processing/extraction.ts
```diff
@@ -96,11 +96,21 @@ export function splitReservedParameters(parameters: ReservedParameters): Reserve
 }
 
 function extractSingleResponse(data: unknown, parameters: ReservedParameters) {
+  const parsedArrayType = parseArrayType(parameters._type);
+  const type = parsedArrayType?.baseType ?? parameters._type;
+
+  if (!isNumericType(type) && parameters._times) {
```
Siegrift, line 102: This is new - but I think it's a good idea to do.

andreogle (approved):

PR #675, opened 2021-11-15
Fix building docker artifacts

Siegrift:
For some strange reason, building docker artifacts was failing for me locally and this seems to fix the issue - not sure how though.


Siegrift:
Actually I'll merge this to https://github.com/api3dao/airnode/pull/669

PR #678, opened 2021-11-16
A3P-39: switch readonly variables from storage to memory

acenolaza:
Realized that I was using storage instead of memory for local vars on those getter functions that I created just to avoid the "stack to deep" error in AirnodeTokenPayment.sol

### DIFF
--- a/packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
+++ b/packages/airnode-protocol/contracts/rrp/authorizers/admin/AirnodeTokenPayment.sol
@@ -294,8 +294,9 @@ contract AirnodeTokenPayment is
         view
         returns (uint64 maximumWhitelistDuration)
     {
-        WhitelistDuration
-            storage whitelistDuration = airnodeToWhitelistDuration[_airnode];
+        WhitelistDuration memory whitelistDuration = airnodeToWhitelistDuration[
+            _airnode
+        ];
         maximumWhitelistDuration = whitelistDuration.maximum != 0
             ? whitelistDuration.maximum
             : DEFAULT_MAXIMUM_WHITELIST_DURATION;
@@ -310,8 +311,9 @@ contract AirnodeTokenPayment is
         view
         returns (uint64 minimumWhitelistDuration)
     {
-        WhitelistDuration
-            storage whitelistDuration = airnodeToWhitelistDuration[_airnode];
+        WhitelistDuration memory whitelistDuration = airnodeToWhitelistDuration[
+            _airnode
+        ];
         minimumWhitelistDuration = whitelistDuration.minimum != 0
             ? whitelistDuration.minimum
             : DEFAULT_MINIMUM_WHITELIST_DURATION;


bbenligiray:
I prefer using `storage` any time the variable is read-only. What are the implications?

acenolaza:
> I prefer using `storage` any time the variable is read-only. What are the implications?

Even though it might be a bit more expensive to use `memory` (since we are making a local copy) I was thinking it is safer and more semantic to do that inside a view or getter function since this function should not be modifying the values in the state var.

Ashar2shahid (approved):
looks good 👍 

> Even though it might be a bit more expensive to use memory (since we are making a local copy) I was thinking it is safer and more semantic to do that inside a view or getter function since this function should not be modifying the values in the state var.

isn't memory cheaper to use ?

bbenligiray:
Actually can you test and report the gas usage with both?

acenolaza:
> Actually can you test and report the gas usage with both?

When using `storage`:
```
·----------------------------------------------------------------------------|----------------------------|-------------|-----------------------------·
|                            Solc version: 0.8.9                             ·  Optimizer enabled: false  ·  Runs: 200  ·  Block limit: 30000000 gas  │
·············································································|····························|·············|······························
|  Methods                                                                                                                                            │
·······································|·····································|··············|·············|·············|···············|··············
|  Contract                            ·  Method                             ·  Min         ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
·······································|·····································|··············|·············|·············|···············|··············
|  AirnodeTokenPayment                 ·  makePayment                        ·           -  ·          -  ·     149452  ·            2  ·          -  │
·······································|·····································|··············|·············|·············|···············|··············
|  Deployments                                                               ·                                          ·  % of limit   ·             │
·············································································|··············|·············|·············|···············|··············
|  AirnodeTokenPayment                                                       ·           -  ·          -  ·    2655495  ·        8.9 %  ·          -  │
·----------------------------------------------------------------------------|--------------|-------------|-------------|---------------|-------------·
```

When using `memory`:
```
·----------------------------------------------------------------------------|----------------------------|-------------|-----------------------------·
|                            Solc version: 0.8.9                             ·  Optimizer enabled: false  ·  Runs: 200  ·  Block limit: 30000000 gas  │
·············································································|····························|·············|······························
|  Methods                                                                                                                                            │
·······································|·····································|··············|·············|·············|···············|··············
|  Contract                            ·  Method                             ·  Min         ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
·······································|·····································|··············|·············|·············|···············|··············
|  AirnodeTokenPayment                 ·  makePayment                        ·           -  ·          -  ·     149452  ·            2  ·          -  │
·······································|·····································|··············|·············|·············|···············|··············
|  Deployments                                                               ·                                          ·  % of limit   ·             │
·············································································|··············|·············|·············|···············|··············
|  AirnodeTokenPayment                                                       ·           -  ·          -  ·    2655495  ·        8.9 %  ·          -  │
·----------------------------------------------------------------------------|--------------|-------------|-------------|---------------|-------------·

Gas usage seems to be negligible.

The main idea of using memory was that a `view` functions should never change state and I thought that a good way to prevent that is by making copies in memory of state vars instead of using pointers but I also understand that higher gas usage might be an issue too.

bbenligiray:
So that 206 gas is the reason why I prefer `storage` over `memory` for view functions. It is indeed error-prone because you may end up updating that variable forgetting that you're also updating storage, but that's what the `view` modifier is there for. So I'm closing this PR, but conceding the fact that using `storage` here is a subjective decision with potential downsides.

PR #680, opened 2021-11-18
AN-261: Add authenticated coinmarketcap example

dcroote:
https://api3dao.atlassian.net/browse/AN-261

```sh
> yarn make-request
yarn run v1.22.11
Making request...
Waiting for fulfillment...
Request fulfilled
ETH price is 4237.325301 USD
```

As an aside, for the local flow, is there any reason not to have the option to use an existing docker client image rather than having to build the artifacts container and Airnode? I tested this and v0.2.2 worked just fine.


packages/airnode-examples/integrations/README.md
```diff
@@ -29,6 +29,7 @@ When adding an integration you need to do a few things:
    the data received by the Airnode
 3. Create all the necessary files (see the section above) for your integration. Some of the files are explained more in
    depth below
+4. Run `yarn build` from the top-level `airnode` directory to build the contract artifacts.
```
Siegrift, line 32: :+1:

packages/airnode-examples/integrations/coinmarketcap/README.md
```diff
@@ -0,0 +1,15 @@
+# CoinMarketCap example integration
```
Siegrift, line 1:
```suggestion
# Authenticated CoinMarketCap example integration
```

or put "authenticated" somewhere (maybe even folder name `authenticated-coinmarketcap`) so it's easier to find for users.

packages/airnode-examples/integrations/coinmarketcap/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+This CoinMarketCap cryptocurrency price example builds upon the foundations of the CoinGecko example by incorporating
+authentication with an `apiKey` security scheme.
+
+In order to focus on authentication, many of the other aspects of this example are similar to the CoinGecko example
+including a single Ethereum price endpoint, the `_times` reserved parameter for handling floating point prices, and much
+of the `secrets.env` file.
+
+New to this example is a `config.json` security scheme that authenticates a request with CoinMarketCap via their
+preferred method: a request header named `X-CMC_PRO_API_KEY`. Your API key is provided as a value via interpolation of
```
Siegrift, line 11:
```suggestion
preferred method - a request header named `X-CMC_PRO_API_KEY`. Your API key is provided as a value via interpolation of
```

packages/airnode-examples/scripts/choose-integration.ts
```diff
@@ -63,6 +63,15 @@ const questions: PromptObject[] = [
       return '';
     },
   },
+  {
+    // skip for all integrations other than coinmarketcap
+    type: (_prev, values) => (values.integration === 'coinmarketcap' ? 'text' : null),
```
Siegrift, line 68: I'd rather like the user to copy the secret to `secrets.env` manually and not pollute this script - as this one should remain as "example-agnostic" as possible

Siegrift (changes requested):
> As an aside, for the local flow, is there any reason not to have the option to use an existing docker client image rather than having to build the artifacts container and Airnode? I tested this and v0.2.2 worked just fine.

Theoretically no reason - but by building the docker locally it's more bulletproof. For example, once we enforce config.json version checks in Airnode client it might throw an error.

packages/airnode-examples/integrations/coinmarketcap/README.md
Siegrift: Also, maybe mention how to obtain this api key (and that it's actually free)?

packages/airnode-examples/scripts/choose-integration.ts
dcroote: I thought about that, but a manual copy step is too easily missed, leading to a poor user experience. I think the tradeoff of a busier script but better user experience is favorable. If you really don't want it here, I could create a `prompt` for this integration's `yarn create-airnode-secrets` step so the user can enter it there.

Siegrift: I'd rather be more conservative about this - in the future there will be many examples (like this https://github.com/vercel/next.js/tree/canary/examples) and I expect this file to grow more once we support GCP or azure.

> If you really don't want it here, I could create a prompt for this integration's yarn create-airnode-secrets step so the user can enter it there.

Yeah, that seems like a better place.

dcroote: That's fair, but one more thought given the discussion of future integrations: if we expect a number of integrations to use API key authentication then we could retain this step and use a standardized `secrets.env` variable name e.g. `API_KEY`. This would avoid duplicating the request for an API key in each integration's `create-secrets.ts` script. This step would only need to be modified to check if the integration is in a list of ones requiring authentication.

Siegrift: It's better if the examples are self contained so I don't think that it's a good idea.

packages/airnode-examples/integrations/coinmarketcap/README.md
dcroote: 👍 title updated and folders renamed

packages/airnode-examples/integrations/coinmarketcap/README.md
dcroote: I left the comma because grammatically it clarifies the preceding independent clause, but I did implement your other suggestion by adding that the user can get an API key with a free Basic CoinMarketCap account.

packages/airnode-examples/scripts/choose-integration.ts
dcroote: I'm still not so sure, but I moved it as you requested. I would appreciate if you could check if what I've done in `check-secrets.ts` is idiomatic.

packages/airnode-examples/integrations/README.md
Siegrift: Actually, this is not enough - also the users need to run this to compile the contracts.

We should add this as a step to the "main" README

packages/airnode-examples/integrations/authenticated-coinmarketcap/create-secrets.ts
```diff
@@ -0,0 +1,25 @@
...
+import { getCommonSecrets, writeSecrets } from '../utils';
+import { runAndHandleErrors } from '../../src';
+
+const questions: PromptObject[] = [
+  {
+    type: 'text',
+    name: 'apiKey',
+    message: [
+      'The CoinMarketCap example requires an API key, which can be acquired with a free, Basic account.',
+      'Please enter your CoinMarketCap API key',
```
Siegrift, line 11:
```suggestion
      'Please enter your CoinMarketCap API key (it will never leave this machine)',
```

Siegrift: I am not sure whetehr we should do this or not. WDYT?

packages/airnode-examples/integrations/authenticated-coinmarketcap/create-secrets.ts
```diff
@@ -0,0 +1,25 @@
...
+const createSecrets = async () => {
+  const secrets = await getCommonSecrets();
+  const response = await prompts(questions);
+  secrets.push(`CMC_PRO_API_KEY=${response.apiKey}`);
+  writeSecrets(secrets);
+};
+
+export default createSecrets;
+
+runAndHandleErrors(createSecrets);
```
Siegrift, line 25: This shouldn't be here because it runs twice (when following the "main" README)

Siegrift (changes requested):
Minor changes, but I've tested the example now and it works :) 

:100: gj!

packages/airnode-examples/integrations/authenticated-coinmarketcap/create-secrets.ts
dcroote: >Actually, this is not enough - also the users need to run this to compile the contracts.
We should add this as a step to the "main" README

Isn't this covered by Setup? - https://github.com/api3dao/airnode/blame/master/packages/airnode-examples/README.md#L26-L29

> I am not sure whetehr we should do this or not. WDYT?

Is it true the API key doesn't leave the machine even if the user chooses the `aws` flow? I admit I don't know the inner workings of how secrets are handled by the deployer.

packages/airnode-examples/integrations/authenticated-coinmarketcap/create-secrets.ts
dcroote: Ah, thanks. I removed `runAndHandleErrors(createSecrets);` and its import.

packages/airnode-examples/integrations/authenticated-coinmarketcap/create-secrets.ts
Siegrift: > Is it true the API key doesn't leave the machine even if the user chooses the aws flow? I admit I don't know the inner workings of how secrets are handled by the deployer.

You're right - they do

Siegrift (approved):
One last thing to fix is the `yarn build`, but otherwise this is good to go :)

dcroote:
> One last thing to fix is the `yarn build`, but otherwise this is good to go :)

The user is actually already instructed to executed `yarn run build` as part of the Setup in the examples README:
> ## Setup
Follow the [repository instructions](https://github.com/api3dao/airnode#instructions). Also, make sure you have `yarn`
installed. If you want to run Airnode as a docker container, you'll also need to have `docker` installed.

Siegrift:
In that case feel free to merge.

PR #681, opened 2021-11-18
[PoC]: Update make request file to allow sending multiple requests at once

Siegrift:
## Intro

I've tried running Airnode locally and modified the `make-request.ts` from the examples package to send multiple requests at once with two versions:

- Same sponsor
- Different sponsor

The results were the same, because hardhat is super fast. So I've implemented https://github.com/Siegrift/hardhat-fastify-gateway to slow it down for certain calls (`eth_call` is the one made by Airnode to fulfilll the transaction). The results showed that Airnode behaves correctly - requests from same sponsor are made sequentially, while from different sponsors in parallel.

## How to run

Follow the steps in `airnode-examples` package - however:
1. During `choose-integration` - choose `http://localhost:9999` as the provider URL
2. Clone and run the server (it runs on port 9999) https://github.com/Siegrift/hardhat-fastify-gateway
3. You may skip step 13 and 14 (they are done by the script now)

It's recommended to run `make-request` (which makes multiple requests now) while airnode is not running, and only run the airnode (by `yarn run-airnode-locally`) after all of those requests were made and are waiting for fulfillment.

### DIFF
--- a/packages/airnode-examples/scripts/make-request.ts
+++ b/packages/airnode-examples/scripts/make-request.ts
@@ -1,5 +1,6 @@
 import { ethers } from 'ethers';
-import { deriveAirnodeXpub, deriveSponsorWalletAddress } from '@api3/airnode-admin';
+import { deriveAirnodeXpub, deriveSponsorWalletAddress, getAirnodeRrp, sponsorRequester } from '@api3/airnode-admin';
+import range from 'lodash/range';
 import {
   getDeployedContract,
   getProvider,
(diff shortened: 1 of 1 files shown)


bbenligiray:
Is this about the stress test? We were talking about if [this](https://github.com/api3dao/airnode/blob/e50dafe57a0a083fd9fc3b2816fe8352a673038d/packages/airnode-node/src/evm/fulfillments/index.ts#L120) was fulfillng requests made with different sponsors in parallel.

Siegrift:
Yeah, I reviewed the stress-test code and was talking to Aaron privately on Slack (so yo didn't have visibility).

This PoC demonstrates that Airnode is doing the intended thing (at least when run locally, assuming I didn't screw up something in this PoC code)

Siegrift:
I guess this doesn't have to be a PR since it never wants to get merged

PR #682, opened 2021-11-18
Improve unit test time

Siegrift:
https://api3workspace.slack.com/archives/C02AYRX8D89/p1637152987184400


packages/airnode-protocol/scripts/run-hardhat-tests.ts
```diff
@@ -0,0 +1,50 @@
+// NOTE: Hardhat tests are terribly slow when there are large number of tests present
```
andreogle, line 1: Where is this script called from? I can't find anything that calls `run-hardhat-test.ts`

Siegrift: I've actually accidentally removed the `test` command (probably during cleanup rebase) :pleading_face: 

I'll add it back - hopefully the test metric I did was done when the `test` command was present :)

andreogle (approved):
Looks good to me 👍🏻 , but CI is failing

PR #683, opened 2021-11-19
AN-222: Update validator readme

drgy:


wkande (approved):
I think this look real good. The link is correct.

PR #684, opened 2021-11-19
Reorganize protocol directories

bbenligiray:
Problems:
- Adminnable contracts are lying around, creating confusion
- "Authorizer admin" contracts were in the wrong directory
- There is no clear distinction between contracts to be audited vs. others

Solutions:
- Moved the archived adminnable contracts to https://github.com/bbenligiray/airnoderrp-archive we won't need them anyway
- Moved the authorizer admin contracts
- Moved the contracts to be audited (and their tests) under `WIP/` where the exact same directory structure is used as `contracts/`

I'll also close the PSP draft PRs and move `psp/` under `WIP/` in an upcoming PR


acenolaza (approved):
LGTM

PR #686, opened 2021-11-20
Add ignored example files to .prettierignore

dcroote:
Running through the `airnode-examples` flow generates files / directories that should be ignored by `prettier`. They are already ignored in the airnode-examples `.gitignore` file, as shown here:
https://github.com/api3dao/airnode/blob/06f14e72c34daeef08ae56b15c7e24b470ad4002/packages/airnode-examples/.gitignore#L8-L11

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -13,6 +13,8 @@
 
 packages/airnode-validator/test/fixtures
 packages/airnode-protocol/src/contracts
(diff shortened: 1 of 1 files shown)


Siegrift (approved):
Thanks - I find it really stupid that prettier doesn't take gitignore into account when formatting files (and doesn't even have an option for it)...

PR #687, opened 2021-11-22
AN-345: Add example with multiple encoded values

dcroote:
This new weather example encodes 4 values, including `timestamp`.

**Updated** example output:

```sh
> yarn make-request
yarn run v1.22.11
Making request...
Waiting for fulfillment...
Request fulfilled
Timestamp of sunset: 1637856008 (Thu, 25 Nov 2021 16:00:08 GMT)
Temperature: 2.5 C
Weather: Clouds
Timestamp of encoded transaction: 1637959322 (Fri, 26 Nov 2021 20:42:02 GMT)
```


packages/airnode-examples/integrations/weather-multi-value/README.md
```diff
@@ -0,0 +1,27 @@
+# Authenticated, Multiple Encoded Value OpenWeather Example Integration
+
+This authenticated OpenWeather historical weather data example builds upon the foundations of the authenticated
+CoinMarketCap example by incorporating the encoding of multiple reserved parameters, which Airnode extracts and converts
+separately.
+
+As this is an authenticated example, you will need to sign up with OpenWeather in order to get a free API key. Like the
```
Siegrift, line 7: 1. I'd link `OpenWeather` to their homepage
2. I'd remove `Like the CoinMarketCap example` - I don't think this is necessary information for the reader

packages/airnode-examples/integrations/weather-multi-value/README.md
```diff
@@ -0,0 +1,27 @@
...
+the `_type` reserved parameter is a comma-separated string with the following "split values":
+
+1. The datetime of sunset encoded as a `uint256`
+2. The temperature encoded as an `int256`
+3. The current weather e.g. "Clouds" encoded as a `string`
+4. The `timestamp`, which corresponds to when a transaction was encoded, encoded as a `uint256`.
+
+Note the comma-separated `_path` reserved parameter only has three "split values" as `timestamp` should not include a
+`_path` value. Lastly, the comma-separated `_times` reserved parameter illustrates that only the temperature (the second
+"split value") is multiple by a value, which is necessary in order to handle the floating point temperature returned by
```
Siegrift, line 20:
```suggestion
"split value") is multiplied by a value, which is necessary in order to handle the floating point temperature returned by
```

packages/airnode-examples/integrations/weather-multi-value/request-utils.ts
```diff
@@ -0,0 +1,33 @@
+import { encode } from '@api3/airnode-abi';
+import { cliPrint, getDeployedContract, readIntegrationInfo } from '../../src';
+
+// OpenWeather historical API allows free access only to the last 5 days
+// Use date for yesterday, excluding milliseconds
+const dt = Math.floor((Date.now() - 86400000) / 1000);
```
Siegrift, line 6: It's a good practice to let the computer compute the result (this better communicates the expression value)

```suggestion
const dt = Math.floor((Date.now() - 24 * 60 * 60 * 1000) / 1000);
```

packages/airnode-examples/integrations/weather-multi-value/request-utils.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const getEncodedParameters = () => {
+  return encode([
+    { name: 'lat', type: 'string', value: '51.507222' },
```
Siegrift, line 10: Maybe add a comment about what is the location :)

packages/airnode-examples/integrations/weather-multi-value/README.md
Siegrift: Also, getting API key seems to be a bit difficult - at least I didn't mange to get a valid one...

When I try:
```
curl http://api.openweathermap.org/data/2.5/onecall/timemachine\&appid\=01b02c3e...
```

I get `{"cod":401, "message": "Invalid API key. Please see http://openweathermap.org/faq#error401 for more info."}`

When I try:

```
curl http://api.openweathermap.org/data/2.5/forecast\?id\=524901\&appid\=01b02c3...
```

it works.

Can you elaborate a bit more how to retrieve the API key for this particular API call? (Or it's just some stupid mistake on my side?)

Siegrift (changes requested):

packages/airnode-examples/integrations/weather-multi-value/README.md
dcroote: The following command works. Perhaps there was a delay in the API key being activated or something slightly off with the curl command itself (note: replace `MYKEY` with the API key and the quotes are necessary):
```sh
curl "https://api.openweathermap.org/data/2.5/onecall/timemachine?lat=51.507222&lon=-0.1275&dt=1637545002&appid=MYKEY"

Comment (dcroote) [reply packages/airnode-examples/integrations/weather-multi-value/README.md]: I addressed suggestions 1 and 2 in the latest commit. I also added the above curl command to the README so that users can ensure their API key is active (works on Windows cmd and Linux)

File: packages/airnode-examples/integrations/weather-multi-value/request-utils.ts (hunk 4 above)
Comment (dcroote) [reply packages/airnode-examples/integrations/weather-multi-value/request-utils.ts:11]: Added.

**Review by Siegrift — APPROVED** [review]
I've tested again and it's working (using the same API key).

One minor thing:
* Myabe we can format the timestamp in a nicer way for the user

E.g. 
```
Timestamp of sunset: 1637856008 (25, 2021 4:00:08 PM)
```

or something similar - but either way this is good to go.

dcroote:
Output updated with timestamp formatting, see my [updated comment above](https://github.com/api3dao/airnode/pull/687#issue-785572853).

PR #688, opened 2021-11-23
Add support for GCP cloud provider

amarthadan:
Issues [AN-307](https://api3dao.atlassian.net/browse/AN-307), [AN-309](https://api3dao.atlassian.net/browse/AN-309)

I was able to include AN-309 as it wasn't that hard to do and testing was much easier that way.

Main changes:
* Added full support for GCP cloud (except HTTP gateway) - you can now deploy Airnode to GCP via deployer (just CLI, there may be some changes needed to support credential passing for containers)
* Extracted common handler to be available in the root of cloud function (needed by GCP)
* Added possible `import` step to the Terraform flow (needed because of the GCP's AppEngine)
* Increase AWS lambdas memory size - both AWS and GCP functions were failing for me on 128 MB memory limit so I set both to 256 MB. This will probably change soon anyway based on the stress test results.

### DIFF
--- a/packages/airnode-deployer/package.json
+++ b/packages/airnode-deployer/package.json
@@ -15,7 +15,7 @@
     "clean": "rimraf -rf ./dist ./.webpack *.tgz",
     "cli": "ts-node bin/deployer.ts",
     "compile": "tsc -p tsconfig.build.json",
-    "copy:terraform": "copyfiles terraform/**/**/*.tf terraform/**/**/*.tpl  dist/",
+    "copy:terraform": "copyfiles terraform/**/**/**/*.tf terraform/**/**/**/*.tpl  dist/",
     "copy:webpack": "copyfiles .webpack/**/*.js .webpack/templates/**/*.json .webpack/conversions/**/*.json dist/",
     "pack": "yarn pack",
     "test": "jest --coverage",
--- a/packages/airnode-deployer/package.json
+++ b/packages/airnode-deployer/package.json
@@ -16,7 +16,7 @@
     "cli": "ts-node bin/deployer.ts",
     "compile": "tsc -p tsconfig.build.json",
     "copy:terraform": "copyfiles terraform/**/**/*.tf terraform/**/**/*.tpl  dist/",
-    "copy:webpack": "copyfiles .webpack/**/**/*.js .webpack/templates/**/*.json .webpack/conversions/**/*.json dist/",
+    "copy:webpack": "copyfiles .webpack/**/*.js .webpack/templates/**/*.json .webpack/conversions/**/*.json dist/",
     "pack": "yarn pack",
     "test": "jest --coverage",
     "webpack": "webpack",
@@ -36,6 +36,7 @@
     "yargs": "^17.0.1"
   },
   "devDependencies": {
+    "@google-cloud/functions-framework": "^2.0.0",
     "@types/lodash": "^4.14.169",
     "@types/node": "^15.12.2",
     "@types/yargs": "^17.0.2",
@@ -47,6 +48,7 @@
     "ts-node": "^10.1.0",
     "typescript": "^4.2.4",
     "webpack": "^5.51.1",
-    "webpack-cli": "^4.8.0"
+    "webpack-cli": "^4.8.0",
+    "webpack-ignore-dynamic-require": "^1.0.0"
   }
 }
--- a/packages/airnode-deployer/src/entrypoint.ts
+++ b/packages/airnode-deployer/src/entrypoint.ts
@@ -1,0 +1,34 @@
+const cloudProvider = process.env.AIRNODE_CLOUD_PROVIDER;
+
+if (!cloudProvider) {
+  throw new Error('Missing cloud specification');
+}
+
+async function cloudHandler() {
+  return await import(/* webpackIgnore: true */ `./handlers/${cloudProvider}`);
+}
+
+export async function startCoordinator(...args: any[]) {
+  const handler = await cloudHandler();
+  return await handler.startCoordinator(...args);
+}
+
+export async function initializeProvider(...args: any[]) {
+  const handler = await cloudHandler();
+  return await handler.initializeProvider(...args);
+}
+
+export async function callApi(...args: any[]) {
+  const handler = await cloudHandler();
+  return await handler.callApi(...args);
+}
+
+export async function processProviderRequests(...args: any[]) {
+  const handler = await cloudHandler();
+  return await handler.processProviderRequests(...args);
+}
+
+export async function testApi(...args: any[]) {
+  const handler = await cloudHandler();
+  return await handler.testApi(...args);
+}
--- a/packages/airnode-deployer/src/entrypoint.ts
+++ b/packages/airnode-deployer/src/entrypoint.ts
@@ -8,27 +8,27 @@ async function cloudHandler() {
   return await import(/* webpackIgnore: true */ `./handlers/${cloudProvider}`);
 }
 
-export async function startCoordinator(...args: any[]) {
+export async function startCoordinator(...args: unknown[]) {
   const handler = await cloudHandler();
   return await handler.startCoordinator(...args);
 }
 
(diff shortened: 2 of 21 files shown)


packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -128,14 +154,29 @@ async function terraformAirnodeManage(
...
+      await execTerraform(
+        { ...execOptions, ignoreError: true },
+        'import',
+        prepareAirnodeManageArguments(cloudProvider, commonArguments),
+        cloudProviderAirnodeImportOptions[cloudProvider.name](cloudProvider as any)
+      );
+    }
+  }
+
+  commonArguments.push('auto-approve');
```
Siegrift, line 177: What is this for?

packages/airnode-deployer/terraform/airnode/aws/modules/function/main.tf
```diff
@@ -47,6 +50,7 @@ resource "aws_lambda_function" "lambda" {
   function_name                  = var.name
   handler                        = var.handler
   runtime                        = "nodejs14.x"
+  memory_size                    = 256
```
Siegrift, line 53: Will we be able to configure this in the future?

packages/airnode-node/src/workers/cloud-platforms/gcp.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    expect(requestMock).toHaveBeenCalledTimes(1);
+    expect(requestMock).toHaveBeenCalledWith({
+      url,
+      method: 'POST',
+      data: parameters.payload,
+    });
+  });
+
+  it('throws an error if the cloud function returns an error', async () => {
+    expect.assertions(5);
```
Siegrift, line 43: I don't like asserting the number of assertions.

Siegrift (approved):
Code (apart from TF, which I skipped) looks good.

packages/airnode-deployer/terraform/airnode/aws/modules/function/main.tf
amarthadan: Yes, I think that's the plan. I'm pretty sure we already have a ticket for that but I can't find it (Jira being Jira...).

packages/airnode-node/src/workers/cloud-platforms/gcp.test.ts
amarthadan: That's actually a suggestion from the [Jest documentation](https://jestjs.io/docs/expect#expectassertionsnumber) when you're testing async functions.

packages/airnode-deployer/src/infrastructure/index.ts
amarthadan: The `import` command uses the same arguments as the following `apply` command except for the `-auto-approve` one so I'm adding it after the `import` is called.

packages/airnode-deployer/src/entrypoint.ts
```diff
@@ -0,0 +1,34 @@
+const cloudProvider = process.env.AIRNODE_CLOUD_PROVIDER;
+
+if (!cloudProvider) {
+  throw new Error('Missing cloud specification');
+}
+
+async function cloudHandler() {
+  return await import(/* webpackIgnore: true */ `./handlers/${cloudProvider}`);
```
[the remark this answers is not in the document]
amarthadan: > Can you explain why we need the `webpackIgnore`?

It's used so that when the webpack is called for the `entrypoint.ts` (or rather `entrypoint.js`) it won't follow the dynamic imports and won't bundle everything into `.webpack/index.js`.

> Also I think `switch` case instead of dynamic import is simpler (and better) solution.

It was done this way for two reasons.
First, this way you won't include AWS-specific node code for GCP and vice versa. Which I guess would not be such a big deal.
Second, I wanted to keep the resulting directory structure (`handlers/aws/index.js`) pretty much the same. Some code expects this structure (when looking for `config.json` and validator templates for example) and I think it's currently easier to keep it than refactoring it. Unfortunately, GCP has a limitation that the main cloud function file must be at the root, that's why I added the `entrypoint.ts` and bundled it separately from actual cloud-specific handlers.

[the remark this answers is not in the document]
andreogle: It's worth adding a code comment to explain this for future readers imo

packages/airnode-deployer/src/entrypoint.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export async function startCoordinator(...args: any[]) {
```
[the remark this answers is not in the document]
andreogle: ^ typo btw if you were planning to commit directly through the Github UI

packages/airnode-deployer/package.json
```diff
@@ -47,6 +48,7 @@
     "ts-node": "^10.1.0",
     "typescript": "^4.2.4",
     "webpack": "^5.51.1",
-    "webpack-cli": "^4.8.0"
+    "webpack-cli": "^4.8.0",
+    "webpack-ignore-dynamic-require": "^1.0.0"
```
andreogle, line 52: Not super excited to add this package with ~1000 weekly downloads and 5 total commits 😬

andreogle (approved):
Also mainly reviewed the code

packages/airnode-node/src/workers/cloud-platforms/gcp.test.ts
Siegrift: But that is for a different use case - it ensures that your callback assertions were called. You don't need to add it just because you write async tests

packages/airnode-deployer/src/entrypoint.ts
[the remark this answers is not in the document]
Siegrift: > First, this way you won't include AWS-specific node code for GCP and vice versa. Which I guess would not be such a big deal.

Mhh, does it work thay way? Don't you need to bundle both because you don't know which will be imported? (But I agree this is not that important).

> and I think it's currently easier to keep it than refactoring it. Unfortunately, GCP has a limitation that the main cloud function file must be at the root

But how does this change things? You can keep the files in the same folders as of now, but use switch statement instead of dynamic import. Or am I missing something?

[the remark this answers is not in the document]
Ashar2shahid: Yeah encountered this GCP limitation when deploying with [serverless](https://www.serverless.com/framework/docs/providers/google/guide/functions)  

> Mhh, does it work thay way? Don't you need to bundle both because you don't know which will be imported? (But I agree this is not that important).

If the provider is specified in the config.json and also in the env variables then do we still need to be able to bundle both?

Ashar2shahid:
I'm still going through the terraform files, but the typescript part looks good. Quick question, are the deployed functions public or private? I remember when I deployed the functions by default they would be private and had to manually change it to public(although I don't think it matters)

packages/airnode-deployer/src/entrypoint.ts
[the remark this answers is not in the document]
amarthadan: > If the provider is specified in the config.json and also in the env variables then do we still need to be able to bundle both?

Yes, because we need to create bundles beforehand so they can be part of the package/Docker image. But during the deployment itself, we want to deploy just the specific one.

packages/airnode-deployer/package.json
amarthadan: I usually don't like it either but this is just a packaged [StackOverflow answer](https://stackoverflow.com/questions/38697321/webpack-require-expression-external) and it's only a few lines. I just used a package instead of copying those.

amarthadan:
> I'm still going through the terraform files, but the typescript part looks good. Quick question, are the deployed functions public or private? I remember when I deployed the functions by default they would be private and had to manually change it to public(although I don't think it matters)

They are private but we don't need them to be public. They are not invoked by anything external. The `startCoordinator` is run periodically by the scheduler and the rest is invoked from within.

PR #689, opened 2021-11-23
AN-364: Fix docker build command - solidity compiler download.

aquarat:
Moving the `wget` commands further down in the `Dockerfile` inserts the files in the appropriate layer for HardHat.
HardHat also requires the appropriate `list.json` file for checksum verification.

Surprisingly opaque problem, easy solution.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -30,11 +30,6 @@ RUN git clone ${repository} ${buildDir} -b ${branch}
 # Production dependencies
 FROM source${build} AS deps
 
-# Download solidity compiler as per: https://github.com/nomiclabs/hardhat/issues/1280#issuecomment-949822371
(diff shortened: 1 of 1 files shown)


Siegrift (approved):
Btw. I forgot to mention to @bbenligiray that we will have to think about this when upgrading solidity version in the future.

bbenligiray:
@aquarat `docker:build:artifacts` still seems to fail for me :thinking:

aquarat:
> @aquarat docker:build:artifacts still seems to fail for me

Wow, I'm surprised; I checked both positive and negative scenarios (swapping in bad checksums in `list.json`). I'll look at it again now. I did see some oddities around HardHat sometimes treating the platform as not belonging to linux-amd64 and defaulting to `wasm` instead (which triggers a download process for a wasm compiler) but I only saw it once over many tests.

PR #691, opened 2021-11-24
AN-364: Further changes to override HardHat automatic solidity compiler download

aquarat:
The compiler downloaded now is JS based, which means it is architecture and OS-agnostic.

I insert a function into HardHat's TASK subsystem so as to force both the compiler type and path.
The platform is no longer detected and used to inform compiler selection.
The reasons for this are:
- Sometimes HardHat switches from `linux-amd64` as a platform to `wasm`
- The `linux-amd64` solidity binary doesn't run on the node docker image downloaded due to a missing library dependency.

While I was at it I swapped hardhat.conf.js for .ts in airnode-protocol.

I'll ask @bbenligiray to test this as his build was still failing after the last _fix_.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -42,10 +42,10 @@ RUN yarn global add lerna && \
 FROM source${build} AS build
 
 # Download solidity compiler as per: https://github.com/nomiclabs/hardhat/issues/1280#issuecomment-949822371
-# The sequencing of this command in this file is important.
-RUN mkdir -p /root/.cache/hardhat-nodejs/compilers/linux-amd64 \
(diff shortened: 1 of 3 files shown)


aquarat:
In response to Burak's comments I'm working on a cleaner alternative.

bbenligiray:
@aquarat Let me know if I need to test. Also, prefer creating the branches in this repo instead of your own.

aquarat:
> @aquarat Let me know if I need to test. Also, prefer creating the branches in this repo instead of your own.

Sorry for the delay. It had crossed my mind @ branches in parent rather than personal, thanks for confirming.
I'm just trying something potentially cleaner as an alternative - I'll keep you updated.

aquarat:
Closing in favour of [a different approach](https://github.com/api3dao/airnode/pull/698).

PR #692, opened 2021-11-24
Exports things needed by Airkeeper and updates beacon reader condition

acenolaza:
@andreogle Airkeeper requires a few functions and types from airnode that are not currently being exported. I've exported only what's needed but we were wondering with Burak if we should be exporting everything by default.

@bbenligiray maybe I should have done this in a separate PR but I've also updated the RrpBeaconServer contract to support reading the beacon value with address(0) using the VoidSigner discussed [here](https://api3workspace.slack.com/archives/C027Y2FAV0S/p1637666522074300).

### DIFF
--- a/packages/airnode-node/src/evm/index.ts
+++ b/packages/airnode-node/src/evm/index.ts
@@ -1,9 +1,15 @@
-import * as encoding from './abi-encoding';
-import * as contracts from './contracts';
-
-export { contracts, encoding };
+export * as authorization from './authorization';
+export * as contracts from './contracts';
+export * as fulfillments from './fulfillments';
+export * as handlers from './handlers';
+export * as requests from './requests';
+export * as templates from './templates';
+export * as verification from './verification';
+export * as encoding from './abi-encoding';
+export * from './evm-provider';
 export * from './gas-prices';
 export * from './networks';
-export * from './evm-provider';
+export * from './transaction-counts';
 export * from './utils';
 export * from './wallet';
+export * from './workers';
--- a/packages/airnode-node/src/index.ts
+++ b/packages/airnode-node/src/index.ts
@@ -1,10 +1,9 @@
-export * as handlers from './handlers';
+export * as adapters from './adapters';
 export * as config from './config';
-export * as providerState from './providers/state';
-export * as logger from './logger';
-export * as promiseUtils from './utils/promise-utils';
 export * as evm from './evm';
-export * as adapterHttpParameters from './adapters/http/parameters';
+export * as handlers from './handlers';
+export * as logger from './logger';
+export * as providers from './providers';
 export * as utils from './utils';
-export * from './version';
 export * from './types';
+export * from './version';
--- a/packages/airnode-node/src/index.ts
+++ b/packages/airnode-node/src/index.ts
@@ -3,5 +3,8 @@ export * as config from './config';
 export * as providerState from './providers/state';
 export * as logger from './logger';
 export * as promiseUtils from './utils/promise-utils';
+export * as evm from './evm';
+export * as adapterHttpParameters from './adapters/http/parameters';
+export * as utils from './utils';
 export * from './version';
 export * from './types';
--- a/packages/airnode-operation/src/config/evm-dev-config.json
+++ b/packages/airnode-operation/src/config/evm-dev-config.json
@@ -17,7 +17,7 @@
             { "type": "bytes32", "name": "to", "value": "USD" },
             { "type": "bytes32", "name": "_type", "value": "int256" },
             { "type": "bytes32", "name": "_path", "value": "result" },
-            { "type": "bytes32", "name": "_times", "value": "100000" }
+            { "type": "bytes32", "name": "_times", "value": "1000000" }
           ]
         }
       }
@@ -64,7 +64,7 @@
         { "type": "bytes32", "name": "to", "value": "USD" },
         { "type": "bytes32", "name": "_type", "value": "int256" },
         { "type": "bytes32", "name": "_path", "value": "result" },
-        { "type": "bytes32", "name": "_times", "value": "100000" }
+        { "type": "bytes32", "name": "_times", "value": "1000000" }
       ]
     },
     {
--- a/packages/airnode-protocol/contracts/rrp/requesters/RrpBeaconServer.sol
+++ b/packages/airnode-protocol/contracts/rrp/requesters/RrpBeaconServer.sol
@@ -297,7 +297,7 @@ contract RrpBeaconServer is
         onlyIfTemplateExists(templateId)
         returns (bool)
     {
-        return userIsWhitelisted(templateId, reader);
+        return userIsWhitelisted(templateId, reader) || reader == address(0);
     }
 
     /// @notice Called to get the detailed whitelist status of the reader for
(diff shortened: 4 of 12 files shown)


packages/airnode-operation/src/config/evm-dev-config.json
```diff
@@ -17,7 +17,7 @@
             { "type": "bytes32", "name": "to", "value": "USD" },
             { "type": "bytes32", "name": "_type", "value": "int256" },
             { "type": "bytes32", "name": "_path", "value": "result" },
-            { "type": "bytes32", "name": "_times", "value": "100000" }
+            { "type": "bytes32", "name": "_times", "value": "1000000" }
```
acenolaza, line 20: Not sure why but this value was not matching the config.json.example value https://github.com/api3dao/airnode/blob/5798e606b070d188344c70b5edb5ad68ad1e16d4/packages/airnode-node/config/config.json.example#L118

packages/airnode-node/src/index.ts
```diff
@@ -3,5 +3,8 @@ export * as config from './config';
 export * as providerState from './providers/state';
 export * as logger from './logger';
 export * as promiseUtils from './utils/promise-utils';
+export * as evm from './evm';
+export * as adapterHttpParameters from './adapters/http/parameters';
```
andreogle, line 7: This file is a bit messy and I think we've just been dumping stuff here without much consideration. We're exporting things like `promiseUtils` and `utils` which will export duplicate functions.

I think this it rather should be something like:

```ts
// src/index.ts
export * as adapters from './adapters';

// src/adapters/index.ts
export * as http from './http';

// src/adapters/http/index.ts
export * as parameters from './parameters';
```

This way, you would access the parameters functions by doing:

```ts
import { adapters } from '@airnode-node';

adapters.http.parameters.somefunction();
```

I think this is a lot cleaner (especially with naming) and allows us to easily grow what we export. What do you think?

andreogle (approved):

packages/airnode-node/src/index.ts
acenolaza: Yes, that is a much cleaner approach. Thanks!

packages/airnode-node/src/index.ts
```diff
@@ -1,7 +1,9 @@
-export * as handlers from './handlers';
+export * as adapters from './adapters';
 export * as config from './config';
-export * as providerState from './providers/state';
+export * as evm from './evm';
+export * as handlers from './handlers';
 export * as logger from './logger';
-export * as promiseUtils from './utils/promise-utils';
```
andreogle, line 5: Changing these imports is going to break stuff. I'm pretty sure things like the deployer depends on these so you'll need to check that out

acenolaza: Build and tests are passing. I also tested the airnode-operation commands and they are working as well. I'll try running the deployer cli.

Siegrift: @andreogle is right, and the build is not passing. Maybe you looked at some outdated build?

packages/airnode-protocol/src/index.ts
```diff
@@ -66,11 +67,18 @@ export {
   RequesterAuthorizerWithAirnodeAddresses,
   AirnodeRrpFactory,
   AccessControlRegistryFactory,
+  RrpBeaconServerFactory,
```
Siegrift, line 70: You don't need to import a thing just to export it, you can export it directly.

```suggestion
export { RrpBeaconServer__factory as RrpBeaconServerFactory } from './contracts';
```

Siegrift (approved):
Exporting the factories is not a problem, actually it's a good thing to as you will receive a typed contract instance when you `connect`.

Regarding exporting everything: I don't think that is a good thing for few reasons
1) Those files are generated and have a strange name (`ContractName__factory`) so you still want to rename it to something nicer - so you can't just use `export *` at that point.
2) Most of the generated files are never going to be imported (e.g. they are just interfaces or contracts that are part of another contract through inheritance)

packages/airnode-node/src/evm/index.ts
```diff
@@ -1,9 +1,15 @@
...
+export * as handlers from './handlers';
+export * as requests from './requests';
+export * as templates from './templates';
+export * as verification from './verification';
+export * as encoding from './abi-encoding';
+export * from './evm-provider';
 export * from './gas-prices';
 export * from './networks';
-export * from './evm-provider';
+export * from './transaction-counts';
```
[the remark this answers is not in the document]
acenolaza: My best guess is that folders where using aliases as well as files with hyphen in their names (ie abi-encoding.ts)

packages/airnode-protocol/src/index.ts
acenolaza: Right. I just tried to apply the same convention already being use (lines 13-19)

packages/airnode-protocol/test/rrp/requesters/RrpBeaconServerWithManager.sol.js
```diff
@@ -746,7 +746,12 @@ describe('readBeacon', function () {
           { gasLimit: 500000 }
         );
       // Read the beacon again
-      const currentBeacon = await rrpBeaconServer.connect(roles.beaconReader).readBeacon(templateId);
+      let currentBeacon = await rrpBeaconServer.connect(roles.beaconReader).readBeacon(templateId);
+      expect(currentBeacon.value).to.be.equal(decodedData);
+      expect(currentBeacon.timestamp).to.be.equal(nextBlockTimestamp);
+      // Read beacon with address(0)
+      const voidSigner = new hre.ethers.VoidSigner(hre.ethers.constants.AddressZero, hre.ethers.provider);
+      currentBeacon = await rrpBeaconServer.connect(voidSigner).readBeacon(templateId);
```
bbenligiray, line 754: VoidSigner is not a whitelisted user so this is misplaced. I will push a fix.

packages/airnode-protocol/contracts/rrp/requesters/RrpBeaconServer.sol
```diff
@@ -297,7 +297,7 @@ contract RrpBeaconServer is
         onlyIfTemplateExists(templateId)
         returns (bool)
     {
-        return userIsWhitelisted(templateId, reader);
+        return userIsWhitelisted(templateId, reader) || reader == address(0);
```
bbenligiray, line 300: I considered doing this in `Whitelist.sol` inside `userIsWhitelisted()` but that feels like a bad idea

packages/airnode-protocol/test/rrp/requesters/RrpBeaconServerWithManager.sol.js
bbenligiray: Done in https://github.com/api3dao/airnode/pull/692/commits/7eda0fdade32fef33d8e295a5172d28dcb0ce35e

bbenligiray (approved):

packages/airnode-protocol/contracts/rrp/requesters/RrpBeaconServer.sol
acenolaza: I thought about doing the same but allowing `address(0)` to read the beacon seems like an edge case of RrpBeaconServer.sol and Whitelist.sol should take a more generic approach IMO.

PR #693, opened 2021-11-24
AN-364: Fix docker build by adding missing library

aquarat:
Further investigation reveals that the reason the libraries solc relies on are incompatible in the docker build process is because Alpine links binaries against "musl" whereas Go binaries link against glibc (like most Linux distributions).

The changes in this commit add those missing libraries. They don't verify the checksum of the added library - but neither does HardHat.


aquarat:
Closed in favour of [a swap out of the base image](https://github.com/api3dao/airnode/pull/698).

PR #694, opened 2021-11-25
AN-363: Optimise handler timeouts and memory using stress test results

aquarat:
* Adjusted timeouts for `initializeProvider` to 12 s from 20 s
* Adjusted timeouts for `processProviderRequests` to 20 s from 10 s
* The above values are in the ratios revealed through stress testing
* Memory allocations for multiple handlers increased to 768 MB
* Memory allocation for `startCoordinator` must be 768 MB - performance is impacted if set lower
* `callAPI` increased to 256 MB (I/O bound task)
* Timeouts in `constants.ts` updated to TF values - 500 ms
* Airnode achieves > 100 requests per cycle with this configuration


amarthadan (approved):

packages/airnode-node/src/constants.ts
```diff
@@ -31,10 +31,10 @@ export const DEFAULT_RETRY_DELAY_MS = 50;
 export const EVM_PROVIDER_TIMEOUT = 10_000;
 
 // The maximum amount of time the "initialize provider" worker is allowed before being timed out
-export const WORKER_PROVIDER_INITIALIZATION_TIMEOUT = 19_500;
+export const WORKER_PROVIDER_INITIALIZATION_TIMEOUT = 11_500;
```
andreogle, line 34: I wonder if we need these 500ms buffers. I originally added them to allow the worker process some time to return a response (even if it's null) instead of having the entire worker promise reject, but I'm not sure this is necessary.

andreogle (approved):

packages/airnode-node/src/constants.ts
aquarat: I'm struggling to find the exact logs now (to quote them) but early on in my testing, before I increased the Terraform timeouts, I encountered quite a few timeouts.

From memory, when the handlers fail due to a Terraform-related timeout, `startCoordinator` emits a JSON unmarshalling error message, along the lines of "bad character 'u' at position x", which I suspect is "undefined" underneath.

Still, the requests got through and the chain is the ultimate reference, so you're probably right, the timing guards are probably unnecessary, but will result in cryptic error messages if the timeouts get hit. I can test that as a scenario quick and get back to you if you want?

andreogle: Yeah sure - that sounds good 👍🏻

bbenligiray:
I think we should reduce `callApi` timeout to ~10s and use the freed up space for `initializeProvider` and `processProviderRequests`

packages/airnode-node/src/constants.ts
aquarat: I'll get round to this - I've just been getting the stress tester working with GCP (so far so good (!) just have to implement GCP log grabbing and processing).

aquarat:
@bbenligiray 
> I think we should reduce `callApi` timeout to ~10s and use the freed up space for `initializeProvider` and `processProviderRequests` 

Sounds good to me, I was eyeing that 30s timeout but I thought it was to accommodate a specific slow API.

Two more considerations:
1. `initializeProvider` cannot partially fail; it either processes all logs or none, whereas `processProviderRequests` can timeout while still having done work that reduces future workloads - so although the ratios would indicate the above change, it increases the risk of Airnode not fulfilling any requests in a cycle.
2. What if Airnode's handlers were allowed to consume up to 1.5 minutes per cycle? Naturally one iteration interval would step over another which would increases costs, but in terms of fulfilment, it would increase the throughput further and Airnode can't double-fulfil requests due to a nonce conflict (worst case resulting in a replacement transaction)?

packages/airnode-node/src/constants.ts
aquarat: @andreogle 
This is how `initializeProvider` times out without buffers/when the Terraform timeout gets hit:
```
3c3a2f91-f062-4ebc-b733-a80726e5eaf5	ERROR	Uncaught Exception 	{
    "errorType": "SyntaxError",
    "errorMessage": "Unexpected token u in JSON at position 0",
    "code": "SyntaxError",
    "message": "Unexpected token u in JSON at position 0",
    "time": "2021-11-26T12:14:31.926Z",
    "stack": [
        "SyntaxError: Unexpected token u in JSON at position 0",
        "    at JSON.parse (<anonymous>)",
        "    at constructor.<anonymous> (/var/task/handlers/aws/index.js:2:1972298)",
        "    at constructor.<anonymous> (/var/task/handlers/aws/index.js:2:711892)",
        "    at constructor.callListeners (/var/task/handlers/aws/index.js:2:729208)",
        "    at constructor.emit (/var/task/handlers/aws/index.js:2:728918)",
        "    at constructor.emitEvent (/var/task/handlers/aws/index.js:2:714767)",
        "    at constructor.e (/var/task/handlers/aws/index.js:2:710272)",
        "    at t.runTo (/var/task/handlers/aws/index.js:2:792563)",
        "    at /var/task/handlers/aws/index.js:2:792769",
        "    at constructor.<anonymous> (/var/task/handlers/aws/index.js:2:710542)"
    ]
}
[ERROR] [1637928871988] LAMBDA_RUNTIME Failed to post handler success response. Http response code: 400.
```
and with the timeout hitting the worker options value from `constants.ts`:
```
3ffa77b0-0531-4fb5-9d94-0f7a1cbfe695	INFO	[2021-11-26 12:33:16.498] ERROR Error: Operation timed out in 2000 ms.
    at Timeout._onTimeout (/var/task/handlers/aws/index.js:2:1970597)
    at listOnTimeout (internal/timers.js:557:17)
    at processTimers (internal/timers.js:500:7)  Coordinator-ID:fc1d5ca1cafb7d60
```
`processProviderRequests` behaves similarly - so it appears there are no problems associated with removing the timing buffers.

andreogle: Ok awesome. I'm happy then if you want to remove them here then. It will also be more intuitive for future readers

aquarat:
Closing in favour of an updated PR and one more test scenario (where Airnode steps on the toes of another run).

aquarat:
AN-363

Timeouts increased and unified to (no timing buffers):
- `startCoordinator`: 65 seconds
- `initializeProvider`: 17 seconds
- `callApi`: 10 seconds
- `processProviderRequests`: 32 seconds

Sum of parts (excl `startCoordinator`): 59 seconds
`startCoordinator` tolerance: 6 seconds (for overheads and heartbeat)

This is pending stress testing.

bbenligiray:
> initializeProvider cannot partially fail; it either processes all logs or none, whereas processProviderRequests can timeout while still having done work that reduces future workloads - so although the ratios would indicate the above change, it increases the risk of Airnode not fulfilling any requests in a cycle.

Yeah, I find it awkward how we ended up with a shorter `initializeProvider` despite reducing `callApi` that much

> What if Airnode's handlers were jointly allowed to consume up to 1.5 minutes per cycle?

As I've mentioned, I don't expect any downsides from doing this, yet it only improves the capacity marginally (I'd assume by ~50%) and it's not obvious to me that we'll need that extra 50%. We'll want to parallelize the fulfillments properly before using this in production, which will allow us to scale up to the account concurrency limit and at that point we will want to restrict the whole cycle to 60 seconds because we don't want the `callApi` invocations from two different cycles to coincide (because that would effectively halve the account concurrency limit). So if we go from 1 minute to 1.5 minutes, it will be a thing that we'll revert in v0.4. It may still be a good idea though. How does this affect the containerized version?

aquarat:
@bbenligiray 
> Yeah, I find it awkward how we ended up with a shorter `initializeProvider` despite reducing `callApi` that much

The ratio ideal duration ratio between `initializeProvider` and `processProviderRequests` is approximately 1:2. If we take 20 seconds away from `callApi` we give (20/3) to `initializeProvider`; about 6 seconds and conversely ~13 seconds to `processProviderRequests`.

> ...it only improves the capacity marginally (I'd assume by ~50%) and it's not obvious to me that we'll need that extra 50%.  ... So if we go from 1 minute to 1.5 minutes, it will be a thing that we'll revert in v0.4. It may still be a good idea though. How does this affect the containerized version?

I haven't tested this against the local(?) docker version, but I'll do so and get back to you.
My PoC for parallelised/multi-handler `processProviderRequests`, which I'll push to this repo later today, wasn't particularly complex and didn't seem to be too inefficient. It's an approach that could probably be introduced in a short space of time. I just replaced the current async call with an invoke call to `processProviderRequests` and included the target `sponsorAddress` in the worker arguments, then filtered for that address in `processProviderRequests`.

bbenligiray:
> The ratio ideal duration ratio between initializeProvider and processProviderRequests is approximately 1:2. If we take 20 seconds away from callApi we give (20/3) to initializeProvider; about 6 seconds and conversely ~13 seconds to processProviderRequests.

Yeah I know the reasoning

> I can improve upon this further and submit for review if you like?

Nah, the quick win is this PR, I think this should be dealt with properly (no bottlenecks whatsoever) in 0.4

PR #695, opened 2021-11-25
Update RrpBeaconServer to expect a data point and a timestamp

bbenligiray:
We want to release this with the multiple response feature in 0.3.

Coverage tests are failing because they take too long and the contract is time-dependent. I'll fix that but the contracts will remain the same and can be reviewed. (Edit: This is done now)

@acenolaza Note that airkeeper will work with this version of the beacon server and [this](https://github.com/api3dao/airkeeper/blob/2f47da330ebc58da6aa9da353c69062cdaf31426/src/index.ts#L131-L133) decoding line will have to change


packages/airnode-protocol/contracts/rrp/requesters/RrpBeaconServer.sol
```diff
@@ -242,24 +246,39 @@ contract RrpBeaconServer is
...
+            "Fulfillment older than beacon"
+        );
+        require(
+            decodedTimestamp + 1 hours > block.timestamp,
+            "Fulfillment stale"
+        );
+        require(
+            decodedTimestamp - 1 hours < block.timestamp,
+            "Fulfillment from future"
+        );
```
bbenligiray, line 272: The tolerance values are not strict (+- 1 hour) assuming that this is not a significant time period in the context of asset prices, and stricter checks are more vulnerable (https://consensys.github.io/smart-contract-best-practices/recommendations/#timestamp-dependence) with no positive tradeoff. The general rule for Ethereum is that you can trust the chain timestamp up to 15 seconds, but note that this contract will also be used in other chains whose node implementation may not be as strict/where consensus attacks are easier to pull off.

Ashar2shahid: I'm not sure how to feel about having the +- 1 being hardcoded but the reasoning is there so I guess that suffices. Is there any practical case in which we get  decodedTimestamp - 1 hours < block.timestamp ?

packages/airnode-protocol/test/rrp/requesters/RrpBeaconServerWithManager.sol.js
```diff
@@ -687,189 +687,346 @@ describe('requestBeaconUpdate', function () {
...
-                signature,
+                firstData,
+                firstSignature,
                 { gasLimit: 500000 }
-              )
-          )
-            .to.emit(rrpBeaconServer, 'UpdatedBeacon')
-            .withArgs(templateId, requestId, decodedData, nextBlockTimestamp);
+              );
+            expect(staticCallResult.callSuccess).to.equal(false);
```
Ashar2shahid, line 1027: doing a `callStatic` will not result in a state update so we expect the .call to fail . Was this the expected result of this test?

Ashar2shahid (approved):
Just some comments

packages/airnode-protocol/contracts/rrp/requesters/RrpBeaconServer.sol
bbenligiray: That shouldn't happen normally, but the consequences are permanent if it happens (https://github.com/api3dao/airnode/pull/695#discussion_r757021322) so it needs to be accounted for

packages/airnode-protocol/test/rrp/requesters/RrpBeaconServerWithManager.sol.js
bbenligiray: `.callStatic` returning false also indicates that a regular transaction attempt would have failed. Maybe we should test both (Airnode testing a fulfillment with a static call will find out that the fulfillment will fail and will not attempt the fulfillment + an already created fulfillment transaction will revert)

Ashar2shahid: yeah, makes sense to test both

bbenligiray: Actually I remember why I'm testing this way. AirnodeRrp `fulfill()` doesn't revert if the target fulfillment function reverts, it only returns a false `callSuccess`, so this is the only/correct way of testing if the RrpBeaconServer `fulfill()` reverts.

bbenligiray: I added non-static calls just in case https://github.com/api3dao/airnode/pull/695/commits/23f61654bd81ede720e4770ac731c98789fe3c19

PR #697, opened 2021-11-26
Replace relay metadata for relay security schemes

Siegrift:
https://api3dao.atlassian.net/browse/AN-347


andreogle (approved):
Looks good to me 👍🏻 

Just to check, are we dropping the rest of the metadata parameters for now and only allowing `chainId`, `chainType` and `requesterAddress`? What about the rest of them?

Siegrift:
> Just to check, are we dropping the rest of the metadata parameters for now and only allowing chainId, chainType and requesterAddress? What about the rest of them?

They will be added later - it will be pretty straightforward.

acenolaza:
Does this change also mean that auth can no longer depend on multiple values? I mean, previously the API could chain multiple values from the metadata but I think now the API will have to rely on single value, am I wrong? 

For example, will the API might want to accept requests only where `chainId=1 AND chainType="evm"` or `requesterAddress="0x123456" OR chainType="evm"`, etc

Siegrift:
When you add the relay security schemes Airnode will just forward (relay)
the value(s) to the API. So you create multiple relay security schemes and
they will all be forwarded.
You can them implement whatever logic you want on the API side.

PR #698, opened 2021-11-26
AN-364: Fix docker build by swapping in Debian for artifacts image base

aquarat:
Debian uses `glibc` vs `musl` for Alpine. Despite the advertising, `solc` relies on the `glibc` binary library.
I've left in the `solc` download as it will help with problematic connections and proxies - but it is no longer necessary.
[Fix relates to this conversation](https://api3workspace.slack.com/archives/C02AYRX8D89/p1637914176266000)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -43,6 +43,7 @@ RUN yarn global add lerna && \
 # Build
 FROM source${build} AS build
 
+# Download solidity compiler as per: https://github.com/nomiclabs/hardhat/issues/1280#issuecomment-949822371
 RUN mkdir -p /root/.cache/hardhat-nodejs/compilers/linux-amd64 && \
   wget -O /root/.cache/hardhat-nodejs/compilers/linux-amd64/solc-linux-amd64-v0.8.9+commit.e5eed63a https://binaries.soliditylang.org/linux-amd64/solc-linux-amd64-v0.8.9+commit.e5eed63a && \
   wget -O /root/.cache/hardhat-nodejs/compilers/linux-amd64/list.json https://solc-bin.ethereum.org/linux-amd64/list.json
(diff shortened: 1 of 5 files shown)


docker/Dockerfile
```diff
@@ -41,11 +43,9 @@ RUN yarn global add lerna && \
 # Build
 FROM source${build} AS build
 
-# Download solidity compiler as per: https://github.com/nomiclabs/hardhat/issues/1280#issuecomment-949822371
```
Siegrift, line 44: I'd preserve this comment as we will have to update this in the future.

aquarat: Done :)

bbenligiray:
Confirming that the artifacts image builds on my machine. However, the admin, deployer and client dockerfiles are still using alpine. I built the alpine admin image using the debian artifacts image and it worked but I assume we should use debian everywhere.

aquarat:
> Confirming that the artifacts image builds on my machine. However, the admin, deployer and client dockerfiles are still using alpine. I built the alpine admin image using the debian artifacts image and it worked but I assume we should use debian everywhere.

I'm happy to swap them all. Debian will increase the size of the images, but it will possibly save fault tracing time in future. I'm not sensitive to larger images. I will submit an update shortly.

aquarat:
I have swapped out Alpine for Debian in all Dockerfiles, as well as modified utilities/commands where appropriate.
I have tested these builds (admin, deployer, local node). :)

bbenligiray:
What is the exact size change? How long your first deployment will take is an extremely important metric.

aquarat:
> What is the exact size change? How long your first deployment will take is an extremely important metric.

Apologies, missed this - already discussed on Slack, but: the Alpine image is 112 MB in size. The Debian image is 898 MB in size.

PR #700, opened 2021-11-27
Limit encoded response size

Siegrift:
https://api3dao.atlassian.net/browse/AN-367

### DIFF
--- a/packages/airnode-adapter/src/response-processing/extraction.test.ts
+++ b/packages/airnode-adapter/src/response-processing/extraction.test.ts
@@ -7,7 +7,9 @@ import {
   escapeAwareSplit,
   unescape,
 } from './extraction';
+import { MAX_ENCODED_RESPONSE_SIZE } from '../constants';
 import { ReservedParameters } from '../types';
+import { exceedsMaximumEncodedResponseSize } from '.';
 
 describe('getRawValue', () => {
   it('returns the data as is if no path is provided', () => {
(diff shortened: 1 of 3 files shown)


packages/airnode-adapter/src/response-processing/extraction.ts
```diff
@@ -139,5 +143,11 @@ export function extractAndEncodeResponse(data: unknown, parameters: ReservedPara
   }
 
   const extractedValue = extractSingleResponse(data, parameters);
-  return { rawValue: data, encodedValue: encodeValue(extractedValue, parameters._type), values: [extractedValue] };
+  const encodedValue = encodeValue(extractedValue, parameters._type);
+
+  if (encodedValue.length > MAX_ENCODED_RESPONSE_SIZE) {
+    throw new Error(`Encoded value exceeds the maximum allowed size (${MAX_ENCODED_RESPONSE_SIZE} bytes)`);
+  }
```
bbenligiray, line 150: It looks like you're checking the stringified length here. I recommend using `ethers.utils.arrayify` and checking the length of its output to get the length in bytes

packages/airnode-adapter/src/response-processing/extraction.test.ts
```diff
@@ -247,3 +247,29 @@ describe('extract and encode multiple values', () => {
     });
   });
 });
+
+describe('respects maximum encoded size', () => {
+  it('throws when the encoded value is too large', () => {
+    const str = 'x'.repeat(417);
```
bbenligiray, line 253: The encoded type can be `bytes` so that this number is 1025 and not this magic number

bbenligiray: Ideally 1024 should be imported from `constants.ts` and this is derived from that

packages/airnode-adapter/src/response-processing/extraction.ts
Siegrift: Yeah, I was wondering whether comparing the bytesString length is actually correct.

packages/airnode-adapter/src/response-processing/extraction.test.ts
```diff
@@ -1,4 +1,5 @@
 import { ethers } from 'ethers';
+import { MAX_ENCODED_RESPONSE_SIZE } from 'src/constants';
```
andreogle, line 2:
```suggestion
import { MAX_ENCODED_RESPONSE_SIZE } from '../constants';
```

andreogle (approved):

packages/airnode-adapter/src/response-processing/extraction.test.ts
Siegrift: My VS code does this a lot - does it happen to you as well?

andreogle: *ahem* I use Neovim

bbenligiray: based

bbenligiray (approved):
:rocket:

PR #702, opened 2021-11-30
AN-364: Download both variants of solc compiler during build

aquarat:
As per [Slack](https://api3workspace.slack.com/archives/C02AYRX8D89/p1637914176266000), HardHat first attempts to download the `linux-amd64` compiler. This always fails to execute, which causes HardHat to always fall back to the solc JS/WASM variant, which is platform and CPU agnostic and does execute. This is also the case in v0.2.2.

The main bug, apart from the Alpine incompatibility, is that the process of downloading solc js *sometimes* results in a corrupt file. I always start from a clean state (`docker system prune -f --all`) when running these builds, but sometimes builds fail continuously. I think a corrupt copy of the compiler is being cached by Amazon CloudFront.

The fault is intermittent: [this commit for the 10th of November](https://github.com/api3dao/airnode/commit/0980a4eb56b0f9ac8c4dbe19c51b2b1428688435), for instance, sometimes succeeds and sometimes fails for me.
`v0.2.2` so far has not failed once, I've built it on a loop more than 30x.

The contents of this PR download both compilers and manifests, expecting HardHat to always fall back to the platform agnostic compiler. This has been built > 20x now and hasn't failed.

This will also fix a scenario where HardHat [can't download the compiler due to a proxy server](https://github.com/nomiclabs/hardhat/issues/1280).

@Siegrift does not believe the e2e tests issue is related.


Siegrift:
> @Siegrift does not believe the e2e tests issue is related.

Siegrift knows :)

bbenligiray (approved):
Works for me :ok_hand:

PR #704, opened 2021-11-30
Remove too restrictive rules from eslint-plugin-functional

Siegrift:
https://api3dao.atlassian.net/browse/AN-190

When I was reading https://github.com/jonaskello/eslint-plugin-functional I came to a conclusion that there is only one worthy rule.
All others have edge cases or just create too much boilerplate. Is there some rule you'd like to preserve?


andreogle (commented):
Looks good.

The only rule that I really like is `functional/immutable-data`. I think all data should be required to be immutable by default with the option to opt out using the `// eslint-disable` comment or prefix it with `mutable`. I think the latter option is pretty reasonable

If we aren't going to use this rule, we might as well remove eslint-plugin-functional. What do you think?

Siegrift:
>  I think all data should be required to be immutable by default with the option to opt out using the // eslint-disable comment or prefix it with mutable.

The problem with this is that mutation in tests is OK imo and I dislike both workarounds. 

> If we aren't going to use this rule, we might as well remove eslint-plugin-functional. What do you think?

Yeah, we can do that as well, but that single rule which I've whitelisted I actually like since they also handled edge cases (at least they say) so it should not produce false positives.

andreogle:
I agree that mutation in tests is definitely fine. Maybe immutability should be disabled for `*.test.ts` files with [ESLint overrides](https://eslint.org/docs/user-guide/configuring/configuration-files#how-do-overrides-work)?

What are your issues with the immutability workarounds otherwise?

Siegrift:
> What are your issues with the immutability workarounds otherwise?

I find them too verbose and ugly.

> I agree that mutation in tests is definitely fine. Maybe immutability should be disabled for *.test.ts files with ESLint overrides?

This is a good point and a good tradeof - I'll try to set it up and will ping you when it's ready.

andreogle (approved):
🤝

PR #705, opened 2021-11-30
Dynamic config.json generation for airnode-examples package

Siegrift:
https://api3dao.atlassian.net/browse/AN-377

**Review commit by commit** - I am trying to be thorough as this will be quite a big PR.


packages/airnode-examples/integrations/authenticated-coinmarketcap/create-config.ts
```diff
@@ -0,0 +1,156 @@
+import { join } from 'path';
+import { Config } from '@api3/airnode-node';
+import { createCloudProviderConfiguration, createNodeVersion, generateConfigFile } from '../config-utils';
+
+const config: Config = {
```
Siegrift: These config objects were based on the original `config.json` files

packages/airnode-examples/integrations/authenticated-coinmarketcap/config.json
```diff
@@ -15,7 +15,9 @@
     }
   ],
   "nodeSettings": {
-    "cloudProvider": "${CLOUD_PROVIDER_TYPE}",
+    "cloudProvider": {
```
Siegrift, line 18: These files were generated by the scripts in this commit.

packages/airnode-examples/README.md
```diff
@@ -98,13 +98,26 @@ After you know the secrets, run the following script to specify them:
 yarn create-aws-secrets
 ```
 
-The command will generate an `aws.env` file with the entered secrets. This file is used only by the deployer and will
-never leave your machine.
+### 6. (Only if deploying to GCP) Create GCP credentials
 
-### 6. Create Airnode secrets
+TODO: Link to docs
```
Siegrift, line 103: Waits for https://github.com/api3dao/api3-docs/pull/444/

packages/airnode-examples/integrations/config-utils.ts
```diff
@@ -1,9 +1,15 @@
 import { writeFileSync } from 'fs';
 import { join } from 'path';
 import { Config, LocalOrCloudProvider } from '@api3/airnode-node';
-import { cliPrint, readIntegrationInfo } from '../src';
+import { cliPrint, getDeployedContract, readChainId, readIntegrationInfo } from '../src';
+
+export const createCloudProviderConfiguration = (generateExampleFile: boolean): LocalOrCloudProvider => {
```
Siegrift, line 6: This was previously a bug - the function always looked into integration-info.json for values.

packages/airnode-examples/scripts/deploy-airnode.ts
```diff
@@ -14,10 +14,13 @@ const main = async () => {
     `docker run -it --rm`,
     `--env-file ${secretsFilePath}`,
```
amarthadan, line 15: You don't need to pass AWS-specific env variables for GCP. It doesn't do any harm but it's going to be inconsistent with the docs which may be confusing.

packages/airnode-examples/README.md
```diff
@@ -130,15 +143,15 @@ by running:
 yarn rebuild-artifacts-container
 ```
 
-### 8. (Only if deploying to AWS) Build deployer container
+### 10. (Only if deploying to AWS) Build deployer container
```
amarthadan, line 146: Or GCP :smile:

packages/airnode-examples/integrations/config-utils.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  const airnodeRrp = await getDeployedContract('@api3/airnode-protocol/contracts/rrp/AirnodeRrp.sol');
+  return airnodeRrp.address;
+};
+
+export const getChainId = async (generateExampleFile: boolean) =>
+  (generateExampleFile ? 31337 : await readChainId()).toString();
+
+export const createNodeVersion = () => {
+  return '0.2.2';
```
amarthadan, line 47: Will this be read from the `package.json` or what's the plan here? Because as soon as 0.3 is out this will break, won't it?

amarthadan (approved):
I'm going to test it now but the code looks good to me.

amarthadan:
I ran multiple examples, one for each cloud provider and local with HardHat. It works great, I found only a few little issues:

* When trying the `weather-multi-value` example I can see that an object was (almost) printed into `secrets.env` after `yarn create-airnode-secrets` run.
Screenshot_20211202_131316
* The example `curl` command in `weather-multi-value` example for testing the API key doesn't work because the timestamp there is older than 5 days. I would suggest (if possible) using an endpoint that doesn't require a timestamp.
* Also, I would suggest adding a warning the it takes a while for API key to become active (still `weather-multi-value` example)
* When I tried to remove an Airnode instance from GCP I got an error saying that that's supported only for AWS
  image
  Also the Docker command for remove is missing GCP credentials.

packages/airnode-examples/integrations/config-utils.ts
Siegrift: It will need to be manually updated for now, before the release. But it's only this one number that needs to be modified.

PR #717, opened 2021-12-01
AN-376: Add cloud provider specific checks to validator template

drgy:


packages/airnode-validator/test/validatorTests/minimal.json
```diff
@@ -23,7 +23,8 @@
     "nodeVersion": "0.3.0",
     "cloudProvider": {
       "type": "aws",
-      "region": "us-east-1"
+      "region": "us-east-1",
+      "name": "aws-us"
```
Siegrift, line 27: aws cloud provider doesn't have a name https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/types.ts#L376

Siegrift (changes requested):

packages/airnode-validator/test/validatorTests/minimal.json
drgy: Has any of the providers `name`?

Siegrift (approved):
LGTM

PR #720, opened 2021-12-02
Parameterized beacons

bbenligiray:
You no longer use `templateId` to read from the beacon server, but a `beaconId` instead, which is the hash of `templateId` and `parameters.

This seems to be compatible with the PSP beacon server (https://github.com/api3dao/airnode/blob/psp-draft-v4/packages/airnode-protocol/contracts/psp/BeaconServer.sol)

The only downside is that the user can't fetch the `templateId`-`parameters` pair by using `beaconId`. I think it's not a good idea to force these to be stored on-chain because `parameters` can potentially be large.

@acenolaza This changes Airkeeper in that job definitions will have to include a `parameters` field in addition to `templateId`. I think it would be nice to have that in decoded form and encode it using `@api3/airnode-abi` on the fly.


packages/airnode-protocol/contracts/rrp/requesters/RrpBeaconServer.sol
```diff
@@ -36,16 +37,8 @@ contract RrpBeaconServer is
...
-    mapping(bytes32 => Beacon) private templateIdToBeacon;
-    mapping(bytes32 => bytes32) private requestIdToTemplateId;
-
-    /// @dev Reverts if the template with the ID is not created
-    /// @param templateId Template ID
-    modifier onlyIfTemplateExists(bytes32 templateId) {
-        (address airnode, , ) = airnodeRrp.templates(templateId);
-        require(airnode != address(0), "Template does not exist");
-        _;
-    }
```
bbenligiray, line 48: We can't check if a beacon ID is valid anymore

packages/airnode-protocol/contracts/rrp/requesters/RrpBeaconServer.sol
```diff
@@ -194,40 +187,46 @@ contract RrpBeaconServer is
...
         emit RequestedBeaconUpdate(
-            templateId,
+            beaconId,
             sponsor,
             msg.sender,
             requestId,
-            sponsorWallet
+            templateId,
+            sponsorWallet,
+            parameters
```
bbenligiray, line 229: I'm not sure if `templateId` and `parameters` are required here because they will also be logged by AirnodeRrp. I still went by my rule of emitting all arguments.

packages/airnode-protocol/contracts/rrp/requesters/RrpBeaconServer.sol
Ashar2shahid: don't see any major drawback to this so seems fine

Ashar2shahid (approved):
Not really anything to add, looks good

acenolaza:
> @acenolaza This changes Airkeeper in that job definitions will have to include a `parameters` field in addition to `templateId`. I think it would be nice to have that in decoded form and encode it using `@api3/airnode-abi` on the fly.

What about calling `rrpBeaconServer.deriveBeaconId()` instead of using the library?

acenolaza (approved):
Created a branch in airkeeper repo where I applied the changes needed to support parameterized beacons: https://github.com/api3dao/airkeeper/tree/parameterized-beacons

bbenligiray:
Thanks, @acenolaza I skimmed through your branch and it looks correct. I left some comments https://github.com/api3dao/airkeeper/commit/bc6c55253be16a3d32368eb40079ae1ebf1711b2

PR #722, opened 2021-12-03
AN-418: Apply timeouts for GCP

aquarat:
- Set memory_size for `startCoordinator` and `callApi` to 256 MB
- Set memory_size for `initializeProvider` and `processProviderRequests` to 1024 MB to enhance performance (GCP does not have a 768 MB option)
- As per @andreogle I've increased the worker option-based timeouts so that the functions can run for as long as the cloud provider will let them. This will affect all cloud providers.


Siegrift (commented):
Do not forget to add changeset https://api3dao.atlassian.net/wiki/spaces/AIRNODE/pages/85229571/Changesets+documentation?parentProduct=JSW&NO_SSR=1#Example

aquarat:
I'll add the changeset (thanks @Siegrift) and I'll also double-check this configuration by running a stress test.

aquarat:
Added changeset :)

I ran stress tests against an Airnode with these changes.
The stress tests indicate that an Airnode instance with these timeouts and memory allocations has a capacity of 119 fulfilments per cycle under ideal conditions. For tracking these were tests keyed as `f690` and `661f` - I'll post these graphs to Jira.

.changeset/wise-cheetahs-play.md
```diff
@@ -0,0 +1,6 @@
+---
+'@api3/airnode-deployer': patch
+'@api3/airnode-node': patch
+---
+
+AN-418: Apply timeouts for GCP based on stress tester results
```
amarthadan, line 6: Can you remove the `AN-418` part? This will be a user-facing changelog and our internal Jira ticket number has no meaning for them.

amarthadan (approved):
LGTM :+1:

.changeset/wise-cheetahs-play.md
aquarat: Sure thing :)

packages/airnode-deployer/terraform/airnode/gcp/main.tf
```diff
@@ -40,7 +43,8 @@ module "startCoordinator" {
   name               = "${local.name_prefix}-startCoordinator"
   entry_point        = "startCoordinator"
   source_dir         = var.handler_dir
-  timeout            = 60
+  memory_size        = 256
+  timeout            = 65
```
andreogle, line 47: 65 seconds?

packages/airnode-node/src/constants.test.ts
```diff
@@ -8,17 +8,17 @@ describe('constants', () => {
...
     expect(constants.BLOCK_COUNT_IGNORE_LIMIT).toEqual(20);
     expect(constants.BLOCK_MIN_CONFIRMATIONS).toEqual(0);
     expect(constants.CONVENIENCE_BATCH_SIZE).toEqual(10);
     expect(constants.DEFAULT_RETRY_DELAY_MS).toEqual(50);
     expect(constants.DEFAULT_RETRY_TIMEOUT_MS).toEqual(5_000);
-    expect(constants.EVM_PROVIDER_TIMEOUT).toEqual(22_000);
-    expect(constants.WORKER_CALL_API_TIMEOUT).toEqual(10_000);
-    expect(constants.WORKER_PROVIDER_INITIALIZATION_TIMEOUT).toEqual(17_000);
-    expect(constants.WORKER_PROVIDER_PROCESS_REQUESTS_TIMEOUT).toEqual(32_000);
+    expect(constants.EVM_PROVIDER_TIMEOUT).toEqual(60_000);
```
andreogle, line 19: I'm not sure about changing all of these values to 60 seconds... I'm worried that this is going to result in weird behaviour for non-serverless deployments.

I think there may have been some miscommunication here. I was only referring to removing the 500ms buffers

andreogle: And for `EVM_PROVIDER_TIMEOUT` specifically, was this changed to 22 seconds because of the `getLogs` call being quite heavy? 22 seconds seems _really_ long for this

aquarat: > I'm worried that this is going to result in weird behaviour for non-serverless deployments

Sorry about, I hadn't considered a local deployment(!), I'll revert those timeouts to the variants without buffers.

> And for EVM_PROVIDER_TIMEOUT specifically, was this changed to 22 seconds because of the getLogs call being quite heavy?

`getLogs` itself isn't heavy, but processing of the results of `getLogs` in `initializeProvider` is. I'll revert this change too.

packages/airnode-deployer/terraform/airnode/gcp/main.tf
aquarat: The sum of the timeouts of the non-`startCoordinator` handlers for this PR are 59s, so my idea was to give `startCoordinator` a few seconds extra to collect the result(s) from `processProviderRequests`, log execution time and also report a heartbeat.

The timeout could be more like 61 or 62 seconds, but there didn't seem any harm in giving it extra headroom.

If you disagree I can reduce the timeouts on the other handlers slightly to allow the heartbeat to be reported.

andreogle: Yeah I think that's _probably_ fine. This will also be slightly different for non-serverless deployments then, since they will still only be allowed 60 seconds. Maybe that's an issue, maybe not - I'm not sure

aquarat: > since they will still only be allowed 60 seconds.
When you say "non-serverless", you mean `yarn run-airnode-locally`?
If so, that's just a cron job, so it won't be restricted, but I'm happy to test.

aquarat: I tested the local variant and it does introduce a problem (Airnode runs every two minutes instead of every minute).
I'll produce a fix for this shortly (by forking Airnode into a child process rather than blocking Cron, which is better behaviour imo anyway).

aquarat: Done - cron now doesn't wait for Airnode to finish executing.

PR #724, opened 2021-12-04
Add sponsor wallet request limit

Siegrift:
https://api3dao.atlassian.net/browse/AN-396

### DIFF
--- a/packages/airnode-node/src/evm/handlers/fetch-pending-requests.ts
+++ b/packages/airnode-node/src/evm/handlers/fetch-pending-requests.ts
@@ -45,10 +45,9 @@ export async function fetchPendingRequests(state: ProviderState<EVMProviderState
     withdrawals: withdrawalRequests,
   };
 
-  // Block any requests that cannot be processed
-  // 1. API calls related to a wallet with a pending withdrawal cannot be processed
-  const [blockLogs, updatedRequests] = blocking.blockRequestsWithWithdrawals(groupedRequests);
-  logger.logPending(blockLogs, baseLogOptions);
+  // Block (filter out) any requests that cannot be processed
+  const [blockRequestsLogs, allowedRequests] = blocking.blockRequests(groupedRequests);
+  logger.logPending(blockRequestsLogs, baseLogOptions);
 
-  return updatedRequests;
+  return allowedRequests;
 }
--- a/packages/airnode-node/src/evm/handlers/fetch-pending-requests.ts
+++ b/packages/airnode-node/src/evm/handlers/fetch-pending-requests.ts
@@ -46,6 +46,7 @@ export async function fetchPendingRequests(state: ProviderState<EVMProviderState
   };
 
   // Block (filter out) any requests that cannot be processed
+  // TODO: Better naming
   const [blockRequestsLogs, allowedRequests] = blocking.blockRequests(groupedRequests);
   logger.logPending(blockRequestsLogs, baseLogOptions);
 
--- a/packages/airnode-node/src/evm/requests/blocking.test.ts
+++ b/packages/airnode-node/src/evm/requests/blocking.test.ts
@@ -1,10 +1,15 @@
+import { range } from 'lodash';
 import * as blocking from './blocking';
 import * as fixtures from '../../../test/fixtures';
-import { GroupedRequests, RequestErrorMessage, RequestStatus } from '../../types';
+import { GroupedRequests, RequestStatus } from '../../types';
+import { MAXIMUM_SPONSOR_WALLET_REQUESTS } from '../../constants';
+
+const buildApiCallsWithSponsor = (count: number, sponsorAddress: string) =>
+  range(count).map((i) => fixtures.requests.buildApiCall({ sponsorAddress, id: `id-${sponsorAddress}-${i}` }));
 
 describe('blockRequestsWithWithdrawals', () => {
   it('blocks API calls with pending withdrawals from the same sponsor', () => {
-    const apiCall = fixtures.requests.buildApiCall({ sponsorAddress: '0x69e2B095fbAc6C3f9E528Ef21882b86BF1595181' });
+    const apiCall = buildApiCallsWithSponsor(1, '0x69e2B095fbAc6C3f9E528Ef21882b86BF1595181')[0];
     const withdrawal = fixtures.requests.buildWithdrawal({
       sponsorAddress: '0x69e2B095fbAc6C3f9E528Ef21882b86BF1595181',
     });
@@ -12,19 +17,20 @@ describe('blockRequestsWithWithdrawals', () => {
       apiCalls: [apiCall],
       withdrawals: [withdrawal],
     };
-    const [logs, res] = blocking.blockRequestsWithWithdrawals(requests);
+    const [logs, res] = blocking.blockRequestsWithWithdrawals([[], requests]);
     expect(logs).toEqual([
-      { level: 'WARN', message: `Ignoring Request ID:${apiCall.id} as it has a pending Withdrawl ID:${withdrawal.id}` },
+      {
+        level: 'WARN',
+        message: `Ignoring Request ID:${apiCall.id} as it has a pending Withdrawal ID:${withdrawal.id}`,
+      },
     ]);
-    expect(res.apiCalls.length).toEqual(1);
-    expect(res.apiCalls[0].status).toEqual(RequestStatus.Ignored);
-    expect(res.apiCalls[0].errorMessage).toEqual(`${RequestErrorMessage.PendingWithdrawal}: ${withdrawal.id}`);
+    expect(res.apiCalls.length).toEqual(0);
     expect(res.withdrawals.length).toEqual(1);
     expect(res.withdrawals[0].status).toEqual(RequestStatus.Pending);
   });
 
   it('does nothing if API call and withdrawal wallet indices do not match', () => {
-    const apiCall = fixtures.requests.buildApiCall({ sponsorAddress: '0x69e2B095fbAc6C3f9E528Ef21882b86BF1595181' });
+    const apiCall = buildApiCallsWithSponsor(1, '0x69e2B095fbAc6C3f9E528Ef21882b86BF1595181')[0];
     const withdrawal = fixtures.requests.buildWithdrawal({
       sponsorAddress: '0x99bd3a5A045066F1CEf37A0A952DFa87Af9D898E',
     });
@@ -32,7 +38,7 @@ describe('blockRequestsWithWithdrawals', () => {
       apiCalls: [apiCall],
       withdrawals: [withdrawal],
     };
-    const [logs, res] = blocking.blockRequestsWithWithdrawals(requests);
+    const [logs, res] = blocking.blockRequestsWithWithdrawals([[], requests]);
     expect(logs).toEqual([]);
     expect(res.apiCalls.length).toEqual(1);
     expect(res.apiCalls[0].id).toEqual(apiCall.id);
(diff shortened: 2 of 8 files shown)


packages/airnode-node/src/evm/handlers/fetch-pending-requests.ts
```diff
@@ -45,10 +45,9 @@ export async function fetchPendingRequests(state: ProviderState<EVMProviderState
     withdrawals: withdrawalRequests,
   };
 
-  // Block any requests that cannot be processed
-  // 1. API calls related to a wallet with a pending withdrawal cannot be processed
-  const [blockLogs, updatedRequests] = blocking.blockRequestsWithWithdrawals(groupedRequests);
-  logger.logPending(blockLogs, baseLogOptions);
+  // Block (filter out) any requests that cannot be processed
+  const [blockRequestsLogs, allowedRequests] = blocking.blockRequests(groupedRequests);
```
Siegrift, line 49: The `allowedRequests` suggest that all of these requests are "pending", but this is not necessarily true, because some requests may be flagged as Ignored in `verifyApiCallIds` (line 37)

packages/airnode-node/src/evm/requests/blocking.test.ts
```diff
@@ -1,20 +1,29 @@
+import { range } from 'lodash';
```
andreogle, line 1: 😉 
```suggestion
import range from 'lodash/range';
```

packages/airnode-node/src/evm/requests/blocking.ts
```diff
@@ -46,6 +45,41 @@ export function blockRequestsWithWithdrawals(requests: GroupedRequests): LogsDat
...
+      allowedApiCalls.push(blockedCall);
+    }
+  });
+
+  return [[...prevLogs, ...logs], { ...requests, apiCalls: allowedApiCalls }];
 }
+
+// TODO: Merge with verification/api-call-verification.ts
+export const blockRequests = (requests: GroupedRequests): LogsData<GroupedRequests> => {
+  const filterRequests = flow(blockRequestsWithWithdrawals, applySponsorRequestLimit);
```
andreogle, line 83: Pretty neat - I didn't know about `flow`. Looks like it's similar to the proposed pipeline operator. I would love to have this available to avoid things like `state1`, `state2` etc. Hopefully one day...

packages/airnode-node/src/evm/requests/blocking.ts
```diff
@@ -46,6 +45,41 @@ export function blockRequestsWithWithdrawals(requests: GroupedRequests): LogsDat
...
-  return [logs, updatedRequests];
+  return [[...prevLogs, ...logs], { ...requests, apiCalls }];
+}
+
+export function applySponsorRequestLimit([prevLogs, requests]: LogsData<GroupedRequests>): LogsData<GroupedRequests> {
+  const requestCountPerSponsor = new Map<string, number>();
+  const allowedApiCalls: Request<ApiCall>[] = [];
+  const logs: PendingLog[] = [];
+
+  requests.apiCalls.forEach((apiCall) => {
```
andreogle, line 56: I see you chose not to `reduce` here 😄

packages/airnode-node/src/evm/requests/blocking.ts
```diff
@@ -46,6 +45,41 @@ export function blockRequestsWithWithdrawals(requests: GroupedRequests): LogsDat
...
+    const sponsorRequests = requestCountPerSponsor.get(apiCall.sponsorAddress)!;
+
+    if (sponsorRequests < MAXIMUM_SPONSOR_WALLET_REQUESTS) {
+      requestCountPerSponsor.set(apiCall.sponsorAddress, sponsorRequests + 1);
+      allowedApiCalls.push(apiCall);
+    } else {
+      logs.push(logger.pend('WARN', `Ignoring Request ID:${apiCall.id} as it exceeded sponsor wallet request limit.`));
+      const blockedCall: Request<ApiCall> = {
+        ...apiCall,
+        status: RequestStatus.Ignored,
```
andreogle, line 71: See comment below about ignored vs blocked. I think this needs to be changed
```suggestion
        status: RequestStatus.Blocked,
```

packages/airnode-node/src/evm/requests/blocking.ts
```diff
@@ -46,6 +45,41 @@ export function blockRequestsWithWithdrawals(requests: GroupedRequests): LogsDat
...
+    if (!requestCountPerSponsor.has(apiCall.sponsorAddress)) {
+      requestCountPerSponsor.set(apiCall.sponsorAddress, 0);
+    }
+    const sponsorRequests = requestCountPerSponsor.get(apiCall.sponsorAddress)!;
+
+    if (sponsorRequests < MAXIMUM_SPONSOR_WALLET_REQUESTS) {
+      requestCountPerSponsor.set(apiCall.sponsorAddress, sponsorRequests + 1);
+      allowedApiCalls.push(apiCall);
+    } else {
+      logs.push(logger.pend('WARN', `Ignoring Request ID:${apiCall.id} as it exceeded sponsor wallet request limit.`));
```
andreogle, line 68:
```suggestion
      logs.push(logger.pend('WARN', `Blocking Request ID:${apiCall.id} as it exceeded sponsor wallet request limit.`));
```

packages/airnode-node/src/evm/requests/blocking.ts
```diff
@@ -1,41 +1,40 @@
...
-  const { logs, apiCalls } = pendingApiCalls.reduce((acc, apiCall) => {
+  const { logs, apiCalls } = requests.apiCalls.reduce((acc, apiCall) => {
     const pendingWithdrawal = withdrawalsBySponsorAddress[apiCall.sponsorAddress];
 
-    if (pendingWithdrawal) {
+    if (apiCall.status === RequestStatus.Pending && pendingWithdrawal) {
       const warningLog = logger.pend(
         'WARN',
-        `Ignoring Request ID:${apiCall.id} as it has a pending Withdrawl ID:${pendingWithdrawal.id}`
+        `Ignoring Request ID:${apiCall.id} as it has a pending Withdrawal ID:${pendingWithdrawal.id}`
```
andreogle, line 28: nice catch

andreogle (changes requested):

packages/airnode-node/src/types.ts
```diff
@@ -37,12 +37,14 @@ export enum RequestErrorMessage {
...
   ResponseValueNotFound = 'Response value not found',
   FulfillTransactionFailed = 'Fulfill transaction failed',
+  SponsorRequestLimitExceeded = 'Sponsor request limit exceeded',
 }
 
 export enum RequestStatus {
   Pending = 'Pending',
   Fulfilled = 'Fulfilled',
   Submitted = 'Submitted',
+  // TODO: Merge Blocked and Ignored status together, since it's confusing which one to use
```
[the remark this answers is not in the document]
Siegrift: Thanks, this makes it much clearer. 

I think these statuses should be explained here in the types definition.

packages/airnode-node/src/evm/requests/blocking.ts
```diff
@@ -46,6 +45,41 @@ export function blockRequestsWithWithdrawals(requests: GroupedRequests): LogsDat
...
+        errorMessage: `${RequestErrorMessage.SponsorRequestLimitExceeded}: ${apiCall.id}`,
+      };
+      allowedApiCalls.push(blockedCall);
+    }
+  });
+
+  return [[...prevLogs, ...logs], { ...requests, apiCalls: allowedApiCalls }];
 }
+
+// TODO: Merge with verification/api-call-verification.ts
```
Siegrift, line 81: @andreogle WDYT?

packages/airnode-node/src/evm/requests/blocking.ts
Siegrift: Yeah, I found the spreading too noisy.

packages/airnode-node/src/evm/handlers/fetch-pending-requests.ts
Siegrift: I will actually need to change this since I ended up "marking" the requests instead of filtering them.

I will think about the naming while doing this.

packages/airnode-node/src/evm/requests/blocking.ts
```diff
@@ -1,41 +1,40 @@
...
+        `Ignoring Request ID:${apiCall.id} as it has a pending Withdrawal ID:${pendingWithdrawal.id}`
       );
       const blockedCall: Request<ApiCall> = {
         ...apiCall,
         status: RequestStatus.Ignored,
+        // TODO: These error messages should be functions accepting id parameter and creating the error string
         errorMessage: `${RequestErrorMessage.PendingWithdrawal}: ${pendingWithdrawal.id}`,
       };
+
+      // TODO: Drop the ignored requests right away (this needs to be done in thw whole codebase)
```
Siegrift, line 37: Decided (in slack) we do not want to do that.

bbenligiray:
Note that the limit is to be applied per sponsor-sponsorWallet pair. Assume we have the following requests

1. sponsor: A sponsorWallet: X
1. sponsor: A sponsorWallet: X
1. sponsor: A sponsorWallet: X
1. sponsor: A sponsorWallet: Y
1. sponsor: A sponsorWallet: Y
1. sponsor: A sponsorWallet: Y
1. sponsor: B sponsorWallet: X
1. sponsor: B sponsorWallet: X
1. sponsor: B sponsorWallet: X
1. sponsor: B sponsorWallet: Y
1. sponsor: B sponsorWallet: Y
1. sponsor: B sponsorWallet: Y

Here, there are 3 of each unique sponsor-sponsorWallet pair. We shouldn't block any of these because the numbers of each pair is less than 5. This is because we didn't verify the sponsorWallet address according to the sponsor address yet so we have to treat each pair as potentially correct/one shouldn't be able to block the usage of a sponsorWallet by making a few requests with incorrect sponsor addresses.

Siegrift:
Thanks, I forgot about that.

packages/airnode-node/src/evm/requests/blocking.ts
```diff
@@ -1,41 +1,40 @@
...
-
+  const withdrawalsBySponsorAddress = keyBy(pendingWithdrawals, 'sponsorAddress');
   const initialState: ApiCallsWithLogs = { logs: [], apiCalls: [] };
 
-  const { logs, apiCalls } = pendingApiCalls.reduce((acc, apiCall) => {
+  const { logs, apiCalls } = requests.apiCalls.reduce((acc, apiCall) => {
     const pendingWithdrawal = withdrawalsBySponsorAddress[apiCall.sponsorAddress];
 
-    if (pendingWithdrawal) {
+    if (apiCall.status === RequestStatus.Pending && pendingWithdrawal) {
```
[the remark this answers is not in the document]
Siegrift: I will return early to make this a tiny bit better.

packages/airnode-node/src/evm/requests/blocking.ts
Siegrift: I think the time for it will come :)

Siegrift (commented):
I have introduced many TODOs and wanted to fix them in a separate PR, but I'll probably postpone that since the team looks busy atm and I will rather focus on the other actionables we (I) have in Jira. 

Created for it https://api3dao.atlassian.net/browse/AN-433

packages/airnode-node/src/evm/requests/blocking.ts
```diff
@@ -1,51 +1,112 @@
...
+  const apiCallsWithSponsorId = requests.apiCalls.map((apiCall) => ({
+    ...apiCall,
+    // The "sponsor id" is a unique combination of sponsor and sponsor wallet
+    sponsorId: `${apiCall.sponsorAddress}-${apiCall.sponsorWalletAddress}`,
+  }));
+  const requestCountPerSponsorId = new Map<string, number>();
+  const allowedApiCalls: Request<ApiCall>[] = [];
+  const logs: PendingLog[] = [];
+
+  // TODO: Consider reduce or map for this as a functional approach
```
andreogle, line 77: Not really necessary

packages/airnode-node/src/evm/requests/blocking.ts
```diff
@@ -1,51 +1,112 @@
...
+      ...apiCall,
+      status: RequestStatus.Blocked,
+      errorMessage: `${RequestErrorMessage.SponsorRequestLimitExceeded}: ${apiCall.id}`,
+    };
+    allowedApiCalls.push(blockedCall);
+  });
+
+  return [
+    [...prevLogs, ...logs],
+    { ...requests, apiCalls: allowedApiCalls.map((apiCall) => omit(apiCall, 'sponsorId')) },
```
andreogle, line 104: Why not just calculate `sponsorId` inside the `forEach` loop? Doesn't seem like it needs to be added to the API calls and then removed again?

andreogle (approved):
It seems a bit unnecessary to map over the API calls to assign `sponsorId` and then map over them again to remove it. Otherwise looks good to me 👍🏻

packages/airnode-node/src/evm/requests/blocking.ts
Siegrift: Yeah, I thought so as well, but wanted feedback.

packages/airnode-node/src/evm/requests/blocking.ts
Siegrift: No reason. I like this suggestion.

PR #726, opened 2021-12-05
AN-403: EIP-1559 Fulfilments - non-configurable version

aquarat:
This has been tested on Ropsten and a HardHat instance:

The transaction itself:
https://ropsten.etherscan.io/tx/0xbec7257ba716ac3e5963c630bea9a1597a39ce80af40eee8a7f3a98774dd4948

The preceding block (from which `maxFeePerGas` was calculated):
https://ropsten.etherscan.io/block/11559286

P.S. I hope I got the changeset right?


.changeset/tall-toes-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@api3/airnode-node': patch
+---
+
+AN-403: Implement EIP-1559 fulfillments
```
aquarat, line 5: I'll modify this to remove the Jira number.

packages/airnode-node/src/evm/gas-prices.ts
```diff
@@ -1,20 +1,49 @@
...
+export const getGasPrice = async (options: FetchOptions): Promise<LogsData<GasTarget | null>> => {
   const { provider } = options;
+  const [logs, getBlockGas] = await (async (): Promise<[PendingLog[], GasTarget | undefined]> => {
+    const operation = () => provider.getBlock('latest');
+    const [err, blockHeader] = await go(operation, { retries: 1, timeoutMs: DEFAULT_RETRY_TIMEOUT_MS });
+    if (err || !blockHeader?.baseFeePerGas) {
+      const log = logger.pend('INFO', 'Failed to get EIP-1559 gas pricing from provider - trying fallback', err);
+
+      return [[log], undefined];
+    } else {
```
andreogle, line 20: Nit: You don't need `else` blocks if you're returning early. A nice benefit of this is that the below code gets indented 2 spaces left, which makes it easier to read and follow (especially when you have lots of `if` statements

packages/airnode-node/src/evm/gas-prices.ts
```diff
@@ -1,20 +1,49 @@
...
-export async function getGasPrice(options: FetchOptions): Promise<LogsData<ethers.BigNumber | null>> {
+export const getGasPrice = async (options: FetchOptions): Promise<LogsData<GasTarget | null>> => {
   const { provider } = options;
+  const [logs, getBlockGas] = await (async (): Promise<[PendingLog[], GasTarget | undefined]> => {
+    const operation = () => provider.getBlock('latest');
+    const [err, blockHeader] = await go(operation, { retries: 1, timeoutMs: DEFAULT_RETRY_TIMEOUT_MS });
+    if (err || !blockHeader?.baseFeePerGas) {
+      const log = logger.pend('INFO', 'Failed to get EIP-1559 gas pricing from provider - trying fallback', err);
+
+      return [[log], undefined];
```
andreogle, line 19: This returns `undefined` but elsewhere returns `null`?

packages/airnode-node/src/evm/gas-prices.ts
```diff
@@ -1,20 +1,49 @@
...
+        [],
+        {
+          maxPriorityFeePerGas,
+          maxFeePerGas,
+        } as GasTarget,
+      ];
+    }
+  })();
+
+  if (getBlockGas) {
```
andreogle, line 36: Nit: maybe just call this variable `blockGas`? The "get" verb implies it's a function to me

packages/airnode-node/src/evm/fulfillments/withdrawals.ts
```diff
@@ -78,7 +78,9 @@ export async function submitWithdrawal(
   // We set aside some ETH to pay for the gas of the following transaction,
   // send all the rest along with the transaction. The contract will direct
   // these funds back to the sponsor wallet.
-  const txCost = paddedGasLimit.mul(options.gasPrice);
+  const txCost = paddedGasLimit.mul(
+    options.gasPrice.gasPrice ? options.gasPrice.gasPrice : options.gasPrice.maxFeePerGas!
```
andreogle, line 82: Nit: The nested "gasPrice.gasPrice" isn't great 😅 Maybe try to rename the object?

packages/airnode-node/src/types.ts
```diff
@@ -166,18 +166,24 @@ export interface EVMContracts {
...
 }
 
 export interface TransactionOptions {
-  readonly gasPrice: number | ethers.BigNumber;
+  readonly gasPrice: GasTarget;
   readonly masterHDNode: ethers.utils.HDNode;
   readonly provider: ethers.providers.JsonRpcProvider;
 }
 
+export interface GasTarget {
```
andreogle, line 181: Maybe it's worth including the transaction type here so that there's a single way of checking what value to use?

packages/airnode-node/src/evm/handlers/process-transactions.ts
```diff
@@ -41,8 +41,18 @@ export async function processTransactions(
     return state2;
   }
 
-  const gweiPrice = utils.weiToGwei(gasPrice);
-  logger.info(`Gas price set to ${gweiPrice} Gwei`, baseLogOptions);
+  if (gasPrice.maxFeePerGas && gasPrice.maxPriorityFeePerGas) {
+    const gweiMaxFee = utils.weiToGwei(gasPrice.maxFeePerGas);
+    const gweiPriorityFee = utils.weiToGwei(gasPrice.maxPriorityFeePerGas);
+    logger.info(
+      `Gas price (EIP-1559) set to a Max Fee of ${gweiMaxFee} Gwei and ` + ` Priority Fee of ${gweiPriorityFee} Gwei`,
```
andreogle, line 48: We don't need the "+" separator here?

packages/airnode-node/src/evm/gas-prices.ts
```diff
@@ -1,20 +1,49 @@
...
-import { LogsData } from '../types';
+import { BASE_FEE_MULTIPLIER, DEFAULT_RETRY_TIMEOUT_MS, PRIORITY_FEE, WEI_PER_GWEI } from '../constants';
+import { GasTarget, LogsData, PendingLog } from '../types';
 
 interface FetchOptions {
   readonly provider: ethers.providers.JsonRpcProvider;
 }
 
-export async function getGasPrice(options: FetchOptions): Promise<LogsData<ethers.BigNumber | null>> {
+export const getGasPrice = async (options: FetchOptions): Promise<LogsData<GasTarget | null>> => {
```
andreogle, line 11: Maybe we want to decide on a consistent way defining functions as a team. I find `export async function getGasPrice` _slightly_ easier to read at a glance because of the "function" keyword. Plus you get function hoisting for free, although I don't think it makes a difference for top level functions. I think @Siegrift also prefers the `expost const getGetPrice = async` approach.

This is obviously a suuuuuuper minor issue though but I do think consistency is important. I don't mind either way

packages/airnode-node/src/constants.ts
```diff
@@ -38,3 +38,12 @@ export const WORKER_PROVIDER_PROCESS_REQUESTS_TIMEOUT = 32_000;
...
 export const WORKER_CALL_API_TIMEOUT = 10_000;
+
+// The priority fee in gwei
+export const PRIORITY_FEE = '3.12';
+
+// The Base Fee to Max Fee multiplier with two decimal places : 2.56 * 100 = 256
+export const BASE_FEE_MULTIPLIER = '266';
+
+// The number of Wei in a Gwei
+export const WEI_PER_GWEI = '1000000000';
```
andreogle, line 49: Doesn't ethers provide some utility functions/constants for this?

packages/airnode-node/src/constants.ts
```diff
@@ -38,3 +38,12 @@ export const WORKER_PROVIDER_PROCESS_REQUESTS_TIMEOUT = 32_000;
 
 // The maximum amount of time the "call API" worker is allowed before being timed out
 export const WORKER_CALL_API_TIMEOUT = 10_000;
+
+// The priority fee in gwei
+export const PRIORITY_FEE = '3.12';
+
+// The Base Fee to Max Fee multiplier with two decimal places : 2.56 * 100 = 256
+export const BASE_FEE_MULTIPLIER = '266';
```
andreogle, line 46: The comment shows `256` (not `266`) ?

andreogle (commented):
A whole lot of minor & nit comments from a quick review 😅 Let me know what you think

packages/airnode-node/src/evm/gas-prices.ts
aquarat: Wow, rookie mistake, thanks :)

packages/airnode-node/src/evm/gas-prices.ts
aquarat: haha yeah, I'm not fussed and I totally agree @ consistency.
I think I prefer the `const` approach because it better reflects that the function is a value and I think it is also newer?

packages/airnode-node/src/constants.ts
bbenligiray, line 46: I think `BASE_FEE_MULTIPLIER` should be a `Number` (where `2` means multiply by two) and used this way
```
  const maxFeePerGas = baseFeePerGas.mul(BASE_FEE_MULTIPLIER * 100).div(100).add(maxPriorityFeePerGas);
```

aquarat: Thank you, I'll fix it.
Do you agree with the `BASE_FEE_MULTIPLIER` being a percentage without a decimal point?

bbenligiray, line 46: Why don't we multiply by 2 here?

packages/airnode-node/src/constants.ts
aquarat: Surprisingly not.
They have [WeiPerEther](https://docs.ethers.io/v5/api/utils/constants/), but no Gwei variants.

packages/airnode-node/src/evm/fulfillments/withdrawals.ts
aquarat: I appreciate the nits, how else will I learn :)
I'll rename the base gasPrice variable to gasTarget perhaps? `gasPrice` is a field in a type specified by `ethers`.

packages/airnode-node/src/constants.ts
```diff
@@ -38,3 +38,12 @@ export const WORKER_PROVIDER_PROCESS_REQUESTS_TIMEOUT = 32_000;
 
 // The maximum amount of time the "call API" worker is allowed before being timed out
 export const WORKER_CALL_API_TIMEOUT = 10_000;
+
+// The priority fee in gwei
+export const PRIORITY_FEE = '3.12';
```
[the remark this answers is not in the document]
aquarat: Acknowledged, I'll update it.

packages/airnode-node/src/constants.ts
aquarat: Acknowledged and will implement.

aquarat: We said that we wanted to be slightly ahead of the herd - I thought that this applied to both Max Fee and Priority Fee, but on reflection it only applies to Priority Fee. I'll set this at 2.

packages/airnode-node/src/evm/gas-prices.ts
aquarat: It won't have a practical effect because the `undefined` value triggers an attempt to retrieve the fallback (which results in a valid value or null), but I'll swap it out for consistency.

packages/airnode-node/src/evm/handlers/process-transactions.ts
aquarat: I'll remove it thanks.

packages/airnode-node/src/types.ts
aquarat: If GasTarget.gasPrice is `!undefined`, it's transaction type is 1, otherwise transaction type 2. 
GasTarget can actually be swapped out with `ethers.providers.FeeData` - which I'll do now.

aquarat:
All suggestions and requests applied. 
`gasPrice` universally renamed to `gasTarget` as `ethers` expects an object with potentially a `gasPrice` field defined.
I have yet to test this on Ropsten but all other tests pass.

packages/airnode-node/src/types.ts
andreogle: Ehh I guess this object is simple enough that it doesn't really matter. Was just a suggestion but I think you're right about it being a bit unnecessary

andreogle (approved):
Looks muuuch better. Nice one 🔥 

Mentioned on Slack too, but I think testing non-EIP-1559 transactions on hardhat is fine

PR #727, opened 2021-12-06
Consolidate protocols

bbenligiray:
This continues from https://github.com/api3dao/airnode/pull/725 (but we can still merge that too) and does the following
- Merges AirnodeRrp and AirnodePsp into a single contract
- Replaces convenience functions with Multicall
- Replaces the RRP withdrawal utility with a protocol-agnostic one (where `protocolId` is the 0 mentioned [here](https://github.com/api3dao/airnode/blob/master/packages/airnode-protocol/README.md#sponsor-wallet-derivation))
- Differentiates between `airnode` and `reporter` in preparation of second-party oracles (`airnode==reporter` for a first-party oracle)
- Removes full requests

I removed full requests and modified the `require(airnode != address(0))` checks to avoid stack too deep errors. `airnode` is also not emitted in logs for the same reason. This doesn't feel ideal (except the removal of full requests, I think that's a good decision because we're very template focused now), let me know if you can think of a workaround.

### DIFF
--- a/packages/airnode-protocol/contracts/WIP/Airnode.sol
+++ b/packages/airnode-protocol/contracts/WIP/Airnode.sol
@@ -29,6 +29,7 @@ contract Airnode is Multicall, WithdrawalUtils, IAirnode {
         bytes32 endpointId,
         bytes calldata parameters
     ) external returns (bytes32 templateId) {
+        require(airnode != address(0), "Airnode address zero");
         templateId = keccak256(
             abi.encodePacked(airnode, endpointId, parameters)
         );
@@ -61,6 +62,7 @@ contract Airnode is Multicall, WithdrawalUtils, IAirnode {
         bytes4 fulfillFunctionId,
         bytes calldata parameters
     ) external returns (bytes32 requestId) {
+        require(templates[templateId].airnode != address(0), "Template does not exist");
         require(fulfillAddress != address(this), "Fulfill address AirnodeRrp");
         require(
             sponsorToRequesterToSponsorshipStatus[sponsor][msg.sender],
--- a/packages/airnode-protocol/contracts/WIP/Airnode.sol
+++ b/packages/airnode-protocol/contracts/WIP/Airnode.sol
@@ -213,6 +213,7 @@ contract Airnode is Multicall, WithdrawalUtils, IAirnode {
         bytes4 fulfillFunctionId,
         bytes calldata parameters
     ) external returns (bytes32 subscriptionId) {
+        require(templates[templateId].airnode != address(0), "Template does not exist");
         subscriptionId = keccak256(
             abi.encodePacked(
                 templateId,
@@ -254,6 +255,7 @@ contract Airnode is Multicall, WithdrawalUtils, IAirnode {
         bytes calldata signature
     ) external returns (bool callSuccess, bytes memory callData) {
         Subscription storage subscription = subscriptions[subscriptionId];
+        require(subscription.templateId != bytes32(0), "Subscription does not exist");
         require(
             (
                 keccak256(abi.encodePacked(subscriptionId, data))
--- a/packages/airnode-protocol/contracts/WIP/Airnode.sol
+++ b/packages/airnode-protocol/contracts/WIP/Airnode.sol
@@ -1,0 +1,273 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "@openzeppelin/contracts/utils/Multicall.sol";
+import "./WithdrawalUtils.sol";
+import "./interfaces/IAirnode.sol";
+
+contract Airnode is Multicall, WithdrawalUtils, IAirnode {
+    using ECDSA for bytes32;
+
+    struct Template {
+        address airnode;
+        bytes32 endpointId;
+        bytes parameters;
+    }
+
+    mapping(bytes32 => Template) public templates;
+
+    mapping(address => mapping(address => bool))
+        public sponsorToRequesterToSponsorshipStatus;
+
+    mapping(address => uint256) public requesterToRequestCountPlusOne;
+
+    mapping(bytes32 => bytes32) private requestIdToFulfillmentParameters;
+
+    function createTemplate(
+        address airnode,
+        bytes32 endpointId,
+        bytes calldata parameters
+    ) external returns (bytes32 templateId) {
+        templateId = keccak256(
+            abi.encodePacked(airnode, endpointId, parameters)
+        );
+        templates[templateId] = Template({
+            airnode: airnode,
+            endpointId: endpointId,
+            parameters: parameters
+        });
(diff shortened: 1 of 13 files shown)


packages/airnode-protocol/contracts/WIP/Airnode.sol
```diff
@@ -0,0 +1,273 @@
...
+    mapping(address => uint256) public requesterToRequestCountPlusOne;
+
+    mapping(bytes32 => bytes32) private requestIdToFulfillmentParameters;
+
+    function createTemplate(
+        address airnode,
+        bytes32 endpointId,
+        bytes calldata parameters
+    ) external returns (bytes32 templateId) {
+        templateId = keccak256(
```
bbenligiray, line 32: No longer reverting if `airnode` is `address(0)` to be consistent with `makeRequest()`

packages/airnode-protocol/contracts/WIP/Airnode.sol
```diff
@@ -0,0 +1,273 @@
...
+    function makeRequest(
+        bytes32 templateId,
+        address reporter,
+        address sponsor,
+        address sponsorWallet,
+        address fulfillAddress,
+        bytes4 fulfillFunctionId,
+        bytes calldata parameters
+    ) external returns (bytes32 requestId) {
+        require(fulfillAddress != address(this), "Fulfill address AirnodeRrp");
```
bbenligiray, line 64: Not checking if `airnode` is `address(0)`

packages/airnode-protocol/contracts/WIP/Airnode.sol
```diff
@@ -0,0 +1,273 @@
...
+                templates[templateId].airnode,
+                reporter,
+                sponsorWallet,
+                fulfillAddress,
+                fulfillFunctionId
+            )
+        );
+        requesterToRequestCountPlusOne[msg.sender]++;
+        emit MadeRequest(
+            reporter,
```
bbenligiray, line 98: It is the `reporter` listening for this event, not `airnode`

packages/airnode-protocol/contracts/WIP/Airnode.sol
```diff
@@ -0,0 +1,273 @@
...
+                sponsorWallet,
+                fulfillAddress,
+                fulfillFunctionId,
+                parameters
+            )
+        );
+        requestIdToFulfillmentParameters[requestId] = keccak256(
+            abi.encodePacked(
+                templates[templateId].airnode,
+                reporter,
```
bbenligiray, line 90: We need to add this to the request ID calculation because it's used in `fulfill()`

packages/airnode-protocol/contracts/WIP/Airnode.sol
```diff
@@ -0,0 +1,273 @@
...
+                    .toEthSignedMessageHash()
+            ).recover(signature) == airnode,
+            "Invalid signature"
+        );
+        delete requestIdToFulfillmentParameters[requestId];
+        (callSuccess, callData) = fulfillAddress.call( // solhint-disable-line avoid-low-level-calls
+            abi.encodeWithSelector(fulfillFunctionId, requestId, data)
+        );
+        if (callSuccess) {
+            emit FulfilledRequest(reporter, requestId, data);
```
bbenligiray, line 145: Again, we're emitting `reporter` instead of `airnode` because it's the `reporter` listening for this event

packages/airnode-protocol/contracts/WIP/Airnode.sol
```diff
@@ -0,0 +1,273 @@
...
+            fulfillAddress,
+            fulfillFunctionId,
+            parameters
+        );
+    }
+
+    function fulfillRequest(
+        bytes32 requestId,
+        address airnode,
+        address reporter,
```
bbenligiray, line 115: We're only passing this to be emitted in the event

packages/airnode-protocol/contracts/WIP/Airnode.sol
```diff
@@ -0,0 +1,273 @@
...
+        isAwaitingFulfillment =
+            requestIdToFulfillmentParameters[requestId] != bytes32(0);
+    }
+
+    // PSP starts here
+
+    struct Subscription {
+        bytes32 templateId;
+        address reporter;
+        address sponsor;
```
bbenligiray, line 194: Same deal here

packages/airnode-protocol/contracts/WIP/Airnode.sol
```diff
@@ -0,0 +1,273 @@
...
+        );
+        (callSuccess, callData) = subscription.fulfillAddress.call( // solhint-disable-line avoid-low-level-calls
+            abi.encodeWithSelector(
+                subscription.fulfillFunctionId,
+                subscriptionId,
+                data
+            )
+        );
+        if (callSuccess) {
+            emit FulfilledSubscription(subscriptionId, data);
```
bbenligiray, line 270: Dropped `airnode` from this event, it didn't feel consistent

packages/airnode-protocol/contracts/WIP/WithdrawalUtils.sol
```diff
@@ -0,0 +1,61 @@
...
+import "./interfaces/IWithdrawalUtils.sol";
+
+contract WithdrawalUtils is IWithdrawalUtils {
+    mapping(address => uint256) public sponsorToWithdrawalRequestCount;
+
+    mapping(bytes32 => bytes32) private withdrawalRequestIdToParameters;
+
+    function requestWithdrawal(
+        address reporter,
+        uint256 protocolId,
```
bbenligiray, line 13: 0 for RRP, 1 for PSP, 12345 for Airkeeper, etc.

packages/airnode-protocol/contracts/WIP/WithdrawalUtils.sol
```diff
@@ -0,0 +1,61 @@
...
+
+import "./interfaces/IWithdrawalUtils.sol";
+
+contract WithdrawalUtils is IWithdrawalUtils {
+    mapping(address => uint256) public sponsorToWithdrawalRequestCount;
+
+    mapping(bytes32 => bytes32) private withdrawalRequestIdToParameters;
+
+    function requestWithdrawal(
+        address reporter,
```
bbenligiray, line 12: It is the `reporter` that we're requesting the withdrawal from, not `airnode`

packages/airnode-protocol/contracts/WIP/interfaces/IAirnode.sol
```diff
@@ -0,0 +1,59 @@
...
+        address indexed reporter,
+        bytes32 indexed requestId,
+        bytes data
+    );
+
+    event FailedRequest(
+        address indexed reporter,
+        bytes32 indexed requestId,
+        string errorMessage
+    );
```
bbenligiray, line 44: It's annoying that you can't parse these to find all the requests related to an `airnode`. It's obvious to me that this is a requirement though.

bbenligiray:
Closed to rename the branch as https://github.com/api3dao/airnode/tree/v1-protocol

PR #728, opened 2021-12-07
Move response signing from `processTransactions` handler to `callApi`

amarthadan:
Issue [AN-397](https://api3dao.atlassian.net/browse/AN-397)

### DIFF
--- a/packages/airnode-node/src/evm/fulfillments/api-calls.ts
+++ b/packages/airnode-node/src/evm/fulfillments/api-calls.ts
@@ -14,20 +14,8 @@ type StaticResponse = { readonly callSuccess: boolean; readonly callData: string
 
 type SubmitResponse = ethers.Transaction | null;
 
-interface ApiCallWithSignature extends ApiCall {
-  readonly signature: string;
-}
-
-async function signResponseMessage(request: Request<ApiCall>, masterHDNode: ethers.utils.HDNode) {
-  const airnodeWallet = ethers.Wallet.fromMnemonic(masterHDNode.mnemonic!.phrase);
-  return await airnodeWallet.signMessage(
-    ethers.utils.arrayify(
-      ethers.utils.keccak256(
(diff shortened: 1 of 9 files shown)


Siegrift (approved):
LGTM :+1:  Doing it this way also feels a bit more natural to me.

andreogle (approved):
Nice one 👍🏻

packages/airnode-node/src/evm/fulfillments/api-calls.ts
```diff
@@ -123,12 +111,8 @@ async function testAndSubmitFulfill(
     return [submitLogs, submitErr, submitData];
   }
 
-  // Compute signature once.
-  // Not computing this beforehand because it might be a very computational expensive operation
-  const signature = await signResponseMessage(request, options.masterHDNode);
-
   // Should not throw
-  const [testLogs, testErr, testData] = await testFulfill(airnodeRrp, { ...request, signature }, options);
+  const [testLogs, testErr, testData] = await testFulfill(airnodeRrp, { ...request }, options);
```
acenolaza, line 115: Late but I was wondering if we really need to spread the `request` object here or if we could just pass it as is to `testFulfill()`

packages/airnode-node/src/evm/fulfillments/api-calls.ts
```diff
@@ -149,7 +133,7 @@ async function testAndSubmitFulfill(
 
   // We expect the transaction to be successful if submitted
   if (testData?.callSuccess) {
-    const [submitLogs, submitErr, submitData] = await submitFulfill(airnodeRrp, { ...request, signature }, options);
+    const [submitLogs, submitErr, submitData] = await submitFulfill(airnodeRrp, { ...request }, options);
```
acenolaza, line 136: same here

packages/airnode-node/src/evm/fulfillments/api-calls.ts
amarthadan: Of course, we can :man_facepalming: I'll add it to my next PR.

PR #732, opened 2021-12-08
Remove WIP contracts

bbenligiray:
Removing all WIP contracts from the master branch. Token lock and payment contracts/tests will be living [here](https://github.com/api3dao/airnode/tree/v1-protocol/packages/airnode-protocol/contracts/WIP/authorizers/admin) for now


Ashar2shahid (approved):
I was wondering if the tests broke, so I pulled it locally and tested it. Everything looks good in this branch, in the v1 branch however, should we keep the files in the WIP directory? I was thinking keeping them as it is

bbenligiray:
In `v1-protocol` I keep all new things in `/WIP` https://github.com/api3dao/airnode/tree/v1-protocol/packages/airnode-protocol/contracts/WIP That directory will be audited as a whole.

PR #733, opened 2021-12-09
Move sponsor wallet check to callApi

Siegrift:
https://api3dao.atlassian.net/browse/AN-394

### DIFF
--- a/packages/airnode-node/src/evm/verification/api-call-verification.test.ts
+++ b/packages/airnode-node/src/evm/verification/api-call-verification.test.ts
@@ -26,7 +26,7 @@ describe('verifyApiCallIds', () => {
       chainId: '31337',
       requesterAddress: '0x7f7d1Aa0792aC39f43C6e7FA2ec31258Fc5FD612',
       sponsorAddress: '0x69e2B095fbAc6C3f9E528Ef21882b86BF1595181',
-      sponsorWalletAddress: '0xd5e6a768f1d23d30B386Bb5c125DBe83A9c40c73',
+      sponsorWalletAddress: '0xd5e6a768f1d23d30B386Bb5c125DBe83A9c40c73', // Invalid sponsor wallet address
       fulfillAddress: '0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512',
       fulfillFunctionId: '0x7c1de7e1',
       encodedParameters: '0x62746f6355534466616d6f756e746131',
@@ -46,7 +46,7 @@ describe('verifyApiCallIds', () => {
       chainId: '31337',
       requesterAddress: '0x7f7d1Aa0792aC39f43C6e7FA2ec31258Fc5FD612',
       sponsorAddress: '0x69e2B095fbAc6C3f9E528Ef21882b86BF1595181',
-      sponsorWalletAddress: '0xd5e6a768f1d23d30B386Bb5c125DBe83A9c40c73',
+      sponsorWalletAddress: '0xd5e6a768f1d23d30B386Bb5c125DBe83A9c40c73', // Invalid sponsor wallet address
       fulfillAddress: '0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512',
       fulfillFunctionId: '0x7c1de7e1',
       encodedParameters: '0x62746f6355534466616d6f756e746131',
@@ -67,7 +67,7 @@ describe('verifyApiCallIds', () => {
       chainId: '31337',
       requesterAddress: '0x7f7d1Aa0792aC39f43C6e7FA2ec31258Fc5FD612',
       sponsorAddress: '0x69e2B095fbAc6C3f9E528Ef21882b86BF1595181',
-      sponsorWalletAddress: '0xd5e6a768f1d23d30B386Bb5c125DBe83A9c40c73',
+      sponsorWalletAddress: '0xd5e6a768f1d23d30B386Bb5c125DBe83A9c40c73', // Invalid sponsor wallet address
       fulfillAddress: '0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512',
       fulfillFunctionId: '0x7c1de7e1',
       encodedParameters: '0x62746f6355534466616d6f756e746131',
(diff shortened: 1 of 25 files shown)


.changeset/friendly-donkeys-sing.md
```diff
@@ -0,0 +1,8 @@
+---
+'@api3/airnode-deployer': minor
+'@api3/airnode-examples': minor
+'@api3/airnode-node': minor
+'@api3/airnode-validator': minor
+---
+
+Initialise packages for v0.4.0
```
andreogle, line 8: Don't we want to include all packages here if we're bumping the version?

packages/airnode-node/src/evm/verification/api-call-verification.test.ts
```diff
@@ -26,7 +26,7 @@ describe('verifyApiCallIds', () => {
       chainId: '31337',
       requesterAddress: '0x7f7d1Aa0792aC39f43C6e7FA2ec31258Fc5FD612',
       sponsorAddress: '0x69e2B095fbAc6C3f9E528Ef21882b86BF1595181',
-      sponsorWalletAddress: '0xd5e6a768f1d23d30B386Bb5c125DBe83A9c40c73',
+      sponsorWalletAddress: '0xd5e6a768f1d23d30B386Bb5c125DBe83A9c40c73', // Invalid sponsor wallet address
```
andreogle, line 29: Does this just mean the test data is technically invalid? Is it possible to update these to use a valid address?

packages/airnode-node/src/evm/verification/request-verification.ts
```diff
@@ -45,6 +51,8 @@ export function verifyRrpTriggers(
 ): LogsData<Request<ApiCall>[]> {
   const logsWithVerifiedApiCalls: LogsData<Request<ApiCall>>[] = apiCalls.map((apiCall) => {
     if (apiCall.status !== RequestStatus.Pending) {
+      // TODO: Do we need to log this? We do not follow the same practice in applySponsorAndSponsorWalletRequestLimit
```
andreogle, line 54: It's a debug log so I would say it's better to log too much information than too little, most users will never see it. I would update the other function to be consistent

packages/airnode-node/src/handlers/call-api.ts
```diff
@@ -1,84 +1,122 @@
 import * as adapter from '@api3/airnode-adapter';
-import { OIS, RESERVED_PARAMETERS } from '@api3/airnode-ois';
+import { RESERVED_PARAMETERS } from '@api3/airnode-ois';
+import { ethers } from 'ethers';
+import { getMasterHDNode } from '../evm';
 import { getReservedParameters } from '../adapters/http/parameters';
 import { API_CALL_TIMEOUT, API_CALL_TOTAL_TIMEOUT } from '../constants';
+import { isValidSponsorWallet } from '../evm/verification';
```
andreogle, line 7: Minor thing, but do we want to import this from `../evm` instead? Like `getMasterNode` above

packages/airnode-node/src/handlers/call-api.ts
```diff
@@ -1,84 +1,122 @@
...
 import { go, retryOnTimeout } from '../utils/promise-utils';
 
-function buildOptions(
-  chain: ChainConfig,
-  ois: OIS,
-  aggregatedApiCall: AggregatedApiCall,
-  apiCredentials: adapter.ApiCredentials[],
-  apiCallOptions: ApiCallOptions | undefined
-): adapter.BuildRequestOptions {
+function buildOptions(payload: CallApiPayload): adapter.BuildRequestOptions {
```
andreogle, line 21: This module refactor looks great 👍🏻

packages/airnode-node/src/types.ts
```diff
@@ -209,24 +210,46 @@ export interface AuthorizationByRequestId {
   readonly [requestId: string]: boolean;
 }
 
-export interface ApiCallResponse {
-  readonly value?: string | boolean;
-  readonly errorMessage?: string;
+export type ApiCallResponse = ApiCallSuccessResponse | ApiCallErrorResponse;
+
+export interface ApiCallSuccessResponse {
+  success: true;
```
andreogle, line 216: Do we want to make these interfaces readonly like the others?

packages/airnode-protocol/contracts/rrp/requesters/RrpBeaconServer.sol
```diff
@@ -245,6 +245,7 @@ contract RrpBeaconServer is
         bytes32 beaconId = requestIdToBeaconId[requestId];
         require(beaconId != bytes32(0), "No such request made");
         delete requestIdToBeaconId[requestId];
+        require(data.length == 64, "Incorrect data length");
```
andreogle, line 248: Is this from another branch?

andreogle (approved):
Nice work. Just a few minor comments

.changeset/friendly-donkeys-sing.md
Siegrift: I dunno why github is showing the diff that badly... The PR is actually way smaller and most of the changes you reviewed were already merged.

Anyway, you are right - That PR which introduced only added validator templates so it's misleading.

packages/airnode-node/src/evm/verification/api-call-verification.test.ts
Siegrift: Yeah, it would - but I think some test would break, since on places which need the correct sponsor wallet we use the overrides for it. 

I think it would be nicer if it was switched (and we passed) the invalid value when it's needed.

Siegrift: Created https://api3dao.atlassian.net/browse/AN-447

packages/airnode-node/src/evm/verification/request-verification.ts
Siegrift: I think this needs to be checked in other places for consistency as well. I am also not sure whether we should not ignore those requests directly.

I've added a bunch of TODOs https://github.com/api3dao/airnode/pull/724/files#diff-bafef992d17b51ddeeecf5c4f67a79342f80c0b28bd383c18310ce0da1a57278R43 and created to https://api3dao.atlassian.net/browse/AN-433 resolve this.

I'd revisit this properly once there is more time.

.changeset/friendly-donkeys-sing.md
Siegrift: I'll double check the changelogs in a separate PR

packages/airnode-node/src/types.ts
Siegrift: I actually dislike that because:
1. It creates unnecessary friction in tests and when using local muration
2. Provides zero benefit since our code is largely functional (in the important places)

and I have to say I dislike it kinda in general :D

packages/airnode-protocol/contracts/rrp/requesters/RrpBeaconServer.sol
Siegrift: Yeah, I think github differ is broken - sorry I didn't notice that there are so many changes not from my PR... 
I'd have rebased it.

packages/airnode-node/src/handlers/call-api.ts
Siegrift: Actually I can't - because it's exported as `verification` object and I think it's not necessary to use it as `verification.isValidSponsorWallet` for that single usage.

PR #734, opened 2021-12-10
Verify sponsorWalletAddress in processTransations instead of initializeProvider for withdrawals

amarthadan:
Issue [AN-400](https://api3dao.atlassian.net/browse/AN-400)

* verification of sponsor wallet address moved from `initializeProvider` handler to `processTransations` handler
* refactor of transaction processing so that changes to the request object are correctly propagated back


packages/airnode-node/src/evm/handlers/initialize-provider.ts
```diff
@@ -109,14 +110,12 @@ export async function initializeProvider(
...
-    state4.masterHDNode
-  );
-  logger.logPending(verifyWithdrawalLogs, baseLogOptions);
-
   const state5 = state.update(state4, {
-    requests: { ...state3.requests, apiCalls: verifiedApiCallsForRrpTriggers, withdrawals: verifiedWithdrawals },
+    requests: {
+      ...state3.requests,
+      apiCalls: verifiedApiCallsForRrpTriggers,
+      withdrawals: state4.requests.withdrawals,
```
acenolaza, line 117: Do we really need to override withdrawals since they haven't changed from state4?

packages/airnode-node/src/evm/handlers/process-transactions.ts
```diff
@@ -58,16 +58,7 @@ export async function processTransactions(
   // =================================================================
   // STEP 4: Submit transactions for each wallet
   // =================================================================
-  const receipts = await fulfillments.submit(state3);
-  const successfulReceipts = receipts.filter((receipt) => !!receipt.data);
-  successfulReceipts.forEach((receipt) => {
-    logger.info(`Transaction:${receipt.data!.hash} submitted for Request:${receipt.id}`, baseLogOptions);
```
acenolaza, line 64: I really like the receipts refactor and the usage of the new `applyFulfillment()` fuction but I'm not finding where this info log entry was moved and I wanted to check that we are still displaying it since I find it useful.

packages/airnode-node/src/evm/fulfillments/index.ts
```diff
@@ -9,141 +7,147 @@ import {
...
-    const updatedRequest = {
-      ...request,
-      fulfillment: { hash: receipt.data.hash },
-      status: RequestStatus.Submitted,
-    };
-    return [...acc, updatedRequest];
-  }, [] as Request<T>[]);
+  const nestedGroupRequests = await Promise.all(promises);
+  return nestedGroupRequests.reduce((merged, groupRequests) => ({
+    apiCalls: [...merged.apiCalls, ...groupRequests.apiCalls],
```
Siegrift, line 150: This will result in a different order of requests that was there previously (before the requests were grouped by sponsor address)

I implemented this with `flatten` previously, which had the same problem, but now with a bit more insight of how the requests flow through Airnode, I think we should respect the order. However, I think the proper way to do this would be to save up indexes of the logs immediately when they are read by `getLogs` of the provider. WDYT?

packages/airnode-node/src/evm/fulfillments/requests.ts
```diff
@@ -0,0 +1,9 @@
+import { ethers } from 'ethers';
+import { Request, RequestStatus } from '../../types';
+export function applyFulfillment<T>(request: Request<T>, data: ethers.Transaction | null) {
```
Siegrift, line 3: \n

Siegrift (changes requested):
LGTM overall, but I think we could improve a readability by refactoring a few things.

packages/airnode-node/src/evm/handlers/process-transactions.ts
amarthadan: It was moved here
https://github.com/api3dao/airnode/pull/734/files#diff-bc443c624cfa9d6ffa22e623859ab1c2923d39f98f02559744469483a234a4dcR123

packages/airnode-node/src/evm/fulfillments/index.ts
```diff
@@ -9,141 +7,147 @@ import {
...
- * Sponsor requests are performed from the sponsor wallet. Each transaction has a fixed nonce and blochchain rejects out
- * of order transactions.
- *
- * This function makes all the sponsor requests sequentially (ordered by transaction nonce increasingly).
- *
- * There is a concept of batched requests, but that doesn't work with transactions. See:
- * https://github.com/ethers-io/ethers.js/issues/892#issuecomment-828897859
- */
-const submitSponsorRequestsSequentially = async (state: ProviderState<EVMProviderState>, sponsorAddress: string) => {
+function baseLogOptions(state: ProviderState<EVMProviderState>) {
```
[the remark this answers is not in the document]
andreogle: This function could maybe be extracted to somewhere more generic (it's not specific to fulfilments). I think it could be re-used in some places

packages/airnode-node/src/evm/fulfillments/index.ts
```diff
@@ -9,141 +7,147 @@ import {
...
     provider: state.provider,
   };
+}
 
-  // Submit transactions for API calls
-  const submittedApiCalls = requests.apiCalls.map((apiCall): OrderedRequest => {
+function submitRequests<T>(
+  state: ProviderState<EVMProviderState>,
+  requests: Request<T>[],
+  type: RequestType,
```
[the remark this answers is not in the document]
amarthadan: We do. I use that to distinguish between API call and withdrawal once the requests are submitted (https://github.com/api3dao/airnode/pull/734/files#diff-bc443c624cfa9d6ffa22e623859ab1c2923d39f98f02559744469483a234a4dcR129). I added a TODO that the type should be probably included in the request object (as I've seen a similar approach in other parts of the code) but that's something for a bigger refactor and I didn't want to add it here.

packages/airnode-node/src/evm/fulfillments/index.ts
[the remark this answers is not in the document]
amarthadan: > This function could maybe be extracted to somewhere more generic (it's not specific to fulfilments). I think it could be re-used in some places

You're right but I would leave that for a separate refactor PR.

packages/airnode-node/src/evm/fulfillments/index.ts
amarthadan: But why is the different order a problem? Isn't the order important only when dealing with nonces (which should be fine)? Anyway, that's probably out of the scope of this PR, isn't it? Or is the current ordering a bug?

packages/airnode-node/src/evm/fulfillments/index.ts
```diff
@@ -9,141 +7,147 @@ import {
...
+): OrderedRequest<T>[] {
+  return requests.map((request) => {
     const makeRequest = async () => {
-      const [logs, err, data] = await submitApiCall(contract, apiCall, txOptions);
-      logger.logPending(logs, baseLogOptions);
-      if (err || !data) {
-        return { id: apiCall.id, type: RequestType.ApiCall, error: err };
-      }
-      return { id: apiCall.id, type: RequestType.ApiCall, data };
+      // NOTE: This err was not actually handled anywhere before, what should we do with it?
```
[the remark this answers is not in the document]
Siegrift: Created https://api3dao.atlassian.net/browse/AN-446

packages/airnode-node/src/evm/fulfillments/api-calls.ts
```diff
@@ -179,6 +179,7 @@ async function submitFail(
 // =================================================================
 // Main functions
 // =================================================================
+// TODO: Use SubmitRequest as a function type once moved to arrow functions
```
Siegrift, line 182: You can create an arrow function, but TODO feels OK to me as well.

Siegrift (approved):
LGTM

packages/airnode-node/src/evm/fulfillments/index.ts
Siegrift: More of an idea of the ideal solution. Whenever the order is important you would have a way to sort them. This way, you have to rely on the ordering being preserved which is not nice. I will create an issue for this.

Siegrift: https://api3dao.atlassian.net/browse/AN-452

PR #735, opened 2021-12-12
Extend the EIP-1559 implementation to be configurable

aquarat:
Things that could be contentious:
- Many tests are now run twice; once for each `txType`. This seems like the safest way to test.
- The use of `switch` vs `if` in some cases, which is more like `match` from other functional languages.
- The naming of the new config object as `chainOptions`...
- `txType` is a string and not a Number, this seems to work better with the validator.
- Is there a procedure for producing a new validator schema folder? For now changes are in v0.3.

This implementation is pending testing on a testnet.

P.S. `ethers` has a means of converting Wei to Gwei, but when working with a decimal number (ie. 3.12 Gwei) it becomes difficult to read, so I've stuck with a `WEI_PER_GWEI` constant - I hope this is okay.

AN-402


packages/airnode-examples/integrations/authenticated-coinmarketcap/config.example.json
```diff
@@ -11,7 +11,12 @@
           "url": "${PROVIDER_URL}"
         }
       },
-      "type": "evm"
+      "type": "evm",
+      "chainOptions": {
```
bbenligiray, line 15: Each chain object has a field called `id` and not `chainId` based on the assumption that there is no room for confusion to the degree that `chain` there is redundant (this is apparent when you consider`config.chains[0].id` vs `config.chains[0].chainId`). So based on this, I'd recommend `options` here.

packages/airnode-examples/integrations/authenticated-coinmarketcap/config.example.json
```diff
@@ -11,7 +11,12 @@
...
+        "txType": "2",
+        "baseFeeMultiplier": "2",
+        "priorityFeeGwei": "3.12"
```
bbenligiray, line 18: `priorityFeeInGwei` imo. Also, what if we had a `priorityFee` instead that expects `"3.12 gwei"` or `3120000000 wei`? I've never seen this, but having it in Gwei by default feels too opinionated but having a lot of optional parameters in the form of `priorityFeeInGwei`, `priorityFeeInWei`, etc. is not nice.

packages/airnode-examples/integrations/authenticated-coinmarketcap/config.example.json
aquarat: That makes sense, I will amend it :)

packages/airnode-examples/integrations/authenticated-coinmarketcap/config.example.json
aquarat: `ethers` has a facility to parse amounts where the unit is specified, so it is feasible to specify the Priority Fee as "3.12 gwei", leaving the unit up to the user. I can prepare an amendment that uses this approach.

bbenligiray: Maybe this is the proper way of doing it though
```
"priorityFee": { "value": "3.12", "unit": "gwei" }
```

packages/airnode-node/src/evm/handlers/process-transactions.test.ts
```diff
@@ -18,114 +19,184 @@ mockEthers({
...
+          chainOptions,
+        },
+      };
+
+      const res = await processTransactions(state);
+      expect(res.requests.apiCalls[0]).toEqual({ ...apiCall, nonce: 79 });
+      expect(res.gasTarget).toEqual(null);
+
+      // API call was NOT submitted
+      expect(contract.fulfill).not.toHaveBeenCalled();
```
Siegrift, line 199: This `contract` is just an instance you created. It is not used anywhere so it will never be submitted. You can mock one of the Airnode internal functions (e.g. `submit`) but I'd like to see it being called in one test and NOT being called in another - but I think it's not necessary.

packages/airnode-node/src/workers/local-handlers.ts
```diff
@@ -49,6 +50,15 @@ export async function callApi({ aggregatedApiCall, logOptions, apiCallOptions }:
   return { ok: true, data: response };
 }
 
+export const logMe = (obj: any) => {
```
Siegrift, line 53: Remove

packages/airnode-node/test/fixtures/config/config.ts
```diff
@@ -30,6 +30,9 @@ export function buildConfig(overrides?: Partial<Config>): Config {
         },
         id: '31337',
         type: 'evm',
+        // We set an invalid chainOptions value to keep us honest in the use of both txType cases in downstream tests.
+        // @ts-ignore
```
Siegrift, line 34: It is better to use `{} as any`

packages/airnode-validator/templates/0.3/config.json
```diff
@@ -61,6 +61,52 @@
           "__message": "__fullPath: Currently only \"evm\" is supported"
         }
       },
+      "chainOptions": {
```
Siegrift, line 64: This is not correct. You should not update the `0.3` templates, but `0.4`.
This means you need to copy all of the templates in `0.3` first and then update them in the new `0.4` directory.

Siegrift: I recommend doing this in a separate PR (since there will be many changes because of the copying)

Siegrift (changes requested):
LGTM, mostly found minor readability issues.

packages/airnode-examples/integrations/authenticated-coinmarketcap/config.example.json
aquarat: Good idea, I'll do this.

packages/airnode-deployer/config/config.json.example
```diff
@@ -11,7 +11,12 @@
           "url": "${PROVIDER_URL}"
         }
       },
-      "type": "evm"
+      "type": "evm",
+            "chainOptions": {
```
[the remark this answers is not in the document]
aquarat: Thank you :), I also came across it while implementing the renaming.

packages/airnode-node/src/constants.ts
```diff
@@ -39,11 +39,14 @@ export const WORKER_PROVIDER_PROCESS_REQUESTS_TIMEOUT = 32_000;
...
 export const PRIORITY_FEE = '3120000000';
 
 // The Base Fee to Max Fee multiplier
-export const BASE_FEE_MULTIPLIER = 2;
+export const BASE_FEE_MULTIPLIER = '2';
 
-// The maximum number of requests for a single sponsor wallet in a single Airnode run cycle
-export const MAXIMUM_SPONSOR_WALLET_REQUESTS = 5;
+// The number of Wei in a Gwei
+export const WEI_PER_GWEI = '1000000000';
```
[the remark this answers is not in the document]
aquarat: I thought it would be better to be consistent in number handling as they all go into BigNumber to avoid range issues.

packages/airnode-node/src/evm/gas-prices.test.ts
```diff
@@ -20,124 +20,157 @@ mockEthers({
 import { BigNumber, ethers } from 'ethers';
 import * as gasPrices from './gas-prices';
 import { BASE_FEE_MULTIPLIER, PRIORITY_FEE } from '../constants';
+import { ChainOptions } from '../types';
 
-describe('getGasPrice', () => {
-  const baseOptions = {
-    provider: new ethers.providers.JsonRpcProvider(),
-  };
+const makeLegacyBaseOptions = () => ({
```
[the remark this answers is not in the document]
aquarat: Noted, thank you

packages/airnode-validator/templates/0.3/config.json
aquarat: Okay, thanks, I'll do this now before any other actions.

packages/airnode-node/src/evm/handlers/process-transactions.test.ts
```diff
@@ -18,114 +19,184 @@ mockEthers({
...
+      const chainOptions = { txType } as ChainOptions;
+      const chains = initialConfig.chains.map((chain) => ({
+        ...chain,
+        chainOptions,
+      }));
+      const config = {
+        ...initialConfig,
+        chains,
+      } as Config;
+      jest.spyOn(fs, 'readFileSync').mockReturnValue(JSON.stringify(config));
```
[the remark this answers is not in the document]
aquarat: haha, wow, yeah, the lack of logging caused me a lot of problems debugging. I eventually wrote a file logger and then `tail -f`'ed the output to see what was going on.

packages/airnode-node/src/evm/handlers/process-transactions.test.ts
aquarat: I don't remember writing this, I think it was part of the test originally, but I will work on it.

packages/airnode-node/src/evm/gas-prices.ts
```diff
@@ -1,47 +1,53 @@
 import { BigNumber, ethers } from 'ethers';
+import { parseEther } from 'ethers/lib/utils';
```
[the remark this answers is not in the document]
andreogle: Agreed, it should be merged with the import above (or even just used as `ethers.utils.parseEther`)

packages/airnode-node/src/evm/gas-prices.ts
```diff
@@ -1,47 +1,53 @@
...
-  }
-
-  // Fallback to pre-EIP-1559
-  const operation = () => provider.getGasPrice();
-  const [err, gasPrice] = await go(operation, { retries: 1, timeoutMs: DEFAULT_RETRY_TIMEOUT_MS });
-  if (err || !gasPrice) {
-    const log = logger.pend('ERROR', 'Failed to get fallback gas price from provider', err);
-    return [[...logs, log], null];
+    case '2': {
+      const operation = () => provider.getBlock('latest');
```
andreogle, line 28: Do we maybe want to extract each case into smaller, descriptive functions?

packages/airnode-node/src/evm/gas-prices.test.ts
```diff
@@ -20,124 +20,157 @@ mockEthers({
...
-    getBlock.mockResolvedValueOnce({});
 
     const getGasPrice = baseOptions.provider.getGasPrice as jest.Mock;
     getGasPrice.mockRejectedValueOnce(new Error('Server is down'));
     getGasPrice.mockResolvedValueOnce(testGasPrice);
 
     const [logs, gasPrice] = await gasPrices.getGasPrice(baseOptions);
 
-    expect(logs.length).toEqual(1);
+    expect(logs.length).toEqual(0);
```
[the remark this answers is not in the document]
aquarat: We will now ;)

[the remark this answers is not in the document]
Siegrift: Actually should we? This happens in the retry operation helper. Maybe raise this in slack for further discussion.

packages/airnode-node/src/evm/gas-prices.ts
aquarat: Re: `parseEther`: this is about to be invalidated due to implementation of Burak's suggestion around a config object for `priorityFee` that will have value and unit fields (which underneath with use a different function).

Re: the cases as separate functions: I'll do that now.

packages/airnode-node/src/evm/handlers/process-transactions.test.ts
```diff
@@ -18,114 +19,184 @@ mockEthers({
...
+      });
+
+      const state = {
+        ...initialState,
+        config,
+        settings: {
+          ...initialState.settings,
+          chainOptions,
+        },
+      } as ProviderState<EVMProviderState>;
```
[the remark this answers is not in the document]
aquarat: This makes it unhappy (it becomes a problem in the call to `processTransactions`.

packages/airnode-node/src/evm/handlers/process-transactions.test.ts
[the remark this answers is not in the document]
aquarat: I checked further, I don't think it's because of this spy; the yarn test command sets an `env` of `SILENCE_LOGGER` which then causes an early return from the `log` function in the logger.

In `packages/airnode-node/src/logger.ts` line 45 onwards:
```
export function log(level: LogLevel, message: string, options: LogOptions) {
  if (process.env.SILENCE_LOGGER) {
    return;
  }
```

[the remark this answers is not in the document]
Siegrift: Well, when you debug you run the tests with `SILENCE_LOGGER` disabled.

packages/airnode-node/src/evm/gas-prices.test.ts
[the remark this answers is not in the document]
aquarat: I only just saw this for some reason. I have implemented this, if it's a problem I can revert.

packages/airnode-node/src/evm/gas-prices.ts
aquarat: Done :)

packages/airnode-node/src/evm/gas-prices.ts
```diff
@@ -11,43 +11,79 @@ export interface FetchOptions {
 
 export const parsePriorityFee = ({ value, unit }: PriorityFee) => ethers.utils.parseUnits(value, unit ? unit : 'wei');
 
-export const getGasPrice = async (options: FetchOptions): Promise<LogsData<GasTarget | null>> => {
+const getLegacyGasPrice = async (options: FetchOptions): Promise<LogsData<GasTarget | null>> => {
+  const { provider } = options;
+  const logs = Array<PendingLog>();
+
+  const operation = (): Promise<BigNumber> =>
```
aquarat, line 18: @Siegrift This is my take on logging the underlying errors without modifying `go`.

packages/airnode-node/src/constants.ts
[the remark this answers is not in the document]
Siegrift: I think it would be a bit better if they were numbers. Range issues is not an issue here.

packages/airnode-node/src/evm/handlers/process-transactions.test.ts
aquarat: This should be fixed now, replaced with a spy.

packages/airnode-node/src/evm/gas-prices.ts
Siegrift: I dislike this. We should either modify the `retryOperation` utility to log out unsuccessful attempts or not implement it at all.

I think not implementing it now is preferable.

packages/airnode-admin/test/setup/init/set-define-property.ts
```diff
@@ -2,15 +2,17 @@
...
-const { defineProperty } = Object;
-Object.defineProperty = function (object, name, meta) {
-  if (meta.get && !meta.configurable) {
-    // it might be an ES6 exports object
-    return defineProperty(object, name, {
-      ...meta,
-      configurable: true, // prevent freezing
-    });
-  }
+{
```
aquarat, line 5: In my IDE, if I ask it to "recompile" it indicates that `defineProperty` is redeclared. Adding some braces fixes it, but my IDE is possibly missing some configuration.

packages/airnode-node/src/evm/handlers/process-transactions.test.ts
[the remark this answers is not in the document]
Siegrift: It is unhappy because `txType` is string and not `'1' | '2'`.
If you do:
- test.each(['1', '2'] as const)... (add as const)
- async (txType) => ... (remove :string)
- do what is in the suggestion

it should work.

packages/airnode-node/src/evm/gas-prices.ts
aquarat: All good, I'll revert.

packages/airnode-admin/test/setup/init/set-define-property.ts
Siegrift: I don't think this should be included in this PR - it is likely an issue of your IDE.
If this is something that is broken - create a new PR for it

packages/airnode-node/src/evm/handlers/process-transactions.test.ts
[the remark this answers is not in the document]
aquarat: Thank you :), I've learnt something, obvious now :)

packages/airnode-admin/test/setup/init/set-define-property.ts
```diff
@@ -2,17 +2,15 @@
 // and use this function as a workaround at the top of your test.
 //
 // Credit: https://github.com/facebook/jest/issues/6914#issuecomment-654710111
-{
```
aquarat: Removed the braces/restored it.

packages/airnode-node/src/constants.test.ts
```diff
@@ -21,7 +21,7 @@ describe('constants', () => {
     expect(constants.WORKER_PROVIDER_INITIALIZATION_TIMEOUT).toEqual(17_000);
     expect(constants.WORKER_PROVIDER_PROCESS_REQUESTS_TIMEOUT).toEqual(32_000);
     expect(constants.MAXIMUM_SPONSOR_WALLET_REQUESTS).toBe(5);
-    expect(constants.PRIORITY_FEE).toEqual('3120000000');
-    expect(constants.BASE_FEE_MULTIPLIER).toEqual('2');
+    expect(constants.PRIORITY_FEE).toEqual(3120000000);
```
aquarat, line 24: Switched to Number.

packages/airnode-node/src/evm/gas-prices.ts
```diff
@@ -13,48 +13,24 @@ export const parsePriorityFee = ({ value, unit }: PriorityFee) => ethers.utils.p
...
-          resolve(result);
-        })
-        .catch((err) => {
-          logs.push(logger.pend('INFO', 'Failed attempting to get legacy gasPrice from provider', err));
-          reject(err);
-        });
-    });
 
-  const [err, gasPrice] = await go(operation, { retries: 1, timeoutMs: DEFAULT_RETRY_TIMEOUT_MS });
+  const [err, gasPrice] = await go(provider.getGasPrice, { retries: 1, timeoutMs: DEFAULT_RETRY_TIMEOUT_MS });
```
aquarat, line 17: Inlined function :)

packages/airnode-node/src/evm/handlers/process-transactions.test.ts
```diff
@@ -1,4 +1,4 @@
-import { mockEthers } from '../../../test/mock-utils';
+import { createAndMockGasTarget, mockEthers } from '../../../test/mock-utils';
```
aquarat, line 1: Hopefully this is in the right place?

.changeset/shaggy-ravens-grab.md
```diff
@@ -0,0 +1,7 @@
+---
+'@api3/airnode-examples': patch
```
Siegrift, line 2: Should be minor (you can edit the file directly).

.changeset/shaggy-ravens-grab.md
```diff
@@ -0,0 +1,7 @@
...
+'@api3/airnode-node': patch
+'@api3/airnode-validator': patch
+---
+
+Extended EIP-1559 implementation to be user-configurable
```
Siegrift, line 7:
```suggestion
Extended EIP-1559 implementation to be user configurable
```

packages/airnode-deployer/config/config.json.example
```diff
@@ -11,7 +11,15 @@
           "url": "${PROVIDER_URL}"
         }
       },
-      "type": "evm"
+      "type": "evm",
+      "options": {
+        "txType": "2",
```
Siegrift, line 16: This should be `legacy` or `eip-1559`

packages/airnode-node/src/evm/gas-prices.ts
```diff
@@ -1,47 +1,53 @@
...
-  const [logs, blockGas] = await (async (): Promise<[PendingLog[], GasTarget | null]> => {
-    const operation = () => provider.getBlock('latest');
-    const [err, blockHeader] = await go(operation, { retries: 1, timeoutMs: DEFAULT_RETRY_TIMEOUT_MS });
-    if (err || !blockHeader?.baseFeePerGas) {
-      const log = logger.pend('INFO', 'Failed to get EIP-1559 gas pricing from provider - trying fallback', err);
-
-      return [[log], null];
+  const { provider, chainOptions } = options;
+
+  switch (chainOptions.txType) {
```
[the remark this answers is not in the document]
Siegrift: Do you want to do that in a separate PR? 
I think a separate commit but this PR would be better.

packages/airnode-node/src/evm/gas-prices.ts
```diff
@@ -1,47 +1,53 @@
...
+
+      return [
+        [],
+        {
+          maxPriorityFeePerGas,
+          maxFeePerGas,
+        } as GasTarget,
+      ];
+    }
+    default:
```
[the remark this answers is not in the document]
Siegrift: This is not resolved still

packages/airnode-node/tsconfig.json
```diff
@@ -6,6 +6,6 @@
     // Enable at root tsconfig.json once all packages have been fixed
     "strict": true
   },
-  "include": ["./src/**/*", "./test/**/*"],
+  "include": ["./src/**/*", "./test/**/*", "../../node_modules/@types/jest/index.d.ts"],
```
Siegrift, line 9: Not sure about this change

packages/airnode-node/src/evm/gas-prices.ts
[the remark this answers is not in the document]
aquarat: I'll do so shortly :)

packages/airnode-node/tsconfig.json
aquarat: I'll remove

packages/airnode-deployer/config/config.json.example
aquarat: Will do in separate commit.

packages/airnode-node/src/evm/gas-prices.ts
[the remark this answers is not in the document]
aquarat: Removing the `default` case exposed a bug in two of the tests, so it was really good that you asked for this to be removed. The fix has made the `config.json` `fs` mock unnecessary (which was in the original tests). :)

.changeset/shaggy-ravens-grab.md
aquarat: I changed all three, I hope this is correct?

Siegrift: Yep.

packages/airnode-node/src/evm/gas-prices.ts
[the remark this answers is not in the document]
aquarat: Done :D

packages/airnode-deployer/config/config.json.example
aquarat: Done

packages/airnode-node/src/evm/gas-prices.ts
```diff
@@ -2,46 +2,63 @@ import { BigNumber, ethers } from 'ethers';
...
+import { ChainOptions, GasTarget, LogsData, PendingLog, PriorityFee } from '../types';
 
-interface FetchOptions {
+export interface FetchOptions {
   readonly provider: ethers.providers.JsonRpcProvider;
+  readonly chainOptions: ChainOptions;
 }
 
-export const getGasPrice = async (options: FetchOptions): Promise<LogsData<GasTarget | null>> => {
+export const parsePriorityFee = ({ value, unit }: PriorityFee) => ethers.utils.parseUnits(value, unit ? unit : 'wei');
```
Siegrift, line 12:
```suggestion
export const parsePriorityFee = ({ value, unit }: PriorityFee) => ethers.utils.parseUnits(value, unit ??  'wei');
```

Siegrift (approved):
:1st_place_medal: LGTM

packages/airnode-node/src/evm/gas-prices.ts
aquarat: !! I didn't know you could do that - awesome :)

PR #737, opened 2021-12-12
New generalized forwarder instead of the previous ACR proxy

bbenligiray:
The owner of the previous `AccessControlManagerProxy` contract could call ACR functions, but they couldn't perform other actions that a regular manager could have been able to, such as [whitelisting](https://github.com/api3dao/airnode/blob/2685f63673b36556315604fdc82246a6c7f369c0/packages/airnode-protocol/contracts/rrp/requesters/RrpBeaconServer.sol#L123).

This PR implements a general ownable contract that forwards a call and ETH value to a target contract. Oddly enough I couldn't find an implementation that we can reuse, they all use `delegatecall`, meaning that they are proxies of contracts rather than contract users.


Ashar2shahid:
If I am understanding this correctly previously the owner(DAO in this case) would initialize roles via the proxy contract. This means the manager is now the proxy contract address and since the proxy contract doesn't have the other functions (like whitelisting for example), the DAO is restricted to just whatever functions are available in the proxy contract?

bbenligiray:
> If I am understanding this correctly previously the owner(DAO in this case) would initialize roles via the proxy contract. This means the manager is now the proxy contract address and the DAO is restricted to just whatever functions are available in the proxy contract?

Yes, so the proxy owner couldn't whitelist addresses directly in the previous contract

> Correct me if I am wrong but in the current context does this mean that the manager for all the roles remain the same i.e the proxy and the owner of the proxy only changes. If that's the case I was wondering if there was a way to add a way to transfer roles that would only be done by the adminRole so there wouldn't be need of a proxy. Then again the manager will not have an adminRole so I guess that's not possible either

Either AccessControlRegistry shouldn't allow role transfers by default and we should implement an additional mechanism to allow that (which is this) or AccessControlRegistry should allow role transfers by default and we should implement an additional mechanism to disallow that. Implementing the former AccessControlRegistry is much easier and this forwarder contract is quite simple so that's why I went with this. The main downside I see with this is that it's not easy to use, the DAO will have to pass proposals with calldata as the parameter and the voters will need to verify that the calldata does what it should.

packages/airnode-protocol/contracts/utils/OwnableCallForwarder.sol
```diff
@@ -0,0 +1,35 @@
...
+    /// @param forwardedCalldata Calldata to be forwarded to the target address
+    /// @return returnedData Data returned by the forwarded call
+    function forwardCall(
+        address forwardTarget,
+        bytes calldata forwardedCalldata
+    ) external payable override onlyOwner returns (bytes memory returnedData) {
+        returnedData = Address.functionCallWithValue(
+            forwardTarget,
+            forwardedCalldata,
+            msg.value
```
acenolaza, line 32: Will this revert if the target is not `payable` or if the function name does not exist and there is no fallback function?

packages/airnode-protocol/contracts/utils/OwnableCallForwarder.sol
```diff
@@ -0,0 +1,35 @@
...
+import "./interfaces/IOwnableCallForwarder.sol";
+
+/// @title Contract that forwards the calls that its owner sends
+/// @dev AccessControlRegistry users that want their access control tables
+/// to be transferrable (e.g., a DAO) will use this forwarder instead of
+/// interacting with it directly. There are cases where this transferrability
+/// is not desired, e.g., if the user is an Airnode and is immutably associated
+/// with a single address, in which case the manager will interact with
+/// AccessControlRegistry directly.
+contract OwnableCallForwarder is Ownable, IOwnableCallForwarder {
```
Ashar2shahid, line 15: the Owner can renounce their ownership, do we want that to happen?

bbenligiray: Yes, I think this was talked about at the call. Say you want to deploy a `RequestAuthorizerWithManager`, deploy `AirnodeTokenLock` , give it the indefinite whitelister role, then renounce your ownership of the `OwnableCallForwarder` that owns `RequestAuthorizerWithManager`. This allows node operators to use the `RequestAuthorizerWithManager` without having to worry about the manager screwing them over. This is a decent solution to serving chains on which we don't have native decentralized governance (though it only works for authorizers, not beacon/dAPI servers).

packages/airnode-protocol/contracts/utils/OwnableCallForwarder.sol
bbenligiray: I extended the tests here https://github.com/api3dao/airnode/pull/748/commits/322e91b620da13fcd02b2667875c333a3c8e0855
If the target is payable and `msg.value` of the call made to `forwardCall()` is zero, the forward doesn't revert. If the target is payable and `msg.value` of the call made to `forwardCall()` is not zero, the forward does revert.
If the function doesn't exist and there is no fallback function, the forward does revert.

PR #738, opened 2021-12-13
Validate fulfillment data size at the beacon server

bbenligiray:
`data` is composed of statically-sized types so we can calculate its length beforehand and validate the incoming fulfillment


Ashar2shahid:
Nice change, atleast a revert string is available now in case the timestamp is not present

Ashar2shahid (approved):

PR #742, opened 2021-12-14
Move testable flag from OIS to triggers

Siegrift:
https://api3dao.atlassian.net/browse/AN-332


drgy:
Validator `0.4.0` template were already added in [this commit](https://github.com/api3dao/airnode/commit/de585e0f7097e1cbf7dffb76652d090ce977068e) on master, perhaps you can merge it first?

packages/airnode-node/src/types.ts
```diff
@@ -325,14 +325,17 @@ export type EVMEventLog =
 // ===========================================
 // Triggers
 // ===========================================
-export interface RrpTrigger {
+export interface BasicTrigger {
```
andreogle, line 328: Why not just call it `Trigger`?

packages/airnode-node/src/handlers/test-api.ts
```diff
@@ -17,20 +17,17 @@ export async function testApi(
 
   const logOptions = logger.buildBaseOptions(config, { requestId: testCallId });
 
-  const rrpTrigger = find(config.triggers.rrp, ['endpointId', endpointId]);
-  if (!rrpTrigger) {
+  const httpTrigger = find(config.triggers.http, ['endpointId', endpointId]);
+  if (!httpTrigger) {
     return [new Error(`No such endpoint with ID '${endpointId}'`), null];
```
andreogle, line 22: The phrasing on this error message is a bit weird to me. I would rather say something like:

```suggestion
    return [new Error(`Unable to find endpoint with ID:'${endpointId}'`), null];
```

I also tried to have messages/errors use colons when printing IDs/value, without spaces. Not in the scope of this PR but maybe we should look at consistency for errors/logs

andreogle (approved):

PR #744, opened 2021-12-15
Document available airnode examples

dcroote:
[AN-387](https://api3dao.atlassian.net/browse/AN-387)


packages/airnode-examples/README.md
```diff
@@ -16,12 +16,26 @@ but you may adapt the configuration to work for your target chain.
...
 
-If you would like to know more about RRP, read the [API3 docs](https://docs.api3.org/airnode/v0.2/concepts/).
+If you would like to know more about RRP, read the [API3 docs](https://docs.api3.org/airnode/v0.3/concepts/).
+
+## Available examples
+
+We call examples "integrations" because they are integrated with some API. With each integration you are offered a
+choice of how you want to run the Airnode and which network to use, as mentioned above. The integrations have been
+designed to highlight various Airnode functionality and use cases, from simple price requests to more complex
+authenticated requests encoding multiple reserved parameters. The following list orders integrations by increasing
```
Siegrift, line 32: Order by complexity is a subjective thing and will be hard to follow as the list grows in size. I think we can just use the folder names in aplhabetical order.

Siegrift: These folder names would ideally be links.

Siegrift: And ideally there would be a unit test that would verify that all of them are listed.

packages/airnode-examples/README.md
```diff
@@ -36,8 +50,7 @@ run the example with Rinkeby.
 
 ### 1. Choose an example
 
-First, you need to choose what example you want to run. We call these examples "integrations" because they are
```
Siegrift, line 39: 👍

Siegrift (changes requested):
Be sure to mention that you need to include the new integration in the root examples readme in https://github.com/api3dao/airnode/tree/master/packages/airnode-examples/integrations#adding-new-integration

packages/airnode-examples/README.md
Siegrift: > And ideally there would be a unit test that would verify that all of them are listed.

This is probably an overkill since this will be migrated in the docs.

dcroote: Thanks for the feedback. I made the requested updates:
- Changed to alphabetical
- Added links
- Mentioned integration addition instructions

Siegrift (approved):
LGTM

PR #746, opened 2021-12-17
Minor correction to validator specification for v0.4

aquarat:
- Corrected validator specification for v0.4 to make gas price fields optional
- Removed optional fields from one example to create a negative test case for the future


Siegrift (approved):
LGTM

PR #747, opened 2021-12-17
Move requestId and templateId verification to callApi handler

Siegrift:
https://api3dao.atlassian.net/browse/AN-393

Verifying the request ID and template ID is computationally expensive task. We should move these checks to `callApi` handler because this handler is spawned for each sponsor wallet (currently `processProvider` verifies these checks in the same handler for all sponsor wallets).


packages/airnode-node/src/handlers/call-api.ts
```diff
@@ -81,28 +81,79 @@ export interface CallApiPayload {
...
+}
+
+function verifyRequestId(payload: CallApiPayload): LogsData<ApiCallErrorResponse> | null {
+  const { aggregatedApiCall } = payload;
+  if (aggregatedApiCall.type === 'testing-gateway') return null;
+
+  if (isValidRequestId(aggregatedApiCall)) return null;
+
+  const message = `${RequestErrorMessage.RequestIdInvalid}. Request ID:${aggregatedApiCall.id}`;
+  const log = logger.pend('ERROR', message);
```
Siegrift, line 111: Logging at the call place is better then propagating the error and logging elsewhere since it's easier to debug (you just find out the error string and you are in the correct place in the code which caused the failure).

However, we return a response as well which is needed for the output of testing gateway.

Siegrift (commented):
I'll refactor the code and resolve TODOs in a separate PR(s)

packages/airnode-node/src/handlers/call-api.ts
andreogle: Yeah I also don't like propagating the logs (especially across worker functions), but it was a trade-off to not have to pass in the entire context to every function

andreogle (approved):

packages/airnode-node/src/handlers/call-api.ts
Siegrift: We actually mean a different thing (I think). I think the pending logs are reasonable and they are still logged at the call site (they are just printed by some other function). What I dislike is this pattern:

```js
const fnA = () => {
try{
const x = someFnCall()
...
return x
} catch (e) {
return null
}
}

const fnB = () => {
const res = fnA()
// The error happens in fnA, but the message is printed by fnB
if (!res) logger.error("An error happened...")
}
```

packages/airnode-node/src/evm/verification/api-call-verification.ts
```diff
@@ -1,47 +1,60 @@
+// TODO: Merge this file with request-verification.ts
 import { ethers } from 'ethers';
-import flatMap from 'lodash/flatMap';
-import * as logger from '../../logger';
-import { ApiCall, Request, LogsData, RequestErrorMessage, RequestStatus } from '../../types';
-
+import { RegularAggregatedApiCall } from '../..';
```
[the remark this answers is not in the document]
Siegrift: Sure - I need to be more cautions with the imports since VS code is not doing this properly for us.

packages/airnode-node/src/handlers/call-api.ts
andreogle: Ah right, my mistake. Yeah I agree with you

PR #749, opened 2021-12-19
Choose validator template version based on the node/deployer version running it

drgy:
https://api3dao.atlassian.net/browse/AN-378

### DIFF
--- a/packages/airnode-deployer/src/cli/commands.ts
+++ b/packages/airnode-deployer/src/cli/commands.ts
@@ -14,10 +14,10 @@ import {
 } from '../utils';
 import * as logger from '../utils/logger';
 
-export async function deploy(configFile: string, secretsFile: string, receiptFile: string, skipVersionCheck: boolean) {
+export async function deploy(configFile: string, secretsFile: string, receiptFile: string, skipValidation: boolean) {
   const secrets = parseSecretsFile(secretsFile);
   const config = nodeConfig.parseConfig(configFile, secrets);
-  validateConfig(config, getNodeVersion(), skipVersionCheck);
+  validateConfig(config, getNodeVersion(), skipValidation);
 
   const mnemonic = config.nodeSettings.airnodeWalletMnemonic;
   if (!validateMnemonic(mnemonic)) {
@@ -69,7 +69,7 @@ export async function remove(airnodeAddressShort: string, stage: string, cloudPr
 }
 
 export async function removeWithReceipt(receiptFilename: string) {
-  const receipt = parseReceiptFile(receiptFilename);
+  const receipt = parseReceiptFile(receiptFilename, getNodeVersion());
   const { airnodeAddressShort, cloudProvider, stage } = receipt.deployment;
   try {
     await remove(airnodeAddressShort, stage, cloudProvider);
--- a/packages/airnode-deployer/src/cli/commands.ts
+++ b/packages/airnode-deployer/src/cli/commands.ts
@@ -14,10 +14,10 @@ import {
 } from '../utils';
 import * as logger from '../utils/logger';
 
-export async function deploy(configFile: string, secretsFile: string, receiptFile: string, skipValidation: boolean) {
+export async function deploy(configFile: string, secretsFile: string, receiptFile: string) {
   const secrets = parseSecretsFile(secretsFile);
   const config = nodeConfig.parseConfig(configFile, secrets);
-  validateConfig(config, getNodeVersion(), skipValidation);
+  validateConfig(config, getNodeVersion());
 
   const mnemonic = config.nodeSettings.airnodeWalletMnemonic;
   if (!validateMnemonic(mnemonic)) {
--- a/packages/airnode-deployer/src/cli/commands.ts
+++ b/packages/airnode-deployer/src/cli/commands.ts
@@ -1,7 +1,7 @@
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
-import { config as nodeConfig, CloudProvider, version as getNodeVersion } from '@api3/airnode-node';
+import { CloudProvider, version as getNodeVersion } from '@api3/airnode-node';
 import { deployAirnode, removeAirnode } from '../infrastructure';
 import {
   deriveAirnodeAddress,
@@ -9,15 +9,14 @@ import {
   parseReceiptFile,
   parseSecretsFile,
   shortenAirnodeAddress,
-  validateConfig,
   validateMnemonic,
+  parseConfig,
 } from '../utils';
 import * as logger from '../utils/logger';
 
 export async function deploy(configFile: string, secretsFile: string, receiptFile: string) {
   const secrets = parseSecretsFile(secretsFile);
-  const config = nodeConfig.parseConfig(configFile, secrets);
-  validateConfig(config, getNodeVersion());
+  const config = parseConfig(configFile, secrets);
 
   const mnemonic = config.nodeSettings.airnodeWalletMnemonic;
   if (!validateMnemonic(mnemonic)) {
--- a/packages/airnode-deployer/src/cli/commands.ts
+++ b/packages/airnode-deployer/src/cli/commands.ts
@@ -1,7 +1,7 @@
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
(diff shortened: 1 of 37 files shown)


packages/airnode-deployer/src/utils/validation.ts
```diff
@@ -2,22 +2,31 @@ import { Config } from '@api3/airnode-node';
...
 import * as logger from '../utils/logger';
 
-export function validateConfig(config: Config, nodeVersion: string, skipVersionCheck: boolean) {
-  if (!skipVersionCheck && nodeVersion !== config.nodeSettings.nodeVersion) {
-    logger.fail(
-      `nodeVersion under nodeSettings in config.json is ${config.nodeSettings.nodeVersion} while the deployer node version is ${nodeVersion}`
-    );
-    throw new Error("Node version specified in config.json does not match the deployer's version");
+export function validateConfig(config: Config, nodeVersion: string) {
+  if (config.nodeSettings.skipValidation) {
```
acenolaza, line 6: Since `skipValidation` is optional I'm guessing that when it's not present in config.json then it means that we don't want to skip config validation.

packages/airnode-deployer/src/utils/validation.ts
```diff
@@ -2,22 +2,31 @@ import { Config } from '@api3/airnode-node';
...
 
   if (config.nodeSettings.cloudProvider.type === 'local') {
     const message = "Deployer can't deploy to 'local' cloud provider";
     logger.fail(message);
     throw new Error(message);
   }
+
+  const result = validateJsonWithTemplate(config, `config@${nodeVersion}`);
+
+  if (!result.valid) {
```
acenolaza, line 19: shouldn't this if sentence follow the same logic as in [node](https://github.com/api3dao/airnode/blob/b79b5f603ad49c4ff3b1a3a5f2b3389e5c85746b/packages/airnode-node/src/config/index.ts#L30) package?

acenolaza (approved):
LGTM 🏅 
Couldn't find any major issues and I just left some comments/questions. Also please remember to update README in deployer if it was making a reference to using `skipValidation` command arg.

packages/airnode-deployer/src/utils/validation.ts
drgy: Yes, `skipValidation` is optional, so this works as well

packages/airnode-deployer/src/utils/validation.ts
drgy: It does, its only split into 2 conditions so that warning is issued in deployer informing user about skipped validation step

drgy:
@acenolaza thanks for the review, I've answered your comments, however I didn't find appropriate place where to mention `skipValidation` in the readme, perhaps it would be better if its documented in the API3 docs wdyt?

acenolaza:
> @acenolaza thanks for the review, I've answered your comments, however I didn't find appropriate place where to mention `skipValidation` in the readme, perhaps it would be better if its documented in the API3 docs wdyt?

Sounds good 👍🏻

packages/airnode-deployer/src/utils/files.ts
```diff
@@ -60,11 +61,47 @@ export function parseReceiptFile(receiptFilename: string) {
...
+  if (config.nodeSettings.skipValidation) {
+    logger.warn('Skipping the config validation');
+    return config;
+  }
+
+  if (config.nodeSettings.cloudProvider.type === 'local') {
+    const message = "Deployer can't deploy to 'local' cloud provider";
+    logger.fail(message);
+    throw new Error(message);
+  }
```
amarthadan, line 95: Skipping the validation is mainly useful for development when you want to try new features regardless of the current released version. But I don't think skipping this cloud provider test is useful in any way. I would keep this test even when the `skipValidation` flag is set.

packages/airnode-deployer/src/cli/commands.ts
```diff
@@ -69,7 +68,7 @@ export async function remove(airnodeAddressShort: string, stage: string, cloudPr
 }
 
 export async function removeWithReceipt(receiptFilename: string) {
-  const receipt = parseReceiptFile(receiptFilename);
+  const receipt = parseReceiptFile(receiptFilename, getNodeVersion());
```
amarthadan, line 71: You're calling `getNodeVersion` inside `parseConfig` but passing the value when calling `parseReceiptFile`. I think it would be better to choose one approach and stick to it if possible.

packages/airnode-deployer/src/utils/files.ts
```diff
@@ -60,11 +61,47 @@ export function parseReceiptFile(receiptFilename: string) {
...
+  const validationResult = validateReceipt(receipt, nodeVersion);
   if (!validationResult.valid) {
     logger.fail('Failed to validate receipt file');
     throw new Error(`Invalid Airnode receipt file: ${JSON.stringify(validationResult.messages)}`);
   }
 
   return receipt as Receipt;
 }
+
+export function parseConfig(configPath: string, secrets: Record<string, string>): Config {
```
amarthadan, line 73: I don't really get what's the reason for having two separate configuration parsing functions (one in `airnode-deployer` and one in `airnode-node`). Can you elaborate on why is that?

amarthadan (changes requested):

packages/airnode-validator/templates/0.2/receipt.json
```diff
@@ -51,7 +51,8 @@
     "nodeVersion": {
       "__regexp": "^0\\.2\\.[0-9]+$",
       "__catch": {
-        "__level": "warning"
+        "__level": "error",
```
Siegrift, line 54: Are we sure we want to change this to error now?

Siegrift:
Please add a changeset.

packages/airnode-deployer/src/utils/files.ts
Siegrift: I think it is because deployer also need to check for the cloud provider type, which is not checked in node. HOwever this check can be moved out of the parseConfig function. I'll implement it that way.

packages/airnode-validator/templates/0.2/receipt.json
Siegrift: I will revert this from the versions below 0.4 as I do not see why they should be changed.

packages/airnode-node/src/config/index.ts
```diff
@@ -23,23 +16,31 @@ function parseOises(oises: OIS[]): OIS[] {
...
+  const parsedConfig: Config = validationResult.specs as Config;
   const ois = parseOises(parsedConfig.ois);
-  return { ...parsedConfig, ois };
+
+  return {
+    config: { ...parsedConfig, ois },
+    shouldSkipValidation: !!parsedConfig.nodeSettings.skipValidation,
+    validationOutput: validationResult,
+  };
 }
```
amarthadan, line 44: I know that this way we can do whatever with the results (throw, use different loggers, etc.) but I wouldn't be against doing that within this function. But I guess that would be a bigger refactor. Just a thought...

Siegrift: The problem is that deployer wants to log it and node would like to do that as well, but for now doesn't.

Fundamentally, they use a different type of loggers (and they probably always will) so it's hard to do and actually I'd say out of scope of this function.

packages/airnode-deployer/src/utils/validation.ts
```diff
@@ -2,22 +2,31 @@ import { Config } from '@api3/airnode-node';
 import { validateJsonWithTemplate } from '@api3/airnode-validator';
 import * as logger from '../utils/logger';
 
-export function validateConfig(config: Config, nodeVersion: string, skipVersionCheck: boolean) {
-  if (!skipVersionCheck && nodeVersion !== config.nodeSettings.nodeVersion) {
-    logger.fail(
-      `nodeVersion under nodeSettings in config.json is ${config.nodeSettings.nodeVersion} while the deployer node version is ${nodeVersion}`
-    );
-    throw new Error("Node version specified in config.json does not match the deployer's version");
+export function validateConfig(config: Config, nodeVersion: string) {
```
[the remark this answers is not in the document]
Siegrift: Yeah. I removed it. Probably the file is now too small and could be merged with the "utils" file - but I think we can leave that for the future.

amarthadan (approved):

PR #750, opened 2021-12-20
AN-441: Fix substitution functions

drgy:
https://api3dao.atlassian.net/browse/AN-441


andreogle (approved):
It would be nice to have some unit tests for the functions defined in this file, but looks good to me otherwise

PR #751, opened 2021-12-25
Update OIS version in validator templates to 1.0.0

drgy:
https://api3dao.atlassian.net/browse/AN-456


Siegrift:
Please add a changeset.

packages/airnode-validator/src/validators/validators.test.ts
```diff
@@ -195,9 +195,7 @@ describe('validators integration test', () => {
         keyFormattingMessage('myPath', ['paths', 'myPath']),
         error('paths.myPath.get : allowed methods are only "get" or "post" not "get "'),
         error("Allowed values in paths.myPath.get .parameters.[0].in are: 'path', 'query', 'header' or 'cookie'"),
-        keyFormattingMessage('mySecurityScheme ', ['components', 'securitySchemes', 'mySecurityScheme ']),
```
Siegrift, line 198: These changes in the test are because of the OIS being updated from pre-alpha to v1.0, right?

packages/airnode-validator/src/validator.test.ts
```diff
@@ -3,13 +3,13 @@ import fs from 'fs';
...
 import { getPath } from './commands/utils';
-import { error } from './utils/logger';
+import { error, warn } from './utils/logger';
 import { Log } from './types';
 
 const messages: Log[] = [];
 
-const endpointsPath = getPath('endpoints.json', messages, 'pre-alpha');
-const oisPath = getPath('ois.json', messages, 'pre-alpha');
+const endpointsPath = getPath('endpoints.json', messages, '1.0');
```
Siegrift, line 11: How will this look like when we release a new version of OIS? Is there a ticket for it?

Siegrift (approved):
:+1: LGTM, just a few questions.

packages/airnode-validator/src/validator.test.ts
amarthadan: I don't think there's any plan for how to deal with this as of now. And I don't know about the ticket either.

packages/airnode-validator/src/validators/validators.test.ts
amarthadan: Yes, I would say so.

amarthadan (approved):

packages/airnode-validator/src/validator.test.ts
Siegrift: Tracked in https://api3dao.atlassian.net/browse/AN-465

PR #753, opened 2021-12-31
Moving the manual build instructions back to the README.

wkande:
As per the discussion in AN-407 the monorepo package should now retain any manual build instructions. The docs will have the content regarding things such as npx and Docker image usage. The manual build instructions have already been removed (see AN-406) and the reference to a manual build now is linked back to the monorepo README.

### DIFF
--- a/packages/airnode-validator/README.md
+++ b/packages/airnode-validator/README.md
@@ -7,7 +7,32 @@ them.
 ## Usage
 
 General usage of validator and convertor is described in the
-[API3 Documentation](https://docs.api3.org/airnode/latest/reference/packages/validator.html). For debugging purposes it
+[API3 Documentation](https://docs.api3.org/airnode/latest/reference/packages/validator.html). 
+
+## Build Manually
+
+You can clone and build the Airnode monorepo then run the validator as a yarn
+script from inside the `packages/airnode-validator` directory.
+
+```sh
+# download and build the airnode monorepo
+git clone git@github.com:api3dao/airnode.git
+cd airnode
+yarn run bootstrap
+yarn run build
+cd packages/airnode-validator
+
+# execute the validator
+yarn run cli:validator --template="config" --specs="exampleSpecs/config.json"
+
+# validator output
+{
+  "valid": true,
+  "messages": []
+}
+```
+
+For debugging purposes it
 might be useful to run the validator with path to template file instead of providing the format name:
 
 ```sh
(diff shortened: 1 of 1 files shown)


dcroote:
@wkande - looks like the CI failed because of Prettier

packages/airnode-validator/README.md
```diff
@@ -32,7 +57,7 @@ specification conversion and validation. Usage of these keywords is described in
 - [optional parameters](docs/optional.md)
 - [nested templates](docs/template.md)
 
-## Documentation
+## Other Documentation
```
drgy, line 60: Perhaps this should be in the monorepo docs only as well, as its instructions for generating the documentation

drgy (approved):

packages/airnode-validator/README.md
wkande: @drgy The section you mention is in the monorepo as you have indicated. Perhaps a better name would be **Source Documentation**? That paragraph is not in the Docs.

drgy: @wkande 
> The section you mention is in the monorepo as you have indicated

Sorry my bad, I mixed up the repos.

> Do you know why the CI prettier failed?

Try running: `yarn run prettier:write`

wkande: @drgy 
Sorry I am still confused. Why would we move the section called **Documentation** (in the monorepo) to the Docs? 
Is it not related to template generation? Should it be called **Template Documentation**? If we do move them then why are we not moving the section named **Templates** as well?

drgy: @wkande I apologize for making things even more confusing. I thought this was the docs repo, I suggested to move the docs generation from api3 docs to airnode readme, after that I noticed that this is the airnode repository, so everything is as it should be, nothing should change, please ignore my suggestion.

wkande: @drgy 
Ok, then I guess this PR can be merged? I do think that the name **Other Documentation** heading that I had added might be better to rename **Template Documentation**. I can change it if you want. Also I will leave the merge to you since this is your project.

packages/airnode-validator/README.md
```diff
@@ -7,8 +7,33 @@ them.
...
+[API3 Documentation](https://docs.api3.org/airnode/latest/reference/packages/validator.html).
+
+## Build Manually
+
+You can clone and build the Airnode monorepo then run the validator as a yarn script from inside the
+`packages/airnode-validator` directory.
+
+```sh
+# download and build the airnode monorepo
+git clone git@github.com:api3dao/airnode.git
```
Siegrift, line 19: I think this shouldn't be there, because adding it here implies adding it to every airnode package which is repetitive. These instructions are already part of https://github.com/api3dao/airnode#instructions

drgy: I agree, opened a [new PR](https://github.com/api3dao/airnode/pull/755#issuecomment-1004054627) for this

PR #756, opened 2022-01-04
Parallelize e2e tests using matrix strategy

dcroote:
Part of AN-462, "Investigate how to improve the CI reliability/speed"

This PR splits the longest step `yarn run test:e2e` into 3 parallel matrix jobs, one for each of the packages that have `e2e` tests. I re-ran the Action 3 times ([1](https://github.com/api3dao/airnode/actions/runs/1651632365/attempts/1), [2](https://github.com/api3dao/airnode/actions/runs/1651632365/attempts/2), [3](https://github.com/api3dao/airnode/actions/runs/1651632365)) and compared the total time to the previous 10 Action runs on master [minutes]:

image


Siegrift (approved):
:+1: LGTM

The only fear I have is when we add e2e tests for a package we might miss updating build-test.yml. Any idea how to prevent that?

dcroote:
That is a genuine potential downside, but if someone is adding e2e tests for a new package I find it reasonable to expect they'll check the CI to confirm their added tests work. In that case they'll see the e2e matrix job doesn't include their new tests and should add the new package to the build matrix array.

I suppose there could be another test that looks in packages for `test:e2e` and confirms that all packages that have it are accounted for in the build matrix, but that sounds like a fair amount of work for this hypothetical.

Siegrift:
It is hard to spot sometimes (https://github.com/api3dao/airnode/pull/749/commits/f36c1cc3c16de01548c7d0cc4f418f897e15485a#diff-0fab668cedab8990ae29c18bfe48e39224faa6071f154cb0d3ba8c28e8643961R28) but I guess someone would notice this at some point...

Feel free to merge.

PR #759, opened 2022-01-06
Default to released images in airnode-examples

dcroote:
Rather than rebuilding containers, which takes a long time and, at least on my machine, can often timeout or have connection issues, we should default to released images while still providing developers instructions for building the images from source should they so choose.

[AN-470](https://api3dao.atlassian.net/browse/AN-470?atlOrigin=eyJpIjoiYjRhOGMwYjE0ZGZkNGEyMTg5NGI3NWNlZWFkOGRiYzkiLCJwIjoiaiJ9)

### DIFF
--- a/packages/airnode-examples/README.md
+++ b/packages/airnode-examples/README.md
@@ -152,26 +152,7 @@ Refer to the
 [documentation](https://docs.api3.org/airnode/v0.2/grp-providers/guides/build-an-airnode/configuring-airnode.html) for
 more details.
 
-### 9. Build docker artifacts
-
-Our docker images are based on a common container which we call "artifacts". This intermediate container is then used by
-both [deployer](https://github.com/api3dao/airnode/tree/master/packages/airnode-deployer) and
-[airnode](https://github.com/api3dao/airnode/tree/master/packages/airnode-node). You can build the artifacts container
-by running:
-
-```sh
-yarn rebuild-artifacts-container
-```
-
-### 10. (Only if deploying to a cloud provider) Build deployer container
-
-```sh
-yarn rebuild-deployer-container
-```
-
-This command will facilitate the previously built artifacts container to build the deployer.
-
-### 11. (Only if deploying to a cloud provider) Deploy Airnode
+### 9. (Only if deploying to a cloud provider) Deploy Airnode
 
 Now you're ready to deploy Airnode on the cloud provider. Just run:
 
@@ -182,16 +163,7 @@ yarn deploy-airnode
 This command will use the [deployer](https://github.com/api3dao/airnode/tree/master/packages/airnode-deployer) package
 to deploy your Airnode. Deployment may take some time so be patient.
 
-### 12. (Only if running Airnode locally) Build Airnode docker container
-
-```sh
-yarn rebuild-airnode-container
-```
-
-This command will facilitate the previously built artifacts container to build the containerized version of Airnode
-which you can run locally.
-
-### 13. (Only if running Airnode locally) Run the Airnode container
+### 10. (Only if running Airnode locally) Run the Airnode container
 
 ```sh
 yarn run-airnode-locally
@@ -200,7 +172,7 @@ yarn run-airnode-locally
 Runs the previously built version of Airnode container. Note that the containerized version runs a cron job which
 triggers every minute - this means that Airnode logs won't start appearing immediately.
 
-### 14. Deploy a requester
+### 11. Deploy a requester
 
 At this point, you have a RRP contract deployed. You also either have a docker running locally or deployed on a cloud
 provider. Airnode is now listening on the events (requests to be made) of the RRP contract. All that is left now, is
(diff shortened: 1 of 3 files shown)


packages/airnode-examples/scripts/deploy-airnode.ts
```diff
@@ -17,7 +17,7 @@ const main = async () => {
     integrationInfo.airnodeType === 'gcp' && `-v "\${HOME}/.config/gcloud:/app/gcloud"`,
     `-v ${integrationPath}:/app/config`,
     `-v ${integrationPath}:/app/output`,
-    `api3/airnode-deployer:latest deploy --skip-version-check`,
+    `api3/airnode-deployer:0.3.1 deploy --skip-version-check`,
```
dcroote, line 20: It would be nice to not have this hardcoded...

packages/airnode-examples/scripts/run-airnode-locally.ts
```diff
@@ -10,7 +10,7 @@ const main = async () => {
 
   const integrationPath = join(__dirname, '../integrations', integrationInfo.integration);
   runShellCommand(
-    `docker run --rm -v ${integrationPath}:/app/config --network="host" --name airnode api3/airnode-client:latest`
+    `docker run --rm -v ${integrationPath}:/app/config --network="host" --name airnode api3/airnode-client:0.3.1`
```
dcroote, line 13: ...hardcoded here too

packages/airnode-examples/scripts/deploy-airnode.ts
dcroote: I realized there was an image with the `latest` tag that can be pulled from docker hub so `latest` can stay, no need for hardcoding versions.

packages/airnode-examples/README.md
```diff
@@ -152,26 +152,7 @@ Refer to the
 [documentation](https://docs.api3.org/airnode/v0.2/grp-providers/guides/build-an-airnode/configuring-airnode.html) for
 more details.
 
-### 9. Build docker artifacts
```
Siegrift, line 155: This is not right. There might be some changes in the deployer on master - so you must run it with latest docker (built from source). This is even a bigger issue for airnode-client - since you will be using outdated code.

Siegrift (changes requested):
I disagree with this change.

bbenligiray:
The examples are for users, not Airnode developers. All users should definitely use the released images for any non-production work, building images suck. The users should use the example projects by pulling the respective tag, they shouldn't ever be on master. I think this is PR improves things as long as this is made clear in the instructions.

Siegrift:
That's valid point, but instructions should mention that (that they should be on the correct tag and use such config.json + secrets.env). They should also mention that when on master branch - they should build docker images from source.

dcroote:
In working on adding clarity to checking out a tag and rebuilding on master for dev work, I ran into the following snag: If we ask users to checkout `v0.3.1` (or any previous tag) they're checking out a README that includes instructions asking them to rebuild the containers...

Siegrift:
You can explain why this is in parentheses? Something like: "If you are reading the docs for a specific version/tag - use the versioned docker images from dockerhub (do this even for version before 0.4, which do not have this information explicitly mentioned)"

Another solution would be to patch the READMEs of the released versions, which I do not think it's worth Buraks time.

dcroote:
If we go the user instruction route, unfortunately we'll also need to specify that after they check out the version tag, they need to update one or both of the following lines to the same version, otherwise they would pull the `latest` image(s) which may be a different version than their tag:

https://github.com/api3dao/airnode/blob/3c2937d573f48444e31954428e57262e703792ee/packages/airnode-examples/scripts/deploy-airnode.ts#L20

https://github.com/api3dao/airnode/blob/3c2937d573f48444e31954428e57262e703792ee/packages/airnode-examples/scripts/run-airnode-locally.ts#L13

Siegrift:
Yeah, that's true. But actually, we can just update this for the previous versions - we don't necessarily have to publish new patch (since this package will never be consumed and used as API).

dcroote:
So after additional consideration, this has more nuance than originally anticipated. As I mentioned above, if we ask users to git checkout `v0.3.1` (or any previous tag) they're checking out a README that includes instructions asking them to rebuild the containers and has `latest` for the docker image version in the commands I mentioned in the comment above. Fixing this retrospectively would require new releases of the previous versions, which is likely not worth the time and effort. Going forward we also have issues: right now we cannot change the version of the docker image command from `latest` to to `0.4.0` because it isn't released yet, however we also cannot leave it as `latest` because it will be wrong for the `0.4` tag when `0.5` gets released. I think the solution is to update the image tag in the scripts when releasing a version.

Siegrift:
I think we have agreed that the current approach is not that easy to fix (and actually impossible to solve in backwards compatible manner since previously released packages and git tags are frozen).

I've been thinking about this a bit more and come up with this: https://api3dao.atlassian.net/browse/AN-494

I am closing this PR for the already mentioned reasons.

PR #760, opened 2022-01-06
Force stage deployment variable to be lowercase

### LINKED ISSUE #757: Deploy to AWS fails with invalid strings in nodeSettings.stage
## Description

When the value of `nodeSettings.stage` in `config.json` is set to the word **Staging** the deployment fails. The deployer stops right after the _dynamoDB_ database is created as well as the _S3 bucket_. It appears that the use of an uppercase letter is the issue.

- dev: works
- Dev: fails
- Staging: fails
- my-bucket-test: works

## Version

Using npx v0.3.1

Change the nodeSettings as follows:
```
"nodeSettings": {
    "cloudProvider": {
      "type": "aws",
      "region": "us-east-1"
    },
    "airnodeWalletMnemonic": "${MNEMONIC}",
    "logFormat": "plain",
    "logLevel": "INFO",
    "nodeVersion": "0.3.1",
    "stage": "as-long-as-is-lowercase",
    "heartbeat": {
      "enabled": false
    },
    "httpGateway": {
      "enabled": true,
      "apiKey": "${HTTP_GATEWAY_API_KEY}"
    }
  },
```

## The Error

The Deployer says the S3 bucket does not exist but it does as I checked the AWS console and it was there.

```
ℹ Deploying Airnode abd9daa Dev to aws us-east-1
✖ Error: Command failed: terraform init -backend-config="region=us-east-1" -backend-config="bucket=airnode-abd9daa-Dev-terraform" -backend-config="dynamodb_table=airnode-abd9daa-Dev-terraform-lock" -from-module=/app/terraform/airnode/aws
╷
│ Error: Failed to get existing workspaces: S3 bucket does not exist.
│ 
│ The referenced S3 bucket must have been previously created. If the S3 bucket
│ was created within the last minute, please wait for a minute or two and try
│ again.
│ 
│ Error: NoSuchBucket: The specified bucket does not exist
│ 	status code: 404, request id: 13PWPE3DP85159DV, host id: xPGgmxL8e2cSvpHQeflnCorRuft1ESYtU6LKLW258uuuo7hoaCb8NDyPVXNSTSzSk/Jp6C2z8k0=
│ 
✖ Failed deploying Airnode abd9daa Dev to aws us-east-1
⚠ Failed deploying configuration, skipping
⚠ Error: Command failed: terraform init -backend-config="region=us-east-1" -backend-config="bucket=airnode-abd9daa-Dev-terraform" -backend-config="dynamodb_table=airnode-abd9daa-Dev-terraform-lock" -from-module=/app/terraform/airnode/aws
ℹ Outputted output/receipt.json
  This file does not contain any sensitive information.
```

## The Docs

stage
(required) - The label used to distinguish between multiple deployments of the same Airnode on a cloud provider. For example, the same Airnode may have multiple deployments with stage set to a different value (dev, public, prod). stage cannot be longer than 16 characters and can only include alphanumeric characters (a–z, A–Z, 0–9) and hyphens (-).

amarthadan:
Close #757

There was a difference between how the bucket name is handled by the state recipes and airnode recipes (one was lowercasing the name and the second one wasn't), that's why it was possible to create a bucket but not find it afterward.

I united it so the bucket name is always lowercase but I would also like to force `stage` to contain only lower case characters (numbers and underscore `_` are ok) just to be on the safe side. I've already modified the validator templates.

@wkande Can you update the docs to reflect that?

### DIFF
--- a/packages/airnode-deployer/src/infrastructure/index.ts
+++ b/packages/airnode-deployer/src/infrastructure/index.ts
@@ -135,6 +135,10 @@ function prepareAirnodeManageArguments(cloudProvider: CloudProvider, commonArgum
   return [...cloudProviderAirnodeManageArguments[cloudProvider.type](cloudProvider as any), ...commonArguments];
 }
 
+function getBucketName(airnodeAddressShort: string, stage: string) {
+  return `airnode-${airnodeAddressShort}-${stage}-terraform`.toLowerCase();
+}
+
 async function terraformAirnodeManage(
   command: string,
   execOptions: child.ExecOptions,
@@ -214,7 +218,7 @@ async function deploy(
   }
 
   const { type: cloudProviderType } = cloudProvider;
-  const bucket = `airnode-${airnodeAddressShort}-${stage}-terraform`;
+  const bucket = getBucketName(airnodeAddressShort, stage);
   const terraformStateCloudProviderDir = path.join(terraformStateDir, cloudProviderType);
 
   if (!(await cloudProviderLib[cloudProviderType].stateExists(bucket, cloudProvider as any))) {
@@ -269,7 +273,7 @@ async function remove(airnodeAddressShort: string, stage: string, cloudProvider:
   }
 
   const { type: cloudProviderType } = cloudProvider;
-  const bucket = `airnode-${airnodeAddressShort}-${stage}-terraform`;
+  const bucket = getBucketName(airnodeAddressShort, stage);
 
   // Remove airnode
   logger.debug('Removing Airnode via Terraform recipes');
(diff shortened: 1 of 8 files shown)


Siegrift:
I guess this deserves a changeset.

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -214,7 +214,7 @@ async function deploy(
   }
 
   const { type: cloudProviderType } = cloudProvider;
-  const bucket = `airnode-${airnodeAddressShort}-${stage}-terraform`;
+  const bucket = `airnode-${airnodeAddressShort}-${stage}-terraform`.toLowerCase();
```
Siegrift, line 217: Maybe extract to a helper fn?

Siegrift (approved):
LGTM.

PR #762, opened 2022-01-06
Fix validator template operationParameter regex

### LINKED ISSUE #758: Deployment fails because an endpoint has a param named project_type
## Description

Deployment is failing because 1 endpoint has a param named `project_type`. See the Slack thread [**here**](https://api3workspace.slack.com/archives/C01N6001YCE/p1641417651248800).
```
Error: Can't have a name of reserved parameter
```

The part `_type`  as a substring of `project_type` trips the error, even though `project_type` should be allowed. See here in the [Docs](https://docs.api3.org/airnode/v0.3/reference/specifications/ois.html#_5-4-reservedparameters). I cannot find (or recall) anything in the Docs that says `project_type` is a reserved string.

## Version

v0.3.1

## Expected Behaviour

Should allow a parameter to be named `project_type`.

## Steps to reproduce

Give an endpoint a warm name of `project_type` or maybe even `abc_type`.

amarthadan:
Close #758

### DIFF
--- a/.changeset/green-dryers-play.md
+++ b/.changeset/green-dryers-play.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-validator': patch
+---
+
+Fix validator template operationParameter regex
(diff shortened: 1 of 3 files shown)


Siegrift (approved):
👍 Do not forget to a patch changeset (finally an use case for patch :D )

andreogle (approved):

PR #763, opened 2022-01-06
Make airnode examples fail early, improve README

Siegrift:
I've tried running the examples on my new laptop (mac) and I came a cross issues:
a) the steps in examples do not fail early
b) I've forgot to change PROVIDER_URL (it's written in the docs, but could be more readable)

I've addressed both points.

### DIFF
--- a/packages/airnode-examples/README.md
+++ b/packages/airnode-examples/README.md
@@ -33,6 +33,9 @@ authenticated requests encoding multiple reserved parameters. The following list
 
 - [CoinGecko](https://github.com/api3dao/airnode/blob/master/packages/airnode-examples/integrations/coingecko) -
   unauthenticated cryptocurrency price request
+- [CoinGecko Testable](https://github.com/api3dao/airnode/tree/master/packages/airnode-examples/integrations/coingecko-testable) -
+  same price request as the CoinGecko example above. However, the endpoint is allowed to be tested using the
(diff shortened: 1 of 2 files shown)


packages/airnode-examples/README.md
```diff
@@ -143,9 +146,9 @@ step. The latter, `secrets.env` can be generated it using:
 yarn create-airnode-secrets
 ```
 
-> If you are not using docker for linux and you want to connect to your local hardhat network, you will need to modify
-> the generated `secrets.env` file found in `integrations/<integration-name>/` by replacing the provider URL with the
-> following: `PROVIDER_URL=http://host.docker.internal:8545`. This is a docker limitation. See:
+> If you are not using docker for Windows/WSL or docker for mac and you want to connect to your local hardhat network,
```
dcroote, line 149: I believe "not" should be removed based both on your comment and the fact that I have to do this for WSL myself
```suggestion
> If you are using docker for Windows/WSL or docker for mac and you want to connect to your local hardhat network,
```

dcroote (approved):
I tested the new fail-early changes and they worked 👍 
One minor comment, otherwise LGTM.

packages/airnode-examples/README.md
Siegrift: Yeah, thanks.

PR #765, opened 2022-01-09
Add per-chain request limits

Siegrift:
https://api3dao.atlassian.net/browse/AN-395

Review commit by commit (since I've implemented a few related refactors at the begining)

### DIFF
--- a/packages/airnode-deployer/config/config.json.example
+++ b/packages/airnode-deployer/config/config.json.example
@@ -1,6 +1,7 @@
 {
   "chains": [
     {
+      "maxConcurrentTransactions": 100,
       "authorizers": [],
       "contracts": {
         "AirnodeRrp": "0x534DfA0d242e7fDCFD096C2B1a3D70F172008778"
--- a/packages/airnode-deployer/config/config.json.example
+++ b/packages/airnode-deployer/config/config.json.example
@@ -1,7 +1,7 @@
 {
   "chains": [
     {
-      "maxConcurrentTransactions": 100,
+      "maxConcurrency": 100,
       "authorizers": [],
       "contracts": {
         "AirnodeRrp": "0x534DfA0d242e7fDCFD096C2B1a3D70F172008778"
--- a/packages/airnode-node/src/coordinator/calls/chain-limits.ts
+++ b/packages/airnode-node/src/coordinator/calls/chain-limits.ts
@@ -1,0 +1,78 @@
+import flatMap from 'lodash/flatMap';
+import groupBy from 'lodash/groupBy';
+import map from 'lodash/map';
+import zip from 'lodash/zip';
+import * as logger from '../../logger';
+import { sortRequests } from '../../requests';
+import {
+  Config,
+  ProviderStates,
+  Request,
+  GroupedRequests,
+  RequestType,
+  ApiCall,
+  Withdrawal,
+  PendingLog,
+} from '../../types';
+
+function groupRequests(flatRequests: Request<any>[]): GroupedRequests {
+  const apiCalls = flatRequests
+    .filter((request) => request.__type === RequestType.ApiCall)
+    .map((request) => request as Request<ApiCall>);
+
+  const withdrawals = flatRequests
+    .filter((request) => request.__type === RequestType.Withdrawal)
+    .map((request) => request as Request<Withdrawal>);
+
+  return { apiCalls, withdrawals };
+}
+
+export function applyChainLimits(config: Config, providerStates: ProviderStates) {
+  const logs: PendingLog[] = [];
+
+  const providerRequests = providerStates.evm
+    .map((evmState, index) => {
+      const { apiCalls, withdrawals } = evmState.requests;
+      return { index, chainId: evmState.settings.chainId, requests: sortRequests([...apiCalls, ...withdrawals]) };
+    })
+    // Sort by number of request ascendingly
+    .sort((p1, p2) => p1.requests.length - p2.requests.length);
+  const groupedProviderRequests = groupBy(providerRequests, 'chainId');
+  const chainRequests = map(groupedProviderRequests, (group, chainId) => ({
+    chainId,
+    chainLimit: config.chains.find((chain) => chain.id === chainId)!.maxConcurrentTransactions,
+    requestsGroups: group,
+  }));
+
+  // Apply the chain request limit by dropping requests which exceed the limit. We do this mutably to increase
+  // coordinator performance
+  chainRequests.forEach((chain) => {
+    const groupsCount = chain.requestsGroups.length;
+    let remainingChainLimit = chain.chainLimit;
+
+    chain.requestsGroups.forEach((group, groupIndex) => {
+      const requestCount = group.requests.length;
+      const unprocessedGroupsCount = groupsCount - groupIndex;
+
+      const allowedRequestCount = Math.min(Math.floor(remainingChainLimit / unprocessedGroupsCount), requestCount);
+      group.requests.slice(allowedRequestCount).forEach((req) => {
+        logs.push(
+          logger.pend(
+            'INFO',
+            `Ignoring request with ID:${req.id} on chain:${chain.chainId} because it exceeded chain limit:${chain.chainLimit}`
+          )
+        );
+      });
+      // eslint-disable-next-line functional/immutable-data
+      group.requests = group.requests.slice(0, allowedRequestCount);
+      remainingChainLimit -= group.requests.length;
+    });
+  });
+
+  const updatedGroups = flatMap(chainRequests, (chain) => chain.requestsGroups).sort((g1, g2) => g1.index - g2.index);
+  const updatedEvmState = zip(providerStates.evm, updatedGroups).map(([evmState, group]) => {
+    return { ...evmState!, requests: groupRequests(group!.requests) };
+  });
+  const processedProviders = { evm: updatedEvmState };
+  return [logs, processedProviders] as const;
+}
--- a/packages/airnode-node/src/coordinator/calls/chain-limits.ts
+++ b/packages/airnode-node/src/coordinator/calls/chain-limits.ts
@@ -3,28 +3,12 @@ import groupBy from 'lodash/groupBy';
 import map from 'lodash/map';
 import zip from 'lodash/zip';
 import * as logger from '../../logger';
-import { sortRequests } from '../../requests';
-import {
-  Config,
-  ProviderStates,
-  Request,
-  GroupedRequests,
-  RequestType,
-  ApiCall,
-  Withdrawal,
-  PendingLog,
-} from '../../types';
+import { Config, ProviderStates, PendingLog, GroupedRequests } from '../../types';
+import { flattenRequests, groupRequests } from '../../requests/grouping';
+import { sortRequests } from '../../requests/sorting';
 
-function groupRequests(flatRequests: Request<any>[]): GroupedRequests {
-  const apiCalls = flatRequests
-    .filter((request) => request.__type === RequestType.ApiCall)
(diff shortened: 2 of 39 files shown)


packages/airnode-node/src/coordinator/calls/chain-limits.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export function applyChainLimits(config: Config, providerStates: ProviderStates) {
+  const logs: PendingLog[] = [];
+
+  const providerRequests = providerStates.evm
+    .map((evmState, index) => {
+      return { index, chainId: evmState.settings.chainId, requests: flattenAndSortRequests(evmState.requests) };
+    })
+    // Sort by number of request ascendingly
```
andreogle, line 21: "ascendingly" is a bit weird. Wasn't sure it was even a word (my browser's spell checker doesn't like it).

https://english.stackexchange.com/a/309294

```suggestion
    // Sort by number of request in an ascending order
```

packages/airnode-node/src/handlers/start-coordinator.ts
```diff
@@ -5,100 +5,176 @@ import * as logger from '../logger';
...
+
+async function coordinator(config: Config): Promise<CoordinatorState> {
   // =================================================================
-  // STEP 3: Group API calls with respect to request IDs
+  // STEP 1: Create a blank coordinator state
   // =================================================================
-  const flatApiCalls = flatMap(state2.providerStates.evm, (provider) => provider.requests.apiCalls);
-  const aggregatedApiCallsById = calls.aggregate(state2.config, flatApiCalls);
-  const state3 = state.update(state2, { aggregatedApiCallsById });
+  let state = createInitialCoordinatorState(config);
```
andreogle, line 139: I assume this is much better for memory usage? I'm not sure exactly how JS garbage collects things like `state1` after they're no longer needed

packages/airnode-node/src/handlers/start-coordinator.ts
```diff
@@ -5,100 +5,176 @@ import * as logger from '../logger';
...
+  // STEP 3: Return early if there are no actionable requests
+  // =================================================================
+  if (hasCoordinatorNoActionableRequests(state)) {
+    const { coordinatorId } = state;
+    const logOptions = logger.buildBaseOptions(config, { coordinatorId });
+    logger.info('No actionable requests detected. Returning', logOptions);
+    return state;
+  }
+
+  // TODO: apply chain request limits
```
andreogle, line 156: Remove?

packages/airnode-node/src/requests/grouping.ts
```diff
@@ -32,3 +33,30 @@ export function groupRequestsBySponsorAddress(requests: GroupedRequests): Reques
 
   return groupedRequests;
 }
+
+export function flattenRequests(groupedRequests: GroupedRequests): Request<AnyRequest>[] {
```
andreogle, line 37: Nice one moving these functions here. Makes a lot more sense

packages/airnode-node/test/e2e/test-api.feature.ts
```diff
@@ -2,6 +2,7 @@ import { testApi } from '../../src/workers/local-handlers';
 import { deployAirnodeAndMakeRequests, increaseTestTimeout } from '../setup/e2e';
 
 it('makes a call to test the API', async () => {
+  console.log('aaa');
```
andreogle, line 5:
```suggestion
```

packages/airnode-node/src/coordinator/calls/chain-limits.ts
```diff
@@ -0,0 +1,61 @@
...
+  // coordinator performance
+  chainRequests.forEach((chain) => {
+    const groupsCount = chain.requestsGroups.length;
+    let remainingChainLimit = chain.chainLimit;
+
+    chain.requestsGroups.forEach((group, groupIndex) => {
+      const requestCount = group.requests.length;
+      const unprocessedGroupsCount = groupsCount - groupIndex;
+
+      const allowedRequestCount = Math.min(Math.floor(remainingChainLimit / unprocessedGroupsCount), requestCount);
```
andreogle, line 40: Would you be able to add a comment here explaining this? It's not so obvious to me and it's a core part of how this works

andreogle (approved):
Took me quite a while to wrap my head around the chain-limits module, but it looks good from what I can tell 👍🏻

bbenligiray:
Quick question without looking at the code: What parameter does the user set? i.e. number of what per-chain? I'm wondering if it will be intuitive.

Siegrift:
> Quick question without looking at the code: What parameter does the user set? i.e. number of what per-chain? I'm wondering if it will be intuitive.

`maxConcurrentTransactions`, but I'll be working on the docs today, so I can let you read those if you'd like.

packages/airnode-node/src/handlers/start-coordinator.ts
```diff
@@ -5,100 +5,176 @@ import * as logger from '../logger';
 import * as providers from '../providers';
 import { reportHeartbeat } from '../reporting';
 import { hasNoActionableRequests } from '../requests/request';
-import * as state from '../coordinator/state';
+import * as coordinatorState from '../coordinator/state';
```
amarthadan, line 8: I like both the `startCoordinator` and `randomString` refactors but I don't think they should have been part of this PR. Especially the `startCoordinator` one since all the other handlers use the previous pattern. It should have been imo changed globally.

amarthadan (approved):

packages/airnode-deployer/config/config.json.example
```diff
@@ -1,6 +1,7 @@
 {
   "chains": [
     {
+      "maxConcurrentTransactions": 100,
```
Siegrift, line 4: Talked with Burak on call and we agreed that this field should be called `maxConcurrency` instead.

packages/airnode-node/src/handlers/start-coordinator.ts
Siegrift: It doesn't matter for performance, but it's a bit more readable and less error prone. In the past we had a potential bug where we had code like this:
```
const state5 = {...state3, ...someOtherThings}
```
where we should use state4 instead. Now there is no way for this bug to occur.

packages/airnode-node/test/e2e/test-api.feature.ts
Siegrift: I've created this issue https://api3dao.atlassian.net/browse/AN-491 since it's hard to cleanup these messages after finishing debugging.

packages/airnode-node/src/coordinator/calls/chain-limits.ts
```diff
@@ -0,0 +1,78 @@
...
+    .map((evmState, index) => {
+      const { apiCalls, withdrawals } = evmState.requests;
+      return { index, chainId: evmState.settings.chainId, requests: sortRequests([...apiCalls, ...withdrawals]) };
+    })
+    // Sort by number of request ascendingly
+    .sort((p1, p2) => p1.requests.length - p2.requests.length);
+  const groupedProviderRequests = groupBy(providerRequests, 'chainId');
+  const chainRequests = map(groupedProviderRequests, (group, chainId) => ({
+    chainId,
+    chainLimit: config.chains.find((chain) => chain.id === chainId)!.maxConcurrentTransactions,
```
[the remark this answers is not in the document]
Siegrift: This tries to group the providers by `chainId`. This assertion is indeed valid since every provider is listed in some under some chain settings inside [config.json](https://docs.api3.org/airnode/v0.4/grp-providers/tutorial/quick-deploy-aws/config-json.html). 

These providers are initialized one by one [here](https://github.com/api3dao/airnode/blob/an395-per-chain-limits/packages/airnode-node/src/providers/actions.ts#L43) and I am basically reverting that mapping.

packages/airnode-node/src/coordinator/calls/chain-limits.ts
```diff
@@ -0,0 +1,78 @@
...
+    requestsGroups: group,
```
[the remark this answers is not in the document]
Siegrift: I've decided to use `requestsGroups: group,`. I understand it's actually pretty complex to understand what is going on, but inside an IDE typescript can help you infer the types which helps a lot.

packages/airnode-node/src/handlers/start-coordinator.ts
Siegrift: Agree and also disagree :D . It shouldn't have been done in this PR, but doing it globally seems like a bigger effort. But I agree we should do this at some point and "play" with the code (globally) to improve it.

packages/airnode-node/src/coordinator/calls/chain-limits.ts
```diff
@@ -0,0 +1,80 @@
...
+//    2. Take the smallest unprocessed group of provider requests
+//    3. Take the maximum number of requests (say M) as you can. Let's say the current provider request count is R and
+//       there are N unprocessed providers (including the one currently processing). Then M = max(A / N, R). This is
+//       because taking M requests of the current provider implies taking M request from the others.
+//    4. Pick the oldest M request and discard other (extraneous) requests.
+//    5. Update A by subtracting M, go step 2.
+//
+// This algorithm naturally favours providers with smaller number of requests, while it also maximizes the number of
+// requests taken. E.g. N = 10, A = 56, R = 10. Then the algorithm picks (5,5,5,5,6,6,6,6,6). In general, if R > A / n,
+// then all providers end up with at least A / N request and last (A % N) end up with +1.
```
Siegrift, line 31: And in the docs I plan to write something along these lines:
```
The "maximumConcurrency" specifies the maximum number of concurrent handler calls per single Airnode invocation. If you set this field to value X, then Airnode will guarantee that:
- At most X api calls are made to the API
- At most X transactions will be made by the blockchain providers of the respective chain

(This only applies to the requests initiated on chain. For example, requests triggered using http geteway are not included in this limit.)
```

Siegrift: Here are the docs if you are interested: https://github.com/api3dao/api3-docs/pull/536

packages/airnode-node/src/coordinator/calls/chain-limits.ts
```diff
@@ -0,0 +1,80 @@
...
+
+function flattenAndSortRequests(grouped: GroupedRequests) {
+  return sortRequests(flattenRequests(grouped));
+}
+
+// Apply the chain request limit by dropping requests which exceed the limit. We do this mutably to increase
+// performance. This algorithm runs in O(P), where P = number of providers across all chains. When the chain limit is
+// exceeded, the latest requests of the largest provider shall be dropped.
+//
+// The algorithm works as following:
```
andreogle, line 18:
```suggestion
// The algorithm works as follows:
```

packages/airnode-node/src/coordinator/calls/chain-limits.ts
andreogle: Looks great, thanks for writing this up 👍🏻 

Is this right?

```suggestion
// then all providers end up with at least A / N request and the last provider (A % N) ends up with +1.
```

andreogle (approved):

packages/airnode-node/src/coordinator/calls/chain-limits.ts
Siegrift: no, this is meant to say that if there is 56 request limit and we have 10 providers with 15 requests each than we end up with the following:
5 5 5 5 6 6 6 6 6 6

A % N = 56 % 10 = 6 so 6 providers end up with +1 (the base amount is A / n = 5)

PR #767, opened 2022-01-11
Do not run the validator in cloud environments.

aquarat:
Cam reported that`callApi` was timing out on AWS for the Amberdata deployment. I investigated and found that the size of the config file had surfaced severe inefficiencies in the `config.json` validator. This appears to only be an issue on cold-start of each handler (`parsedConfig` appears to effectively be memo’ised), but the inefficiency resulted in a start-up delay of 14-16 seconds, enough to prevent callApi from ever running.

After some investigation I found that the validator was largely responsible for this delay; disabling the validator in cloud environments reduces total execution time to the point that `callApi` no longer times out before the config has been parsed and memo'ised.

This PR implements disabling the validator in cloud environments.

### DIFF
--- a/packages/airnode-deployer/src/handlers/aws/index.ts
+++ b/packages/airnode-deployer/src/handlers/aws/index.ts
@@ -1,10 +1,9 @@
 import * as path from 'path';
 import { handlers, logger, utils, providers, WorkerResponse } from '@api3/airnode-node';
 import { loadConfig } from '../../utils';
-import { isCloudFunction } from '../../utils/infrastructure';
 
 const configFile = path.resolve(`${__dirname}/../../config-data/config.json`);
-const parsedConfig = loadConfig(configFile, process.env, !isCloudFunction());
+const parsedConfig = loadConfig(configFile, process.env, false);
 
 function encodeBody(data: WorkerResponse): string {
   return JSON.stringify(data);
--- a/packages/airnode-deployer/src/handlers/aws/index.ts
+++ b/packages/airnode-deployer/src/handlers/aws/index.ts
@@ -1,9 +1,10 @@
 import * as path from 'path';
 import { handlers, logger, utils, providers, WorkerResponse } from '@api3/airnode-node';
 import { loadConfig } from '../../utils';
+import { isCloudFunction } from '../../utils/infrastructure';
 
 const configFile = path.resolve(`${__dirname}/../../config-data/config.json`);
-const parsedConfig = loadConfig(configFile, process.env);
+const parsedConfig = loadConfig(configFile, process.env, !isCloudFunction());
 
 function encodeBody(data: WorkerResponse): string {
   return JSON.stringify(data);
--- a/packages/airnode-deployer/src/utils/infrastructure.ts
+++ b/packages/airnode-deployer/src/utils/infrastructure.ts
@@ -17,3 +17,8 @@ export function formatTerraformArguments(args: CommandArg[]) {
     })
     .map((arg) => `-${arg}`);
 }
+
+/**
+ * Checks if the environment is a GCP or AWS cloud function
+ */
+export const isCloudFunction = () => process.env.LAMBDA_TASK_ROOT || process.env.FUNCTION_TARGET;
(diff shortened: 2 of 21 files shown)


aquarat:
And apologies, I will add a changeset in a moment.

bbenligiray:
Since the node is almost always supposed to run on a cloud environment, shouldn't we limit validation to the deployer instead?

packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -1,9 +1,10 @@
 import * as path from 'path';
 import { handlers, logger, utils, providers, WorkerResponse } from '@api3/airnode-node';
 import { loadConfig } from '../../utils';
+import { isCloudFunction } from '../../utils/infrastructure';
 
 const configFile = path.resolve(`${__dirname}/../../config-data/config.json`);
-const parsedConfig = loadConfig(configFile, process.env);
+const parsedConfig = loadConfig(configFile, process.env, !isCloudFunction());
```
amarthadan, line 7: Do we even need the `isCloudFunction` function? Neither this, nor the GCP handler are called locally (there's a local handler for that in `airnode-node` package). So we can pass `false` directly.

amarthadan (approved):
I'd say this is a quick fix for now but not a long-term solution as having validation for each validator run is imo useful.

amarthadan:
> Since the node is almost always supposed to run on a cloud environment, shouldn't we limit validation to the deployer instead?

But deployer is run locally so there's no time limitation for execution. Or have I misunderstood?

bbenligiray:
I was saying that we should only validate during deployment and be done with it, assuming that the validator is inherently too processing-heavy to ever be run every minute. Because if that's the case, I'm not sure if it's a good idea to do this at every iteration of the local node (instead this can be done once when the user runs the container).

Siegrift:
I strongly agree with Burak on this.

packages/airnode-deployer/src/utils/infrastructure.ts
```diff
@@ -17,3 +17,8 @@ export function formatTerraformArguments(args: CommandArg[]) {
     })
     .map((arg) => `-${arg}`);
 }
+
+/**
+ * Checks if the environment is a GCP or AWS cloud function
+ */
+export const isCloudFunction = () => process.env.LAMBDA_TASK_ROOT || process.env.FUNCTION_TARGET;
```
Siegrift, line 24: I wonder how reliable this check is. Will these always be present?

packages/airnode-node/src/config/index.ts
```diff
@@ -56,7 +60,17 @@ export function getEnvValue(envName: string) {
   return process.env[envName];
 }
 
-function validateConfig(supposedConfig: unknown, secrets: Record<string, string | undefined>): Result {
+function validateConfig(
+  supposedConfig: unknown,
+  secrets: Record<string, string | undefined>,
+  shouldValidate = true
```
Siegrift, line 66: Why one of these has default `true` and the other `false` - it feels inconsistent.

Siegrift: Do we actually need the default values?

Siegrift (commented):
An unit test would be nice, but LGTM otherwise.

packages/airnode-deployer/src/handlers/aws/index.ts
aquarat: Ahhh, wow, yes, thank you, I'll amend :)

packages/airnode-deployer/src/utils/infrastructure.ts
aquarat: According to the Google and AWS docs, yes:
[AWS](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html)
[GCP](https://cloud.google.com/functions/docs/configuring/env-var)
A library to do this would be ideal, but checking for those ENVs is quick, should be reliable and doesn't require any additional dependencies.

Siegrift (approved):
:+1: LGTM

PR #769, opened 2022-01-11
Remove non existent flag from deployer in airnode examples

Siegrift:
I forgot to remove this one from examples when finishing https://github.com/api3dao/airnode/pull/749


dcroote (approved):
LGTM 👍

PR #772, opened 2022-01-13
Make the protocol IDs start from 1

Siegrift:
https://api3dao.atlassian.net/browse/AN-453


packages/airnode-operation/src/config/evm-dev-config.json
```diff
@@ -17,7 +17,7 @@
             { "type": "bytes32", "name": "to", "value": "USD" },
             { "type": "bytes32", "name": "_type", "value": "int256" },
             { "type": "bytes32", "name": "_path", "value": "result" },
-            { "type": "bytes32", "name": "_times", "value": "1000000" }
+            { "type": "bytes32", "name": "_times", "value": "100000" }
```
Siegrift, line 20: I edited this back to `100000` because it would end up changing more things inside Airnode. This was introduced when creating fixtures for Airkeeper in https://github.com/api3dao/airnode/commit/5798e606b070d188344c70b5edb5ad68ad1e16d4#diff-04917c5d42958ab67a1a84d91bbab5dee575960c2fc7751f86ed6283682ca5d4L67

packages/airnode-operation/src/evm/requests/make-requests.ts
```diff
@@ -72,7 +72,8 @@ export async function makeWithdrawal(state: State, request: Request) {
 
   const { AirnodeRrp } = state.contracts;
 
-  await AirnodeRrp.connect(signer).requestWithdrawal(airnode, sponsorWallet.address);
+  const tx = await AirnodeRrp.connect(signer).requestWithdrawal(airnode, sponsorWallet.address);
```
Siegrift, line 75: This is part of the patch that @acenolaza shared to me which I've used to regenerate the fixtures. I didn't include it since it should be done more properly and its tracked in https://api3dao.atlassian.net/browse/AN-154. Gladly these fixtures do not change very often.

packages/airnode-node/src/handlers/start-coordinator.test.ts
```diff
@@ -30,6 +30,10 @@ import * as validator from '@api3/airnode-validator';
 import { startCoordinator } from './start-coordinator';
 import * as fixtures from '../../test/fixtures';
 
+beforeEach(() => {
+  console.log('s');
```
Siegrift, line 34: TODO: Remove

packages/airnode-node/src/evm/handlers/fetch-pending-requests.test.ts
```diff
@@ -18,19 +18,20 @@ describe('fetchPendingRequests', () => {
...
+          id: '0x8d6d6ce1ac5020c4db26c34ae42fa8318a51fa21faef9a65663724641b369e63',
           metadata: {
             address: '0x5FbDB2315678afecb367f032d93F642f64180aa3',
-            blockNumber: 12,
+            blockNumber: 13,
             currentBlock: null,
             ignoreBlockedRequestsAfterBlocks: 20,
-            transactionHash: '0x0ba1b3c4b6710fffeb6c2d1e62f11fa695177a456e4e895f0666ec54b2e7f8bf',
+            transactionHash:
+              '0xba69a4584ca8e383412741f9f75b351d34d5eff9e14b38d8ba8b994bb175e3c5549594040952fc82817cbfb3eaf4593aa8858a34b682886737aac15f759dc1d41c',
```
amarthadan, line 34: Why is it that this transaction hash is much longer than other transaction hashes?

Siegrift: It's probably a bug - at one place I incorrectly renamed one hash and then I reverted that for failing tests. This is probably never used anywhere so the tests pass without it.

bbenligiray (commented):
Looks good. You could have parameterized protocol ID as in `deriveWalletPathFromSponsorAddress(protocolId, sponsorAddress)` but we won't need this any time soon.

packages/airnode-admin/src/implementation.ts
```diff
@@ -30,7 +30,7 @@ export const deriveWalletPathFromSponsorAddress = (sponsorAddress: string): stri
     const shiftedSponsorAddressBN = sponsorAddressBN.shr(31 * i);
     paths.push(shiftedSponsorAddressBN.mask(31).toString());
   }
-  return `0/${paths.join('/')}`;
```
acenolaza, line 33: Basically same comment I left here: https://github.com/api3dao/airnode-stress-tester/pull/5#discussion_r786139114

packages/airnode-node/src/evm/wallet.ts
```diff
@@ -24,7 +24,7 @@ export const deriveWalletPathFromSponsorAddress = (sponsorAddress: string): stri
     const shiftedSponsorAddressBN = sponsorAddressBN.shr(31 * i);
     paths.push(shiftedSponsorAddressBN.mask(31).toString());
   }
-  return `0/${paths.join('/')}`;
```
acenolaza, line 27: same here: https://github.com/api3dao/airnode-stress-tester/pull/5#discussion_r786139114

packages/airnode-operation/src/config/evm-dev-config.json
acenolaza: The thing is that I changed that value because of the number of decimals the mocked api is [currently](https://github.com/api3dao/airnode/blob/f717c568e08c7fe896f2973bb75223eeff32ef23/packages/airnode-operation/src/server.ts#L16) returning. I had to add that extra 0 to be able to get an integer representation of the full number so it doesn't get truncated. What things need to change in Airnode? Should we take out a decimal in server.ts instead?

packages/airnode-operation/src/evm/requests/make-requests.ts
acenolaza: I see you've been having fun with test fixtures 😄

packages/airnode-operation/src/evm/utils.ts
```diff
@@ -39,5 +39,5 @@ export const deriveWalletPathFromSponsorAddress = (sponsorAddress: string): stri
     const shiftedSponsorAddressBN = sponsorAddressBN.shr(31 * i);
     paths.push(shiftedSponsorAddressBN.mask(31).toString());
   }
-  return `0/${paths.join('/')}`;
```
acenolaza, line 42: again, we should at least have a `protocolId` as arg here

packages/airnode-protocol/test/test-utils.js
```diff
@@ -7,7 +7,7 @@ function deriveWalletPathFromSponsorAddress(sponsorAddress) {
     const shiftedSponsorAddressBN = sponsorAddressBN.shr(31 * i);
     paths.push(shiftedSponsorAddressBN.mask(31).toString());
   }
-  return `0/${paths.join('/')}`;
```
acenolaza, line 10: and here 😄

acenolaza (changes requested):

packages/airnode-operation/src/config/evm-dev-config.json
Siegrift: >  What things need to change in Airnode?

The template IDs are changed.

> Should we take out a decimal in server.ts instead?

Yeah, I'd rather do that.

packages/airnode-admin/src/implementation.ts
Siegrift: I'd rather postpone that and revisit it properly by moving this function to a separate package or when the use case for parameterizing this occurs.

packages/airnode-operation/src/evm/utils.ts
Siegrift: Do we need it anytime soon? I'd rather postpone this.

packages/airnode-admin/src/implementation.ts
acenolaza: 👍🏻

packages/airnode-operation/src/config/evm-dev-config.json
acenolaza: It might brake some tests but I'm ok with that 👍🏻

packages/airnode-operation/src/evm/utils.ts
acenolaza: Let's postpone it since Burak already said so https://github.com/api3dao/airnode/pull/772#pullrequestreview-854572547

packages/airnode-admin/src/implementation.ts
```diff
@@ -30,7 +30,7 @@ export const deriveWalletPathFromSponsorAddress = (sponsorAddress: string): stri
...
+  return `1/${paths.join('/')}`;
```
acenolaza, line 33: Might as well update the comment in line 21

acenolaza (approved):
Just a minor comment update but LGTM 👍🏻

packages/airnode-admin/src/implementation.ts
Siegrift: Thanks. I was sure I got them all... Not sure how this one slipped :/

PR #773, opened 2022-01-14
Update GCP authentication process & API management

amarthadan:
[Issue AN-464](https://api3dao.atlassian.net/browse/AN-464)

Users will now need to enable just one API in GCP console and create a service account with the access key that can be downloaded as a JSON file to use for authentication. No need to install `gcloud` tools.

Docs update and further explanation of the new flow in [PR #532](https://github.com/api3dao/api3-docs/pull/532)

### DIFF
--- a/.changeset/violet-terms-unite.md
+++ b/.changeset/violet-terms-unite.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-deployer': minor
+---
+
+Update GCP authentication process & APIs management
--- a/packages/airnode-deployer/terraform/airnode/gcp/modules/function/main.tf
+++ b/packages/airnode-deployer/terraform/airnode/gcp/modules/function/main.tf
@@ -9,14 +9,6 @@ resource "random_string" "function_service_account_id" {
   number  = false
 }
 
-resource "random_string" "scheduler_service_account_id" {
-  length  = 20
-  lower   = true
-  upper   = false
-  special = false
-  number  = false
-}
-
 resource "null_resource" "fetch_function_files" {
   provisioner "local-exec" {
     command = <<EOC
@@ -85,7 +77,16 @@ resource "google_cloudfunctions_function" "function" {
   timeout               = var.timeout
   max_instances         = var.max_instances
   environment_variables = merge(merge(var.environment_variables, fileexists(var.secrets_file) ? jsondecode(file(var.secrets_file)) : {}), { AIRNODE_CLOUD_PROVIDER = "gcp" })
+}
 
+resource "random_string" "scheduler_service_account_id" {
+  count = var.schedule_interval == 0 ? 0 : 1
+
+  length  = 20
+  lower   = true
+  upper   = false
+  special = false
+  number  = false
 }
 
 resource "google_app_engine_application" "app" {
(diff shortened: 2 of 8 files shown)


packages/airnode-deployer/terraform/airnode/gcp/modules/function/main.tf
```diff
@@ -85,7 +77,16 @@ resource "google_cloudfunctions_function" "function" {
   timeout               = var.timeout
   max_instances         = var.max_instances
   environment_variables = merge(merge(var.environment_variables, fileexists(var.secrets_file) ? jsondecode(file(var.secrets_file)) : {}), { AIRNODE_CLOUD_PROVIDER = "gcp" })
+}
 
+resource "random_string" "scheduler_service_account_id" {
```
Siegrift, line 82: Why is it called `random_string`?

Siegrift (changes requested):
You probably also need to update the airnode-examples documenation.

Btw. this means that we no longer support the current GCP auth scheme right?

packages/airnode-deployer/terraform/airnode/gcp/modules/function/main.tf
amarthadan: I guess because it's a string that's random :smile: It's a Terraform resource name, that's not my naming.

amarthadan:
> You probably also need to update the airnode-examples documentation.

Right, totally forgot about those.

> Btw. this means that we no longer support the current GCP auth scheme right?

No, you can still use that as well. The only difference is that I've changed the default path where the Docker container looks for the credentials file. So the command will look like `... -v "${HOME}/.config/gcloud:/app/gcloud" -e GOOGLE_APPLICATION_CREDENTIALS=/app/gcloud/application_default_credentials.json ...` with the `GOOGLE_APPLICATION_CREDENTIALS` addition.

.changeset/violet-terms-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+'@api3/airnode-deployer': minor
+---
+
+Update GCP authentication process & APIs management
```
Siegrift, line 5: Nit: 
```suggestion
Update GCP authentication process and APIs management
```

packages/airnode-examples/README.md
```diff
@@ -119,11 +119,12 @@ yarn create-aws-secrets
 
 ### 6. (Only if deploying to GCP) Create GCP credentials
 
-If you intend to deploy Airnode on GCP, you will need to sign in using the GCP CLI tool. If you are not sure how to do
-this or how to create a GCP account, see
+If you intend to deploy Airnode on GCP, you will need to create a service account for your project and add and download
```
Siegrift, line 122: You've mentioned that I can use the previous auth mathod as well, but I am not sure how since it is not mentioned in the docs nor examples.

Siegrift (approved):
LGTM code wise, but I didn't test it myself.

packages/airnode-examples/README.md
amarthadan: I'd say that's fine. We're preferring this auth method and encourage users to use it. We're also not describing multiple auth methods you can use for AWS.

PR #774, opened 2022-01-14
Add string32 and bool options to Airnode ABI

Siegrift:
https://api3dao.atlassian.net/browse/AN-459


Siegrift:
@acenolaza @amarthadan I am sorry, but since the template encoded parameters changed (and thus templateID too). I had to once again update a lot of the test fixtures. The relevant changes, however are done in the first commit.

bbenligiray:
There should be a `bytes32` type represented as `b`, a `string32` type represented as `s`, and a `bool` type represented as `?`

packages/airnode-adapter/e2e/index.ts
```diff
@@ -274,4 +274,9 @@ describe('Extraction, encoding and simple on chain decoding', () => {
       );
     });
   });
+
+  it.only('wow', async () => {
```
andreogle, line 278: I assume this can be removed?

packages/airnode-abi/src/encoding.ts
```diff
@@ -1,36 +1,33 @@
...
     return [...acc, shortType!];
   }, []);
 
-  // '1' must be the first character as it indicates the version
-  return `1${selectedShortTypes.join('')}`;
+  return `${VERSION}${selectedShortTypes.join('')}`;
 }
 
-function buildNameValuePairs(parameters: InputParameter[]): string[] {
+function buildNameValuePairs(parameters: InputParameter[]): unknown[] {
```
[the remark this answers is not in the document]
Siegrift: because the value may also be `boolean` and the value can techincally be anything

packages/airnode-abi/src/constants.ts
```diff
@@ -0,0 +1,21 @@
+import { ValueOf, TypeTransformation } from './types';
+
+// Upper case letters refer to dynamically sized types, lower case letters refer to statically sized types
+export const PARAMETER_SHORT_TYPES = {
+  B: 'bytes',
+  b: 'bytes32',
+  S: 'string',
+  s: 'string32', // This is not officially supported type in solidity and we use bytes32 on chain instead
```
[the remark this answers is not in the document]
Siegrift: Yeah, I've explained this more with this comment:
```
  // Type "string32" is not supported in solidity and we use bytes32 on chain instead. We need to create this
  // artificially made up type so that users can decide if they want their data encoded as strings or as raw bytes
  // value.
```

packages/airnode-adapter/e2e/index.ts
Siegrift: Yeah, sorry. This was incorrectly squashed when I was cleaning up after myself.

packages/airnode-abi/src/constants.ts
```diff
@@ -0,0 +1,24 @@
...
+  // artificially made up type so that users can decide if they want their data encoded as strings or as raw bytes
+  // value.
+  s: 'string32',
+  a: 'address',
+  i: 'int256',
+  u: 'uint256',
+  f: 'bool',
+} as const;
+
+export type ParameterType = ValueOf<typeof PARAMETER_SHORT_TYPES>;
```
Siegrift, line 18: @andreogle @amarthadan btw. this is a cool trick to create unions of keys and values of an object. This is preferred over enums in Typescript as they are actually poorly designed. See this for more info: https://medium.com/flow-type/typescript-enums-vs-flow-enums-83da2ca4a9b4

andreogle (approved):

amarthadan (approved):

PR #787, opened 2022-01-19
Implement sponsorAddress and sponsorWalletAddress relay security schemes

Siegrift:
https://api3dao.atlassian.net/browse/AN-496

### DIFF
--- a/packages/airnode-adapter/src/request-building/authentication.ts
+++ b/packages/airnode-adapter/src/request-building/authentication.ts
@@ -104,8 +104,10 @@ function getSchemeAuthentication(
       return getRelayAuthSchemeFromMetadata(apiSecurityScheme, options, 'chainType');
     case 'relayRequesterAddress':
       return getRelayAuthSchemeFromMetadata(apiSecurityScheme, options, 'requesterAddress');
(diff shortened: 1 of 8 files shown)


packages/airnode-adapter/src/request-building/authentication.ts
```diff
@@ -104,8 +104,10 @@ function getSchemeAuthentication(
       return getRelayAuthSchemeFromMetadata(apiSecurityScheme, options, 'chainType');
     case 'relayRequesterAddress':
       return getRelayAuthSchemeFromMetadata(apiSecurityScheme, options, 'requesterAddress');
-    default:
-      return {};
```
amarthadan, line 108: Is it ok to get rid of the `default`?

amarthadan (approved):

packages/airnode-adapter/src/request-building/authentication.ts
Siegrift: Yeah, it should be verified by validator.
Also stripping away the default is cleaner, since TS will use checked switch statement and warn you when you whn you add a new auth scheme.

PR #789, opened 2022-01-22
Add SelfServeRrpBeaconServerWhitelister Contract

Ashar2shahid:
Main purpose of this contract is for readers to whitelist themselves. The contract is assumed to have the `indefiniteWhitelisterRole` and `WhitelistExpirationSetterRole`. 

There are two ways to whitelist:
1. With an expirationTimestamp
2. Indefinitely

A set of beaconIds are available for readers to be able to whitelist themselves for either 1 or 2 (different sets). The deployer(owner) can add/remove from the sets. 
 
I wasn't sure where to put the contracts, so I put it under `/utils`.

### DIFF
--- a/packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
+++ b/packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
@@ -1,0 +1,126 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "../rrp/requesters/RrpBeaconServer.sol";
+import "./interfaces/ISelfServeRrpBeaconServerWhitelister.sol";
+
+/// @title Contract that allows to whitlist readers on the RrpBeaconServer
+/// @dev The SelfServerRrpBeaconServerWhitelister contract has the WhitelistExpirationSetterRole
+/// and the IndefiniteWhitelisterRole of the RrpBeaconServer contract. The deployer of this contract
+/// can specify the beaconIds for which readers can whitelist themselves. The deployer(also the owner)
+/// can also add new beaconIds later for readers to self whitelist themselves.
+contract SelfServeRrpBeaconServerWhitelister is
+    Ownable,
+    ISelfServeRrpBeaconServerWhitelister
+{
+    address public rrpBeaconServer;
+
+    mapping(bytes32 => uint64) public beaconIdToExpirationTimestamp;
+    mapping(bytes32 => bool) public beaconIdToIndefiniteWhitelistStatus;
+
+    /// @param _rrpBeaconServer The RrpBeaconServer contract to whitelist readers on
+    /// @param _beaconIdsExpirationTimestamp Array of beaconIds that have an associated expiration timestamp
+    /// @param _expirationTimestamps Array of expiration timestamps that correspond to the beaconIds
+    /// @param _beaconIdsIndefiniteWhitelistStatus Array of beaconIds that have an associated indefinite whitelist status
+    /// @param _indefiniteWhitelistStatuses Array of indefinite whitelist status that correspond to the beaconIds
+    constructor(
+        address _rrpBeaconServer,
+        bytes32[] memory _beaconIdsExpirationTimestamp,
+        uint64[] memory _expirationTimestamps,
+        bytes32[] memory _beaconIdsIndefiniteWhitelistStatus,
+        bool[] memory _indefiniteWhitelistStatuses
+    ) {
+        require(_rrpBeaconServer != address(0), "RrpBeaconServer address zero");
+        rrpBeaconServer = _rrpBeaconServer;
+        for (
+            uint256 ind = 0;
+            ind < _beaconIdsExpirationTimestamp.length;
+            ind++
+        ) {
+            beaconIdToExpirationTimestamp[
+                _beaconIdsExpirationTimestamp[ind]
+            ] = _expirationTimestamps[ind];
+        }
+        for (
+            uint256 ind = 0;
+            ind < _beaconIdsIndefiniteWhitelistStatus.length;
+            ind++
+        ) {
+            beaconIdToIndefiniteWhitelistStatus[
+                _beaconIdsIndefiniteWhitelistStatus[ind]
+            ] = _indefiniteWhitelistStatuses[ind];
+        }
+    }
+
+    /// @notice Adds a new beaconId with an expiration timestamp
+    /// that can be whitelisted by readers
+    /// @param _beaconId The beaconId to set an expiration timestamp for
+    /// @param _expirationTimestamp The expiration timestamp for the beaconId
+    function setBeaconIdWithExpirationTimestamp(
+        bytes32 _beaconId,
+        uint64 _expirationTimestamp
+    ) external override onlyOwner {
+        beaconIdToExpirationTimestamp[_beaconId] = _expirationTimestamp;
+        emit SetBeaconIdWithExpirationTimestamp(
+            _beaconId,
+            _expirationTimestamp
+        );
+    }
+
+    /// @notice Adds a new beaconId with an indefinite whitelist status
+    /// that can be whitelisted by readers indefinetly
+    /// @param _beaconId The beaconId to set an indefinite whitelist status for
+    /// @param _indefiniteWhitelistStatus The indefinite whitelist status for the beaconId
+    function setBeaconIdWithIndefiniteWhitelistStatus(
+        bytes32 _beaconId,
+        bool _indefiniteWhitelistStatus
+    ) external override onlyOwner {
+        beaconIdToIndefiniteWhitelistStatus[
+            _beaconId
+        ] = _indefiniteWhitelistStatus;
+        emit SetBeaconIdWithIndefiniteWhitelistStatus(
+            _beaconId,
+            _indefiniteWhitelistStatus
+        );
+    }
+
+    /// @notice Whitelists a reader on the RrpBeaconServer with an expiration timestamp
+    /// @param _beaconId The beaconId to whitelist
+    /// @param _reader The reader to whitelist on the beaconId
+    function whitelistReaderWithExpiration(bytes32 _beaconId, address _reader)
+        external
+        override
+    {
+        uint64 expirationTimestamp = beaconIdToExpirationTimestamp[_beaconId];
+        require(
+            expirationTimestamp > block.timestamp,
+            "Cannot whitelist for beacon"
+        );
+        IRrpBeaconServer(rrpBeaconServer).setWhitelistExpiration(
+            _beaconId,
+            _reader,
+            expirationTimestamp
+        );
+        emit WhitelistedReaderWithExpiration(_beaconId, _reader);
+    }
+
+    /// @notice Whitelists a reader on the RrpBeaconServer with an indefinite whitelist status
+    /// @param _beaconId The beaconId to whitelist
+    /// @param _reader The reader to whitelist on the beaconId
+    function whitelistReaderIndefinitely(bytes32 _beaconId, address _reader)
+        external
+        override
+    {
+        require(
+            beaconIdToIndefiniteWhitelistStatus[_beaconId],
+            "Cannot whitelist indefinitely for beacon"
+        );
+        IRrpBeaconServer(rrpBeaconServer).setIndefiniteWhitelistStatus(
+            _beaconId,
+            _reader,
+            true
+        );
+        emit WhitelistedReaderIndefinitely(_beaconId, _reader);
+    }
+}
--- a/packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
+++ b/packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
@@ -6,9 +6,9 @@ import "../rrp/requesters/RrpBeaconServer.sol";
 import "./interfaces/ISelfServeRrpBeaconServerWhitelister.sol";
 
 /// @title Contract that allows to whitlist readers on the RrpBeaconServer
-/// @dev The SelfServerRrpBeaconServerWhitelister contract has the WhitelistExpirationSetterRole
+/// @dev The SelfServeRrpBeaconServerWhitelister contract has the WhitelistExpirationSetterRole
 /// and the IndefiniteWhitelisterRole of the RrpBeaconServer contract. The deployer of this contract
-/// can specify the beaconIds for which readers can whitelist themselves. The deployer(also the owner)
+/// can specify the beaconIds for which readers can whitelist themselves. The deployer (also the owner)
 /// can also add new beaconIds later for readers to self whitelist themselves.
 contract SelfServeRrpBeaconServerWhitelister is
     Ownable,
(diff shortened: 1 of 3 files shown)


packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /// @notice Whitelists a reader on the RrpBeaconServer with an indefinite whitelist status
+    /// @param _beaconId The beaconId to whitelist
+    /// @param _reader The reader to whitelist on the beaconId
+    function whitelistReaderIndefinitely(bytes32 _beaconId, address _reader)
+        external
+        override
+    {
+        require(
```
Siegrift, line 115: This needs to be shortened, right?

packages/airnode-protocol/test/utils/SelfServeRrpBeaconServerWhitelister.sol.js
```diff
@@ -0,0 +1,363 @@
...
+
+  selfServeRrpBeaconServerWhitelister = await selfServeRrpBeaconServerWhitelisterFactory.deploy(
+    rrpBeaconServer.address,
+    [beaconId],
+    [beaconIdExpirationTimeStamp],
+    [beaconId2],
+    [true]
+  );
+
+  const managerRootRole = await accessControlRegistry.deriveRootRole(roles.manager.address);
```
Siegrift, line 174: Figuring out this step would take me a lot of time. It was good, that you took over this for me.

packages/airnode-protocol/test/utils/SelfServeRrpBeaconServerWhitelister.sol.js
```diff
@@ -0,0 +1,363 @@
...
+      expect(whitelistStatus.expirationTimestamp).to.equal(beaconIdExpirationTimeStamp);
+      expect(whitelistStatus.indefiniteWhitelistCount).to.equal(0);
+
+      const beaconResponse = await rrpBeaconServer.connect(roles.beaconReader).readBeacon(beaconId);
+      expect(beaconResponse.value).to.equal(123);
+      expect(beaconResponse.timestamp).to.equal(encodedTimestamp);
+    });
+  });
+  context('beaconId expirationTimestamp is invalid', function () {
+    it('reverts', async function () {
```
Siegrift, line 319: Shouldn't we have a similar test that tries to whitelist a reader, but doesn't have the permissions for it?

packages/airnode-protocol/test/utils/SelfServeRrpBeaconServerWhitelister.sol.js
```diff
@@ -0,0 +1,363 @@
...
+    to: sponsorWalletAddress,
+    value: hre.ethers.utils.parseEther('1'),
+  });
+  sponsorWallet = testUtils.deriveSponsorWallet(airnodeMnemonic, roles.sponsor.address).connect(hre.ethers.provider);
+  endpointId = testUtils.generateRandomBytes32();
+  endpointId2 = testUtils.generateRandomBytes32();
+  templateParameters = testUtils.generateRandomBytes();
+  await airnodeRrp.connect(roles.randomPerson).createTemplate(airnodeAddress, endpointId, templateParameters);
+  await airnodeRrp.connect(roles.randomPerson).createTemplate(airnodeAddress, endpointId2, templateParameters);
+  templateId = hre.ethers.utils.keccak256(
```
Siegrift, line 55: Nit: Shouldn't this be moved to some utils file?

packages/airnode-protocol/test/utils/SelfServeRrpBeaconServerWhitelister.sol.js
```diff
@@ -0,0 +1,363 @@
...
+  );
+  beaconId2 = hre.ethers.utils.keccak256(
+    hre.ethers.utils.solidityPack(['bytes32', 'bytes'], [templateId2, beaconParameters])
+  );
+
+  await airnodeRrp.connect(roles.sponsor).setSponsorshipStatus(rrpBeaconServer.address, true);
+  await rrpBeaconServer.connect(roles.sponsor).setUpdatePermissionStatus(roles.updateRequester.address, true);
+
+  // Compute the expected request ID
+  const requestId = hre.ethers.utils.keccak256(
```
Siegrift, line 74: Nit: Shouldn't this be moved to some utils file?

Siegrift (approved):
LGTM, but my contract experience is limited so better wait for someone elses approval.

packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
```diff
@@ -0,0 +1,126 @@
...
+            _reader,
+            expirationTimestamp
+        );
+        emit WhitelistedReaderWithExpiration(_beaconId, _reader);
+    }
+
+    /// @notice Whitelists a reader on the RrpBeaconServer with an indefinite whitelist status
+    /// @param _beaconId The beaconId to whitelist
+    /// @param _reader The reader to whitelist on the beaconId
+    function whitelistReaderIndefinitely(bytes32 _beaconId, address _reader)
```
acenolaza, line 111: So we can't really set the indefinite whitelist status back to `false` using this contract, correct? Wouldn't it be better to have a setter redirect instead?

acenolaza (commented):
Do we really need to keep track of the `expirationTimestamps` and  the `indefiniteWhitelistStatuses` in this contract? Shouldn't we only store the `beaconIds` that are allowed to be managed by this contract?

Then when we call the whitelisting function we can pass the `expirationTimestamp` or the `indefiniteWhitelistStatus`.

packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
```diff
@@ -0,0 +1,126 @@
...
+    mapping(bytes32 => bool) public beaconIdToIndefiniteWhitelistStatus;
+
+    /// @param _rrpBeaconServer The RrpBeaconServer contract to whitelist readers on
+    /// @param _beaconIdsExpirationTimestamp Array of beaconIds that have an associated expiration timestamp
+    /// @param _expirationTimestamps Array of expiration timestamps that correspond to the beaconIds
+    /// @param _beaconIdsIndefiniteWhitelistStatus Array of beaconIds that have an associated indefinite whitelist status
+    /// @param _indefiniteWhitelistStatuses Array of indefinite whitelist status that correspond to the beaconIds
+    constructor(
+        address _rrpBeaconServer,
+        bytes32[] memory _beaconIdsExpirationTimestamp,
```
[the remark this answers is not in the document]
Ashar2shahid: yeah I forgot we don't do this anymore, this was how we used to do it for function argument variables.

packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
```diff
@@ -0,0 +1,126 @@
...
+        uint64[] memory _expirationTimestamps,
+        bytes32[] memory _beaconIdsIndefiniteWhitelistStatus,
+        bool[] memory _indefiniteWhitelistStatuses
+    ) {
+        require(_rrpBeaconServer != address(0), "RrpBeaconServer address zero");
```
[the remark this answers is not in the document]
Ashar2shahid: its usually standard practice to check for address(0) for contract addresses. It comes up in audits so I added the check here

packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
Ashar2shahid: I don't follow ?

packages/airnode-protocol/test/utils/SelfServeRrpBeaconServerWhitelister.sol.js
Ashar2shahid: do you mean the contract itself doesn't have the permissions ?

packages/airnode-protocol/test/utils/SelfServeRrpBeaconServerWhitelister.sol.js
Ashar2shahid: I think this is fine to remain here

packages/airnode-protocol/test/utils/SelfServeRrpBeaconServerWhitelister.sol.js
Ashar2shahid: this one definently can be moved to utils

Ashar2shahid:
> Do we really need to keep track of the `expirationTimestamps` and the `indefiniteWhitelistStatuses` in this contract? Shouldn't we only store the `beaconIds` that are allowed to be managed by this contract?
> 
> Then when we call the whitelisting function we can pass the `expirationTimestamp` or the `indefiniteWhitelistStatus`.

So the purpose of this contract is for readers to be able to whitelist themselves, for the hackathon we want users to be able to whitelist themselves upto an expiration, so we will add those beaconIds with their respective expirationIds to the mapping and allow users to whitelist themselves. 

We also want certain beaconIds to always be available for testing purposes (think testnets) in this case we can add these beaconIds to the indefinite mapping and users of this beaconIds can be whitelisted indefinetely.

packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
acenolaza: "Error message for require is too long" lint warning
Maybe just change it to "Cannot whitelist indefinitetly"

packages/airnode-protocol/contracts/utils/interfaces/ISelfServeRrpBeaconServerWhitelister.sol
```diff
@@ -0,0 +1,37 @@
...
+        bytes32 _beaconId,
+        bool _indefiniteWhitelistStatus
+    ) external;
+
+    function whitelistReaderWithExpiration(bytes32 _beaconId, address _reader)
+        external;
+
+    function whitelistReaderIndefinitely(bytes32 _beaconId, address _reader)
+        external;
+}
```
acenolaza, line 37: Should we also add the 2 public getters for `beaconIdToExpirationTimestamp` and `beaconIdToIndefiniteWhitelistStatus` mappings?

acenolaza (approved):

packages/airnode-protocol/contracts/utils/interfaces/ISelfServeRrpBeaconServerWhitelister.sol
Ashar2shahid: don't public mappings generate getters automatically?

acenolaza: they do but if a contract only wants to import the interface to call any of the getters then they wouldn't be available

packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
```diff
@@ -0,0 +1,123 @@
...
+    /// @param beaconIdsExpirationTimestamp Array of beaconIds that have an associated expiration timestamp
+    /// @param expirationTimestamps Array of expiration timestamps that correspond to the beaconIds
+    /// @param beaconIdsIndefiniteWhitelistStatus Array of beaconIds that have an associated indefinite whitelist status
+    /// @param indefiniteWhitelistStatuses Array of indefinite whitelist status that correspond to the beaconIds
+    constructor(
+        address rrpBeaconServerAddress,
+        bytes32[] memory beaconIdsExpirationTimestamp,
+        uint64[] memory expirationTimestamps,
+        bytes32[] memory beaconIdsIndefiniteWhitelistStatus,
+        bool[] memory indefiniteWhitelistStatuses
```
bbenligiray, line 32: This would always be all `true`s so it's weird. Do this instead:
```
        address rrpBeaconServerAddress,
        bytes32[] memory beaconIds,
        uint64[] memory expirationTimestamps,
        bool[] memory indefiniteWhitelistStatuses
```

bbenligiray: Also validate that the argument lengths are equal

packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
```diff
@@ -0,0 +1,123 @@
...
+import "../rrp/requesters/RrpBeaconServer.sol";
+import "./interfaces/ISelfServeRrpBeaconServerWhitelister.sol";
+
+/// @title Contract that allows to whitlist readers on the RrpBeaconServer
+/// @dev The SelfServeRrpBeaconServerWhitelister contract has the WhitelistExpirationSetterRole
+/// and the IndefiniteWhitelisterRole of the RrpBeaconServer contract. The deployer of this contract
+/// can specify the beaconIds for which readers can whitelist themselves. The deployer (also the owner)
+/// can also add new beaconIds later for readers to self whitelist themselves.
+contract SelfServeRrpBeaconServerWhitelister is
+    Ownable,
```
bbenligiray, line 14: The point of initializing the parameters in the constructor was that this contract was going to be immutable / we wouldn't need to worry about who the owner is and if the wallet would get compromised. So this contract should either be ownable and not do anything in the constructor, or not be ownable and do everything in the constructor and never again.

packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
```diff
@@ -0,0 +1,123 @@
...
+        emit SetBeaconIdWithIndefiniteWhitelistStatus(
+            beaconId,
+            indefiniteWhitelistStatus
+        );
+    }
+
+    /// @notice Whitelists a reader on the RrpBeaconServer with an expiration timestamp
+    /// @param beaconId The beaconId to whitelist
+    /// @param reader The reader to whitelist on the beaconId
+    function whitelistReaderWithExpiration(bytes32 beaconId, address reader)
```
bbenligiray, line 91: I think it's simpler to have a single `whitelistReader()` function that both sets the whitelist expiration and the indefinite whitelist status

bbenligiray:
I don't want this contract in the airnode-protocol package tbh. We will be using a lot of such one-use contracts and we can't clutter the monorepo with them.

packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
Ashar2shahid: I see what you mean, more like this then:
```    
bytes32[] memory beaconIdsExpirationTimestamp,
uint64[] memory expirationTimestamps,
bytes32[] memory beaconIdsIndefiniteWhitelistStatus
```

and not bool

packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
Ashar2shahid: The owner of this contract is going to be the OwnableCallForwarder after deployment, who is also the manager of the `RequesterAuthorizerWithManager`. So I think its fine for this to be a ownable contract because if that is compromised then the `RequesterAuthorizer` is compromised which is a much bigger issue.

Yes, I agree probably better to remove from the constructor then

packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
Ashar2shahid: okay makes sense

packages/airnode-protocol/contracts/utils/SelfServeRrpBeaconServerWhitelister.sol
bbenligiray: I mean you don't even need two lists of beacons, it could go like
```
[beaconId1, beaconId2, beaconId3]
[12345678, 0, 987654312]
[false, true, false]
```
Also `beaconIdsExpirationTimestamp` and `beaconIdsIndefiniteWhitelistStatus` are bad variable names. It should be something like `temporarilyWhitelistableBeaconIds` / `indefinitelyWhitelistableBeaconIds` (if you're going to go with https://github.com/api3dao/airnode/pull/789#discussion_r792637995)

bbenligiray:
Moved to https://github.com/api3dao/utility-contracts

PR #790, opened 2022-01-24
Add HTTP gateway support for GCP

amarthadan:
[Issue AN-308](https://api3dao.atlassian.net/browse/AN-308)

### DIFF
--- a/packages/airnode-deployer/src/handlers/gcp/index.ts
+++ b/packages/airnode-deployer/src/handlers/gcp/index.ts
@@ -1,6 +1,6 @@
 import * as path from 'path';
 import { Request, Response } from '@google-cloud/functions-framework/build/src/functions';
-import { handlers, logger, utils, providers } from '@api3/airnode-node';
+import { handlers, logger, utils, providers, config } from '@api3/airnode-node';
 import { loadConfig } from '../../utils';
 
 const configFile = path.resolve(`${__dirname}/../../config-data/config.json`);
@@ -51,4 +51,30 @@ export async function processProviderRequests(req: Request, res: Response) {
 
   const body = { ok: true, data: providers.scrub(updatedState) };
   res.status(200).send(body);
+}
+
+export async function testApi(req: Request, res: Response) {
+  // We need to check for an API key manually because GCP HTTP Gateway
+  // doesn't support managing API keys via API
+  const apiKey = req.header('x-api-key');
+  if (!apiKey || apiKey !== config.getEnvValue('HTTP_GATEWAY_API_KEY')) {
+    res.status(401).send({ error: 'Wrong API key' });
+  }
+
+  const { parameters } = req.body;
+  const { endpointId } = req.query;
+
+  if (!endpointId) {
+    res.status(400).send({ error: 'Missing endpointId' });
+    return;
+  }
+
+  const [err, result] = await handlers.testApi(parsedConfig, endpointId as string, parameters);
+  if (err) {
+    res.status(400).send({ error: err.toString() });
+    return;
+  }
+
+  // NOTE: We do not want the user to see {"value": <actual_value>}, but the actual value itself
+  res.status(200).send(result!.value);
 }
--- a/packages/airnode-deployer/terraform/airnode/aws/main.tf
+++ b/packages/airnode-deployer/terraform/airnode/aws/main.tf
@@ -76,7 +76,7 @@ module "testApi" {
   handler            = "index.testApi"
   source_dir         = var.handler_dir
   memory_size        = 256
-  timeout            = 30
+  timeout            = 15
   configuration_file = var.configuration_file
   secrets_file       = var.secrets_file
 
--- a/packages/airnode-deployer/terraform/airnode/gcp/modules/apigateway/main.tf
+++ b/packages/airnode-deployer/terraform/airnode/gcp/modules/apigateway/main.tf
@@ -1,0 +1,74 @@
+resource "random_string" "api_gateway_service_account_id" {
+  length  = 20
+  lower   = true
+  upper   = false
+  special = false
+  number  = false
+}
+
+resource "google_service_account" "api_gateway_service_account" {
+  account_id   = random_string.api_gateway_service_account_id.result
+  display_name = "${var.name}-service-account"
+}
+
+resource "google_project_iam_member" "api_gateway_monitoring_writer_role" {
+  role    = "roles/monitoring.metricWriter"
+  member  = "serviceAccount:${google_service_account.api_gateway_service_account.email}"
+  project = var.project
+}
+
+resource "google_project_iam_member" "api_gateway_logging_writer_role" {
+  role    = "roles/logging.logWriter"
+  member  = "serviceAccount:${google_service_account.api_gateway_service_account.email}"
+  project = var.project
+}
+
+resource "google_cloudfunctions_function_iam_member" "invoker" {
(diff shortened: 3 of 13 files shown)


packages/airnode-deployer/src/handlers/gcp/index.ts
```diff
@@ -52,3 +52,27 @@ export async function processProviderRequests(req: Request, res: Response) {
...
+    return;
+  }
+
+  const [err, result] = await handlers.testApi(parsedConfig, endpointId as string, parameters);
+  if (err) {
+    res.status(400).send({ error: err.toString() });
+    return;
+  }
+
+  // NOTE: We do not want the user to see {"value": <actual_value>}, but the actual value itself
```
andreogle, line 76: Why is this?

andreogle (approved):

packages/airnode-deployer/src/handlers/gcp/index.ts
amarthadan: That was a change introduced by @Siegrift to an AWS HTTP gateway, I just wanted to keep it consistent. I think he'll be able to answer.

Siegrift: It's as the comment say. If you did a request in http gw you would have a response:
```
{"value": {"rawValue": ..., "encodedValue": ..., "values": [...]}}
```

that is unnecessary an you isntead want the reponse flattened, like described in docs https://docs.api3.org/airnode/v0.4/grp-providers/guides/build-an-airnode/http-gateway.html#using-curl

packages/airnode-deployer/src/handlers/gcp/index.ts
```diff
@@ -52,3 +52,27 @@ export async function processProviderRequests(req: Request, res: Response) {
   const body = { ok: true, data: providers.scrub(updatedState) };
   res.status(200).send(body);
 }
+
+export async function testApi(req: Request, res: Response) {
+  const apiKey = req.header('x-api-key');
+  if (!apiKey || apiKey !== config.getEnvValue('HTTP_GATEWAY_API_KEY')) {
```
Siegrift, line 58: This is here because we can't whitelist by this header in the gateway, right? (in aws this is done by aws)

Siegrift: It would deserve a comment.

packages/airnode-deployer/src/handlers/gcp/index.ts
Siegrift: It could be refactored so that this wrapping is removed and the flattened value is only returned.

packages/airnode-deployer/terraform/airnode/gcp/modules/apigateway/main.tf
```diff
@@ -0,0 +1,74 @@
...
+
+resource "google_api_gateway_api_config" "api_gateway_api_config" {
+  provider = google-beta
+
+  api           = google_api_gateway_api.api_gateway_api.api_id
+  # Generating a suffix manually because the one generated with `api_config_id_prefix` is far too long
+  api_config_id = lower("${var.name}-config-${substr(uuid(), 0, 8)}")
+  display_name  = "${var.name}-config"
+
+  openapi_documents {
```
Siegrift, line 50: What is this for?

packages/airnode-deployer/terraform/airnode/gcp/templates/apigateway.yaml.tpl
```diff
@@ -0,0 +1,54 @@
+swagger: "2.0"
```
Siegrift, line 1: I guess this is for the open API file referenced in terraform, right?

Siegrift (commented):
It would be nice (at least for me) to know how this works and why are some things needed (e.g. the openapi spec). It might be worth adding a bit of documentation to the teraform stuff

packages/airnode-deployer/src/handlers/gcp/index.ts
amarthadan: Yes, that's why we have to do it manually here. I'll add a comment.

packages/airnode-deployer/terraform/airnode/aws/main.tf
```diff
@@ -76,7 +76,7 @@ module "testApi" {
   handler            = "index.testApi"
   source_dir         = var.handler_dir
   memory_size        = 256
-  timeout            = 30
+  timeout            = 15
```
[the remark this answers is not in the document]
amarthadan: I set `testApi` handler timeout (both AWS and GCP) to 15 because the only thing it does is call the `callAPI` handler which has a timeout of 10. So, anything higher is unnecessary.

packages/airnode-deployer/terraform/airnode/gcp/modules/apigateway/main.tf
```diff
@@ -0,0 +1,74 @@
...
+  api_id       = lower("${var.name}-api")
+  display_name = "${var.name}-api"
+}
+
+resource "google_api_gateway_api_config" "api_gateway_api_config" {
+  provider = google-beta
+
+  api           = google_api_gateway_api.api_gateway_api.api_id
+  # Generating a suffix manually because the one generated with `api_config_id_prefix` is far too long
+  api_config_id = lower("${var.name}-config-${substr(uuid(), 0, 8)}")
```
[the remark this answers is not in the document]
amarthadan: This is used to uniquely identify an API config. During the deployment of a new version, first, a new config is created and then the old one is destroyed, so, for a few moments, there are two of them, that's why we need to distinguish them.
The ID can have a max length of 64 characters. Taking into consideration our naming convention, I don't really want to make the identifying part longer as it may become a problem. But anyway, the first 8 characters from a UUID should be unique enough to avoid collisions between just two resources.

packages/airnode-deployer/terraform/airnode/gcp/modules/apigateway/main.tf
amarthadan: This is the gateway definition. The OpenAPI definition describes what endpoints of what type should be available on the gateway. Pretty much the same as for AWS.

packages/airnode-deployer/terraform/airnode/gcp/templates/apigateway.yaml.tpl
amarthadan: Yes.

amarthadan:
> It would be nice (at least for me) to know how this works and why are some things needed (e.g. the openapi spec). It might be worth adding a bit of documentation to the teraform stuff

I created an [issue](https://api3dao.atlassian.net/browse/AN-508) for writing such documentation.

packages/airnode-deployer/terraform/airnode/gcp/modules/apigateway/main.tf
Siegrift: I forgot (or didn't know) we have that for AWS as well.

Siegrift (approved):
LGTM

amarthadan:
Not merging this yet since the issue is for v0.5 (I can create a v0.5 branch and put it there but it's the only v0.5 issue so far so just holding off the merge is good enough for now). But moving the issue to testing in Jira.

amarthadan:
We've decided to include this in v0.4 after all so I'm merging it.

PR #791, opened 2022-01-26
Trim error message sent on-chain

dcroote:
Per [AN-484](https://api3dao.atlassian.net/browse/AN-484), removes an `err.message` and introduces trimming to protect against long (expensive) error messages sent on-chain. However, removal of `err.message` did require updating a number of tests in which error context was lost, so please confirm this is the desired behavior.


packages/airnode-node/src/evm/fulfillments/api-calls.ts
```diff
@@ -168,7 +168,7 @@ async function submitFail(
       request.airnodeAddress,
       request.fulfillAddress,
       request.fulfillFunctionId,
-      errorMessage,
+      errorMessage.length > 100 ? errorMessage.substring(0, 97).concat('...') : errorMessage,
```
Siegrift, line 171: Can we maybe move this constant here: https://github.com/api3dao/airnode/blob/an459-string32-bool-to-abi/packages/airnode-node/src/constants.ts ?

Siegrift (approved):
I think the constant should be moved, but otherwise LGTM

Siegrift (commented):
And yeah, we do not want to send error messages (or parts of them) on chain, since they can contain some private information.

packages/airnode-node/src/evm/fulfillments/api-calls.ts
dcroote: Done- thanks for the suggestion

PR #792, opened 2022-01-26
Revert v0.4 contract changes

bbenligiray:
This reverts all the v0.4 contract changes, most notably the addition of the optimizer. After merging this, we will no longer need to redeploy contracts for the v0.4 release.

These changes are all applied in protocol-v1, which we'll use in production.


bbenligiray:
You can go to https://github.com/api3dao/airnode/compare/v0.3.1...revert-v-0-4-contract-changes and Ctrl-F "packages/airnode-protocol" for an easier time

packages/airnode-protocol/test/utils.js
```diff
@@ -7,7 +7,7 @@ function deriveWalletPathFromSponsorAddress(sponsorAddress) {
     const shiftedSponsorAddressBN = sponsorAddressBN.shr(31 * i);
     paths.push(shiftedSponsorAddressBN.mask(31).toString());
   }
-  return `1/${paths.join('/')}`;
+  return `0/${paths.join('/')}`;
```
aquarat, line 10: Changes like this will have to be made in a number of places, I'm just confirming that reverting those changes will be a separate task?
I count that exact path string in at least 11 places in airnode.

aquarat (approved):
LGTM with two queries :)

packages/airnode-protocol/test/utils.js
Siegrift: This one should actually not be changed (it should be 1)

packages/airnode-protocol/README.md
```diff
@@ -4,8 +4,8 @@
 
 > The contracts that implement the Airnode protocols
 
-**_This documents the protocol for v0.3. We have also published and documented the pre-alpha version widely. Pre-alpha
-and v0.3 are very different in implementation and the terminology they use is contradictory. If you are referring to any
+**_This documents the protocol for v0. We have also published and documented the pre-alpha version widely. Pre-alpha and
+v0 are very different in implementation and the terminology they use is contradictory. If you are referring to any
```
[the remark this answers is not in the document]
bbenligiray: Yeah, I figured we don't want to update this at every minor release

packages/airnode-protocol/test/utils.js
bbenligiray: Oops, good find

bbenligiray:
I added OwnableCallForwarder and its tests back because it's used in https://github.com/api3dao/beacon-setup-guide and I don't want people to have to go to the protocol-v1 branch to find it

PR #793, opened 2022-01-26
Remove DynamoDB locks from AWS Terraform recipes

amarthadan:
[Issue AN-499](https://api3dao.atlassian.net/browse/AN-499)


Siegrift (approved):
LGTM

PR #797, opened 2022-01-28
Remove `callApi` call from `testApi` handler

amarthadan:
[Issue AN-513](https://api3dao.atlassian.net/browse/AN-513)


packages/airnode-node/src/api/index.ts
```diff
@@ -0,0 +1,233 @@
...
+  // Each API call is allowed API_CALL_TIMEOUT ms to complete, before it is retried until the
+  // maximum timeout is reached.
+  const adapterConfig: adapter.Config = { timeout: API_CALL_TIMEOUT };
+  // If the request times out, we attempt to call the API again. Any other errors will not result in retries
+  const retryableCall = retryOnTimeout(API_CALL_TOTAL_TIMEOUT, () =>
+    adapter.buildAndExecuteRequest(options, adapterConfig)
+  );
+
+  const [err, res] = await go(() => retryableCall);
+  if (err) {
```
andreogle, line 183: Do you not want to check `!res` here as well?

packages/airnode-node/src/api/index.ts
```diff
@@ -0,0 +1,233 @@
...
+    [log],
+    {
+      success: false,
+      errorMessage: message,
+    },
+  ];
+}
+
+export function verifyTemplateId(
+  config_: Config,
```
andreogle, line 123: Why the underscore at the end of the variable name? Shouldn't it be at the start for ESLint?
```suggestion
  _config: Config,
```

packages/airnode-node/src/api/index.ts
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  if (!template) {
+    const message = `Ignoring Request:${id} as the template could not be found for verification`;
+    const log = logger.pend('ERROR', message);
+    return [
+      [log],
+      {
+        success: false,
+        errorMessage: message,
```
andreogle, line 138: Could save a few lines by moving everything to one line?

```ts
return [[log], { success: false, errorMessage: message }];
```

1 line vs 7 lines

andreogle (approved):
I realise some of this code where I commented was existing but just added some thoughts anyway. Looks good otherwise 👍🏻

Siegrift (changes requested):

packages/airnode-node/src/api/index.test.ts
```diff
@@ -0,0 +1,92 @@
+import * as adapter from '@api3/airnode-adapter';
```
Siegrift, line 1: Is the `api` folder the right place? Maybe better move this folder to handlers?

packages/airnode-node/src/api/index.ts
amarthadan: I'm not sure what happened here TBH :sweat_smile: And how that passed a CI check...

Siegrift: I think eslint is satisfied with both begin and end underscore.

packages/airnode-node/src/api/index.ts
```diff
@@ -0,0 +1,233 @@
...
+  if (templateId === expectedTemplateId) return null;
+
+  const message = `Invalid template ID:${templateId} found for Request:${id}. Expected template ID:${expectedTemplateId}`;
+  const log = logger.pend('ERROR', message);
+  return [[log], { success: false, errorMessage: message }];
+}
+
+function verifyCallApiPayload(payload: CallApiPayload) {
+  const { config, aggregatedApiCall } = payload;
+  const verifications =
```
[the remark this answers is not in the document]
amarthadan: I don't really agree. Currently, when doing an HTTP gateway request, the flow goes through 3 verification function calls just to figure out it should not do anything. I know it probably won't have an effect on performance but still.

If I look at the current code (line 153) I see that all requests go through 3 validations, which is technically not true, because there are exceptions for `testing-gateway` requests in the validation.
If I look at the proposed code, I immediately see that `testing-gateway` requests are treated a bit differently from normal requests.

But I agree that the current solution is better type-wise and it will have to be changed if verification will be added for `testing-gateway` requests.

I don't really feel strongly about this so I will happily revert it if you insist.

packages/airnode-node/src/api/index.test.ts
amarthadan: I didn't really find the right place that's why I created an `api` directory. I was kinda following the pattern used in `initializeProvider` and `processTransactions` handlers - there's a "stub" code in `handlers` directory calling an actual code from `evm` directory. Since API calls are not related to EVM I created a separate directory.

packages/airnode-node/src/api/index.test.ts
```diff
@@ -0,0 +1,92 @@
...
+import { RequestErrorMessage } from '../types';
+import * as fixtures from '../../test/fixtures';
+import { callApi } from '.';
+
+describe('callApi', () => {
+  const OLD_ENV = process.env;
+
+  beforeEach(() => {
```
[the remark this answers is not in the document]
amarthadan: It looks like they're not needed, I'll remove them.

packages/airnode-node/src/api/index.test.ts
Siegrift: Let's keep it this way - we can think about moving this to some other place later.

packages/airnode-node/src/api/index.ts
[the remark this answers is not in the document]
Siegrift: I think it would be better.

Siegrift (approved):
LGTM. I'd still like the validations refactor to be reverted though.

Siegrift:
btw. do not forget about changeset

amarthadan:
I'm merging this to a new branch `v0.5` so it won't pollute `master` before `v0.4` release.

PR #799, opened 2022-01-30
Revert In-lined Gas Price Call

### LINKED ISSUE #798: Arbitrum Integration Issues
## Description

While testing Airnode on Arbitrum some small issues came up.

## Version

0.4.0 Airnode

## Present Behaviour

The first error occured on getting legacy gas price and is shown below. I checked getGasPrice on Arbitrum through Ethers and it worked fine so I just tweaked the code a little to work.

```
TypeError: Cannot read property 'getNetwork' of undefined
    at /home/ec2-user/arbi/node_modules/@ethersproject/providers/src.ts/base-provider.ts:1194:20
    at step (/home/ec2-user/arbi/node_modules/@ethersproject/providers/lib/base-provider.js:48:23)
    at Object.next (/home/ec2-user/arbi/node_modules/@ethersproject/providers/lib/base-provider.js:29:53)
    at /home/ec2-user/arbi/node_modules/@ethersproject/providers/lib/base-provider.js:23:71
    at new Promise (<anonymous>)
    at __awaiter (/home/ec2-user/arbi/node_modules/@ethersproject/providers/lib/base-provider.js:19:12)
    at BaseProvider.getGasPrice (/home/ec2-user/arbi/node_modules/@ethersproject/providers/lib/base-provider.js:1271:16)
    at /home/ec2-user/arbi/packages/airnode-node/src/utils/promise-utils.ts:67:30
    at /home/ec2-user/arbi/node_modules/@lifeomic/attempt/src/index.js:121:17
    at new Promise (<anonymous>)
```

The second error occured on transaction fulfillment and was a "not enough funds for gas" error when in fact my wallet had plenty, it was because the hardcoded gas limit in the api-calls.ts file was too low. I changed it from 500k to 5 million and the fulfillment was successful. 

## Steps to reproduce

I created a branch in Airnode called Arbitrum where all the code changes I needed to make can be seen

aquarat:
This adds a function wrapper around the gas pricing call, without this the underlying code references the wrong context.

Subject to testing by @jgcogsystematictrading

Fixes #798

### DIFF
--- a/packages/airnode-node/src/evm/gas-prices.ts
+++ b/packages/airnode-node/src/evm/gas-prices.ts
@@ -14,7 +14,7 @@ export const parsePriorityFee = ({ value, unit }: PriorityFee) => ethers.utils.p
 const getLegacyGasPrice = async (options: FetchOptions): Promise<LogsData<GasTarget | null>> => {
   const { provider } = options;
 
-  const [err, gasPrice] = await go(provider.getGasPrice, { retries: 1, timeoutMs: DEFAULT_RETRY_TIMEOUT_MS });
+  const [err, gasPrice] = await go(() => provider.getGasPrice(), { retries: 1, timeoutMs: DEFAULT_RETRY_TIMEOUT_MS });
   if (err || !gasPrice) {
     const log = logger.pend('ERROR', 'All attempts to get legacy gasPrice from provider failed');
     return [[log], null];


packages/airnode-node/src/evm/gas-prices.ts
```diff
@@ -14,7 +14,7 @@ export const parsePriorityFee = ({ value, unit }: PriorityFee) => ethers.utils.p
 const getLegacyGasPrice = async (options: FetchOptions): Promise<LogsData<GasTarget | null>> => {
   const { provider } = options;
 
-  const [err, gasPrice] = await go(provider.getGasPrice, { retries: 1, timeoutMs: DEFAULT_RETRY_TIMEOUT_MS });
+  const [err, gasPrice] = await go(() => provider.getGasPrice(), { retries: 1, timeoutMs: DEFAULT_RETRY_TIMEOUT_MS });
```
Siegrift, line 17: Mhh, I don't understand how this solves the issue - this looks like noop to me.

jgcogsystematictrading:
This change fixed the issue correctly

packages/airnode-node/src/evm/gas-prices.ts
aquarat: We're learning ;)
The failure is to do with the context in which the function is called and that this function is a class-based function (relies on a field in the `this` object). The field in question is the RPC URL.
Most of our stuff is pure/functional/not class-based, but ethers is class based.
The automated tests didn't pick up on it because we mock the return value of the `getGasPrice` function. Only an E2E test against a node instance would have picked up on it *if* both an eip1559 and legacy scenario were used.

Siegrift: Mhh, I'll need to lookup the `this` rules again...

andreogle: Yeah this was also really confusing to me. I still don't really get how `this` will be different between `provider.getGasPrice` and `() => provider.getGasPrice()` 🤔 

JS is a weird language 😄

Siegrift: From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this

```
In strict mode, however, if the value of this is not set when entering an execution context, it remains as undefined, as shown in the following example:
```

The `this` in javascript is based on a dynamic scope, not lexical. E.g. this works:
```
'use strict';

function fn() {
  console.log(this.x);
}

const obj = { x: 'abc' };
obj.fn = fn;
// By calling the fn with "obj." the this is bound to the "obj"
obj.fn();
```

Siegrift (approved):

PR #806, opened 2022-02-02
Add unit test for error message trimming

dcroote:
[AN-523](https://api3dao.atlassian.net/browse/AN-523)

### DIFF
--- a/packages/airnode-node/src/evm/fulfillments/api-calls.test.ts
+++ b/packages/airnode-node/src/evm/fulfillments/api-calls.test.ts
@@ -19,6 +19,7 @@ import * as fixtures from '../../../test/fixtures';
 import * as wallet from '../wallet';
 import { GasTarget, RequestErrorMessage, RequestStatus } from '../../types';
 import { AirnodeRrp } from '../contracts';
+import { MAXIMUM_ONCHAIN_ERROR_LENGTH } from '../../constants';
 
 const createAirnodeRrpFake = () => new ethers.Contract('address', ['ABI']) as unknown as AirnodeRrp;
 const config = fixtures.buildConfig();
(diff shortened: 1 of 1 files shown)


packages/airnode-node/src/evm/fulfillments/api-calls.test.ts
```diff
@@ -477,6 +478,51 @@ describe('submitApiCall', () => {
...
+        const provider = new ethers.providers.JsonRpcProvider();
+        const longError = 'This very long error message should get trimmed'.repeat(10);
+        const trimmedError = longError.substring(0, MAXIMUM_ONCHAIN_ERROR_LENGTH - 3).concat('...');
+        failMock.mockResolvedValueOnce({ hash: '0xfailtransaction' });
+        const apiCall = fixtures.requests.buildApiCall({
+          errorMessage: longError,
+          status: RequestStatus.Errored,
+          nonce: 5,
+        });
+        const [logs, err, data] = await apiCalls.submitApiCall(createAirnodeRrpFake(), apiCall, {
```
Siegrift, line 494: Add `\n` for clarity before and after this call.

We do not have a strict convention for this, but one of the generic conventions for tests is "Arange/act/assert" where each section is separated by a new line (optionally preceeded by comment, which I hate).

Anyway, separating the sections is nice to do as the whole test reads better. But ofc you can create more groups when necessary or comment some non trivial stuff.

Siegrift (approved):
LGTM

packages/airnode-node/src/evm/fulfillments/api-calls.test.ts
dcroote: Makes sense, thanks.

PR #807, opened 2022-02-02
Revise timeouts and resource limits based on stress testing

aquarat:
Revised timeouts based on [the recent stress testing](https://monitor.api3.org/d/qW6nM2xnk/v0-4-0?orgId=1):
- The report is still in progress but the graphs speak for themselves. This can be discussed during the call.
- GCP is less efficient than AWS, hence the larger RAM allocations (which impacts processing speed).
- Throughput with this configuration is greater than 200 requests per minute (~240 on AWS).
- There's a need for larger API timeouts, hence a bias towards callApi having more processing time.


packages/airnode-deployer/terraform/airnode/aws/main.tf
```diff
@@ -9,7 +9,7 @@ module "initializeProvider" {
   handler            = "index.initializeProvider"
   source_dir         = var.handler_dir
   memory_size        = 768
-  timeout            = 17
+  timeout            = 20
```
andreogle, line 12: This will have conflicts with https://github.com/api3dao/airnode/pull/808 All of the workers have been combined into a single "run" worker. I don't know how this will affect memory_size and overall performance though

.changeset/pretty-tools-bathe.md
```diff
@@ -0,0 +1,6 @@
+---
+'@api3/airnode-deployer': patch
```
andreogle, line 2: I would say this should be `minor` instead? `patch` should just be for bug fixes

packages/airnode-node/src/constants.ts
```diff
@@ -22,22 +22,22 @@ export const BLOCK_MIN_CONFIRMATIONS = 0;
 export const CONVENIENCE_BATCH_SIZE = 10;
 
 // The default amount of time before a "retryable" promise is timed out and retried
-export const DEFAULT_RETRY_TIMEOUT_MS = 5_000;
+export const DEFAULT_RETRY_TIMEOUT_MS = 30_000;
```
andreogle, line 25: Why change this one? I think this is used in quite a lot of places throughout the app where a promise is retried and could change behaviour quite significantly. I would rather call the places we want to set something other than the default of 5 seconds

packages/airnode-node/src/constants.ts
```diff
@@ -1,8 +1,8 @@
 // The maximum TOTAL time an API call has before it is timed out (including retries).
-export const API_CALL_TOTAL_TIMEOUT = 10_000;
+export const API_CALL_TOTAL_TIMEOUT = 30_000;
 
 // The maximum time an API call has before it is timed out (and retried).
-export const API_CALL_TIMEOUT = 10_000;
+export const API_CALL_TIMEOUT = 30_000;
```
andreogle, line 5: I didn't realise that these times were the same (`API_CALL_TIMEOUT` and `API_CALL_TOTAL_TIMEOUT`). The original idea was that the worker itself gets (for example) 30 seconds to run, while the call gets less time - both per call and in total, as the name suggests. That way, if the call hangs for whatever reason, it will at least get retried once.

Now, it looks like we just execute a single API call and hope it succeeds? Not sure how big of a deal this is, just thought I would point it out

packages/airnode-node/src/constants.ts
Siegrift: I agree, but I'd actually refactor this constant to be more contextual. From what I see we use it only to limit blockchain provider calls. Maybe renaming it to `BLOCKCHAIN_PROVIDER_TOTAL_TIMEOUT` ?

Actually I think we should not use go utils for this at all and use https://github.com/lifeomic/attempt#readme instead (which we already use).

I've created https://api3dao.atlassian.net/browse/AN-526

packages/airnode-node/src/constants.ts
```diff
@@ -22,22 +22,22 @@ export const BLOCK_MIN_CONFIRMATIONS = 0;
...
-export const EVM_PROVIDER_TIMEOUT = 22_000;
+export const EVM_PROVIDER_TIMEOUT = 20_000;
 
 // The maximum amount of time the "initialize provider" worker is allowed before being timed out
-export const WORKER_PROVIDER_INITIALIZATION_TIMEOUT = 17_000;
+export const WORKER_PROVIDER_INITIALIZATION_TIMEOUT = 20_000;
 
 // The maximum amount of time the "process requests" worker is allowed before being timed out
-export const WORKER_PROVIDER_PROCESS_REQUESTS_TIMEOUT = 32_000;
+export const WORKER_PROVIDER_PROCESS_REQUESTS_TIMEOUT = 10_000;
```
Siegrift, line 37: I find it strange that 10s is enough for this, but if the stress test says so, we can have it.

packages/airnode-node/src/constants.ts
aquarat: Hey
As I understand it some APIs could take a long time to respond, so I thought making them the same and a high-ish value would serve the greatest selection of potential APIs. Worst case the API gets re-called a minute later, and assuming the first failure was due to transient network events, succeeds and results in a fulfilment.

packages/airnode-deployer/terraform/airnode/aws/main.tf
```diff
@@ -23,10 +23,10 @@ module "callApi" {
   name               = "${local.name_prefix}-callApi"
   handler            = "index.callApi"
   source_dir         = var.handler_dir
-  timeout            = 10
+  timeout            = 30
   configuration_file = var.configuration_file
   secrets_file       = var.secrets_file
-  memory_size        = 256
+  memory_size        = 768
```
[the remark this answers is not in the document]
aquarat: AWS charges based on Memory multiplied by Time usage, so increasing RAM uses more free tier capacity.
The thing is that memory capacity non-linearly impacts compute performance. The sweet spot for single-threaded performance is 768 MB, so compute heavy functions get this and `startCoordinator` gets 256 MB.
On GCP `startCoordinator` underperforms _a lot_ if it's given only 256 MB, so it gets 512 MB.
GCP does not have a 768 MB option.

packages/airnode-node/src/constants.ts
aquarat: It actually needs much less than 10s... most requests were processed in under 3s :) 
I was going to submit an update to this PR to make it 6 seconds (and give 4 more seconds to `initializeProvider`).

.changeset/pretty-tools-bathe.md
aquarat: I will update it :)

packages/airnode-node/src/constants.ts
aquarat: I'll revert this change.

packages/airnode-deployer/terraform/airnode/aws/main.tf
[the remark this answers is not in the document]
Siegrift: I think you wrote this in the analysis report, right?

packages/airnode-deployer/terraform/airnode/aws/main.tf
```diff
@@ -39,7 +39,7 @@ module "processProviderRequests" {
   handler            = "index.processProviderRequests"
   source_dir         = var.handler_dir
   memory_size        = 768
-  timeout            = 32
+  timeout            = 6
```
Siegrift, line 42: This feels too little to me. I would not take the stress test as the absolute guideline for the timeouts (it's still a limited test scenario). I'd maybe go with 10 here (so that it has a little more time)

Siegrift (approved):
I had one more comment, but otherwise LGTM.

packages/airnode-deployer/terraform/airnode/aws/main.tf
aquarat: All good, I agree that 10 is safer and certainly would be necessary for slow nodes.
This timing still gives > 250 requests per cycle.

packages/airnode-node/src/constants.ts
andreogle: Yeah I figured that's probably the better option. We would prefer to support APIs that have very slow response times over a second (retried) API call in the same cycle.

aquarat:
I've tested this exact commit and I was able to get 238 requests/min on AWS and 250 requests per minute on GCP.
Airnode could do more, but the bottleneck now is the target chain (Ropsten/HardHat, they both have ceilings).
Next step, when we have time again is probably a small PoA test network based on Parity/OpenEthereum.

PR #808, opened 2022-02-02
Merge serverless handlers to share the concurrency reservation

amarthadan:
[Issue AN-514](https://api3dao.atlassian.net/browse/AN-514)

### DIFF
--- a/packages/airnode-deployer/src/handlers/aws/index.ts
+++ b/packages/airnode-deployer/src/handlers/aws/index.ts
@@ -26,20 +26,18 @@ export async function startCoordinator() {
 }
 
 export async function run(payload: WorkerPayload): Promise<AWSLambda.APIGatewayProxyResult> {
-  if (payload.functionName === 'initializeProvider') {
-    return initializeProvider(payload);
+  switch (payload.functionName) {
+    case 'initializeProvider':
+      return initializeProvider(payload);
+    case 'callApi':
+      return callApi(payload);
+    case 'processTransactions':
+      return processTransactions(payload);
   }
-  if (payload.functionName === 'callApi') {
-    return callApi(payload);
-  }
-  if (payload.functionName === 'processTransactions') {
-    return processTransactions(payload);
-  }
-
-  return { statusCode: 400, body: JSON.stringify({ error: 'Unknown function', payload }) };
 }
 
 // TODO: Refactor handlers so they are common for all the cloud providers
+// https://api3dao.atlassian.net/browse/AN-527
 
 async function initializeProvider(payload: InitializeProviderPayload) {
   const stateWithConfig = { ...payload.state, config: parsedConfig };
--- a/packages/airnode-deployer/src/handlers/aws/index.ts
+++ b/packages/airnode-deployer/src/handlers/aws/index.ts
@@ -1,5 +1,15 @@
 import * as path from 'path';
-import { handlers, logger, utils, providers, WorkerResponse } from '@api3/airnode-node';
+import {
+  handlers,
+  logger,
+  utils,
+  providers,
+  WorkerResponse,
+  WorkerPayload,
+  InitializeProviderPayload,
+  ProcessTransactionsPayload,
+  CallApiPayload,
+} from '@api3/airnode-node';
 import { loadConfig } from '../../utils';
 
 const configFile = path.resolve(`${__dirname}/../../config-data/config.json`);
@@ -15,8 +25,24 @@ export async function startCoordinator() {
   return { statusCode: 200, body: encodeBody(response) };
 }
 
-export async function initializeProvider(event: any) {
-  const stateWithConfig = { ...event.state, config: parsedConfig };
+export async function run(payload: WorkerPayload): Promise<AWSLambda.APIGatewayProxyResult> {
+  if (payload.functionName === 'initializeProvider') {
+    return initializeProvider(payload);
+  }
+  if (payload.functionName === 'callApi') {
+    return callApi(payload);
+  }
+  if (payload.functionName === 'processProviderRequests') {
+    return processProviderRequests(payload);
+  }
+
+  return { statusCode: 400, body: JSON.stringify({ error: 'Unknown function', payload }) };
+}
+
+// TODO: Refactor handlers so they are common for all the cloud providers
+
+async function initializeProvider(payload: InitializeProviderPayload) {
+  const stateWithConfig = { ...payload.state, config: parsedConfig };
 
   const [err, initializedState] = await utils.go(() => handlers.initializeProvider(stateWithConfig));
   if (err || !initializedState) {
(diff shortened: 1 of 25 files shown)


packages/airnode-deployer/terraform/airnode/aws/main.tf
```diff
@@ -2,41 +2,11 @@
...
-module "processProviderRequests" {
-  source = "./modules/function"
-
-  name               = "${local.name_prefix}-processProviderRequests"
-  handler            = "index.processProviderRequests"
+  name               = "${local.name_prefix}-run"
+  handler            = "index.run"
   source_dir         = var.handler_dir
   memory_size        = 768
   timeout            = 32
```
andreogle, line 12: Does this mean they all have the same timeout?

packages/airnode-node/src/types.ts
```diff
@@ -287,11 +287,26 @@ export interface WorkerOptions {
...
 }
 
-export type WorkerFunctionName = 'initializeProvider' | 'callApi' | 'processProviderRequests';
+export interface InitializeProviderPayload {
+  readonly functionName: 'initializeProvider';
+  readonly state: ProviderState<EVMProviderState>;
+}
+
+export interface ProcessTransactionsPayload {
+  readonly functionName: 'processProviderRequests';
```
andreogle, line 296: Maybe a good time to rename to `processTransactions`? The difference has bothered me for a while
```suggestion
  readonly functionName: 'processTransactions';
```

andreogle (approved):
This change seems so obvious in hindsight 😅 

Doesn't it also have the added benefits of:
1. All cloud provider logs are now in one place?
2. We can easily rename/change functions without having to update existing functions

?

packages/airnode-deployer/terraform/airnode/aws/main.tf
amarthadan: Unfortunately, yes. I used the highest timeout of those three handlers. But as far as I know, we still have a "manual" timeout in the code, so, it shouldn't be a problem.

amarthadan:
> Doesn't it also have the added benefits of:
> 
>     1. All cloud provider logs are now in one place?

Most of them. `startCoordinator` and `testApi` (for HTTP gateway) still have separate logs.

>     2. We can easily rename/change functions without having to update existing functions

Yes, we can. We probably could have done that even before but nobody tried it :smile:

packages/airnode-deployer/terraform/airnode/aws/main.tf
andreogle: Ok cool. We'll probably want to test this on both cloud providers then

amarthadan: I think @aquarat tested that when he was doing stress tests. Or do you mean writing additional tests for that?

packages/airnode-node/src/types.ts
```diff
@@ -287,11 +287,26 @@ export interface WorkerOptions {
...
 }
 
-export type WorkerFunctionName = 'initializeProvider' | 'callApi' | 'processProviderRequests';
+export interface InitializeProviderPayload {
+  readonly functionName: 'initializeProvider';
+  readonly state: ProviderState<EVMProviderState>;
+}
+
+export interface ProcessTransactionsPayload {
+  readonly functionName: 'processTransactions';
```
andreogle, line 296: One other thing I noticed is the timeout constant is still called `WORKER_PROVIDER_PROCESS_REQUESTS_TIMEOUT` in constants.ts which might be confusing

packages/airnode-deployer/terraform/airnode/aws/main.tf
andreogle: Actually I think that should be fine - don't think we need additional tests. It's nice having the timeout in the function because then it means that there should just be a single place where the values need to be updated.

andreogle (approved):

packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -53,7 +79,15 @@ export async function processProviderRequests(event: any) {
   return { statusCode: 200, body };
 }
 
-export async function testApi(event: any) {
+export async function testApi(event: AWSLambda.APIGatewayProxyEvent): Promise<AWSLambda.APIGatewayProxyResult> {
+  if (!event.body) {
```
Siegrift, line 83: Nit: Move this to a separate validation function (as we will use for GCP as well)

packages/airnode-deployer/src/handlers/gcp/index.ts
```diff
@@ -12,8 +21,29 @@ export async function startCoordinator(_req: Request, res: Response) {
   res.status(200).send(response);
 }
 
-export async function initializeProvider(req: Request, res: Response) {
-  const stateWithConfig = { ...req.body.state, config: parsedConfig };
+export async function run(req: Request, res: Response) {
+  const payload = req.body;
+
+  if (payload.functionName === 'initializeProvider') {
```
Siegrift, line 27: use TS checked switch. You can also use the `return fnCall()` as the return value is not important for this function.

packages/airnode-node/src/workers/index.ts
```diff
@@ -12,6 +12,16 @@ export function spawn(params: WorkerParameters): Promise<WorkerResponse> {
       return gcp.spawn(params);
 
     case 'local':
-      return localHandlers[params.functionName](params.payload);
+      if (params.payload.functionName === 'initializeProvider') {
```
Siegrift, line 15: TS checked switch

Siegrift (changes requested):
LGTM, but I think we can do even better.

packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -15,8 +25,22 @@ export async function startCoordinator() {
...
+export async function run(payload: WorkerPayload): Promise<AWSLambda.APIGatewayProxyResult> {
+  switch (payload.functionName) {
+    case 'initializeProvider':
+      return initializeProvider(payload);
+    case 'callApi':
+      return callApi(payload);
+    case 'processTransactions':
+      return processTransactions(payload);
+  }
+  return { statusCode: 400, body: JSON.stringify({ error: 'Unknown function', payload }) };
```
amarthadan, line 37: I know this is an unreachable statement according to TS, but since it's a payload of a serverless function that can contain basically anything I'd leave it there.

packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -15,8 +25,24 @@ export async function startCoordinator() {
...
+    return callApi(payload);
+  }
+  if (payload.functionName === 'processTransactions') {
+    return processTransactions(payload);
+  }
+
+  return { statusCode: 400, body: JSON.stringify({ error: 'Unknown function', payload }) };
+}
+
+// TODO: Refactor handlers so they are common for all the cloud providers
```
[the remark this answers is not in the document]
amarthadan: Logged as https://api3dao.atlassian.net/browse/AN-527

packages/airnode-deployer/src/handlers/aws/index.ts
Siegrift: Not sure I understand. How is it posible to cintain anything? It's ustriggering the lambda functions

amarthadan: Yes, it is, but it's not like calling a typed function, TS won't help us much. If we make a mistake and send an invalid function name nothing will catch that really. That's why I want to keep the check.

Siegrift: I see. But you can cast the function name so that TS knows its type. I think having TS tell you to update this function when you add a new lambda its better to just preventing invalid fn name

packages/airnode-deployer/src/handlers/aws/index.ts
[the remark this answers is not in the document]
Siegrift: Do you want to remove the TODO? It's good idea to link Jira issue there directly in cae you wanna keep it.

packages/airnode-node/src/workers/index.ts
```diff
@@ -12,6 +12,13 @@ export function spawn(params: WorkerParameters): Promise<WorkerResponse> {
       return gcp.spawn(params);
 
     case 'local':
-      return localHandlers[params.functionName](params.payload);
+      switch (params.payload.functionName) {
```
Siegrift, line 15: You needed to rewrite this because of TS complaints right?

amarthadan: Yes.

PR #809, opened 2022-02-04
Add relay security schemes example integration

dcroote:
[AN-511](https://api3dao.atlassian.net/browse/AN-511)

Note the name `relay-security-schemes` was chosen as this is the nomenclature [used in the docs](https://docs.api3.org/airnode/v0.4/grp-providers/guides/build-an-airnode/api-security.html#supported-security-schemes).

Ignore the failing docs link check in CI- this will pass once merged.

Here is how the output looks:

```
> yarn make-request
yarn run v1.22.11
Making request...
Waiting for fulfillment...
Request fulfilled
The following was successfully relayed:
    requesterAddress: 0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512
    sponsorAddress: 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
    sponsorWalletAddress: 0x68F7F1197a9FD5f2EC79B4eBff2879A1dc1fe78f
    chainId: 31337
    chainType: evm
Done in 20.31s.
```

### DIFF
--- a/packages/airnode-examples/integrations/relay-security-schemes/README.md
+++ b/packages/airnode-examples/integrations/relay-security-schemes/README.md
@@ -2,9 +2,9 @@
 
 This example highlights how Airnode can relay request metadata, like the chain ID or sponsor address, to the API
 endpoint via security schemes. In order to demonstrate this functionality in a realistic way, the API endpoint used is
-httpbin.org, which simply responds with request headers and query parameters in JSON format. This behavior allows
-request metadata returned in the API response to be fulfilled by Airnode on-chain as proof that the metadata is being
-relayed as expected.
+[httpbin.org](https://httpbin.org/), which simply responds with request headers and query parameters in JSON format.
+This behavior allows request metadata returned in the API response to be fulfilled by Airnode on-chain as proof that the
+metadata is being relayed as expected.
 
 The following security scheme types illustrate the request metadata being relayed:
 
--- a/packages/airnode-examples/integrations/relay-security-schemes/README.md
+++ b/packages/airnode-examples/integrations/relay-security-schemes/README.md
@@ -1,0 +1,18 @@
+# Relay security schemes example integration
+
+This example highlights how Airnode can relay request metadata, like the chain ID or sponsor address, to the API
+endpoint via security schemes. In order to demonstrate this functionality in a realistic way, the API endpoint used is
+httpbin.org, which simply responds with request headers and query parameters in JSON format. This behavior allows
+request metadata returned in the API response to be fulfilled by Airnode on-chain as proof that the metadata is being
+relayed as expected.
+
+The following security scheme types illustrate the request metadata being relayed:
+
+- relayRequesterAddress
+- relaySponsorAddress
+- relaySponsorWalletAddress
+- relayChainId
+- relayChainType
+
+For more information on supported security schemes, refer to the
+[docs](https://docs.api3.org/airnode/latest/grp-providers/guides/build-an-airnode/api-security.html).
--- a/packages/airnode-examples/integrations/relay-security-schemes/README.md
+++ b/packages/airnode-examples/integrations/relay-security-schemes/README.md
@@ -6,7 +6,8 @@ endpoint via security schemes. In order to demonstrate this functionality in a r
 This behavior allows request metadata returned in the API response to be fulfilled by Airnode on-chain as proof that the
 metadata is being relayed as expected.
 
-The following security scheme types illustrate the request metadata being relayed:
+The following security scheme types illustrate the request metadata being relayed and, as shown in `config.json`,
+demonstrate how the value can be sent via `query`, `header`, and `cookie`:
 
 - relayRequesterAddress
 - relaySponsorAddress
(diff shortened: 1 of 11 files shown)


packages/airnode-examples/integrations/relay-security-schemes/README.md
```diff
@@ -0,0 +1,18 @@
+# Relay security schemes example integration
+
+This example highlights how Airnode can relay request metadata, like the chain ID or sponsor address, to the API
+endpoint via security schemes. In order to demonstrate this functionality in a realistic way, the API endpoint used is
+httpbin.org, which simply responds with request headers and query parameters in JSON format. This behavior allows
```
Siegrift, line 5: Maybe link that service?

packages/airnode-examples/integrations/relay-security-schemes/config.example.json
```diff
@@ -0,0 +1,135 @@
...
+              "type": "relaySponsorAddress",
+              "name": "sponsorAddress"
+            },
+            "relaySponsorWalletAddress": {
+              "in": "query",
+              "type": "relaySponsorWalletAddress",
+              "name": "sponsorWalletAddress"
+            },
+            "relayChainId": {
+              "in": "query",
```
Siegrift, line 88: Nit: Consider using cookie? (Not sure if the mock API returns that as well)

packages/airnode-examples/integrations/relay-security-schemes/create-config.ts
```diff
@@ -0,0 +1,150 @@
...
+              fixed: 'address,address,address,uint256,string32',
+            },
+            {
+              name: '_path',
+              fixed:
+                'headers.Requesteraddress,args.sponsorAddress,args.sponsorWalletAddress,args.chainId,args.chainType',
+            },
+            {
+              name: '_times',
+              fixed: '',
```
Siegrift, line 134: Mhh, does this work? I think it should error out and the correct value should be `,,,,`

Siegrift (approved):
LGTM, but the `_times` value looks suspicious to me.

packages/airnode-examples/integrations/relay-security-schemes/config.example.json
dcroote: I can if there is a way to parse `31337` out of this httpbin.org JSON response. My question is what should `_path` be? For example `headers.Cookie.???`? The challenge is the `_type` for `chainId` that I have is `uint256`. I suppose I could switch it to `string32` and use just `headers.Cookie`, but then you have to decode and split the text on `=` and `;` which feels like a hack.
```json
"headers": {
  "Accept":"application/json, text/plain, */*",
  "Cookie":"chainId=31337;",
  "Host":"httpbin.org",
```

packages/airnode-examples/integrations/relay-security-schemes/create-config.ts
dcroote: Yea that actually does work. I've done some more testing:
Works in `config.json`:
- `"fixed": ""`
- `"fixed": ",,,,"`
- `"fixed": ",,,10,"`  (since the 4th parameter is chainId, this returns `313370`)

Does not work (more to prove I'm testing correctly):
- `"fixed": ",,,,999"`  (since the 5th parameter is not an int type)

I've gone ahead and updated it to `,,,,` anyway as a precaution

packages/airnode-examples/integrations/relay-security-schemes/config.example.json
dcroote: Actually since `chainType` is `string32` already anyway, I'll use cookie there.

dcroote: Ok I implemented cookie for `relayChainType`. I've updated the README to mention values being passed by `query`, `header`, and `cookie`.

dcroote:
@Siegrift  - the [unit tests were actually failing](https://github.com/api3dao/airnode/runs/5075780354?check_suite_focus=true#step:4:225) because the validator hadn't been updated to expect security schemes `relaySponsorAddress` and `relaySponsorWalletAddress`. See the latest commit in this PR for the fix.

Siegrift (approved):
Oh, totally forgot about validator. Thanks for this.

packages/airnode-examples/integrations/relay-security-schemes/create-config.ts
Siegrift: Mhh, it seems this doesn't work then https://github.com/api3dao/airnode/blob/master/packages/airnode-adapter/src/response-processing/extraction.ts#L87

can you investigate? I find it strange, since there are even tests for it https://github.com/api3dao/airnode/blob/master/packages/airnode-adapter/src/response-processing/extraction.test.ts#L197

dcroote: I'll merge this and investigate separately as I've already updated it to `,,,,` as a precaution, but interestingly none of the tests in the block you linked to include `_times`.

Siegrift: > I'll merge this and investigate separately

Makes sense.

> but interestingly none of the tests in the block you linked to include _times.

Yeah, but the logic is that it checks the number of values in `_type` and enforces _path and _times values to have same length - which seems correct to me.

PR #811, opened 2022-02-07
Clarify the need to build airnode-examples

Siegrift:
https://api3dao.atlassian.net/browse/AN-511

Feel free to close if not applicable.

### DIFF
--- a/packages/airnode-examples/README.md
+++ b/packages/airnode-examples/README.md
@@ -277,6 +277,8 @@ This will use the deployer to remove the Airnode lambdas from the cloud provider
 
 ## For developers
 
+When pulling or creating a new integration you need to run `yarn build` to create necessary contract artifacts.
+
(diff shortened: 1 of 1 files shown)


dcroote (approved):
While it still might be missed, I think it's worth adding.

dcroote:
Actually I also think the user-facing `Setup` could be made simpler i.e. just give them the commands rather than them having to follow a link to "follow the repository instructions".
```markdown
## Setup
1. Make sure you have `yarn` installed.
2. At the **repository root**, run `yarn run bootstrap` followed by `yarn build`
3. If you want to run Airnode as a docker container, you'll also need to have `docker` installed.
```

Siegrift:
> Actually I also think the user-facing Setup could be made simpler i.e. just give them the commands rather than them having to follow a link to "follow the repository instructions".

I dislike that since it's basically copy paste of the commands in the root. It could go out of sync if the root command change.

PR #814, opened 2022-02-09
Add concurrency reservation for HTTP gateway serverless function

amarthadan:
[Issue AN-522](https://api3dao.atlassian.net/browse/AN-522)


.changeset/itchy-cycles-develop.md
```diff
@@ -0,0 +1,8 @@
+---
+'@api3/airnode-deployer': minor
+'@api3/airnode-examples': minor
+'@api3/airnode-node': minor
+'@api3/airnode-validator': minor
+---
+
+Add option to set maximal concurrency for HTTP gateway
```
Siegrift, line 8: [Very nit](https://www.quora.com/What-are-the-differences-between-%E2%80%9Cmaximum%E2%80%9D-and-%E2%80%9Cmaximal%E2%80%9D-Especially-in-scientific-papers#:~:text=K%20answer%20views-,Originally%20Answered%3A%20What%20is%20the%20difference%20between%20maximum%20and%20maximal,or%20a%20noun%20or%20adverb.): 

```suggestion
Add option to set maximum concurrency for HTTP gateway
```

packages/airnode-deployer/config/config.json.example
```diff
@@ -33,11 +33,12 @@
...
     "httpGateway": {
       "enabled": true,
-      "apiKey": "${HTTP_GATEWAY_API_KEY}"
+      "apiKey": "${HTTP_GATEWAY_API_KEY}",
+      "maxConcurrency": 20
     },
     "logFormat": "plain",
     "logLevel": "INFO",
-    "nodeVersion": "0.4.0",
+    "nodeVersion": "0.5.0",
```
Siegrift, line 41: Shouldn't this be changed for all `nodeVersion` we have in the project at the begginning of v0.5 (and validator templates copied)?

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -161,8 +161,11 @@ async function terraformAirnodeManage(
     'no-color',
   ];
 
-  if (httpGatewayApiKey) {
-    commonArguments.push(['var', 'api_key', httpGatewayApiKey]);
+  if (httpGateway?.enabled) {
```
Siegrift, line 164: This function handles both Airnode deploy and removal - It would be nicer if there were two functions for this and they would use a function for common parts (common arguments).

You could also better TS type this.

packages/airnode-deployer/terraform/airnode/aws/variables.tf
```diff
@@ -38,3 +38,8 @@ variable "api_key" {
   type        = string
   default     = null
 }
+
+variable "api_max_concurrency" {
+  description = "Maximum amount of concurrent executions for Airnode Test Gateway Lambda"
+  default = -1
```
Siegrift, line 44: Shouldn't this be 0?

packages/airnode-deployer/config/config.json.example
Siegrift: EDIT: I see you are doing this in this PR, but shouldn't also these change:
- https://github.com/api3dao/airnode/blob/concurrency-reservation-gateway/packages/airnode-validator/exampleSpecs/config.specs.json#L32
- https://github.com/api3dao/airnode/blob/concurrency-reservation-gateway/packages/airnode-validator/exampleSpecs/secrets.env#L4
- https://github.com/api3dao/airnode/blob/concurrency-reservation-gateway/packages/airnode-validator/src/validator.test.ts#L400

Siegrift (changes requested):

packages/airnode-deployer/config/config.json.example
amarthadan: Yeah, I guess so. TBH updating `config.json` is such a painful process. We should do something about that.

packages/airnode-deployer/src/infrastructure/index.ts
amarthadan: Yes, I was thinking about that too. I'll make a refactor issue and do that during the 0.6 window.

packages/airnode-deployer/config/config.json.example
Siegrift: Why is it painful? Isn't it just find and replace? (with some copying)?

packages/airnode-deployer/terraform/airnode/aws/variables.tf
amarthadan: No, in AWS 0 means disabled (no instance can be run) and -1 means no reservations. In GCP you can't disable a function with this option, so 0 means no reservations. :man_shrugging:

packages/airnode-deployer/config/config.json.example
amarthadan: Depending on what you're changing in the `config.json` it can go through multiple packages, source files, and tests. And as you can see I still missed a few :smile: It was actually worse in https://github.com/api3dao/airnode/pull/820/commits/bdeb56a9eb47e23724295dd2289bbb4b3e86b70c.

packages/airnode-deployer/src/infrastructure/index.ts
amarthadan: https://api3dao.atlassian.net/browse/AN-545

packages/airnode-deployer/terraform/airnode/aws/variables.tf
```diff
@@ -41,5 +41,6 @@ variable "api_key" {
 
 variable "api_max_concurrency" {
   description = "Maximum amount of concurrent executions for Airnode Test Gateway Lambda"
+  # https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function#reserved_concurrent_executions
```
Siegrift, line 44: That link only mentions AWS, but we also want this parameter for GCP?

Siegrift (commented):
Btw. how can I test this? Can we add unit tests for this (not sure what to test though)?

packages/airnode-deployer/terraform/airnode/aws/variables.tf
amarthadan: That's in the `gcp` directory. I added the link for AWS because of the non-intuitive -1. I think that's enough. Otherwise I could add links to all the parameters.

amarthadan:
> Btw. how can I test this? Can we add unit tests for this (not sure what to test though)?

Deployer tests are another thing I was thinking about for 0.6. But it's going to be tricky and potentially a lot of mocking :confused: You can check whether the values are passed correctly but that's about it what you can do with unit tests really.

As for how you can test it, you can run deployment with that `maxConcurrency` set and then check in the cloud provider console that it is set correctly. But I did that and it will be done during testing so you probably don't need to do that.

Siegrift (approved):
Code-wise looks good, but I didn't test.

Issue #752, opened 2021-12-30
Single level _path fails?


An-dz:
## Description

It seems that a `_path` with a single level doesn't work on v0.3.1.

## Version

Airnode v0.3.1

## Present Behaviour

I have created an Airnode where the API responds with a simple JSON with a single key and I configured the endpoint `reservedParameters` like this:

```json
"reservedParameters": [
  {
    "name": "_type",
    "fixed": "int256"
  },
  {
    "name": "_path",
    "fixed": "scores"
  }
]
```

But the following witchcraft happens:

```
ERROR Unable to find response value from 
{
    "scores": "57896044889312237195750096412266029227848665016659598748611696315871396439764"
}
. Path: scores
```

## Expected Behaviour

I expect a single level `_path` to work. I have not tested on this Airnode for multi-level but I assume they work considering the examples you guys created have multi-level `_path`.

## Steps to reproduce

I believe the only thing needed is to use the Coingecko example but change the path to something in the first level.

dcroote:
@An-dz  - I wasn't able to reproduce your error with the `0.3.1` client docker image and a coingecko single level path (`coingecko_rank` for the API endpoint `/coins/{id}`). Could you share your full `config.json`?

bbenligiray:
I replaced the number [here](https://github.com/api3dao/airnode/blob/v0.3.1/packages/airnode-adapter/e2e/index.ts#L40) with `'57896044889312237195750096412266029227848665016659598748611696315871396439764'` and ran `yarn test:adapter` and got
```
Error: value out-of-bounds (argument=null, value="57896044889312237195750096412266029227848665016659598748611696315871396439764", code=INVALID_ARGUMENT, version=abi/5.5.0)
```

Experimenting a bit more, when I enter `'57896044618658097711785492504343953926634992332820282019728792003956564819967'` (`2^255 - 1`) there, the tests pass. When I enter `'57896044618658097711785492504343953926634992332820282019728792003956564819968'` (`2^255`) there, the tests fail with the error message above.

So the value your API returns doesn't fit into an `int256`. Failure in this case is the preferred behavior, but the error message is indeed a bit cryptic. I'll leave the issue open for you to confirm.

An-dz:
Thank @bbenligiray, that was it.

The endpoint was creating a ***u***int256 since it deals with bit manipulation, now it creates a negative decimal number.

[...]

PR #818, opened 2022-02-11
Add more errors to airnode-adapter response extraction

vponline:
Show better error messages when the extraction or encoding of responses in airnode-adapter. [AN428](https://api3dao.atlassian.net/jira/software/projects/AN/boards/18?selectedIssue=AN-428)


packages/airnode-adapter/src/types.ts
```diff
@@ -78,3 +78,48 @@ export interface ExtractedAndEncodedResponse {
 }
 
 export type ReservedParametersDelimeter = typeof MULTIPLE_PARAMETERS_DELIMETER | typeof PATH_DELIMETER;
+
+export class SizeLimitError extends Error {
```
Siegrift, line 82: I dislike these custom error messages and we do not use this pattern in Airnode (and I think it's not that common in JS/TS world in general). I think we should just use the `Error` with a different error messages.

packages/airnode-node/src/handlers/call-api.ts
```diff
@@ -211,9 +211,30 @@ async function processSuccessfulApiCall(
     const signature = await signResponseMessage(aggregatedApiCall.id, response.encodedValue, config);
     return [[], { success: true, value: response.encodedValue, signature }];
   } catch (e) {
-    const data = JSON.stringify(rawResponse.data);
-    const log = logger.pend('ERROR', `Unable to find response value from ${data}. Path: ${reservedParameters._path}`);
-    return [[log], { success: false, errorMessage: RequestErrorMessage.ResponseValueNotFound }];
+    if (e instanceof adapter.MissingValueError) {
```
Siegrift, line 214: I think we can assume that the `e` will always be and `Error`. And I think this checking is overkill, since we ultimately only care about the user receiving a good error message.

Siegrift (changes requested):
btw. do not forget to add a changeset (I think patch is enough).

packages/airnode-adapter/src/types.ts
vponline: I noticed that this pattern hasn't been used but how would you prefer to check for the type of error in the `call-api.ts` `extractAndEncodeResponse` catch-block in this case? Checking for the error (message) string content doesn't seem like a great alternative? Would you have a better suggestion?

packages/airnode-node/src/handlers/call-api.ts
vponline: This could be changed to just pass the `error.message` into both the logger `message` and `errorMessage` fields. But in this case how would we specify `RequestErrorMessage` enums?

packages/airnode-adapter/src/response-processing/encoding.ts
```diff
@@ -21,12 +21,20 @@ export function getSolidityType(type: ResponseType) {
 
 export function encodeValue(value: ValueType, type: ResponseType): string {
   const solidityType = getSolidityType(type);
-
-  return ethers.utils.defaultAbiCoder.encode([solidityType], [value]);
+  try {
+    const encodedValue = ethers.utils.defaultAbiCoder.encode([solidityType], [value]);
+    return encodedValue;
```
acenolaza, line 26: Maybe just return here instead of creating a new var?

packages/airnode-adapter/src/response-processing/encoding.ts
```diff
@@ -21,12 +21,20 @@ export function getSolidityType(type: ResponseType) {
...
+  } catch (e) {
+    throw new EncodingError((e as Error).message);
```
acenolaza, line 28: How certain are we that we can cast e as Error? Should we check it's type first like in [here](https://github.com/api3dao/airnode/blob/v1-protocol/packages/airnode-admin/src/cli.ts#L479-L480)?

packages/airnode-node/src/handlers/call-api.ts
```diff
@@ -211,9 +211,30 @@ async function processSuccessfulApiCall(
...
+      const data = JSON.stringify(rawResponse.data);
+      const log = logger.pend('ERROR', `Unable to find response value from ${data}. Path: ${reservedParameters._path}`);
+      return [[log], { success: false, errorMessage: RequestErrorMessage.ResponseValueNotFound }];
+    } else if (e instanceof adapter.SizeLimitError) {
+      console.log('HERE SizeLimitError');
```
acenolaza, line 219: Kill it with 🔥

packages/airnode-adapter/src/types.ts
acenolaza: Dealing and parsing strings is something that I have regretted many times in the past. A alternative could be to return a new AirnodeError object that has an error code and the Error object inside. Then you could use the error code to determine which error message to display to the user. Just an idea.

acenolaza (commented):
I would add @andreogle to this PR. Maybe they are doing something else to handle this in ChainAPI

packages/airnode-adapter/src/types.ts
vponline: I like this solution as well @acenolaza. @Siegrift What do you think, should we go with that?

dcroote:
I think it'd be great to include a test specifically for #752 here.

packages/airnode-adapter/src/response-processing/encoding.ts
```diff
@@ -21,12 +21,26 @@ export function getSolidityType(type: ResponseType) {
...
 export function encodeValue(value: ValueType, type: ResponseType): string {
   const solidityType = getSolidityType(type);
-
-  return ethers.utils.defaultAbiCoder.encode([solidityType], [value]);
+  try {
+    return ethers.utils.defaultAbiCoder.encode([solidityType], [value]);
+  } catch (e) {
+    if (e instanceof Error) {
+      throw new EncodingError((e as Error).message);
+    } else {
```
andreogle, line 29: Nit: there's no need for the else block here. It could be written as

```ts
if (e instanceof Error) {
  throw new EncodingError(...);
}
throw new EncodingError(...);
```

packages/airnode-adapter/src/types.ts
andreogle: I agree with @Siegrift I'm also not really sure about this approach... 

If we're going to be relying on error responses so much, maybe it could be better to wrap the responses in an object instead of throwing and then matching on different error types (?)

```ts
{ 
  error: { 
    type: 'SizeLimitError', // Some kind of enum
    message: 'optional custom message'
  },
  value: null
}
```

Just an idea though. What do you think?

packages/airnode-adapter/src/response-processing/encoding.ts
andreogle: In general, the more right indents, the harder the code becomes to read (at least in my opinion 😁 ). If you return (or throw) from inside the `if` then you don't need the `else`. A side effect of this is that you're often forced to write smaller functions. This is why you'll find very few `else` blocks in the Airnode repo.

packages/airnode-adapter/src/types.ts
vponline: This also seems like a good alternative to me. I guess it depends on which solution everyone finds fits the best for this issue.

packages/airnode-adapter/src/response-processing/encoding.ts
vponline: I will remove the `else` blocks :+1:

packages/airnode-adapter/e2e/index.ts
```diff
@@ -38,6 +38,10 @@ const apiResponse = {
   bytes: '0x72616e646f6d20737472696e6700000000000000000000000000000000000000',
   big: {
     decimal: '11223344556677889900',
+    // The maximum number that fits in int256
```
vponline, line 41: > I think it'd be great to include a test specifically for #752 here.

@dcroote I added these tests. What do you think?

For some reason using the powers (`2 ** 255`, `2 ** 255 - 1`) results in a `Error: invalid BigNumber string`, so I included them as strings instead.

packages/airnode-adapter/src/types.ts
vponline: Alternatively an even shorter solution (assuming we don't really care about specifying the errors in detail). We could do the following:

1. all errors thrown are instances of a normal `Error` with a string `message`
2. and then in _call-api.ts_:

```typescript
} catch (e) {
  if (e instanceof Error) {
      const log = logger.pend('ERROR', e.message);
      return [[log], { success: false, errorMessage: e.message }];
  }
  const log = logger.pend('ERROR', String(e));
  return [[log], { success: false, errorMessage: String(e) }];
}
```

Siegrift: I'd do exactly that (just provider better error message). I'd also drop the instanceof check and always assume that we did throw an error (this holds in practice).

I am not a fan of the suggestion with custom error message nor wrapping, since they both require some centralized enum with allowed error codes, which doesn't work well with monorepo.

packages/airnode-adapter/e2e/index.ts
```diff
@@ -273,5 +292,19 @@ describe('Extraction, encoding and simple on chain decoding', () => {
         "Unable to find value from path: 'json.strange.key'"
       );
     });
+
+    it('throws on int256 max number', async () => {
+      try {
```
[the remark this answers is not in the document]
vponline: I am not able to make it work with that example, were you able to do that?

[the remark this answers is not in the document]
Siegrift: Oh, these are hardhat tests. That pattern is only possible in jest.

Siegrift (approved):
👍 LGTM.

PR #821, opened 2022-02-14
Add check for trailing bytes in airnode-abi

vponline:
Throw error if encoded data has trailing bytes

Jira: https://api3dao.atlassian.net/jira/software/projects/AN/boards/18?selectedIssue=AN-469


.changeset/sharp-news-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'@api3/airnode-abi': patch
```
Siegrift, line 2: I think this could be minor.

.changeset/sharp-news-run.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+throw error if abi encoded data has trailing bytes
```
Siegrift, line 5:
```suggestion
Throw error if abi encoded data has trailing bytes
```

(it looks nicer)

packages/airnode-abi/src/decoding.ts
```diff
@@ -63,6 +73,10 @@ export function decode(encodedData: string): DecodedMap {
   // exactly what you got from the contract, including the header.
   const decodedData = ethers.utils.defaultAbiCoder.decode(decodingTypes, encodedData);
 
+  if (!reEncodedLengthMatches(decodingTypes, decodedData, encodedData.length)) {
```
Siegrift, line 76: Why don't we compare the encoded data directly (but only its length)?

vponline: It seemed that comparing the length would be enough based on [this](https://api3dao.atlassian.net/browse/AN-469?focusedCommentId=13795), but would you prefer to compare the string?

packages/airnode-abi/src/decoding.test.ts
```diff
@@ -102,4 +102,10 @@ describe('decode', () => {
   it('throws an error when invalid data is provided', () => {
     expect(() => decoding.decode('0x123456')).toThrowError();
   });
+
```
acenolaza, line 105: I would maybe also add a test for the example in Buraks comment

acenolaza (approved):
LGTM

packages/airnode-abi/src/decoding.test.ts
vponline: > I would maybe also add a test for the example in Buraks comment

Added these

packages/airnode-abi/src/decoding.ts
Siegrift: I would require an exact match. This is a stricter requirement then checking length (but is the same in practice).

Also, should we say `Trailing bytes in encoded data are not allowed`? This assumes that the length of the newly encoded data is shorter (which is true, but it's not apparent from reading the code).

vponline: @bbenligiray Would you have any input on this? Should we just compare length here or the entire string?

bbenligiray: Assuming that the ethers implementation is correct, we will always have an exact match if the lengths are identical, and the only reason the lengths won't match is if there are trailing bytes. So if we were checking if there is an exact match in addition to a length match, we would be validating the ethers implementation.

I don't see anything wrong with requiring an exact match though, but we would have to have a more general error message. We never expect to have trailing bytes either, so we probably don't need a specialized error message for that specifically.

Siegrift (approved):
👍 LGTM

PR #831, opened 2022-02-15
Remove Submitted variant of RequestStatus enum

dcroote:
Part 1/n for [AN-482](https://api3dao.atlassian.net/browse/AN-482) - Remove “RequestStatus” enum. This removes the Submitted variant which, as noted in the existing TODO comment, wasn't important.


packages/airnode-node/src/evm/fulfillments/api-calls.test.ts
```diff
@@ -246,7 +245,7 @@ describe('submitApiCall', () => {
         expect(data).toEqual({
           ...apiCall,
           fulfillment: { hash: '0xfailtransaction' },
-          status: RequestStatus.Submitted,
+          status: RequestStatus.Errored,
```
Siegrift, line 248: Why is this errored? Is it because we were rewriting the status in fulfillments/requests.ts?

Siegrift (approved):
This one LGTM.

packages/airnode-node/src/evm/fulfillments/api-calls.test.ts
dcroote: Yes, as you said this was a small bug- the test is submitting a fail transaction so `Errored` is appropriate.

PR #834, opened 2022-02-16
Add transaction overrides to admin CLI and SDK

vponline:
Enables use of Ethers transaction overrides (mainly for gas options) to admin CLI and SDK

Jira: [https://api3dao.atlassian.net/jira/software/projects/AN/boards/18?selectedIssue=AN-149](AN-149)
Docs PR: [#631](https://github.com/api3dao/api3-docs/pull/631)


packages/airnode-admin/src/sdk.ts
```diff
@@ -48,49 +52,62 @@ export class AdminSdk {
   sponsorToWithdrawalRequestCount = (sponsorAddress: string) =>
     admin.sponsorToWithdrawalRequestCount(this.airnodeRrp, sponsorAddress);
 
-  fulfillWithdrawal = (requestId: string, airnodeAddress: string, sponsorAddress: string, amount: string) =>
-    admin.fulfillWithdrawal(this.airnodeRrp, requestId, airnodeAddress, sponsorAddress, amount);
+  // fulfillWithdrawal = (requestId: string, airnodeAddress: string, sponsorAddress: string, amount: string) =>
+  //   admin.fulfillWithdrawal(this.airnodeRrp, requestId, airnodeAddress, sponsorAddress, amount);
+  // TODO: check overrides here that amount (i.e. overrides.value) is passed on correctly
```
Siegrift, line 57: The implementation looks good. Or is there something I am missing from this comment?

packages/airnode-admin/src/implementation.ts
```diff
@@ -206,13 +256,15 @@ export async function setWhitelistExpiration(
...
+  expirationTimestamp: number,
+  overrides?: ethers.Overrides
 ) {
   const tx = await requesterAuthorizerWithAirnode.setWhitelistExpiration(
     airnodeAddress,
     endpointId,
     requesterAddress,
-    expirationTimestamp
+    expirationTimestamp,
+    { ...overrides }
```
Siegrift, line 267: I think we can just pass `overrides` (we don't need to destructure)

packages/airnode-admin/src/implementation.ts
```diff
@@ -1,13 +1,50 @@
...
   }
 };
 
+/**
+ * Parses Ethers transaction override options from CLI arguments.
+ * @param args The yargs inferred CLI arguments object
+ * @returns The parsed overrides object with values compatible with ethers
+ */
+export const parseTransactionOverrides = (
+  args: Arguments | { [key: string]: string }
```
Siegrift, line 18:
```suggestion
  args: Arguments
```

packages/airnode-admin/src/implementation.ts
```diff
@@ -1,13 +1,50 @@
...
+): ethers.Overrides | ethers.PayableOverrides => {
+  if ((args['max-fee'] || args['max-priority-fee']) && args['gas-price'])
+    throw new Error('EIP-1559 transactions cannot have gas-price argument');
+  if ((args['max-fee'] && !args['max-priority-fee']) || (!args['max-fee'] && args['max-priority-fee']))
+    throw new Error('EIP-1559 transactions require both max-fee and max-priority-fee arguments');
+  if (args['value'] && ethers.utils.parseEther(args['value'] as string).lt(0))
```
Siegrift, line 24: Do we need to check this? We are not checking this for the other ones (and they need to be non zero as well).
I think it's fine to let ethers check this.

Siegrift (changes requested):

packages/airnode-admin/src/cli.ts
```diff
@@ -104,6 +104,32 @@ const COMMON_COMMAND_ARGUMENTS = {
...
+    },
+    'max-fee': {
+      type: 'string',
+      describe: 'Transaction maximum fee (in gwei)',
+    },
+    'max-priority-fee': {
+      type: 'string',
+      describe: 'Transaction maximum priority fee (in gwei)',
+    },
+    value: {
```
vponline, line 124: I'm not sure about adding `value` as an argument to all (or any) commands as it is only needed in `Payable` transactions and it seems it's only used for airnode's `fulfillWithdrawal` method which is not included in the CLI as a command.

packages/airnode-admin/src/sdk.ts
vponline: I had forgotten to remove this

packages/airnode-admin/src/implementation.ts
vponline: I used destructuring here because the SDK wrapped methods would pass in `overrides`  as `undefined` if they were omitted, e.g.
```js
  sponsorRequester = (requesterAddress: string, overrides?: ethers.Overrides) =>
    admin.sponsorRequester(this.airnodeRrp, requesterAddress, overrides); // omitted overrides are undefined when passed in to admin.sponsorRequester
```
Ethers seems to throw an error if overrides are undefined, but an empty object works.

Please correct me if I'm wrong or you would have a better solution here.

packages/airnode-admin/src/implementation.ts
vponline: We could probably drop all of these checks if it's fine to rely on ethers checking them?

packages/airnode-admin/src/cli.ts
Siegrift: Yeah, I see no reason in adding it there. Maybe we could also remove it from `fulfillWithdrawal` ? Since that value is not needed for any other command.

packages/airnode-admin/src/implementation.ts
Siegrift: Do they also check that you can't mix the eip1559 and legacy params?

vponline: Yes it also checks that :+1:

packages/airnode-admin/test/e2e/cli.feature.ts
```diff
@@ -50,7 +50,14 @@ describe('CLI', () => {
       return execSync(`node ${CLI_EXECUTABLE} ${formattedCommand}`).toString().trim();
     } catch (e: any) {
       // rethrow the output of the CLI
-      throw new Error(e.stdout.toString().trim());
+      const error = e.stdout.toString().trim();
+      // Checking for ethers errors and throwing a custom error message here because
+      // the ethers error will contain transaction information which is not constant,
+      if (error.includes('eip-1559 transaction do not support gasPrice')) {
```
Siegrift, line 56: I don't think you need this. Afaik, jest .toThrow(str) only check if `str` is a substring of the error message.

vponline: That works, thanks!

packages/airnode-admin/test/e2e/cli.feature.ts
```diff
@@ -211,6 +211,27 @@ describe('CLI', () => {
       expect(sponsored).toBe(true);
     });
 
+    it('uses transaction overrides', async () => {
```
Siegrift, line 214: Could you also write a test that uses the admin SDK (for any command) and passes undefined as the third argument (because when used as a CLI, the `parseTransactionOverrides` will always return `{}` if none override is passed)

vponline: I was thinking that here: [sponsorRequester](https://github.com/api3dao/airnode/pull/834/files#diff-df88b69abee08b2eb5c21f2807960839c363faa0260009f82eef94880ad6b217R231)  is called with omitted overrides, but I'll add a test for admin SDK as well.

packages/airnode-admin/test/e2e/cli.feature.ts
```diff
@@ -335,7 +356,9 @@ describe('CLI', () => {
 
       const balanceBefore = await sponsorBalance();
       airnodeRrp = airnodeRrp.connect(sponsorWallet);
-      await admin.fulfillWithdrawal(airnodeRrp, withdrawalRequestId, airnodeWallet.address, sponsor.address, '0.8');
+      await admin.fulfillWithdrawal(airnodeRrp, withdrawalRequestId, airnodeWallet.address, sponsor.address, {
```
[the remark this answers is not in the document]
vponline: Nice catch, will add

packages/airnode-admin/src/sdk.ts
```diff
@@ -48,49 +52,60 @@ export class AdminSdk {
...
     admin.sponsorToWithdrawalRequestCount(this.airnodeRrp, sponsorAddress);
 
-  fulfillWithdrawal = (requestId: string, airnodeAddress: string, sponsorAddress: string, amount: string) =>
-    admin.fulfillWithdrawal(this.airnodeRrp, requestId, airnodeAddress, sponsorAddress, amount);
+  fulfillWithdrawal = (
+    requestId: string,
+    airnodeAddress: string,
+    sponsorAddress: string,
+    amount: string,
+    overrides?: ethers.Overrides
```
vponline, line 60: I changed the overrides type to normal overrides instead payable since `value` is handled through the `amount` parameter.

packages/airnode-admin/test/e2e/sdk.feature.ts
```diff
@@ -1,35 +1,81 @@
...
+    airnodeRrp = await new AirnodeRrpFactory(deployer).deploy();
+    accessControlRegistry = await new AccessControlRegistryFactory(deployer).deploy();
+    requesterAuthorizerWithAirnode = await new authorizers.RequesterAuthorizerWithAirnodeFactory(deployer).deploy(
+      accessControlRegistry.address,
+      'RequesterAuthorizerWithAirnode admin'
+    );
+    sdk = new AdminSdk(airnodeRrp, requesterAuthorizerWithAirnode);
+  });
+
+  it('provides same API', async () => {
```
Siegrift, line 44: This one doesn't need the beforeEach hooks, but this is fine as is.

Siegrift (approved):
👍 LGTM

PR #835, opened 2022-02-16
[PoC]: Validator v2

Siegrift:
## Recommended way to read this (messy) PR

Read all code changes at once, read my own comments. You may then try to review commit by commit, but I didn't pay too much attention to splitting it nicely (sorry) :/

Main properties of this PoC:
- The biggest difference to the older validator is that you can easily understand and edit the dynamic validation logic (since it's JS)
- The error handling is abstracted away by zod, the schema is mostly declarative
- The schema generated TS typings, but these should not be used directly. Imo and we should maintain our own types like we did (I am a bit disappointed by this)
- I suggest removing the `skipValidation` parameter and always validate the config.json
- No need to copy templates after new version release, but it should be done just before the release (this is related to removing the skipValidation param)

## Checklist

Notable checklist items (this list is a bit chaotic, some items are only in TODOs of the code):
- [x] created a schema for config.json and receipt.json based on zod
- [x] wrote a basic unit test (to test if the validation passes for current config.json format we use)
- [x] used the TS types from the schema in other packages (to see if the things remain compatible)
- [x] Try out dynamic refinements
- [x] Add an "unsafeAPI" which only interpolates secrets into config json (useful to parse and interpolate config.json in Airnode - since it will be verified by validator at deploy time).
- [x] Deprecate skipValidation (removal tracked in https://api3dao.atlassian.net/browse/AN-556)
- [ ] More tests
- [ ] Fix the TODOs in code
- [ ] Move exampleSpecs to test (as fixtures)
- [ ] In “ApiSpecification.paths” the path parameters must be listed in the parameters section
- [ ] All  “ApiSpecification.paths[someEndpoint]” must be either fixed or or user parameters
- [ ] Allowed security scheme (https://docs.api3.org/airnode/v0.4/grp-providers/guides/build-an-airnode/api-security.html#step-2-turn-on-the-defined-security-schemes) should exist
- [ ] Any parameter in parameters or fixedOperationalParameters should be listed in apiSpeicfication.paths
- [ ] Implement old validator rules in config.json
- [ ] Endpoint ID refinement
- [ ] Trigger must point to an OIS endpoint
- [ ] Eip1559 validations
- [ ] Extra field / missing field tests
- [ ] Implement regex validator rules (https://github.com/colinhacks/zod#strings)
- [ ] Revisit how validator is bundled (nit)
- [ ] Remove old validator
- [ ] Docs update
- [ ] Verify with chainAPI guys
- [ ] Revisit how validator is bundled (and double check it is built correctly)
- [ ] disableConcurrencyReservations check (implies both ways) aws|gcp
- [ ] Validate invisible characters in secrets
- [ ] Create validations for GCP and AWS secrets
- [ ] Revisit the CLI
- [ ] Create a script which bumps the noodeVersion in examples
- [ ] Interpolation in OAS path should check whether the parameter is defined

## Design decisions

- Validator will not support warnings
- The skipValidationParameter will be removed from config.json or reworked as a deployer/airnode-client flag
- Node and OIS define types manually, since the inferred types are hard to use.


packages/airnode-examples/integrations/authenticated-coinmarketcap/config.example.json
```diff
@@ -36,7 +36,7 @@
     },
     "logFormat": "plain",
     "logLevel": "INFO",
-    "nodeVersion": "0.5.0",
+    "nodeVersion": "0.4.0",
```
Siegrift, line 39: Basically, the plan is to update these before releasing a new version (will be added as a step in the versioning flow). Gladly, this will only need to be updated in examplatory `config.json` files (in deployer, node and examples).

packages/airnode-node/src/types.ts
```diff
@@ -9,6 +7,43 @@ import {
   RequestedWithdrawalEvent,
   FulfilledWithdrawalEvent,
 } from '@api3/airnode-protocol';
+// TODO: Make this nicer
```
Siegrift, line 10: @amarthadan Suggested importing the types directly from validator, which is a better approach to re-exporting.

However, the inferred schemas from zod have a disadvantage that you don't really see the type. For example, when looking at the `Config` type here, you will see the following:
```
export declare type Config = SchemaType<typeof configSchema>;
```

ofc, the `configSchema` is in the same file, but it looks like this:
```
export declare const configSchema: z.ZodObject<{
    chains: z.ZodArray<z.ZodObject<{
        authorizers: z.ZodArray<z.ZodString, "many">;
        blockHistoryLimit: z.ZodOptional<z.ZodNumber>;
        contracts: z.ZodObject<{
            AirnodeRrp: z.ZodString;
        }, "strip", z.ZodTypeAny, {
            AirnodeRrp: string;
        }, {
            AirnodeRrp: string;
        }>;
        id: z.ZodString;
        ignoreBlockedRequestsAfterBlocks: z.ZodOptional<z.ZodNumber>;
        minConfirmations: z.ZodOptional<z.ZodNumber>;
        type: z.ZodLiteral<"evm">;
        options: z.ZodObject<{
            txType: z.ZodUnion<[z.ZodLiteral<"legacy">, z.ZodLiteral<"eip1559">]>;
            baseFeeMultiplier: z.ZodOptional<z.ZodString>;
            priorityFee: z.ZodOptional<z.ZodObject<{
                value: z.ZodString;
                unit: z.ZodOptional<z.ZodUnion<[z.ZodLiteral<"wei">, z.ZodLiteral<"kwei">, z.ZodLiteral<"mwei">, z.ZodLiteral<"gwei">, z.ZodLiteral<"szabo">, z.ZodLiteral<"finney">, z.ZodLiteral<"ether">]>>;
            }, "strip", z.ZodTypeAny, {
                unit?: "wei" | "kwei" | "mwei" | "gwei" | "szabo" | "finney" | "ether" | undefined;
                value: string;
            }, {
                unit?: "wei" | "kwei" | "mwei" | "gwei" | "szabo" | "finney" | "ether" | undefined;
                value: string;
            }>>;
        }, "strip", z.ZodTypeAny, {
            baseFeeMultiplier?: string | undefined;
            priorityFee?: {
                unit?: "wei" | "kwei" | "mwei" | "gwei" | "szabo" | "finney" | "ether" | undefined;
                value: string;
            } | undefined;
            txType: "legacy" | "eip1559";
        }, {
...
```

I think the better solution is to write those manually here - the zod schema will be compatible (but we will have to update two locations when making changes to config.json). I also thought about generating these schemas, but that has the same problem as with inferring (TS is not able to extract subtypes). 

Any better ideas?

packages/airnode-validator/exampleSpecs/config.specs.json
```diff
@@ -2,138 +2,182 @@
   "chains": [
```
Siegrift, line 2: This file was not valid, so I replaced it with the one from airnode-examples.

packages/airnode-validator/src/v2/api/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    );
+
+    return {
+      success: true,
+      data: interpolated,
+    };
+  } catch (e) {
+    return {
+      success: false,
+      error: new ValidatorError('Error interpolating secrets. Make sure the secrets format is correct'),
```
Siegrift, line 63: It is possible to implement custom validations (outside of zod schema + refinements). You just have to throw `ValidatorError`.

packages/airnode-validator/src/v2/config/config.ts
```diff
@@ -0,0 +1,150 @@
...
+import { oisSchema } from '../ois';
+import { SchemaType } from '../types';
+import { zodDiscriminatedUnion } from '../zod-discriminated-union';
+
+export const triggerSchema = z.object({
+  endpointId: z.string(),
+  endpointName: z.string(),
+  oisTitle: z.string(),
+});
+export type Trigger = SchemaType<typeof triggerSchema>;
```
Siegrift, line 12: Exporting these types might not be necessary (especially if we end up preserving the types in ois and node packages as I've suggested in the github comment above).

It would make the schemas a bit more compact.

packages/airnode-deployer/src/cli/__tests__/commands.test.ts
```diff
@@ -22,8 +22,8 @@ import { Receipt } from '../../types';
 describe('deployer commands', () => {
   it('can deploy Airnode', async () => {
     await deploy(
-      join(__dirname, '../../../config/config.json.example'),
-      join(__dirname, '../../../config/secrets.env.example'),
+      join(__dirname, '../../../config/config.example.json'),
```
andreogle, line 25: I suppose it's better to have it this way for syntax highlighting?

packages/airnode-node/config/config.example.json
```diff
@@ -33,7 +33,7 @@
     },
     "logFormat": "plain",
     "logLevel": "INFO",
-    "nodeVersion": "0.5.0",
+    "nodeVersion": "0.4.0",
```
andreogle, line 36: Maybe we need some kind of script we can run to update all of these values. I assume you're doing this manually now?

packages/airnode-validator/src/v2/receipt/receipt.ts
```diff
@@ -0,0 +1,33 @@
+import { z } from 'zod';
+import { cloudProviderSchema } from '../config';
+import { SchemaType } from '../types';
+
+// TODO: Implement old validator rules
```
andreogle, line 5: Do you mean improving the types to enforce strings conforming to a certain regex pattern as well?

packages/airnode-deployer/src/cli/__tests__/commands.test.ts
Siegrift: Yeah, this is tracked in https://api3dao.atlassian.net/browse/AN-424 and I thought I will do this to simplify the rename from 0.5 to 0.4 (I thought replacing in *.json files would be enough). 

It turned out that it isn't enough though.

packages/airnode-node/config/config.example.json
Siegrift: Yes, but it will be pretty easy to do so - after a few minor changes you will be able to just globally find and replace them (e.g in IDE) - doing it with script is likely an overkill.

packages/airnode-validator/src/v2/receipt/receipt.ts
Siegrift: Yeah, I will implement those - but Zod has already helper functions for most of what we need -  https://github.com/colinhacks/zod#strings

Siegrift: The good part again is that I trust they probide a pretty good error messages that have been reviewed an bikeshed by number of people :)

packages/airnode-node/src/types.ts
amarthadan: I don't agree. I would still prefer keeping the types in one place, updated automatically when the schema is updated. I agree that the generated types are a bit harder to read but it's still quite straightforward. In my opinion, the advantage of having the types in one place and automatically updated with the schema outweighs the readability issue.

Siegrift: > but it's still quite straightforward

I disagree. Take a look at `chains.options` and try to understand what is its type.

> the advantage of having the types in one place

This is a good point, and we check only that zod schema is compatible with the manually defined one, but not the check other way around (because it doesn't hold, since zod types are more precise). This means that inconsistencies can occur. However, we will likely notice, because we will be missing the functionality in Airnode.

> and automatically updated with the schema outweighs the readability issue.

To me not, because the schema is updated only from time to time.

packages/airnode-node/config/config.example.json
amarthadan: I would definitely welcome a script for that. You need to update:
* All the examples in airnode-examples
* Actual code in examples
* Example in airnode-deployer
* Example in airnode-node
* airnode-validator multiple places (guess that won't be an issue anymore)

I did it few times and I almost always missed at least one place :smile:

Siegrift: Makes sense. I'll add a TODO to github description.

amarthadan (approved):
I was a bit worried when I saw `Files changed 90` but the actual validator code is like 6 files and it's pretty much just the schema. Super simple, I very much like it :+1:

aquarat (approved):
Looks great, much easier to read than the existing validator :smile: 
Cloud vs deployer config parsing is very neat.

packages/airnode-node/src/types.ts
Siegrift: I will keep this comment unresolved and we can discuss that in the google document.

packages/airnode-validator/src/v2/api/api.ts
```diff
@@ -0,0 +1,78 @@
...
+    );
+
+    return {
+      success: true,
+      data: interpolated,
+    };
+  } catch (e) {
+    return {
+      success: false,
+      error: new ValidatorError('Error interpolating secrets. Make sure the secrets format is correct'),
```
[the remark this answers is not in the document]
Siegrift: Yeah, there is pretty good error message and there is somewhere a TODO item rework this.

PR #837, opened 2022-02-17
AN428-fix

vponline:
Fixes an error from #818 ([AN428](https://api3dao.atlassian.net/jira/software/projects/AN/boards/18?selectedIssue=AN-428)) by removing a variable from airnode-adapter extraction error message


Siegrift (commented):
We also need to change the error handling in casting (we also interpolate request information there - see screenshot in jira issue).

I think for now we can just wrap https://github.com/api3dao/airnode/blob/master/packages/airnode-adapter/src/response-processing/extraction.ts#L117 and throw an error saying `Unable to cast value to ${type}`.

PR #839, opened 2022-02-18
Fix heartbeat API key location & name

amarthadan:
Based on [slack discussion](https://api3workspace.slack.com/archives/C02AYRX8D89/p1635066629013200).
I tested it with the deployment created on staging and it became `Active` :slightly_smiling_face:


amarthadan:
> Why not merge this one as well?

Because the tests were still running (they failed on a falsy link check...) and I went away from the PC :see_no_evil:

PR #842, opened 2022-02-18
AN443: Refactor Jest fs, 'readFileSync'

vponline:
Refactors `jest.spyOn(fs, 'readFileSync').mockReturnValue` to use `mockImplementation` to fix affected tests

Jira: https://api3dao.atlassian.net/jira/software/projects/AN/boards/18?selectedIssue=AN-443


packages/airnode-node/src/handlers/start-coordinator.test.ts
```diff
@@ -43,7 +43,8 @@ describe('startCoordinator', () => {
         },
       })),
     };
-    jest.spyOn(fs, 'readFileSync').mockReturnValue(JSON.stringify(config));
+    jest.spyOn(console, 'log').mockImplementation();
```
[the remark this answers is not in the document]
Siegrift: Mhh, but what will happen when someone would want to print out something? I guess it would do nothing, right?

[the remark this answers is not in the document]
vponline: Yes indeed, I'm not sure why it works either but I found quite a few people having console problems with fs. I'll keep looking for other fixes.

packages/airnode-node/test/mock-utils.ts
```diff
@@ -61,3 +61,16 @@ export const createAndMockGasTarget = (txType: 'legacy' | 'eip1559') => {
 
   return { gasTarget: { maxPriorityFeePerGas, maxFeePerGas }, blockSpy, gasPriceSpy };
 };
+
+/**
+ * Mocks the console for silenced tests and otherwise enables it for local development.
+ */
+export const mockConsole = () => {
```
vponline, line 68: This solution seems to allow using `console.log` locally and still passes Github tests. Other solutions (mostly) completely disable the console, and then try to restore it if logging is needed, or implement a custom logger (`process.stdout.write` seems to still work).

I think this `mockConsole` approach may be enough for this task. What do you think @Siegrift ?

packages/airnode-node/src/handlers/start-coordinator.test.ts
```diff
@@ -43,6 +44,7 @@ describe('startCoordinator', () => {
         },
       })),
     };
+
     jest.spyOn(fs, 'readFileSync').mockReturnValue(JSON.stringify(config));
```
vponline, line 48: It's not entirely clear to me how the original idea of changing `mockReturnValue` to `mockReturnValueOnce`  would help with the `console` issue?

It looks like we cannot change it here or in `actions.test.ts` as  the tests don't pass in Github (although it works locally).

Siegrift: The problem is that jest uses it's own logger and replaces the standard nodeJS console.log. The problem is that jest internally initializes this logger by a config file. To load this file it uses `readFileSync` which we mock and the return value is not what the logger expects and throws.

vponline: Ah, I see! Although using mockReturnValueOnce seems to work locally, it doesn't pass on Github, weird.

packages/airnode-deployer/src/utils/validation.test.ts
```diff
@@ -1,17 +1,20 @@
+import { mockConsole } from '../../test/mock-utils';
 import fs from 'fs';
 import { loadConfig } from './files';
 import * as fixtures from '../../test/fixtures';
 
+mockConsole();
+
 describe('deployer-validation', () => {
-  it('loads the config without validation', () => {
+  it('loads the config without validation', async () => {
```
vponline, line 9: `console.log `seems to only  work here if this is `async` which seems weird

packages/airnode-deployer/test/mock-utils.ts
```diff
@@ -0,0 +1,14 @@
+// NOTE: This file is referenced as a pathGroup pattern in .eslintrc (import/order)
+
+/**
+ * Mocks the console for silenced tests and otherwise enables it for local development.
+ */
+export const mockConsole = () => {
+  if (process.env.SILENCE_LOGGER) {
+    // Mock console log for github
```
Siegrift, line 8: Github shouldn't behave differently to local environment. Are you sure you didn't have some wierd setup locally?

packages/airnode-node/src/handlers/start-coordinator.test.ts
Siegrift: That's strange. Maybe some cache thing or something?

packages/airnode-deployer/src/utils/validation.test.ts
Siegrift: That is very wierd.

Siegrift (commented):
I do not like the `mockConsole` solution, can you investigate more why this happens? (I'll try to take a look as well)

packages/airnode-deployer/test/mock-utils.ts
vponline: Yes definitely I can look into this more

packages/airnode-node/test/setup/e2e/testing.ts
```diff
@@ -20,7 +21,13 @@ export const deployAirnodeAndMakeRequests = async (filename: string, requests?:
   });
   // TODO: This is caused by duplicated mnemonic in Airnode state
   (config.nodeSettings as any).airnodeWalletMnemonic = deployConfig.airnodes.CurrencyConverterAirnode.mnemonic;
-  jest.spyOn(fs, 'readFileSync').mockReturnValue(JSON.stringify(config));
+  jest.spyOn(fs, 'readFileSync').mockImplementation((...args) => {
```
vponline, line 24: @Siegrift Your solution works everywhere so I've updated this to use `mockImplementation`.

packages/airnode-node/src/coordinator/calls/coordinated-execution.test.ts
```diff
@@ -28,7 +30,13 @@ describe('callApis', () => {
 
   it('returns each API call with the response if successful', async () => {
     const config = fixtures.buildConfig();
-    jest.spyOn(fs, 'readFileSync').mockReturnValueOnce(JSON.stringify(config));
+    jest.spyOn(fs, 'readFileSync').mockImplementation((...args) => {
```
Siegrift, line 33: Can you define this function in https://github.com/api3dao/airnode/blob/master/packages/airnode-node/test/mock-utils.ts and import it in the tests?

Siegrift: You can create similar file in deployer `test` folder.

Siegrift (changes requested):

.changeset/four-apricots-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+'@api3/airnode-node': patch
+---
+
+Refactors `jest.spyOn(fs, 'readFileSync').mockReturnValue` to use `mockImplementation` to fix affected tests
```
Siegrift, line 5: Nit:
```suggestion
Refactor `jest.spyOn(fs, 'readFileSync').mockReturnValue` to use `mockImplementation` to fix affected tests
```

but very nice description 💪

packages/airnode-deployer/src/cli/__tests__/commands.test.ts
```diff
@@ -62,7 +62,7 @@ describe('deployer commands', () => {
         stage: 'stage',
       },
     };
-    jest.spyOn(fs, 'readFileSync').mockImplementationOnce(() => JSON.stringify(receipt, null, 2));
+    mockReadFileSync('mockedReceiptFile', receipt, { replacer: null, space: 2 });
```
Siegrift, line 65: The replacer and space doesn't matter actually.

packages/airnode-deployer/test/mock-utils.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export const mockReadFileSync = (
+  filePathSubstr: string,
+  mockValue: unknown,
+  options?: { replacer?: (number | string)[] | null; space?: string | number }
+) => {
+  return jest.spyOn(fs, 'readFileSync').mockImplementation((...args) => {
+    const path = args[0].toString();
+    if (path.includes(filePathSubstr)) {
+      return JSON.stringify(mockValue, options?.replacer, options?.space);
```
Siegrift, line 20: Now that I think of it - we should probably not stringify here as that is the detail that can vary (e.g. you might want to mock this for secrets.env and those should not be stringified).

It would be a bit better if `mockValue` was string and we would call `JSON.stringify` when calling this method.

Siegrift (changes requested):
👍 Looks very good now.

I think this is my final change request :D

packages/airnode-deployer/test/mock-utils.ts
vponline: This makes sense, can also get rid of the options then. Will update

packages/airnode-deployer/src/cli/__tests__/commands.test.ts
```diff
@@ -62,7 +62,7 @@ describe('deployer commands', () => {
         stage: 'stage',
       },
     };
-    jest.spyOn(fs, 'readFileSync').mockImplementationOnce(() => JSON.stringify(receipt, null, 2));
+    mockReadFileSync('mockedReceiptFile', JSON.stringify(receipt, null, 2));
```
Siegrift, line 65: Nit: 
```suggestion
    mockReadFileSync('mockedReceiptFile', JSON.stringify(receipt));
```

Siegrift (approved):
👍 LGTM

PR #833, opened 2022-02-15
Implement endpoint for getting signed relayed data

Siegrift:

PR #843, opened 2022-02-18
Add signed relayed HTTP gateway

amarthadan:
Issues [AN-352](https://api3dao.atlassian.net/browse/AN-352) and [AN-493](https://api3dao.atlassian.net/browse/AN-493)
Merged with the first half from https://github.com/api3dao/airnode/pull/833 so it can be tested as a whole.


packages/airnode-node/src/api/index.ts
```diff
@@ -76,14 +86,36 @@ async function signResponseMessage(requestId: string, responseValue: string, con
...
+  requestId: string,
+  timestamp: string,
+  relayerAddress: string,
+  responseValue: string,
+  config: Config
+) {
+  const masterHDNode = getMasterHDNode(config);
+  const airnodeWallet = ethers.Wallet.fromMnemonic(masterHDNode.mnemonic!.phrase);
+
+  return await airnodeWallet.signMessage(
```
Siegrift, line 99: Based on https://github.com/api3dao/airnode/blob/v1-protocol/packages/airnode-protocol-v1/contracts/dapis/DapiServer.sol#L369

packages/airnode-deployer/src/cli/commands.ts
```diff
@@ -54,6 +54,13 @@ export async function deploy(configPath: string, secretsPath: string, receiptFil
     }
   }
 
+  const httpSignedRelayedGateway = config.nodeSettings.httpSignedRelayedGateway;
+  if (httpSignedRelayedGateway.enabled) {
+    if (httpSignedRelayedGateway.maxConcurrency !== undefined && httpSignedRelayedGateway.maxConcurrency <= 0) {
```
Siegrift, line 59: This should be handled by validator, can you add a TODO?

Siegrift (commented):
The gateway part LGTM. I wonder what will happen when one tries to remove airnode v0.4 with deployer with the TF updates (I am not sure if we disallow removals when there is deployer/node version mismatch)

Siegrift (commented):
@bbenligiray please review the signature computation part and the docs.

packages/airnode-deployer/src/cli/commands.ts
amarthadan: We do disallow that, at least when you're removing with the `receipt.json` as the airnode version from receipt is checked.

As for what would actually happen if you tried it, I think that it would succeed anyway. The TF state is stored in the bucket and you're doing a delete operation, so that would need to delete those resources that are actually deployed, so the ones from that state file.

packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -89,7 +91,32 @@ export async function testApi(event: AWSLambda.APIGatewayProxyEvent): Promise<AW
   const parameters = JSON.parse(event.body).parameters;
   const endpointId = event.pathParameters.endpointId;
 
-  const [err, result] = await handlers.testApi(parsedConfig, endpointId, parameters);
+  const [err, result] = await handlers.processHttpRequest(parsedConfig, endpointId, parameters);
+  if (err) {
+    return { statusCode: 400, body: JSON.stringify({ error: err.toString() }) };
```
andreogle, line 96: Is 400 (bad request) the right status code? This implies a request/client error

andreogle (approved):
This looks great to me 👍🏻

packages/airnode-deployer/src/handlers/aws/index.ts
amarthadan: Those error codes need to be revisited for sure. There's already an issue for that https://api3dao.atlassian.net/browse/AN-506

andreogle: Happy with that then. I've added an explicit comment to the ticket about reviewing the status codes as well

packages/airnode-node/src/api/index.ts
bbenligiray: I think you mean https://github.com/api3dao/airnode/blob/v1-protocol/packages/airnode-protocol-v1/contracts/protocol/AirnodeProtocol.sol#L300 and the order is wrong

bbenligiray:
@Siegrift This implements relayed RRP, it should have implemented [this](https://github.com/api3dao/airnode/blob/v1-protocol/packages/airnode-protocol-v1/contracts/dapis/DapiServer.sol#L402).

packages/airnode-node/src/api/index.ts
Siegrift: 🤦 sorry for the order mistake - I was careless.

amarthadan:
I think this can be merged now. Can you @bbenligiray, @andreogle take a look once more at key sections?

packages/airnode-node/src/api/index.ts
```diff
@@ -190,14 +219,21 @@ async function processSuccessfulApiCall(
...
+        // to compute it, since it is performance heavy operation.
+        return [[], { success: true, value: JSON.stringify(response), signature: 'not-yet-supported' }];
+      case 'regular': {
+        const signature = await signWithRequestId(aggregatedApiCall.id, response.encodedValue, config);
+        return [[], { success: true, value: response.encodedValue, signature }];
+      }
+      case 'http-signed-data-gateway': {
+        const timestamp = Math.floor(Date.now() / 1000).toString();
+        const signature = await signWithTemplateId(parameters._templateId, timestamp, response.encodedValue, config);
+        return [[], { success: true, value: JSON.stringify({ timestamp, value: response.encodedValue }), signature }];
```
bbenligiray, line 234: The timestamp will be used while pushing the data to the chain so we need to return it to the caller

amarthadan: The timestamp is there, you can see the resulting object example here https://github.com/api3dao/airnode/pull/870#pullrequestreview-893537683.
But as @Siegrift says, the format of different response types will be united in #871. I just don't want to add it to this PR as it's big enough as it is.

bbenligiray: Yeah sorry, I didn't see it for some reason

bbenligiray:
I only checked the signature implementation, looks good

PR #848, opened 2022-02-20
Use the same version of dependencies, run dependabot more often

Siegrift:
https://api3dao.atlassian.net/browse/AN-226


package.json
```diff
@@ -76,15 +77,12 @@
     "test:node:watch": "cd packages/airnode-node && yarn run test:watch",
     "test:validator": "(cd packages/airnode-validator && yarn run test)"
   },
-  "dependencies": {
-    "ethers": "^5.4.5",
```
Siegrift, line 80: These are not needed for the root package

Siegrift (commented):
I am explicitely not updating the packages to latest version, but I want to see if dependabt starts to be more active... Currently when you run `yarn outdated` you will see a lot of outdated dependecies...

packages/airnode-examples/package.json
```diff
@@ -35,23 +35,23 @@
...
-    "@types/node": "^16.10.3",
+    "@types/node": "^15.12.2",
     "@types/prompts": "^2.0.14",
     "chalk": "^4.1.2",
     "dotenv": "^10.0.0",
     "ethereum-waffle": "^3.4.0",
-    "hardhat": "^2.6.0",
-    "jest": "^27.2.5",
+    "hardhat": "^2.6.5",
+    "jest": "^26.6.3",
```
andreogle, line 51: Quite a few packages seem to have been downgraded (sometimes a major version like this). Any reason for that?

.github/dependabot.yml
```diff
@@ -3,7 +3,7 @@ updates:
   - package-ecosystem: npm
     directory: '/'
     schedule:
-      interval: monthly
+      interval: weekly
```
andreogle, line 6: I can't wait for more frequent dependabot PRs 😂

andreogle (approved):
Looks good. Just wondering about some of the downgrades

packages/airnode-examples/package.json
Siegrift: I've added a check to use the same versions of shared dependencies in the monorepo (see: https://github.com/JamieMason/syncpack#fix-mismatches). The advantage is that:
a) less space reserved in node modules (no one cares)
b) "Predictability" - e.g. I know that jest in package X behaves the same as in Y (this is pretty good for example for typescript).

And I did the "easy" way and used the version that was used in majority of the packages.

andreogle: Oh right - I missed the syncpack part. That's been a concern of mine for a while and it's great to have a new check in place 💯

package.json
amarthadan: Also some packages from `devDependencies` like:
* `@types/jest`
* `@types/lodash`
* `dotenv`
* `rimraf`

amarthadan (approved):

PR #851, opened 2022-02-21
AN136: Move default gasLimit to constants

vponline:
Move default gasLimit to constants, Jira: [AN136](https://api3dao.atlassian.net/jira/software/projects/AN/boards/18?selectedIssue=AN-136)


packages/airnode-node/src/evm/fulfillments/api-calls.test.ts
```diff
@@ -19,7 +19,7 @@ import * as fixtures from '../../../test/fixtures';
 import * as wallet from '../wallet';
 import { GasTarget, RequestErrorMessage, RequestStatus } from '../../types';
 import { AirnodeRrp } from '../contracts';
-import { MAXIMUM_ONCHAIN_ERROR_LENGTH } from '../../constants';
+import { MAXIMUM_ONCHAIN_ERROR_LENGTH, GAS_LIMIT } from '../../constants';
```
Siegrift, line 22: Why importing it in this test and not the others (e.g. process-transactions.test)

vponline: That was fast! I was actually going to ask if we should use the same constant in every place with the same `gasLimit` value, or only in the places where it had been defined as a constant previously.

Siegrift: Feel free to use it as variable everywhere.

packages/airnode-node/src/constants.ts
```diff
@@ -50,3 +50,6 @@ export const BASE_FEE_MULTIPLIER = 2;
 
 // The maximum character length of an error message sent on-chain
 export const MAXIMUM_ONCHAIN_ERROR_LENGTH = 100;
+
+// The default gas limit
+export const GAS_LIMIT = 500_000;
```
andreogle, line 55: Nit: It's not clear what this "gas limit" applies to. I would prefer a more descriptive name here if possible. 

```suggestion
export const DEFAULT_FULFILLMENT_GAS_LIMIT = 500_000;
```
or maybe `API_CALL_FULFILLMENT_GAS_LIMIT` (?)

Also, the comment above doesn't really add much in it's current state

andreogle (approved):

packages/airnode-node/src/constants.ts
```diff
@@ -50,3 +50,6 @@ export const BASE_FEE_MULTIPLIER = 2;
 
 // The maximum character length of an error message sent on-chain
 export const MAXIMUM_ONCHAIN_ERROR_LENGTH = 100;
+
+// The default gas limit for API fulfillment transactions
+export const API_CALL_FULFILLMENT_GAS_LIMIT = 500_000;
```
andreogle, line 55: Nice 👍🏻

andreogle (approved):

PR #865, opened 2022-02-21
Change structure of parameters for the http signed endpoint

Siegrift:
https://github.com/api3dao/airnode/pull/833#discussion_r809133540

### DIFF
--- a/packages/airnode-node/test/e2e/http-signed-relayed.feature.ts
+++ b/packages/airnode-node/test/e2e/http-signed-relayed.feature.ts
@@ -12,11 +12,9 @@ it('makes a call for signed relayed API data', async () => {
   await deployAirnodeAndMakeRequests(__filename);
 
   const parameters = {
-    from: 'ETH',
-    _type: 'int256',
-    _path: 'result',
-    _relayer: '0xA7b4c9bf0AF030a171c49400d3299703d3E97706',
-    _id: '0xcf2816af81f9cc7c9879dc84ce29c00fe1e290bcb8d2e4b204be1eeb120811bf',
+    parameters: { from: 'ETH', _type: 'int256', _path: 'result' },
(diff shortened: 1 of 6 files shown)


packages/airnode-node/test/e2e/http-signed-relayed.feature.ts
```diff
@@ -12,11 +12,9 @@ it('makes a call for signed relayed API data', async () => {
   await deployAirnodeAndMakeRequests(__filename);
 
   const parameters = {
-    from: 'ETH',
-    _type: 'int256',
-    _path: 'result',
-    _relayer: '0xA7b4c9bf0AF030a171c49400d3299703d3E97706',
-    _id: '0xcf2816af81f9cc7c9879dc84ce29c00fe1e290bcb8d2e4b204be1eeb120811bf',
+    parameters: { from: 'ETH', _type: 'int256', _path: 'result' },
```
Siegrift, line 15: @amarthadan I actually dislike this change. Why are the `relayer` and `id` parameters passed as root, but the `parameters` still contain the the `_type` and `_path` (it would make sense if they are all in the root of the object)

I'd instead do what I suggested in https://github.com/api3dao/airnode/pull/833#discussion_r810603816 and left kept using `_relayer` and `_id` for now. WDYT?

amarthadan: Oh, right, now I understand what you meant by that comment. Ok, in that case, I agree, let's do that then.

Siegrift: wanna do it now or as a separate issue? (because this also changes http gateway so we might do it separately)

amarthadan: Yeah, let's do it separately.

Siegrift:
Will be reworked once the signed http endpoint is merged.

PR #866, opened 2022-02-22
Formatting for Terraform files

amarthadan:
Issue [AN-543](https://api3dao.atlassian.net/browse/AN-543)

There is no Terraform Prettier plugin so I'm using an inbuilt `terraform fmt` command. I'm also ignoring the checks if someone doesn't have the Terraform binary available (the same goes for the CI checks). I don't want to make Terraform a dev dependency as it's an external tool and not everybody uses it.

The yarn scripts for formatting are now called `format:check` for checking and `format:write` for fixing the formatting issues.

I'm not sure this will work on Windows (actually, I think it won't), so, if someone can help me with that I'll be thankful. Maybe @aquarat?

Edit: I stand corrected, CI image contains Terraform binary so the checks will happen. Holy cow, there's so much [stuff in that image](https://github.com/actions/virtual-environments/blob/ubuntu20/20220220.1/images/linux/Ubuntu2004-Readme.md) :sweat_smile:


scripts/terraform-fmt.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const BINARY = 'terraform';
+const DIR = 'packages/airnode-deployer/terraform';
+const CMD_CHECK = `${BINARY} fmt -list=true -write=false -recursive ${DIR}`;
+const CMD_WRITE = `${BINARY} fmt -list=true -write=true -recursive ${DIR}`;
+
+if (process.argv.length !== 3) {
+  console.error('Wrong script usage!');
+  console.error('terraform-fmt.ts check | write');
+  exit(3);
```
andreogle, line 12: Is `exit(3)` right? Is it using these codes https://nodejs.org/api/process.html#process_exit_codes?

This also applies to the other places where `exit()` is called with a code of 2 or 3

amarthadan: Oh, I didn't know about these. I just wanted to distinguish between formatting error (1), processing error (2) and CLI error (3). I'll try to accommodate them even though I don't know if it will map well to those in the documentation.

amarthadan: Hmm, it looks like those are internal Node.js exit codes, I'm not sure those are intended to be used by an external code. I would rather suggest using codes let's say 21, 22, and 23 (higher than 13) so it's clear that those are codes returning from the script itself and not from the Node.js. What do you think, does that make more sense?

scripts/terraform-fmt.ts
```diff
@@ -0,0 +1,66 @@
+import { exec, ExecException } from 'child_process';
```
Siegrift, line 1: WDYT about rewriting this using https://github.com/google/zx ?

amarthadan: I mean, I can but I don't really see any additional value, especially for such a small linting script :man_shrugging:

Siegrift: I really like that package and how concise it is. I think it would be a good idea to start using it even for such a small script - but up to you. I do not feel strongly about this.

scripts/terraform-fmt.ts
amarthadan: I've changed the invalid argument to 22 but then I just followed with 23 and 24 as there were no appropriate codes in the errno list. And afaik that list is not commonly used, scripts and tools have usually their own error codes.

But we will avoid colliding with the internal Node.js error codes and at least one will be compatible with the Unix errno codes :man_shrugging:

scripts/terraform-fmt.ts
andreogle: I would lean towards keeping this as a regular TS script

aquarat: > WDYT about rewriting this using https://github.com/google/zx ?

This is very cool, I might use this for personal stuff.

scripts/terraform-fmt.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const command = process.argv[2];
+
+if (!['check', 'write'].includes(command)) {
+  console.error(`Unknown command '${command}'`);
+  exit(EC_ARGUMENTS);
+}
+
+exec(`command -v ${BINARY}`, (err, _stdout, _stderr) => {
```
aquarat, line 26: As per your Slack DM/my response, it may be better to stat the file first and/or just directly run it so that this command is compatible with Windows/PowerShell/`cmd.com`.

aquarat (approved):
I can confirm it works on Windows.

PR #867, opened 2022-02-22
Prohibit use of console and refactor where required

aquarat:
- Created `airnode-utilities` to reduce duplication
- Refactored `console.log`/`warn`/`error` calls to point to the new package
- Configured ESLint rule to error on use of `console`


.changeset/bright-hairs-camp.md
```diff
@@ -0,0 +1,12 @@
...
+'@api3/airnode-deployer': minor
+'@api3/airnode-examples': minor
+'@api3/airnode-node': minor
+'@api3/airnode-operation': minor
+'@api3/airnode-protocol': minor
+'@api3/airnode-utilities': minor
+'@api3/airnode-validator': minor
+---
+
+Refactored console.log/warn/error calls to point to an abstracted version of the function in a new package, airnode-utilities
```
Siegrift, line 12: Nit:

```suggestion
Refactored console calls to point to an abstracted version of the function in a new package, airnode-utilities
```

packages/airnode-utilities/src/logging.ts
```diff
@@ -0,0 +1,19 @@
+// @ts-ignore
```
Siegrift, line 1: Do we need this ts ignore?

packages/airnode-admin/package.json
```diff
@@ -26,6 +26,7 @@
   "dependencies": {
     "@api3/airnode-abi": "^0.4.0",
     "@api3/airnode-protocol": "^0.4.0",
+    "@api3/airnode-utilities": "^0.5.0",
```
Siegrift, line 29: Why is this 0.5?

package.json
```diff
@@ -27,6 +27,7 @@
     "build:operation": "(cd packages/airnode-operation && yarn run build)",
     "build:protocol": "(cd packages/airnode-protocol && yarn run build)",
     "build:validator": "(cd packages/airnode-validator && yarn run build)",
+    "build:utilities": "(cd packages/airnode-utilities && yarn run build)",
```
Siegrift, line 30: I wonder if we want to use a separate pacakge for this 🤔

packages/airnode-examples/src/cli.ts
```diff
@@ -32,7 +33,7 @@ export const runShellCommand = (command: string) => {
 };
 
 export const cliPrint = {
-  info: (text: string) => console.log(chalk.bold.white(text)),
-  warning: (text: string) => console.log(chalk.bold.hex('#FFA500')(text)), // Orange color
-  error: (text: string) => console.log(chalk.bold.red(text)),
+  info: (text: string) => log(chalk.bold.white(text)),
```
andreogle, line 36: I've been thinking about extracting other common utilities from Airnode and was thinking that it would be nice to extract the entire logger from the node package. So this would include the timestamp, log level, message and any metadata.

I was also thinking that a nice API would look like:

```ts
export const logger = {
  log: (...) => { ... },
  debug: (...) => { ... },
  info: (...) => { ... },
  warn: (...) => { ... },
  error: (...) => { ... }
};

import { logger } from '@api3/airnode-utilities';

logger.info('Hello world');
```

To me this is quite a nice intuitive, "traditional" logging API. What do you think? cc @Siegrift

andreogle: The logger would also be able to format as either JSON or regular text like the current node logger does

packages/airnode-utilities/src/logging.ts
```diff
@@ -0,0 +1,19 @@
...
+      }
+
+      return arg.toString();
+    })
+    .join(' ');
+
+export const log = (...args: any[]) => callConsoleFn('log', createLogMessage(args));
+export const logWarn = (...args: any[]) => callConsoleFn('warn', createLogMessage(args));
+export const logError = (...args: any[]) => callConsoleFn('error', createLogMessage(args));
+export const logTrace = (...args: any[]) => callConsoleFn('trace', createLogMessage(args));
```
amarthadan, line 19: TBH I don't like this approach much. I would rather have something more standard, e.g.
```js
import * as log from './logging';

log.info("Something to log")
```

amarthadan (commented):
I think it's fine in general but I would actually combine this with the [AN-478](https://api3dao.atlassian.net/browse/AN-478) issue and use an existing logging library instead of writing our own (either as part of monorepo or as a separate package). I don't think we have that specific needs so no library will cover it.

packages/airnode-examples/src/cli.ts
Siegrift: I'd rather use an existing solution like winston or pino. The logging is actually pretty hard concept (e.g. performance) to get right and I'd rather use some battle tested solution.

packages/airnode-utilities/src/logging.ts
Siegrift: I'd also prefer something like https://github.com/api3dao/airnode/blob/master/packages/airnode-examples/src/cli.ts#L34 (but I am biased since I wrote that part)

packages/airnode-examples/src/cli.ts
aquarat: > I've been thinking about extracting other common utilities from Airnode and was thinking that it would be nice to extract the entire logger from the node package. So this would include the timestamp, log level, message and any metadata.

I should have given more context to implementing this as a separate package, but my idea was exactly this, to move all the logging over to utilities eventually. Airkeeper has been using these facilities.

packages/airnode-deployer/package.json
```diff
@@ -26,6 +26,7 @@
     "@api3/airnode-node": "^0.4.0",
     "@api3/airnode-ois": "^0.4.0",
     "@api3/airnode-protocol": "^0.4.0",
+    "@api3/airnode-utilities": "^0.5.0",
```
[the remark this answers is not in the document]
aquarat: erf, thank you you for noticing :)

package.json
aquarat: We can discuss it further on the call later? I'm flexible.

packages/airnode-admin/package.json
aquarat: It can be 0.4.0, I just figured that's the version we're going to next.

package.json
Siegrift: Yeah, essentially what exactly to put into this package and what do we want from it in the long term...

packages/airnode-examples/src/cli.ts
aquarat: > I'd rather use an existing solution like winston or pino.

pino looks nice. My only concern is another 3rd-party dependency and then a supply chain attack like Log4j.

packages/airnode-utilities/src/logging.ts
```diff
@@ -0,0 +1,36 @@
+import { TextEncoder } from 'util';
```
aquarat, line 1: This is the new approach.
This is just for CLI utilities. I'm separately investigating/implementing Pino for Airnode's main logging facilities.

packages/airnode-deployer/src/cli/index.ts
```diff
@@ -6,31 +6,33 @@ import isEmpty from 'lodash/isEmpty';
 import uniq from 'lodash/uniq';
 import { hideBin } from 'yargs/helpers';
 import { CloudProvider, version as getNodeVersion } from '@api3/airnode-node';
+import { logger as loggerUtils } from '@api3/airnode-utilities';
```
andreogle, line 9: This file now has two loggers? Can we consolidate into a single one?

packages/airnode-deployer/src/handlers/gcp/index.ts
```diff
@@ -44,7 +44,6 @@ async function initializeProvider(payload: InitializeProviderPayload, res: Respo
   const [err, initializedState] = await utils.go(() => handlers.initializeProvider(stateWithConfig));
   if (err || !initializedState) {
     const msg = `Failed to initialize provider: ${stateWithConfig.settings.name}`;
-    console.log(err!.toString());
```
andreogle, line 47: Removing this is changing behaviour. Maybe it should be added back?

packages/airnode-admin/test/e2e/cli.feature.ts
```diff
@@ -9,6 +9,7 @@ import {
   AccessControlRegistryFactory,
```
andreogle, line 9: I can't comment on it but there's an empty file above `packages/airnode-admin/src/utils.ts` ?

packages/airnode-admin/src/cli.ts
```diff
@@ -510,15 +512,16 @@ yargs
...
     }
   )
   .demandCommand(1)
   .strict()
   .fail((message, err) => {
-    if (message) console.log(message);
-    else if (err instanceof Error) console.log(`Command failed with unexpected error:\n\n${err.message}`);
-    else console.log(`Command failed with unexpected error:\n\n${err}`);
+    if (message) logger.log(message);
+    else {
```
andreogle, line 522: This is a weird way of writing this - braces for the `else` only because the line is too long. Can we make it consistent?

packages/airnode-node/src/logger.ts
```diff
@@ -1,4 +1,5 @@
-import { formatDateTimeMs } from './utils/date-utils';
+import { logger as loggerUtils } from '@api3/airnode-utilities';
```
andreogle, line 1: Is this aliased as `loggerUtils` to reduce confusion with this file name? `logger` doesn't seem to be used anywhere else in this file

packages/airnode-protocol/dev-scripts/delete-flattened-licenses.js
```diff
@@ -1,9 +1,10 @@
 const replace = require('replace-in-file');
+const { log } = require('@api3/airnode-utilities');
```
andreogle, line 2: Maybe more consistent to important `logger` instead?

packages/airnode-utilities/README.md
```diff
@@ -0,0 +1,3 @@
+# `@api3/airnode-utilities`
+
+airnode-utilities provides common utilities for Airnode.
```
andreogle, line 3: haha I guess this is a WIP 😄

packages/airnode-utilities/src/logging.ts
```diff
@@ -0,0 +1,36 @@
...
+
+type LogType = 'DEBUG' | 'INFO' | 'WARN' | 'ERROR';
```
andreogle, line 3: Maybe this is a bit clearer than "type"?

```suggestion
type LogLevel = 'DEBUG' | 'INFO' | 'WARN' | 'ERROR';
```

packages/airnode-utilities/src/logging.ts
```diff
@@ -0,0 +1,36 @@
...
+      }
+
+      return arg.toString();
+    })
+    .join(' ');
+
+const writeLogMessage = (logType: LogType, args: any[]) => {
+  const formattedArgs = createLogMessage(args);
+
+  if (logType === 'ERROR') {
```
andreogle, line 21: So the level doesn't really matter for now apart from if it's set to `ERROR`?

andreogle (changes requested):

packages/airnode-deployer/src/cli/index.ts
aquarat: Sure, that was my next step but I figured it'd be best to start with this in this PR. I'm happy to do this now.

packages/airnode-admin/test/e2e/cli.feature.ts
aquarat: Thank you, I'll remove it :small_airplane:

packages/airnode-admin/src/cli.ts
aquarat: I dislike one-line ifs anyway (ternaries excluded).

packages/airnode-utilities/README.md
aquarat: I wasn't sure if a new monorepo package was a supported approach :stuck_out_tongue:

packages/airnode-deployer/src/cli/index.ts
```diff
@@ -6,31 +6,33 @@ import isEmpty from 'lodash/isEmpty';
...
 
 async function runCommand(command: () => Promise<void>) {
   try {
     await command();
   } catch (err) {
-    console.error(err);
-    // eslint-disable-next-line functional/immutable-data
-    process.exitCode = 1;
+    loggerUtils.error(err);
+    process.exit(1);
```
Siegrift, line 35: I think this was written like that for a reason I do not remember. Maybe @amarthadan does?

packages/airnode-deployer/src/handlers/gcp/index.ts
Siegrift: Did you? 👀

packages/airnode-utilities/README.md
Siegrift: Nit: Prefix it with ` >` so it's a bit nicer (but we are not consistent in Airnode even now, so...)|

Also create a JIRA ticket to enhance this a bit. Also, create issues for the things that should belong in here (I forgot now, but I remember mentioning them at the call yesteday)

Siegrift: Nit: Prefix it with ` >` so it's a bit nicer (but we are not consistent in Airnode even now, so...)|

Also create a JIRA ticket to enhance this a bit. Also, create issues for the things that should belong in here (I forgot now, but I remember mentioning them at the call yesteday)

packages/airnode-utilities/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@api3/airnode-utilities",
+  "version": "0.4.0",
+  "description": "Airnode utilities",
+  "homepage": "https://github.com/api3dao/airnode#readme",
```
Siegrift, line 5: Is this correct? Also other monorepo pacakges do not have this.

packages/airnode-utilities/src/logging.ts
Siegrift: Why using the TextEncoder instead of plain old console.log?

packages/airnode-utilities/src/logging.ts
```diff
@@ -0,0 +1,36 @@
...
+
+type LogType = 'DEBUG' | 'INFO' | 'WARN' | 'ERROR';
+
+const encoder = new TextEncoder();
+
+const createLogMessage = (args: any[]) =>
+  args
+    .map((arg) => {
+      if (arg instanceof Object) {
+        return JSON.stringify(arg, null, 2);
```
Siegrift, line 11: I'd say we don't want it pretty printed.

aquarat:
I will follow up with substantial changes shortly; I'm moving all the logging and utilities from airnode-node into airnode-utilities.
Currently it's just the e2e tests that need fixing.

packages/airnode-deployer/src/cli/index.ts
amarthadan: Yes, `yargs` didn't play well with calling `process.exit()`, I instead had to set the exit code and let them call the `exit` method.

aquarat: I'll revert it :+1:

packages/airnode-utilities/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "license": "MIT",
+  "files": [
+    "dist"
+  ],
+  "types": "dist/index",
```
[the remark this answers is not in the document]
aquarat: It works :shrug: 
and in airnode-validator we have:
```
  "main": "./dist/src/index.js",
  "types": "./dist/src/index",
```

packages/airnode-utilities/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "main": "dist/index",
+  "scripts": {
+    "build": "yarn clean && yarn compile",
+    "clean": "rimraf -rf ./dist *.tgz",
+    "compile": "tsc -p tsconfig.build.json",
+    "pack": "yarn pack"
```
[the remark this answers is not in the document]
aquarat: Removed :)

packages/airnode-utilities/src/logging.ts
aquarat: Removed :)

packages/airnode-deployer/src/handlers/gcp/index.ts
Siegrift: ping

packages/airnode-utilities/src/index.ts
```diff
@@ -0,0 +1,7 @@
+export * from './logging';
+export * from './utils/object-utils';
```
Siegrift, line 2: I really think these should have waited and be implemented as a separate issue.

packages/airnode-utilities/src/types.ts
```diff
@@ -0,0 +1,32 @@
+// ===========================================
```
Siegrift, line 1: I'd remove this and call this filename `logging-types.ts`

packages/airnode-utilities/src/utils/promise-utils.ts
```diff
@@ -1,5 +1,7 @@
 import { AttemptOptions, retry, sleep } from '@lifeomic/attempt';
-import { DEFAULT_RETRY_DELAY_MS } from '../constants';
+// import { DEFAULT_RETRY_DELAY_MS } from '../constants';
```
Siegrift, line 2: Remove.

packages/airnode-utilities/test/setup/init/set-define-property.ts
```diff
@@ -0,0 +1,18 @@
+// Sometimes Jest can have issues when trying to spy on an internal module. Import
+// and use this function as a workaround at the top of your test.
+//
+// Credit: https://github.com/facebook/jest/issues/6914#issuecomment-654710111
```
Siegrift, line 4: I actually wonder if we still need this file 🤔

Siegrift (commented):
I'll take a closer look after lunch, but I have to say that I like this very much. :)

packages/airnode-utilities/src/logging.ts
```diff
@@ -0,0 +1,133 @@
...
+
+const logLevels: { readonly [key in LogLevel]: number } = {
+  DEBUG: 0,
+  INFO: 1,
+  WARN: 2,
+  ERROR: 3,
+};
+
+export const logger = {
+  buildBaseOptions: (config: LogConfig, meta: LogMetadata) => ({
```
Siegrift, line 12: Maybe this could be a regular function outside the logger object?

packages/airnode-utilities/src/logging.ts
```diff
@@ -0,0 +1,133 @@
...
+    format: config.nodeSettings.logFormat,
+    level: config.nodeSettings.logLevel,
+    meta,
+  }),
+  log: (message: string, options?: LogOptions) => {
```
Siegrift, line 17: I wonder if we could allow creating a logger with options only once, so they don't have to be passed on every call - but that can be done as a separate issue.

Siegrift (commented):
👍 LGTM - there are a few minor comments, but after that I think this is good to go.

packages/airnode-utilities/src/logging.ts
aquarat: This would be wonderful.

packages/airnode-utilities/src/logging.ts
aquarat: That makes sense :+1:

aquarat:
> I'll take a closer look after lunch, but I have to say that I like this very much. :)

I am surprise but I won't complain :stuck_out_tongue:

packages/airnode-utilities/test/setup/init/set-define-property.ts
aquarat: I'll try removing it...

packages/airnode-utilities/src/utils/promise-utils.ts
aquarat: :see_no_evil: :+1:

packages/airnode-utilities/test/setup/init/set-define-property.ts
aquarat: It appears that it is not necessary in this package.

packages/airnode-deployer/src/handlers/gcp/index.ts
aquarat: Yes, sorry, it'll appear in the next commit (hence why I haven't marked this as resolved yet).

packages/airnode-utilities/test/setup/init/set-define-property.ts
aquarat: or any package except airnode-node :muscle:

packages/airnode-utilities/src/types.ts
```diff
@@ -0,0 +1,32 @@
...
+// Logging
+// ===========================================
+export type LogLevel = 'DEBUG' | 'INFO' | 'WARN' | 'ERROR';
+
+export type LogFormat = 'json' | 'plain';
+
+export interface LogMetadata {
+  readonly coordinatorId?: string;
```
andreogle, line 9: I don't know if we should use `readonly` for types on a utilities library... There might be valid cases where we want edit the properties. I know @Siegrift doesn't like these 😬 

I think a better option might be to use `Readonly<T>` where these types are used in the node/other packages.

andreogle: But this PR is already pretty big so I wouldn't do this as part of this branch (if you guys agree).

andreogle (approved):

packages/airnode-deployer/src/handlers/gcp/index.ts
aquarat: It has been replaced with `logger.log`.

PR #868, opened 2022-02-24
AN419: Display airnode address and xpub with generate-mnemonic

vponline:
Jira: [AN419](https://api3dao.atlassian.net/jira/software/projects/AN/boards/18?selectedIssue=AN-419)

### DIFF
--- a/packages/airnode-admin/src/cli.ts
+++ b/packages/airnode-admin/src/cli.ts
@@ -504,7 +504,7 @@ yargs
         '',
         mnemonic,
         '',
-        `The airnode address for this mnemonic is: ${airnodeAddress}`,
+        `The Airnode address for this mnemonic is: ${airnodeAddress}`,
         `The airnode xpub for this mnemonic is: ${airnodeXpub}`,
         '',
       ];
--- a/packages/airnode-admin/src/cli.ts
+++ b/packages/airnode-admin/src/cli.ts
@@ -505,7 +505,7 @@ yargs
         mnemonic,
         '',
         `The Airnode address for this mnemonic is: ${airnodeAddress}`,
-        `The airnode xpub for this mnemonic is: ${airnodeXpub}`,
+        `The Airnode xpub for this mnemonic is: ${airnodeXpub}`,
         '',
       ];
       lines.forEach((line) => console.log(line));
--- a/packages/airnode-admin/src/cli.ts
+++ b/packages/airnode-admin/src/cli.ts
@@ -495,11 +495,18 @@ yargs
     'Generates a random mnemonic. Uses "ethers.Wallet.createRandom" under the hood',
     async () => {
       const mnemonic = await admin.generateMnemonic();
+      const airnodeAddress = await admin.deriveAirnodeAddress(mnemonic);
+      const airnodeXpub = admin.deriveAirnodeXpub(mnemonic);
+
       const lines = [
         'This mnemonic is created locally on your machine using "ethers.Wallet.createRandom" under the hood.',
         'Make sure to back it up securely, e.g., by writing it down on a piece of paper:',
         '',
         mnemonic,
+        '',
+        `The airnode address for this mnemonic is: ${airnodeAddress}`,
+        `The airnode xpub for this mnemonic is: ${airnodeXpub}`,
+        '',
       ];
       lines.forEach((line) => console.log(line));
     }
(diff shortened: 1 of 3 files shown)


packages/airnode-admin/test/e2e/cli.feature.ts
```diff
@@ -412,23 +412,28 @@ describe('CLI', () => {
...
     expect(words).toHaveLength(12);
     words.forEach((word) => expect(word).toMatch(/\w+/));
+
+    const airnodeAddress = await admin.deriveAirnodeAddress(mnemonic);
+    expect(out[5]).toEqual(`The airnode address for this mnemonic is: ${airnodeAddress}`);
+
+    const airnodeXpub = admin.deriveAirnodeXpub(mnemonic);
+    expect(out[6]).toEqual(`The airnode xpub for this mnemonic is: ${airnodeXpub}`);
+
+    expect(() => admin.verifyAirnodeXpub(airnodeXpub, airnodeAddress)).not.toThrow();
```
vponline, line 436: This was probably not necessary, but I added it anyway

packages/airnode-admin/src/cli.ts
```diff
@@ -495,11 +495,18 @@ yargs
...
+      const airnodeAddress = await admin.deriveAirnodeAddress(mnemonic);
+      const airnodeXpub = admin.deriveAirnodeXpub(mnemonic);
+
       const lines = [
         'This mnemonic is created locally on your machine using "ethers.Wallet.createRandom" under the hood.',
         'Make sure to back it up securely, e.g., by writing it down on a piece of paper:',
         '',
         mnemonic,
+        '',
+        `The airnode address for this mnemonic is: ${airnodeAddress}`,
```
andreogle, line 507: Nit: I think we should titleize `Airnode` when outputting anything an end user would see. But I realise this is probably much a larger issue than just these two lines
```suggestion
        `The Airnode address for this mnemonic is: ${airnodeAddress}`,
```

packages/airnode-admin/test/e2e/cli.feature.ts
```diff
@@ -412,23 +412,28 @@ describe('CLI', () => {
     });
   });
 
-  it('generates mnemonic', () => {
-    const out = execCommand('generate-mnemonic');
+  it('generates mnemonic', async () => {
```
andreogle, line 415: Weird - how did this not break before? The generate-mnemonic function was always async 🤔

packages/airnode-admin/test/e2e/cli.feature.ts
andreogle: Nice one - I like it! Might be better to test the actual result though instead of just checking that it doesn't throw?

andreogle (approved):
Nice work! 🔥

packages/airnode-admin/test/e2e/cli.feature.ts
vponline: Good idea, added

packages/airnode-admin/test/e2e/cli.feature.ts
Siegrift: but it's used through CLI using the synchronous `execCommand` so async is not required here.

Siegrift (approved):
👍 LGTM, but remove the unncessary async keyword

packages/airnode-admin/test/e2e/cli.feature.ts
```diff
@@ -412,23 +412,30 @@ describe('CLI', () => {
...
-    const mnemonic = out.split(explanationInfo)[1];
+    ];
+    expect(out.slice(0, 3)).toEqual(explanationInfo);
 
+    const mnemonic = out[3];
     const words = mnemonic.split(' ');
     expect(words).toHaveLength(12);
     words.forEach((word) => expect(word).toMatch(/\w+/));
+
+    const airnodeAddress = await admin.deriveAirnodeAddress(mnemonic);
```
vponline, line 430: @Siegrift We need it `async` for deriving the address here though

Siegrift: Sorry, I missed this line and got confused by comment from @andreogle (who I think also missed this line :D )

All good here.

Siegrift (approved):

PR #869, opened 2022-02-24
Add CLI examples to admin and deployer packages

vponline:
Left out the validator CLI as it may change, but can be added as needed.

### DIFF
--- a/packages/airnode-admin/src/cli.ts
+++ b/packages/airnode-admin/src/cli.ts
@@ -3,6 +3,7 @@ import { exit } from 'process';
 import * as yargs from 'yargs';
 import * as evm from './evm';
 import * as admin from './implementation';
+import { cliExamples } from './cli-examples';
 
 const COMMON_COMMAND_ARGUMENTS = {
   airnodeRrpCommands: {
@@ -520,13 +521,7 @@ yargs
       console.log(`Airnode address: ${airnodeAddress}`);
     }
   )
-  .example(
-    [
-      '$0 derive-airnode-address --airnode-mnemonic "test test test test test test test test test test test junk"',
-      '$0 derive-sponsor-wallet-address --airnode-xpub xpub6Ce9NcJvTk36xtLSrJLZqE7wtgA5deCeYs7rSQtreh4cj6ByPtrg9sD7V2FNFLPnf8heNP3FGkeV9qwfzvZNSd54JoNXVsXFYSYwHsnJxqP --airnode-address 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --sponsor-address 0x70997970c51812dc3a010c7d01b50e0d17dc79c8',
-      '$0 derive-endpoint-id --ois-title "title" --endpoint-name "endpoint"',
-    ].map((line) => [line + '\n'])
-  )
+  .example(cliExamples.map((line) => [`$0 ${line}\n`]))
   .demandCommand(1)
   .strict()
   .fail((message, err) => {
--- a/packages/airnode-admin/src/cli.ts
+++ b/packages/airnode-admin/src/cli.ts
@@ -523,7 +523,7 @@ yargs
   .example([
     ['$0 derive-airnode-address --airnode-mnemonic "test test test test test test test test test test test junk"\n'],
     [
-      '$0 derive-sponsor-wallet-address --airnode-xpub xpub6Ce9NcJvTk36xtLSrJLZqE7wtgA5deCeYs7rSQtreh4cj6\nByPtrg9sD7V2FNFLPnf8heNP3FGkeV9qwfzvZNSd54JoNXVsXFYSYwHsnJxqP --airnode-address 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --sponsor-address 0x70997970c51812dc3a010c7d01b50e0d17dc79c8\n',
+      '$0 derive-sponsor-wallet-address --airnode-xpub xpub6Ce9NcJvTk36xtLSrJLZqE7wtgA5deCeYs7rSQtreh4cj6ByPtrg9sD7V2FNFLPnf8heNP3FGkeV9qwfzvZNSd54JoNXVsXFYSYwHsnJxqP --airnode-address 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --sponsor-address 0x70997970c51812dc3a010c7d01b50e0d17dc79c8\n',
     ],
     ['$0 derive-endpoint-id --ois-title "title" --endpoint-name "endpoint"\n'],
   ])
--- a/packages/airnode-admin/src/cli.ts
+++ b/packages/airnode-admin/src/cli.ts
@@ -521,18 +521,11 @@ yargs
     }
   )
   .example([
+    ['$0 derive-airnode-address --airnode-mnemonic "test test test test test test test test test test test junk"\n'],
     [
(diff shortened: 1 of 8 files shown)


packages/airnode-admin/src/cli.ts
```diff
@@ -520,6 +520,13 @@ yargs
       console.log(`Airnode address: ${airnodeAddress}`);
     }
   )
+  .example([
+    ['$0 derive-airnode-address --airnode-mnemonic "test test test test test test test test test test test junk"\n'],
```
Siegrift, line 524: Let's define this array without new lines and then do:
```
[...].map(line => line + '\n')
```

packages/airnode-admin/src/cli.ts
```diff
@@ -520,6 +520,13 @@ yargs
...
+    [
+      '$0 derive-sponsor-wallet-address --airnode-xpub xpub6Ce9NcJvTk36xtLSrJLZqE7wtgA5deCeYs7rSQtreh4cj6\nByPtrg9sD7V2FNFLPnf8heNP3FGkeV9qwfzvZNSd54JoNXVsXFYSYwHsnJxqP --airnode-address 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --sponsor-address 0x70997970c51812dc3a010c7d01b50e0d17dc79c8\n',
```
Siegrift, line 526:
```suggestion
      '$0 derive-sponsor-wallet-address --airnode-xpub xpub6Ce9NcJvTk36xtLSrJLZqE7wtgA5deCeYs7rSQtreh4cj6ByPtrg9sD7V2FNFLPnf8heNP3FGkeV9qwfzvZNSd54JoNXVsXFYSYwHsnJxqP --airnode-address 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --sponsor-address 0x70997970c51812dc3a010c7d01b50e0d17dc79c8\n',
```

otherwise the command doesn't work.

packages/airnode-admin/test/e2e/__snapshots__/cli.feature.ts.snap
```diff
@@ -30,5 +30,14 @@ Commands:
 
 Options:
   --version  Show version number                                                                               [boolean]
-  --help     Show help                                                                                         [boolean]"
+  --help     Show help                                                                                         [boolean]
+
+Examples:
```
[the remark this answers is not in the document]
Siegrift: I agree we do not need the descriptions.

packages/airnode-deployer/src/cli/index.ts
```diff
@@ -158,6 +158,11 @@ yargs(hideBin(process.argv))
       }
     }
   )
+  .example([
+    ['$0 deploy -c myConfig/config.json -s myConfig/secrets.env -r myOutput/receipt.json'],
```
Siegrift, line 162:
```suggestion
    ['$0 deploy -c pathTo/config.json -s pathTo/secrets.env -r myOutput/receipt.json'],
```

reads a bit better

packages/airnode-admin/src/cli.ts
Siegrift, line 524: Could you define these commands in some variable and write e2e tests for these. It would be nice to maintain the validity of those examples. (and here it's easy to do - you can skip the deployer commands since they are more challenging).

Siegrift (changes requested):

packages/airnode-admin/test/e2e/cli.feature.ts
```diff
@@ -660,4 +660,24 @@ describe('CLI', () => {
       expect(out).toEqual(`Airnode address: ${airnodeWallet.address}`);
     });
   });
+
+  describe('has valid examples', () => {
+    const examples = [
+      'derive-airnode-address --airnode-mnemonic "test test test test test test test test test test test junk"',
+      'derive-sponsor-wallet-address --airnode-xpub xpub6Ce9NcJvTk36xtLSrJLZqE7wtgA5deCeYs7rSQtreh4cj6ByPtrg9sD7V2FNFLPnf8heNP3FGkeV9qwfzvZNSd54JoNXVsXFYSYwHsnJxqP --airnode-address 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --sponsor-address 0x70997970c51812dc3a010c7d01b50e0d17dc79c8',
+      'derive-endpoint-id --ois-title "title" --endpoint-name "endpoint"',
+    ];
```
vponline, line 669: It looks like these examples cannot be exported from `cli.ts` (maybe because it's the `yargs` command file? and it ended up running the commands instead) so I just copied in the array, but these could be moved to a common `cli-examples.ts` file if that's better.

packages/airnode-deployer/src/cli/index.ts
vponline: :+1: Ill make the corresponding updates to docs as well

packages/airnode-admin/test/e2e/cli.feature.ts
Siegrift: Yeah, that would be better.

packages/airnode-admin/test/e2e/cli.feature.ts
```diff
@@ -660,4 +661,19 @@ describe('CLI', () => {
...
   });
+
+  describe('has valid examples', () => {
+    const exampleOutcomes = [
+      'Airnode address: 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266',
+      'Sponsor wallet address: 0x61cF9Eb3691A715e7B2697a36e9e60FdA40A8617',
+      'Endpoint ID: 0x901843fb332b24a9a71a2234f2a7c82214b7b70e99ab412e7d1827b743f63f61',
+    ];
+
+    test.each(cliExamples)(`example command: %# is valid`, (example: string) => {
```
Siegrift, line 672: Won't `%#` result in an extremely long test message? I'd just use index instead.

Siegrift: I actually dislike a bit the `test.each` helper and would prefer this instead:

```
cliExamples.forEach((command, index) => {
it('tests example command ${index}', () => {...})
})
```

packages/airnode-deployer/README.md
```diff
@@ -91,4 +94,9 @@ Options:
   -c, --cloud-provider         Cloud provider                                                    [choices: "aws", "gcp"]
   -e, --region                 Region                                                                           [string]
   -p, --project-id             Project ID (GCP only)                                                            [string]
+
+Examples:
+  deployer.ts remove -r myOutput/receipt.json
```
Siegrift, line 99: Btw. this `deployer.ts` is there only because the filename is `deployer.ts` and you probably ran `yarn cli --help`.

Also, I'd probably remove this `### Commands` section. WDYT: @amarthadan ?

Siegrift (changes requested):

packages/airnode-deployer/README.md
vponline: Thanks, fixed!

packages/airnode-admin/test/e2e/cli.feature.ts
Siegrift: Oh, I thought it's interpolating the command. (this is also a reason I don't like it - I find it hard to remember what is interpolated...)

Siegrift (approved):
👍 LGTM

PR #870, opened 2022-02-24
Add http signed relay gateway example to simplify testing

Siegrift:
https://api3dao.atlassian.net/browse/AN-557


Siegrift (commented):
I just tested the new updates (rename of the handler) and got this response:

```
{"data":"{\"timestamp\":\"1645786263\",\"value\":\"0x00000000000000000000000000000000000000000000000000000008fd651ac0\"}","signature":"0xbc017f87fe39795e5ab89511b4a76185ee2426316a259b4dc1448fea4dc7991373ec35b428529f25a25067a05d7dd6b2abd37b8441e9051a17631fa0d78bde751c"}
```

The `data` contains stringified response (which is not nice), but that will be resolved in https://github.com/api3dao/airnode/pull/871

packages/airnode-examples/integrations/coingecko-signed-data/README.md
```diff
@@ -0,0 +1,28 @@
+# Coingecko-signed-data example integration
+
+This example contains the same API integration as `../coingecko`.
+
+The only difference is that this integration allows the endpoint to return signed data that can be submitted on chain by
+the to update a beacon.
```
vponline, line 6: This sentence seems to be missing a word :thinking:

amarthadan: Yes, looks like it. Shorten it to just
`The only difference is that this integration allows the endpoint to return signed data that can be then submitted on chain.`

amarthadan (approved):
I am a bit confused by the number and content of your commits. It should have been just one commit with a few lines changed in the `airnode-examples` package. What happened there? :smile:

But the result seems fine and it's building so the renaming probably went well. I'll leave the testing up to @dcroote.

vponline:
Ok this was a misunderstanding on my part, I renamed `signedData` to `httpSignedData` (and affected variables/functions) everywhere. My bad :pray:

packages/airnode-examples/contracts/coingecko-signed-data/Requester.sol
```diff
@@ -0,0 +1,41 @@
...
+    function fulfill(bytes32 requestId, bytes calldata data)
+        external
+        onlyAirnodeRrp
+    {
+        require(incomingFulfillments[requestId], "No such request made");
+        delete incomingFulfillments[requestId];
+        int256 decodedData = abi.decode(data, (int256));
+        fulfilledData[requestId] = decodedData;
+    }
+}
```
vponline, line 41: I added this (copy of Coingecko)

amarthadan (approved):

PR #872, opened 2022-02-25
Signed data endpoint rework

Siegrift:
I implemented the endpoint logic wrong and misunderstood how the contracts work.
This is the rework based on the following comment: https://api3dao.atlassian.net/browse/AN-352?focusedCommentId=14092


packages/airnode-node/src/api/index.ts
```diff
@@ -69,34 +69,25 @@ function buildOptions(payload: CallApiPayload): adapter.BuildRequestOptions {
...
 }
 
-async function signRelayedResponseMessage(
-  requestId: string,
-  timestamp: string,
-  relayerAddress: string,
-  responseValue: string,
-  config: Config
-) {
+async function signDataForBeaconUpdate(templateId: string, timestamp: string, data: string, config: Config) {
```
bbenligiray, line 83: Don't refer to beacons specifically anywhere in the Airnode code, it's just an arbitrary requester contract implementation. This is signing with the template ID so could practically be used for anything.

packages/airnode-node/src/api/index.ts
```diff
@@ -69,34 +69,25 @@ function buildOptions(payload: CallApiPayload): adapter.BuildRequestOptions {
...
   const airnodeWallet = ethers.Wallet.fromMnemonic(masterHDNode.mnemonic!.phrase);
 
   return await airnodeWallet.signMessage(
     ethers.utils.arrayify(
       ethers.utils.keccak256(
-        ethers.utils.solidityPack(
-          ['bytes32', 'uint256', 'address', 'bytes'],
-          [requestId, timestamp, relayerAddress, responseValue || '0x']
-        )
+        ethers.utils.solidityPack(['bytes32', 'uint256', 'bytes'], [templateId, timestamp, data || '0x'])
```
bbenligiray, line 90: Looks good

amarthadan (approved):

PR #873, opened 2022-02-26
AN-160 | Implement the "require minimum confirmations" functionality

Ashar2shahid:
minimum confirmations refers to the number of blocks that have elapsed since the current confirmed block according to [this](https://ethereum.stackexchange.com/a/7437) and a few other sources. The changes to the code were simple enough, so let me know if I've missed anything.


packages/airnode-node/src/evm/requests/event-logs.ts
```diff
@@ -33,12 +34,13 @@ export function parseAirnodeRrpLog<T = { readonly args: any }>(log: ethers.provi
 }
 
 export async function fetch(options: FetchOptions): Promise<EVMEventLog[]> {
-  // Protect against a potential negative fromBlock value
+  // Protect against a potential negative fromBlock and toBlock value
   const fromBlock = Math.max(0, options.currentBlock - options.blockHistoryLimit);
+  const toBlock = Math.max(0, options.currentBlock - options.minConfirmations);
```
andreogle, line 39: `toBlock` should always be >= `fromBlock`.
```suggestion
  const toBlock = Math.max(fromBlock, options.currentBlock - options.minConfirmations);
```

packages/airnode-node/src/evm/requests/event-logs.ts
```diff
@@ -33,12 +34,13 @@ export function parseAirnodeRrpLog<T = { readonly args: any }>(log: ethers.provi
 }
 
 export async function fetch(options: FetchOptions): Promise<EVMEventLog[]> {
-  // Protect against a potential negative fromBlock value
+  // Protect against a potential negative fromBlock and toBlock value
```
andreogle, line 37: This sentence isn't syntactically correct anymore. Maybe revert to what it was and add a new comment above `const toBlock`. Something along the lines of `// toBlock should always be >= fromBlock`

packages/airnode-node/src/evm/requests/event-logs.ts
andreogle: I think there should also be a new unit test that covers this scenario

andreogle (changes requested):

packages/airnode-node/src/evm/requests/event-logs.ts
Ashar2shahid: right, i'll add it in

packages/airnode-node/src/evm/requests/event-logs.ts
Ashar2shahid: I was thinking more along the lines of `Protect against a potential toBlock value that is lower than fromBlock value`  but that also works.

.changeset/plenty-emus-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@api3/airnode-node': patch
```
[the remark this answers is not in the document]
Ashar2shahid: hmm, I was thinking this was a patch since the field already existed and all we did was "patch" in the functionality but yeah the case of minor also exists.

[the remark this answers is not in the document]
andreogle: I think we should stick to semver which dictates that "minor" releases are for new, non-breaking functionality and "patch" releases are for backwards compatible bug fixes. https://semver.org/#summary

I think I would consider this new functionality rather than a bug fix. cc @Siegrift @amarthadan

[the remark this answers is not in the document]
amarthadan: @andreogle is correct. Patch is only for fixing. If there is a bug or security issue that needs to be fixed that's patch. Adding new functionality or changing some functionality (not fixing the problem but changing correct behavior to a different correct behavior) should be minor.

I would also consider this a new functionality.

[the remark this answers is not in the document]
Ashar2shahid: got it changing it to fit that description

PR #874, opened 2022-02-27
Remove Ignored variant of RequestStatus enum

dcroote:
[AN-482](https://api3dao.atlassian.net/browse/AN-482), part 2


packages/airnode-node/src/evm/verification/request-verification.ts
```diff
@@ -16,32 +16,34 @@ export function verifySponsorWallets<T>(
...
       const message = `Invalid sponsor wallet:${request.sponsorWalletAddress} for Request:${request.id}. Expected:${expectedSponsorWalletAddress}`;
       const log = logger.pend('ERROR', message);
-      const updatedRequest: Request<T> = {
-        ...request,
-        status: RequestStatus.Ignored,
-        errorMessage: `${RequestErrorMessage.SponsorWalletInvalid}: ${request.sponsorWalletAddress}`,
-      };
-      return [[log], updatedRequest];
+      logs.push(log);
+      return;
```
dcroote, line 37: Note here that only the `log` is pushed and not the `request` i.e. the `request` is dropped.

Using two separate arrays for `logs` and `verifiedRequests` in conjunction with `forEach` is less appealing than the clean `map` that existed before, but I needed a way to drop the request while still pushing the log for when `request.sponsorWalletAddress !== expectedSponsorWalletAddress` and I couldn't get around the map return type constraint.

packages/airnode-node/src/evm/requests/blocking.ts
```diff
@@ -32,21 +32,16 @@ export function blockRequestsWithWithdrawals([
...
-      status: RequestStatus.Ignored,
-      // TODO: These error messages should be functions accepting id parameter and creating the error string
-      errorMessage: `${RequestErrorMessage.PendingWithdrawal}: ${pendingWithdrawal.id}`,
-    };
 
     return {
       ...acc,
       logs: [...acc.logs, warningLog],
-      apiCalls: [...acc.apiCalls, blockedCall],
+      apiCalls: [...acc.apiCalls],
```
Siegrift, line 44:
```suggestion
      apiCalls,
```

packages/airnode-node/src/evm/requests/blocking.ts
```diff
@@ -32,21 +32,16 @@ export function blockRequestsWithWithdrawals([
       return { ...acc, apiCalls: [...acc.apiCalls, apiCall] };
     }
 
+    // Drop request, but log that it is being ignored
```
Siegrift, line 35: I'd remove this comment - this is apparent from the code.

packages/airnode-node/src/evm/verification/request-verification.ts
```diff
@@ -16,32 +16,34 @@ export function verifySponsorWallets<T>(
   requests: Request<T>[],
   masterHDNode: ethers.utils.HDNode
 ): LogsData<Request<T>[]> {
-  const logsWithVerifiedRequests: LogsData<Request<T>>[] = requests.map((request) => {
+  const logs: PendingLog[] = [];
+  const verifiedRequests: Request<T>[] = [];
+
+  requests.forEach(function (request) {
     if (request.status !== RequestStatus.Pending) {
```
Siegrift, line 23: This will have to go eventually as well.

packages/airnode-node/src/evm/verification/request-verification.ts
Siegrift: I'd keep an `logs` array just as you use now and would use the `.filter` to filter the actionable requests.

packages/airnode-node/src/requests/request.ts
```diff
@@ -1,9 +1,9 @@
 import isEmpty from 'lodash/isEmpty';
 import { ApiCall, Request, GroupedRequests, RequestStatus, Withdrawal } from '../types';
 
-export function blockedOrIgnored<T>(request: Request<T>): RequestStatus.Blocked | RequestStatus.Ignored {
+export function shouldDropAfterBlockLimit<T>(request: Request<T>): boolean {
```
Siegrift, line 4: This name is confusing. I'd suggest `isRequestBlocked`.

Siegrift (commented):
From a quick pass LGTM

packages/airnode-node/src/evm/requests/blocking.ts
dcroote: Reverting as the suggestion fails CI

packages/airnode-node/src/evm/verification/request-verification.ts
dcroote: Yes indeed 😅

packages/airnode-node/src/requests/request.ts
dcroote: Fair- `shouldDrop` doesn't pertain to what the function does. I wouldn't say `isRequestBlocked` is right for what the function does either though given it operates on blocked requests and returns a bool- how about `hasExceededIgnoredBlockLimit` - this is specific and clearly returns a bool.

packages/airnode-node/src/evm/requests/blocking.ts
Siegrift: Does it? It probably should be `apiCalls: acc.apiCalls` but destrucutre is not needed.

dcroote: Here is the failure: https://github.com/api3dao/airnode/runs/5351085862?check_suite_focus=true#step:6:133

But actually since the accumulator isn't being updated the whole line can be deleted 😄

packages/airnode-node/src/evm/verification/request-verification.ts
dcroote: `.filter` worked nicely, thanks

packages/airnode-node/src/requests/request.ts
andreogle: I like `hasExceededIgnoredBlockLimit`

packages/airnode-node/src/evm/requests/blocking.ts
```diff
@@ -36,17 +36,10 @@ export function blockRequestsWithWithdrawals([
...
-      ...apiCall,
-      status: RequestStatus.Ignored,
-      // TODO: These error messages should be functions accepting id parameter and creating the error string
-      errorMessage: `${RequestErrorMessage.PendingWithdrawal}: ${pendingWithdrawal.id}`,
-    };
 
     return {
       ...acc,
       logs: [...acc.logs, warningLog],
-      apiCalls: [...acc.apiCalls, blockedCall],
```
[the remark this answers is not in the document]
dcroote: Yes- [this test](https://github.com/api3dao/airnode/blob/0c0e2b2f7202926bbf7a665d3288f36de0142087/packages/airnode-node/src/evm/requests/blocking.test.ts#L10-L30) now is an example of exactly your scenario where a request with a pending withdrawal is dropped. 

My understanding in that removing the `RequestStatus` enum (only the `Ignored` variant is addressed in this PR) is meant to be an explicit departure from the previous behavior by instead "drop blocked/ignored requests immediately" ([AN-482](https://api3dao.atlassian.net/browse/AN-482)).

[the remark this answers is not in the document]
dcroote: I've opened a docs PR to make this explicit: https://github.com/api3dao/api3-docs/pull/653 and updated the `changeset` as well.

packages/airnode-node/src/evm/requests/blocking.ts
```diff
@@ -26,19 +26,12 @@ export function blockRequestsWithWithdrawals([
 
     const warningLog = logger.pend(
       'WARN',
-      `Ignoring Request ID:${apiCall.id} as it has a pending Withdrawal ID:${pendingWithdrawal.id}`
+      `Dropping Request ID:${apiCall.id} as it has a pending Withdrawal ID:${pendingWithdrawal.id}`
```
andreogle, line 29: Thanks for updating this 👍🏻

packages/airnode-node/src/evm/verification/request-verification.ts
```diff
@@ -16,32 +16,30 @@ export function verifySponsorWallets<T>(
...
 ): LogsData<Request<T>[]> {
-  const logsWithVerifiedRequests: LogsData<Request<T>>[] = requests.map((request) => {
+  const logs: PendingLog[] = [];
+
+  const verifiedRequests = requests.filter((request) => {
     if (request.status !== RequestStatus.Pending) {
       const message = `Sponsor wallet verification skipped for Request:${request.id} as it has status:${request.status}`;
       const log = logger.pend('DEBUG', message);
-      return [[log], request];
+      logs.push(log);
```
andreogle, line 25: I can understand that `filter` is a bit easy to read than the current `map` implementation, but I don't think we should have side effects from a `.filter` callback.

packages/airnode-node/src/requests/nonces.ts
```diff
@@ -28,12 +28,14 @@ function assignWalletNonces(flatRequests: Request<AnyRequest>[], transactionCoun
     }
 
     if (request.status === RequestStatus.Blocked) {
-      const status = blockedOrIgnored(request);
-      const assignmentBlocked = status === RequestStatus.Blocked;
-
+      if (hasExceededIgnoredBlockLimit(request)) {
+        return {
```
andreogle, line 32: `return acc;` No need to spread here

andreogle (changes requested):

packages/airnode-node/src/evm/verification/request-verification.ts
dcroote: Agreed it's not ideal, but `map` no longer works, see [here](https://github.com/api3dao/airnode/pull/874#discussion_r815377836). I had a loop with two arrays, then `filter` was suggested so I went with that.

packages/airnode-node/src/evm/verification/request-verification.ts
andreogle: I prefer `.reduce` to pushing to the array as a side effect from `.filter`, but if you guys prefer the side effect approach then I'm ok with that. It's less readable to use `.reduce` but I would definitely consider it "safer"

dcroote: I implemented it with `reduce`, but particularly with prettier formatting it became very easy to miss how requests were being filtered. I'll leave as is, but this code block will be revisited when other `RequestStatus` enum variants are dropped, at which point I'll have another look if there is a way to simplify and/or use `.reduce`.

dcroote: @andreogle - if my above comment works for you, could you approve so I can merge?

andreogle: Sure - happy with that then. Thanks for taking a look 👍🏻

andreogle (approved):

PR #890, opened 2022-02-28
Update EIP1559 config values to numbers

vponline:
Minor refactoring to config types and affected tests


packages/airnode-validator/src/v2/config/config.ts
```diff
@@ -44,7 +44,7 @@ export const priorityFeeSchema = z.object({
 
 export const chainOptionsSchema = z.object({
   txType: z.union([z.literal('legacy'), z.literal('eip1559')]),
-  baseFeeMultiplier: z.string().optional(),
+  baseFeeMultiplier: z.number().int().optional(),
```
vponline, line 47: Zod also has more options for numbers, see [here](https://github.com/colinhacks/zod#numbers) (e.g. only positive numbers).

packages/airnode-validator/templates/0.4/config.json
```diff
@@ -135,16 +135,14 @@
...
-            "__regexp": "^[0-9]$",
+            "__type": "number",
             "__catch": {
               "__message": "__fullPath: Base Fee multiplier must be a whole number"
             }
           },
           "priorityFee": {
             "value": {
-              "__type": "string",
-              "__regexp": "^\\d*\\.?\\d*$",
```
vponline, line 147: I simply removed the `regexp` since Zod has more options for numbers that can be used.

aquarat (approved):
It LGTM :tractor:, nice use of zod :)

packages/airnode-examples/integrations/relay-security-schemes/config.example.json
```diff
@@ -15,9 +15,9 @@
       "type": "evm",
       "options": {
         "txType": "eip1559",
-        "baseFeeMultiplier": "2",
+        "baseFeeMultiplier": 2,
         "priorityFee": {
-          "value": "3.12",
```
andreogle, line 20: What's the advantage of changing this? We could lose precision by using JS numbers?

vponline: Tbh, I'm not sure besides perhaps keeping things more consistent with other config fields (see this [task](https://api3dao.atlassian.net/browse/AN-489))

aquarat: I originally made them strings when I implemented eip1559 pricing, I think because of the imprecision issue and I think the value was being ingested underneath by ethers as a string but I didn't feel strongly either way. There was an extended discussion on the topic and the outcome was that they be changed to `number`. Discussion here: https://api3workspace.slack.com/archives/C02AYRX8D89/p1641981419038200?thread_ts=1641981419.038200&cid=C02AYRX8D89

aquarat:
Good job anticipating the timeout revert - hopefully once the other PR is merged this branch should build properly.

packages/airnode-examples/integrations/relay-security-schemes/config.example.json
andreogle: Sorry I missed that thread. I still think strings would be slightly safer but don't feel strongly enough about it. As you mentioned, these numbers should always be really small and not require much precision.

bbenligiray: Yeah it's a bit of an imaginary problem

PR #891, opened 2022-02-28
Temporarily disable link checking in CI actions

aquarat:
The new redirect system in the API3 docs, as a byproduct of the way Github Pages serves responses for unknown pages, has broken the link checker in Airnode's CI system.

This PR disables the link checker by commenting it out in the Github Actions workflow.

Extended detail:
The API3 docs are a single-page application implemented in Vue. If a browser visits a redirect page, for example `/airnode/latest`, the browser is first served a 404 (in that the status code of the response is 404) and then the Javascript redirects the user. Unfortunately the link checker doesn't wait for the redirect and only sees the 404.

Currently we preview docs builds on Firebase and Firebase has the ability to configure a rewrite rule, which they recommend be used for SPAs. In my tests this worked well, so perhaps we should switch production docs to Firebase too?


dcroote:
Reenabled in #901

PR #893, opened 2022-03-01
Add back yarn pack to fix docker build process

aquarat:
`yarn pack` is an important command because it is used by the artifacts build process.
The lack of this command meant that `airnode-utilities` didn't make its way into the airnode-client image and hence e2e tests started failing.


vponline (approved):
LGTM :smile:

PR #894, opened 2022-03-01
Add `describe` to protocol tests

vponline:
This wraps `airnode-protocol` tests with a `describe` to improve test times (based on [this](https://github.com/NomicFoundation/hardhat/issues/2076)) It also removes @Siegrift's custom test script from the package as it is no longer needed.

Here are the test times I got locally: 
- without describe: 335.96s
- test-script: 57.41s
- with describe: 41.90s


vponline:
> Wow that is a seriously impressive speed up. I'm not sure about converting the `function() {}` blocks to arrow functions though... [Mocha discourages this](https://mochajs.org/#arrow-functions)

Good to know, I wasn't aware. It doesn't look like `this` is used for now, but perhaps I should convert back in case it is needed?

bbenligiray:
> perhaps I should convert back in case it is needed

Sounds like a good idea to me

andreogle:
Yeah I think just sticking with `function() {}` everywhere is just more consistent and it makes it less likely that future devs won't run into hard to diagnose issues

PR #897, opened 2022-03-01
Revert Merge in Change EIP1559 config value string type switch

aquarat:
This revert _should_ fix a bug introduced by an eip1559-related type change, originally merged from PR #890


vponline (approved):
Thanks for reverting, I'll investigate.

PR #900, opened 2022-03-02
Add esbuild-loader to deployer

vponline:
This adds `esbuild-loader` to deployer webpack to improve the build time

Github build times were
- without 63s
- with 19s

Local
- without 18s
- with 9s

Further improvements may be possible by replacing webpack with esbuild completely.


amarthadan:
Have you also tried deploying the airnode to see whether the new bundle works?

vponline:
Yes (coingecko example) and it worked as well :+1:

amarthadan (approved):
Works like a charm :+1:

PR #902, opened 2022-03-03
Airnode caching PoC

aquarat:
This is a draft PR for an Airnode caching PoC.

This is not intended to be merged into main, just for discussion.

This PR causes Airnode to:
- Cache blocked requests (fs-based)
- Cache wallet derivations in-memory (variable based)
- Cache API responses per requestId (fs based)

TODO:
- Limit wallet derivation cache size/sweeping
- Investigate blocked requests edge cases
- Stress test
- Investigate use of a dedicated caching function to centralise the cache
- Investigate multi-region caching using above function (push and forget via http gateway)
- Investigate making everything async (we never mutate, only ever append)


aquarat:
Closed due to additional requirements.

PR #904, opened 2022-03-03
Use standard packing for subscription ID derivation

bbenligiray:
The subscription struct has two dynamically sized fields, which makes the usage of `abi.encodePacked` inappropriate. It was replaced by `abi.encode` as a solution.

An alternative to this is to always use `abi.encode` as done in the pre-alpha contracts. Since we never store the encoded bytes, we don't need to care about its length.

### DIFF
--- a/packages/airnode-protocol-v1/contracts/dapis/DapiServer.sol
+++ b/packages/airnode-protocol-v1/contracts/dapis/DapiServer.sol
@@ -267,7 +267,7 @@ contract DapiServer is
         require(relayer != address(0), "Relayer address zero");
         require(sponsor != address(0), "Sponsor address zero");
         subscriptionId = keccak256(
-            abi.encodePacked(
+            abi.encode(
                 block.chainid,
                 airnode,
                 templateId,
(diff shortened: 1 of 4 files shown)


Ashar2shahid (approved):
More context on why it is inappropriate to use `abi.encodePacked()`.
> If you use keccak256(abi.encodePacked(a, b)) and both a and b are dynamic types, it is easy to craft collisions in the hash value by moving parts of a into b and vice-versa. More specifically, abi.encodePacked("a", "bc") == abi.encodePacked("ab", "c"). If you use abi.encodePacked for signatures, authentication or data integrity, make sure to always use the same types and check that at most one of them is dynamic. Unless there is a compelling reason, abi.encode should be preferred.

Apart from the increased gas costs, I don't see any downsides. I don't suppose there is any use case where instead of `encode()` `encodeWithSignature()` would be better, is there?

bbenligiray:
> I don't suppose there is any use case where instead of encode() encodeWithSignature() would be better, is there?

No

bbenligiray:
> Apart from the increased gas costs

From Table G at https://ethereum.github.io/yellowpaper/paper.pdf

> Gkeccak256 30 Paid for each KECCAK256 operation.
Gkeccak256word 6 Paid for each word (rounded up) for input data to a KECCAK256 operation.

PR #907, opened 2022-03-04
AN 576 | Template ID verification for the signed data endpoint

Ashar2shahid:
Initially I was adding the templateId as a parameter to the `processHttpSignedDataRequest` function but thought it would be good to get other opinions first. With the way its currently implemented there is very little need to change any of the methods in `index.ts` .


packages/airnode-node/src/api/index.ts
```diff
@@ -135,7 +134,7 @@ function verifyRequestId(payload: CallApiPayload): LogsData<ApiCallErrorResponse
 
 export function verifyTemplateId(payload: CallApiPayload): LogsData<ApiCallErrorResponse> | null {
   const { aggregatedApiCall } = payload;
-  if (aggregatedApiCall.type !== 'regular') return null;
+  if (aggregatedApiCall.type === 'http-gateway') return null;
```
Ashar2shahid, line 137: We should be able to use this method now for Singed data requests, so even if we pass in a templateId to the http request an invalid template would raise an error (if we decide to go that route).

packages/airnode-node/src/handlers/process-http-signed-data-request.ts
```diff
@@ -17,29 +19,43 @@ export async function processHttpSignedDataRequest(
...
-    ];
+  if (!decodedParameters) {
+    return [new Error(`Request contains invalid encodedParameters: ${parameters}`), null];
   }
 
   const requestId = randomHexString(16);
   const logOptions = buildBaseOptions(config, { requestId });
   const airnodeAddress = wallet.getAirnodeWallet(config).address;
+
+  const template = <ApiCallTemplate>{
```
andreogle, line 37: I haven't seen this syntax used before (?)

```suggestion
  const template: ApiCallTemplate = {
```

packages/airnode-node/src/handlers/process-http-signed-data-request.ts
```diff
@@ -17,29 +19,43 @@ export async function processHttpSignedDataRequest(
 
   const endpoints = find(config.ois, ['title', trigger.oisTitle])?.endpoints;
   const endpoint = find(endpoints, ['name', trigger.endpointName]);
+  const decodedParameters = evm.encoding.safeDecode(parameters);
```
andreogle, line 22: I think it would be slightly better to do this after the `if (!endpoint)` check? i.e. above the `if (!decodedParameters)` part

packages/airnode-node/src/handlers/process-http-signed-data-request.ts
```diff
@@ -17,29 +19,43 @@ export async function processHttpSignedDataRequest(
...
     airnodeAddress,
     endpointId,
     endpointName: trigger.endpointName,
     oisTitle: trigger.oisTitle,
-    parameters,
+    parameters: decodedParameters,
+    templateId: templateId,
+    template: {
+      airnodeAddress,
+      endpointId,
```
andreogle, line 55: It's weird to me to have some of these fields duplicated. Is this necessary?

andreogle (commented):
I don't feel like I have all of the context for this PR so just added some minor suggestions/comments

packages/airnode-node/src/api/index.ts
```diff
@@ -135,7 +134,7 @@ function verifyRequestId(payload: CallApiPayload): LogsData<ApiCallErrorResponse
 
 export function verifyTemplateId(payload: CallApiPayload): LogsData<ApiCallErrorResponse> | null {
```
amarthadan, line 135: So, this check will basically never fail for the HTTP signed data request, right? But I'm ok with doing the check anyway, I just want to make sure I understand.

amarthadan (changes requested):
I like this approach :+1:

So, just to be clear, the difference between HTTP gateway parameters and HTTP signed data gateway parameters is that the first are representing the values for the parameters but the second are describing the parameters themselves, hence the encoding, is that correct?
In that case, I would suggest renaming the field for HTTP signed data gateway to `encodedParameters`, so the difference is obvious. I think it would be confusing if one example curl command would contain something like `'{"parameters": {"param1": "string", "param2": 5}}'` but the other one would be `{"parameters": "0x3173737300....000"}`.

packages/airnode-node/src/handlers/process-http-signed-data-request.ts
amarthadan: I guess it depends on where do you want to construct the apiCall object. You can do it inside `callApi` function but I like this better even with some duplicity.

packages/airnode-node/src/api/index.ts
```diff
@@ -23,10 +23,9 @@ function buildOptions(payload: CallApiPayload): adapter.BuildRequestOptions {
 
   switch (aggregatedApiCall.type) {
```
amarthadan, line 24: The `http-signed-data-gateway` section and `http-gateway` section look the same, can you merge them?

amarthadan (changes requested):

packages/airnode-node/src/handlers/process-http-signed-data-request.ts
Ashar2shahid: so the `getExpectedTemplateId()`  expects a `ApiCallTemplate ` type . The problem being this type needs an `id` which we are trying to generate. I wasn't sure if I should be creating a new type so I just type casted the existing ApiCallTemplate without the `id` . this is functionally identical to:

```
  const template = {
    airnodeAddress,
    endpointId,
    encodedParameters: parameters,
  } as ApiCallTemplate;
  ```
  
  Maybe I should create a new type without id ?

File: packages/airnode-node/src/handlers/process-http-signed-data-request.ts (hunk 4 above)
Comment (Ashar2shahid) [reply packages/airnode-node/src/handlers/process-http-signed-data-request.ts:55]: I think this duplication happens in the regular `apiCalls` as well so it should be fine to do this since these fields are used in the `callApi` methods (verifyTemplate)

File: packages/airnode-node/src/types.ts [hunk 7]
```diff
@@ -280,6 +280,8 @@ export interface HttpGatewayAggregatedApiCall extends BaseAggregatedApiCall {
 
 export interface HttpSignedDataAggregatedApiCall extends BaseAggregatedApiCall {
   type: 'http-signed-data-gateway';
+  templateId: string;
+  template?: ApiCallTemplate;
```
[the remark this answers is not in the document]
Ashar2shahid: right

packages/airnode-node/src/api/index.ts
Ashar2shahid: yes, with the way it is currently it will always pass. If we decide to pass in templateId at some point then the check will be useful.

Ashar2shahid:
> In that case, I would suggest renaming the field for HTTP signed data gateway to encodedParameters, so the difference is obvious. I think it would be confusing if one example curl command would contain something like '{"parameters": {"param1": "string", "param2": 5}}' but the other one would be {"parameters": "0x3173737300....000"}.

Yeah that makes sense to me as well, although onchain the encodedParameters are referred to as parameters so maybe its fine?

amarthadan:
> > In that case, I would suggest renaming the field for HTTP signed data gateway to encodedParameters, so the difference is obvious. I think it would be confusing if one example curl command would contain something like '{"parameters": {"param1": "string", "param2": 5}}' but the other one would be {"parameters": "0x3173737300....000"}.
> 
> Yeah that makes sense to me as well, although onchain the encodedParameters are referred to as parameters so maybe its fine?

I'd still say we should rename it to convey the distinction.

packages/airnode-node/src/handlers/process-http-signed-data-request.ts
andreogle: I'm not sure I like this approach... So this would loses any type safety we get from the type? As a general rule, I think if a function expects a certain type, then we shouldn't try to force our way around that. Would changing the input type on `getExpectedTemplateId` be a lot of work?

@amarthadan @Siegrift what do you guys think?

amarthadan: I agree with @andreogle. If you would use
```ts
type ApiCallTemplateForValidation = Omit<ApiCallTemplate, 'id'>
```
as an argument type for `getTemplateIdValidationFields` and `getExpectedTemplateId` it should work without any other changes.

Feel free to come up with a better name for that type :sweat_smile:

amarthadan (approved):
LGTM :+1:

packages/airnode-deployer/src/handlers/gcp/index.ts
```diff
@@ -114,15 +114,19 @@ export async function processHttpSignedDataRequest(req: Request, res: Response)
     res.status(401).send({ error: 'Wrong API key' });
   }
 
-  const { parameters } = req.body;
+  const { encodedParameters } = req.body;
```
andreogle, line 117: Nit: it's kinda weird that here we assign by destructuring but above `aws/index.ts` it's done with regular assignment like `const encodedParameters = JSON.parse(event.body).encodedParameters;`.

I kinda think we should just make it consistent while we're modifying these files

andreogle (approved):
Nice one 👍🏻

amarthadan:
@bbenligiray You can review this, it should be the final implementation of the sign data gateway.

Ashar2shahid:
@amarthadan I think we can go ahead and merge this and burak can take a look at the documentation once that's been updated

PR #908, opened 2022-03-06
Fix coingecko-signed-data integration command typo

dcroote:
Removes the `\`, which was causing the curl command to return an internal server error.


vponline (approved):
Nice :+1:

PR #924, opened 2022-03-07
Allow Airnode to opt back in

bbenligiray:
My initial plan for behavior was to only allow
```
(Any status) => OptedOut
```
or
```
OptedOut => Inactive
```
if the caller was `airnode` (but apparently I forgot to implement the second part). This felt a bit like unnecessarily complex, so I decided to allow the following instead
```
(Any status) => Inactive
```
or
```
(Any status) => OptedOut
```
This means that differently from the first version, `airnode` can set change its status as `Active => Inactive`, which is not really an issue (because `airnode` could already do that by opting out and opting back in). This made the implementation a lot simpler.


bbenligiray:
I wanted to prevent `airnode` from being able to flip between `Active` and `Inactive`/`OptedOut` because opting out is serious business

PR #926, opened 2022-03-08
Implement optional lead time requirement for withdrawals

bbenligiray:
Implemented a lead time requirement to prevent front-running the block transaction, but kept it disabled by default assuming we will never need it. The lead time is capped at 30 days to prevent its abuse.

### DIFF
--- a/packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
+++ b/packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
@@ -219,12 +219,12 @@ contract RequesterAuthorizerWhitelisterWithTokenDeposit is
                 earliestWithdrawalTime != 0,
                 "Withdrawal intent not signaled"
             );
-        }
-        if (earliestWithdrawalTime != 0) {
             require(
                 earliestWithdrawalTime <= block.timestamp,
                 "Not withdrawal time yet"
             );
+        }
+        if (earliestWithdrawalTime != 0) {
             uint256 tokenDepositsCount = tokenDeposits.count;
             emit WithdrewTokens(
                 airnode,
--- a/packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
+++ b/packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
@@ -288,24 +288,38 @@ contract RequesterAuthorizerWhitelisterWithTokenDeposit is
         ];
         require(tokenWithdrawAmount != 0, "Depositor has not deposited");
         tokenDeposits.depositorToAmount[depositor] = 0;
-        uint256 tokenDepositsCount = --tokenDeposits.count;
-        emit WithdrewTokensDepositedForBlockedRequester(
-            airnode,
-            chainId,
-            endpointId,
-            requester,
-            depositor,
-            tokenDepositsCount,
-            tokenWithdrawAmount
-        );
-        if (tokenDepositsCount == 0) {
-            IRequesterAuthorizer(getRequesterAuthorizerAddress(chainId))
-                .setIndefiniteWhitelistStatus(
-                    airnode,
-                    endpointId,
-                    requester,
-                    false
-                );
+        if (tokenDeposits.depositorToEarliestWithdrawalTime[depositor] == 0) {
+            uint256 tokenDepositsCount = --tokenDeposits.count;
+            emit WithdrewTokensDepositedForBlockedRequester(
+                airnode,
+                chainId,
+                endpointId,
+                requester,
+                depositor,
+                tokenDepositsCount,
+                tokenWithdrawAmount
+            );
+            if (tokenDepositsCount == 0) {
+                IRequesterAuthorizer(getRequesterAuthorizerAddress(chainId))
+                    .setIndefiniteWhitelistStatus(
+                        airnode,
+                        endpointId,
+                        requester,
+                        false
+                    );
+            }
+        } else {
+            uint256 tokenDepositsCount = tokenDeposits.count;
+            tokenDeposits.depositorToEarliestWithdrawalTime[depositor] = 0;
+            emit WithdrewTokensDepositedForBlockedRequester(
+                airnode,
+                chainId,
+                endpointId,
+                requester,
+                depositor,
+                tokenDepositsCount,
+                tokenWithdrawAmount
+            );
         }
         assert(
             IERC20(token).transfer(proceedsDestination, tokenWithdrawAmount)
--- a/packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
+++ b/packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
@@ -13,9 +13,16 @@ contract RequesterAuthorizerWhitelisterWithTokenDeposit is
     IRequesterAuthorizerWhitelisterWithTokenDeposit
 {
     struct TokenDeposits {
-        mapping(address => uint256) depositorToAmount;
         uint256 count;
+        mapping(address => uint256) depositorToAmount;
+        mapping(address => uint256) depositorToEarliestWithdrawalTime;
     }
+
+    /// @notice Time the token depositors have to wait after signaling
+    /// withdrawal intent to be able to withdraw
+    /// @dev The depositors can withdraw tokens without signaling intent if the
+    /// withdrawal lead time is zero
+    uint256 public override withdrawalLeadTime = 0;
 
     /// @notice Token deposits made for Airnode, chain, endpoint, requester
     mapping(address => mapping(uint256 => mapping(bytes32 => mapping(address => TokenDeposits))))
@@ -56,6 +63,21 @@ contract RequesterAuthorizerWhitelisterWithTokenDeposit is
             _proceedsDestination
         )
     {}
+
+    /// @notice Called by the manager to set the withdrawal lead time
+    /// @param _withdrawalLeadTime Withdrawal lead time
+    function setWithdrawalLeadTime(uint256 _withdrawalLeadTime)
+        external
+        override
+        onlyMaintainerOrManager
+    {
+        require(
+            _withdrawalLeadTime <= 30 days,
+            "Withdrawal lead time too long"
+        );
+        withdrawalLeadTime = _withdrawalLeadTime;
+        emit SetWithdrawalLeadTime(_withdrawalLeadTime, msg.sender);
+    }
 
     /// @notice Deposits tokens for the requester to be whitelisted for the
     /// Airnode–endpoint pair on the chain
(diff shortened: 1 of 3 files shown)


packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
```diff
@@ -158,6 +189,73 @@ contract RequesterAuthorizerWhitelisterWithTokenDeposit is
...
+                earliestWithdrawalTime != 0,
+                "Withdrawal intent not signaled"
+            );
+        }
+        if (earliestWithdrawalTime != 0) {
+            require(
+                earliestWithdrawalTime <= block.timestamp,
+                "Not withdrawal time yet"
+            );
+            uint256 tokenDepositsCount = tokenDeposits.count;
```
Ashar2shahid, line 228: I don't think we should be setting `tokenDeposits.count` in `signalWithdrawalIntent()` instead it should be done here
 ```
uint256 tokenDepositsCount = --tokenDeposits.count;
```

File: packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol [hunk 2]
```diff
@@ -158,6 +189,73 @@ contract RequesterAuthorizerWhitelisterWithTokenDeposit is
...
+                airnode
+            ][chainId][endpointId][requester];
+        uint256 tokenWithdrawAmount = tokenDeposits.depositorToAmount[
+            msg.sender
+        ];
+        require(tokenWithdrawAmount != 0, "Sender has not deposited tokens");
+        tokenDeposits.depositorToAmount[msg.sender] = 0;
+        uint256 earliestWithdrawalTime = tokenDeposits
+            .depositorToEarliestWithdrawalTime[msg.sender];
+        if (withdrawalLeadTime != 0) {
```
Ashar2shahid, line 217: only ` if (withdrawalLeadTime != 0)` is probably sufficient here along with the `require()` I think we don't need the if below. The else could lead this if

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
```diff
@@ -158,6 +189,73 @@ contract RequesterAuthorizerWhitelisterWithTokenDeposit is
...
+            uint256 tokenDepositsCount = tokenDeposits.count;
+            emit WithdrewTokens(
+                airnode,
+                chainId,
+                endpointId,
+                requester,
+                msg.sender,
+                tokenDepositsCount,
+                tokenWithdrawAmount
+            );
```
Ashar2shahid, line 237: again we probably want to `setIndefiniteWhitelistStatus()` here as well.

Ashar2shahid (changes requested):
`withdrawTokens` code could be much cleaner if we move over the `setIndefiniteWhitelistStatus` to only withdrawToken. I don't really see any negatives to doing it except the requester would have access to the airnode until he withdraws which is fine imo.

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
```diff
@@ -59,6 +66,21 @@ contract RequesterAuthorizerWhitelisterWithTokenDeposit is
...
 
+    /// @notice Called by the manager to set the withdrawal lead time
+    /// @param _withdrawalLeadTime Withdrawal lead time
+    function setWithdrawalLeadTime(uint256 _withdrawalLeadTime)
+        external
+        override
+        onlyMaintainerOrManager
+    {
+        require(
+            _withdrawalLeadTime <= 30 days,
```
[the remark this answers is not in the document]
bbenligiray: This is only there to prevent `withdrawalLeadTime` from being set to 10 years and someone accidentally signaling withdrawal intent, locking the tokens practically forever. We will probably set it at something like 1 day if we ever use it.

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
bbenligiray: This doesn't work because then the users that expect to be blocked will signal withdrawal intent immediately after depositing, and would be able to frontrun the blocking transaction.
The DAO staking pool takes away your voting power and stops you from receiving staking rewards as soon as you schedule an unstake (and not at the time that you actually unstake) for the same reason.

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
bbenligiray: https://github.com/api3dao/airnode/pull/926/files/eee4256371ee411278e8f83c4bebbcdf7bd403cd#r824507853

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
bbenligiray: I don't understand this. Would that implement the same logic?

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
Ashar2shahid: I don't really see how that's a problem. Even if the user signals withdrawIntent() he won't be able to withdraw because of the onlyNonBlockedRequester modifier. We would have to withdraw the blocked tokens to revoke the whitelisting but I think that's fine.

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
Ashar2shahid: in a way, take this for example:
```
if (withdrawalLeadTime != 0) {
    require(
            earliestWithdrawalTime <= block.timestamp,
            "Intent not signaled or not withdrawal time yet"
        );
        uint256 tokenDepositsCount = tokenDeposits.count;
        emit WithdrewTokens(
            airnode,
            chainId,
            endpointId,
            requester,
            msg.sender,
            tokenDepositsCount,
            tokenWithdrawAmount
        );
} else 
```
Essentially only check `earliestWithdrawalTime` if there is a lead time otherwise don't

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
Ashar2shahid: Also there is another problem with this approach, say for example a user is blocked and he signals `withdrawIntent()` so `--tokenDeposits.count` happens once. If we call `withdrawFundsDepositedForBlockedRequester()` it will trigger another `--tokenDeposits.count` for the same user.

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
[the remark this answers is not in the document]
Ashar2shahid: referring to this

> Implemented a lead time requirement to prevent front-running the block transaction, but kept it disabled by default assuming we will never need it. The lead time is capped at **1 week** to prevent its abuse.

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
bbenligiray: > I don't really see how that's a problem. Even if the user signals withdrawIntent() he won't be able to withdraw because of the onlyNonBlockedRequester modifier in withdrawTokens().

The only reason for this PR is that the call to withdrawTokens() can frontrun the blocking call. Read the audit report again.

> Also there is another problem with this approach...

Good point, I'll fix that

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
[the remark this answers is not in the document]
bbenligiray: Right, I don't know what I was thinking

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
bbenligiray: Fixed in https://github.com/api3dao/airnode/pull/926/commits/91938b143cf14a0d8dcbd2b1d58101f26a51ce80

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
bbenligiray: There, I'm also checking `earliestWithdrawalTime` to see if I need to decrement the deposit count or not. So if there was a `withdrawalLeadTime` and someone signaled withdrawal intent but then `withdrawalLeadTime` was set to zero and the user wants to withdraw, we shouldn't decrement the deposit count twice.

bbenligiray: I think I addressed this in https://github.com/api3dao/airnode/pull/926/commits/cde8b1b2d23fc547cce2197054a5bb1c03a34caa

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
Ashar2shahid: reread your initial comment and I think I understand what you mean. I agree that won't work.

packages/airnode-protocol-v1/test/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol.js
```diff
@@ -1241,8 +1302,10 @@ describe('withdrawTokens', function () {
           expect(whitelistStatus.indefiniteWhitelistCount).to.equal(0);
         });
       });
-      context('Withdrawn deposit was not the last one for the requester-endpoint pair', function () {
-        it('decrements the number of times tokens were deposited for the requester-endpoint pair and withdraws tokens', async function () {
+      context('Not signaling intent to withdraw the last deposit for the requester-endpoint pair', function () {
```
Ashar2shahid, line 1305: kind of ambiguous what this means maybe something like:
`signaling intent to withdraw but not for the last deposit for the requester-endpoint pair`
is better

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
Ashar2shahid: > Fixed in [91938b1](https://github.com/api3dao/airnode/commit/91938b143cf14a0d8dcbd2b1d58101f26a51ce80)

we should probably update the tests to cover both the cases.

Ashar2shahid (approved):
Just a few comments regarding the tests.

packages/airnode-protocol-v1/contracts/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol
bbenligiray: Yeah, but I think we can do that later on. This contract will probably not be used in this form anyway and I have a lot of time-critical work.

packages/airnode-protocol-v1/test/monetization/RequesterAuthorizerWhitelisterWithTokenDeposit.sol.js
bbenligiray: Fixed in https://github.com/api3dao/airnode/pull/926/commits/8c97ab69b9784d88131ff5e9444d70db975aa910

PR #927, opened 2022-03-08
Use the SafeERC20 library

bbenligiray:
Using SafeERC20 in case the token in question doesn't conform to ERC20 exactly


Ashar2shahid:
@bbenligiray any idea why the unit tests keep failing ?

bbenligiray:
I think it's just bad luck, I'm rerunning again. This should be resolved with https://github.com/api3dao/airnode/pull/894 but I didn't apply it on this branch to limit the diff since the audit.

Ashar2shahid (approved):

PR #928, opened 2022-03-08
Update dependencies

bbenligiray:
After running `yarn bootstrap`, `yarn audit` in `packages/airnode-protocol-v1` should only display 4 moderate vulnerabilities.

This broke TS eslint so I disabled it for a line as a quick fix. Master dependencies are up to date and we won't merge this branch to master anyway, so this is fine.


Ashar2shahid (approved):
I didn't find any breaking changes between the versions per se but having someone from the airnode team chime in on this would be good.

bbenligiray:
> I didn't find any breaking changes between the versions per se but having someone from the airnode team chime in on this would be good.

I actually copied over these dependency versions from the master branch so it should be fine

Siegrift:
Btw. take the output of yarn audit with a grain of salt. It is a bit broken https://overreacted.io/npm-audit-broken-by-design/

bbenligiray:
Gotcha, I used it because I assume that's what the auditors used

PR #929, opened 2022-03-08
Admin EIP1559 transactions

vponline:
This updates `airnode-admin` to use the **gas-prices** implementation from `airnode-node` for improved transaction fee settings if the user does not specify any.


packages/airnode-admin/src/implementation.ts
```diff
@@ -35,6 +36,27 @@ export const parseTransactionOverrides = (args: Arguments): ethers.Overrides =>
...
+  if (gasPrice && !maxFeePerGas && !maxPriorityFeePerGas && feeData.maxFeePerGas) {
+    return {
+      maxFeePerGas: gasPrice,
+      maxPriorityFeePerGas: ethers.BigNumber.from(PRIORITY_FEE),
+      ...rest,
+    };
+  }
+
+  return overrides;
+};
```
vponline, line 58: Is this approach what was intended? Or would you prefer that we import the gas-prices implementation from `airnode-node` to set `maxFeePerGas`?

aquarat: The `if` statement scares me a bit and may be unnecessary:
The user shouldn't be specifying both eip1559 and legacy pricing override options, so I think we should error if both are specified? This is how ethers does it internally (see below).

eg.
```
if (gasPrice && (maxFeePerGas || priorityFeePerGas)) { // brackets may not be necessary
throw 'Both legacy and EIP1559 override pricing options specified - ambiguous';
}

// if the user is specifying eip1559 pricing options and the network doesn't support it,
// they shouldn't be specifying those options
if (!feeData.maxFeePerGas && (maxFeePerGas || priorityFeePerGas)) {
throw 'EIP1559 override pricing specified on legacy network'
}

/* ethers defaults to a fee of maxPriorityFeePerGas = BigNumber.from("2500000000");, 
which I think is fine for the admin utilities. The reason we went with a larger default in 
Airnode was to get in front of the herd for tx reliability, but here that's less important because
there's a human operator.
*/
return {
...feeData,
...overrides,
};
```

Ethers pricing implementation around overrides:
```
        // Do not allow mixing pre-eip-1559 and eip-1559 properties
        const hasEip1559 = (tx.maxFeePerGas != null || tx.maxPriorityFeePerGas != null);
        if (tx.gasPrice != null && (tx.type === 2 || hasEip1559)) {
            logger.throwArgumentError("eip-1559 transaction do not support gasPrice", "transaction", transaction);
        } else if ((tx.type === 0 || tx.type === 1) && hasEip1559) {
            logger.throwArgumentError("pre-eip-1559 transaction do not support maxFeePerGas/maxPriorityFeePerGas", "transaction", transaction);
        }
```

and the ethers getFeeData implementation:

```
        const { block, gasPrice } = await resolveProperties({
            block: this.getBlock("latest"),
            gasPrice: this.getGasPrice().catch((error) => {
                // @TODO: Why is this now failing on Calaveras?
                //console.log(error);
                return null;
            })
        });

        let maxFeePerGas = null, maxPriorityFeePerGas = null;

        if (block && block.baseFeePerGas) {
            // We may want to compute this more accurately in the future,
            // using the formula "check if the base fee is correct".
            // See: https://eips.ethereum.org/EIPS/eip-1559
            maxPriorityFeePerGas = BigNumber.from("2500000000");
            maxFeePerGas = block.baseFeePerGas.mul(2).add(maxPriorityFeePerGas);
        }

        return { maxFeePerGas, maxPriorityFeePerGas, gasPrice };
```

packages/airnode-admin/src/implementation.ts
```diff
@@ -83,7 +105,11 @@ export async function deriveSponsorWalletAddress(airnodeXpub: string, airnodeAdd
 }
 
 export async function sponsorRequester(airnodeRrp: AirnodeRrp, requesterAddress: string, overrides?: ethers.Overrides) {
-  await airnodeRrp.setSponsorshipStatus(requesterAddress, true, overrides ?? {});
+  await airnodeRrp.setSponsorshipStatus(
+    requesterAddress,
+    true,
+    parseOverrides(await airnodeRrp.provider.getFeeData(), overrides)
```
aquarat, line 111: I feel like you could reduce a lot of duplication by supplying the overrides up-front to these functions?
So wherever these functions are called, the caller should be responsible for calling `parseOverrides`?
For example [`const overrides = admin.parseCliOverrides(args);`](https://github.com/api3dao/airnode/blob/230f9112bb8476df0a1d8ab6a2f05421320fc9f6/packages/airnode-admin/src/cli.ts#L207) could be a good place to do this and then it's a small diff.

aquarat (commented):
I've left some comments to discuss :smile:

packages/airnode-admin/src/implementation.ts
vponline: Thanks Aaron, great points to discuss. I probably should have added some context initially to how I wrote it :sweat_smile: : 

> The user shouldn't be specifying both eip1559 and legacy pricing override options, so I think we should error if both are specified? This is how ethers does it internally (see below).

I intentionally allowed the mixed overrides to pass through `parseOverrides` because this approach was also used in [834](https://github.com/api3dao/airnode/pull/834#discussion_r809034853). (Of course, since the `parsedCliOverrides` will now also pass through `parseOverrides` here so I'm happy to add our own checks here.)

> ethers defaults to a fee of maxPriorityFeePerGas = BigNumber.from("2500000000");

Sorry, my bad I should have linked the task [AN-454](https://api3dao.atlassian.net/jira/software/projects/AN/boards/18?selectedIssue=AN-454). 3.1 was specified in the description so I think the same idea for the higher priority fee was intended here.

> I feel like you could reduce a lot of duplication by supplying the overrides up-front to these functions?
So wherever these functions are called, the caller should be responsible for calling parseOverrides?
For example [const overrides = admin.parseCliOverrides(args);](https://github.com/api3dao/airnode/blob/230f9112bb8476df0a1d8ab6a2f05421320fc9f6/packages/airnode-admin/src/cli.ts#L207) could be a good place to do this and then it's a small diff.

Hmm the reason I put it inside the `implementation.ts` methods is so that it works the same way for the CLI and SDK. Otherwise we'd need to expose parseOverrides as an SDK function to the user? 

Happy to make the changes so please let me know what you think.

aquarat: Hey

Thank you for the context and detailed responses.

I see @ "3.1". I don't think we're tied to that value.

My main concern is the if statement, which is complex and difficult to read. I'm mostly concerned with how we can improve the readability of that statement :relaxed:  

That makes sense @ `parseOverrides`. No problem.

packages/airnode-admin/src/implementation.ts
```diff
@@ -35,6 +36,62 @@ export const parseTransactionOverrides = (args: Arguments): ethers.Overrides =>
...
+ */
+export const parseOverrides = async (
+  provider: ethers.providers.Provider,
+  overrides: ethers.Overrides = {}
+): Promise<ethers.Overrides> => {
+  const { gasPrice, maxFeePerGas, maxPriorityFeePerGas } = overrides;
+
+  const feeData = await provider.getFeeData();
+  const supportsEip1559 = !!(feeData.maxFeePerGas && feeData.maxPriorityFeePerGas);
+  const hasEip1559Overrides = !!(maxFeePerGas || maxPriorityFeePerGas);
```
vponline, line 53: Note that with this check using `maxFeePerGas || maxPriorityFeePerGas` instead of `maxFeePerGas && maxPriorityFeePerGas` will not cause `parseOverrides` to throw, but ethers will end up throwing because `maxFeePerGas` has not been defined?

I used `||` so we can use the same check in the if-blocks below.

packages/airnode-admin/package.json
```diff
@@ -25,6 +25,7 @@
   },
   "dependencies": {
     "@api3/airnode-abi": "^0.4.0",
+    "@api3/airnode-node": "^0.4.0",
```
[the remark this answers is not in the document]
aquarat: Great idea :)  I've asked on Slack to make sure no one disagrees,

aquarat (approved):
I think this is a big improvement.
I like the way you've split up the decisions in `parseOverrides` and commented them.
I don't think there's any _beautiful_ way of dealing with the overrides due to the way `ethers` works, but this is the closest we can get.
As you noted on Slack, the new approach will cause a few extra RPC calls, but I don't believe that's a problem as this package is used infrequently (as opposed to Airnode which runs every minute indefinitely - so efficiency is a bigger concern).

Well done :trophy: :smile:

packages/airnode-admin/src/implementation.ts
aquarat: If the user is specifying overrides then they should understand the effect of those overrides :shrug: :laughing:

PR #932, opened 2022-03-10
Miscellaneous ToB audit fixes

bbenligiray:
> To reduce gas costs, make variables immutable wherever possible

Addressed in https://github.com/api3dao/airnode/commit/53f2a33594f53bad819a10bde01d4a20ae107564

> To improve readability make contracts abstract if they should not be deployed

If a contract doesn't implement the whole interface the compiler errors. Using `abstract` as described above disables this check, so `abstract` shouldn't be used unless the contract actually doesn't implement the whole interface. As a compromise, the title strings of such contracts are edited to all start with `"Contract to be inherited..."` (some contracts that are not planned to deployed in standalone fashion but potentially could have been are omitted) https://github.com/api3dao/airnode/commit/a0df928212111d7f49ffa443d736f32819dd3435

> Improve the NatSpec comments

Addressed in https://github.com/api3dao/airnode/commit/69174d9ff0f1832a4e6c779f98c3019cd52b9805

> Remove the onlyRegistrarOrManager modifier from _registerAddress

It didn't feel right for `_registerAddress()` to not check if the sender is allowed to register an address. Therefore, I removed the `onlyRegistrarOrManager` from the exposed function instead https://github.com/api3dao/airnode/commit/3ed253e99c6f5a211f8a15c3ee6ea28ff9c55dab

> Replace the deprecated _setupRole with _grantRole

The audited contracts use OpenZeppelin contracts 4.3.2 where _setupRole isn't deprecated yet and _grantRole is private so this suggestion doesn't work. Updated the OpenZeppelin contracts to 4.4.2 anyway and implemented this https://github.com/api3dao/airnode/commit/c954ad6883989f7982ca2f3ba21b9f5968e5e919

> Remove the require(...) check

The redundant check was added on purpose so that Sort.sol doesn't get reused overseeing this requirement, so leaving this in.

### DIFF
--- a/packages/airnode-protocol-v1/contracts/access-control-registry/AccessControlRegistry.sol
+++ b/packages/airnode-protocol-v1/contracts/access-control-registry/AccessControlRegistry.sol
@@ -32,7 +32,7 @@ contract AccessControlRegistry is
         require(manager != address(0), "Manager address zero");
         bytes32 rootRole = deriveRootRole(manager);
         if (!hasRole(rootRole, manager)) {
-            _setupRole(rootRole, manager);
+            _grantRole(rootRole, manager);
             emit InitializedManager(rootRole, manager);
         }
     }
--- a/packages/airnode-protocol-v1/contracts/access-control-registry/AccessControlRegistryAdminned.sol
+++ b/packages/airnode-protocol-v1/contracts/access-control-registry/AccessControlRegistryAdminned.sol
@@ -5,8 +5,8 @@ import "./RoleDeriver.sol";
 import "./AccessControlRegistryUser.sol";
 import "./interfaces/IAccessControlRegistryAdminned.sol";
 
-/// @title Contract that should be inherited by contracts whose adminship
-/// functionality will be implemented using AccessControlRegistry
+/// @title Contract to be inherited by contracts whose adminship functionality
+/// will be implemented using AccessControlRegistry
 contract AccessControlRegistryAdminned is
     RoleDeriver,
     AccessControlRegistryUser,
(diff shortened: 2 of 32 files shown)


Ashar2shahid (commented):
> If a contract doesn't implement the whole interface the compiler errors. Using abstract as described above disables this check, so abstract shouldn't be used unless the contract actually doesn't implement the whole interface. As a compromise, the title strings of such contracts are edited to all start with "Contract to be inherited..." (some contracts that are not planned to deployed in standalone fashion but potentially could have been are omitted) https://github.com/api3dao/airnode/commit/a0df928212111d7f49ffa443d736f32819dd3435            

I think the ambiguity is still sort of there even with this change, would having an additional `dev` in the NatSpec comment stating that `This contract will be inherited from and not deployed` be a good change?

 From the TOB Audit:
 > Remove the onlyRegistrarOrManager modifier from registerAddress . The
registerAddress function is **only called from
registerChainRequesterAuthorizer** which already has the
onlyRegistrarOrManager .               

But from the [commit](https://github.com/api3dao/airnode/commit/3ed253e99c6f5a211f8a15c3ee6ea28ff9c55dab) I see we made changes in `AirnodeEndpointPriceRegistry.sol` as well. I'm guessing TOB missed that?

bbenligiray:
> would having an additional dev in the NatSpec comment stating that This contract will be inherited from and not deployed be a good change

That was the first thing I tried but the `@dev` comments are already so crowded that it didn't have any practical use

> But from the [commit](https://github.com/api3dao/airnode/commit/3ed253e99c6f5a211f8a15c3ee6ea28ff9c55dab) I see we made changes in AirnodeEndpointPriceRegistry.sol as well. I'm guessing TOB missed that?

Yeah

Ashar2shahid (approved):

PR #933, opened 2022-03-11
Move gas-prices to airnode-utilities

vponline:
This moves the gas-prices implementation to `airnode-utilities` and updates imports in affected places.


packages/airnode-utilities/src/gas-prices/gas-prices.ts
```diff
@@ -1,15 +1,21 @@
...
-import { go, logger, PendingLog } from '@api3/airnode-utilities';
-import { BASE_FEE_MULTIPLIER, DEFAULT_RETRY_TIMEOUT_MS, PRIORITY_FEE } from '../constants';
-import { GasTarget, LogsData } from '../types';
-import { ChainOptions, PriorityFee } from '../config/types';
+import { ChainOptions, PriorityFee, GasTarget } from './types';
+import { go, logger, PendingLog } from '../';
+import { LogsData } from '../logging';
+
+// The default amount of time before a "retryable" promise is timed out and retried
+export const DEFAULT_RETRY_TIMEOUT_MS = 5_000;
```
aquarat, line 7: We might want to move these constants into a `constants.ts` like we've done in other packages, eg. `constants.ts` in [airnode-node](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/constants.ts)

packages/airnode-utilities/src/gas-prices/gas-prices.ts
```diff
@@ -1,15 +1,21 @@
...
+// The Base Fee to Max Fee multiplier
+export const BASE_FEE_MULTIPLIER = 2;
 
 export interface FetchOptions {
   readonly provider: ethers.providers.JsonRpcProvider | ethers.providers.Provider;
   readonly chainOptions: ChainOptions;
 }
 
-export const parsePriorityFee = ({ value, unit }: PriorityFee) =>
+export const parsePriorityFee = ({ value, unit }: PriorityFee): BigNumber =>
```
aquarat, line 18: It appears the return type isn't required.

packages/airnode-utilities/src/gas-prices/gas-prices.ts
```diff
@@ -1,15 +1,21 @@
...
   ethers.utils.parseUnits(value.toString(), unit ?? 'wei');
 
 export const getLegacyGasPrice = async (options: FetchOptions): Promise<LogsData<GasTarget | null>> => {
```
aquarat, line 21: We should probably make `DEFAULT_RETRY_TIMEOUT_MS` configurable with an optional argument. An optional argument won't break existing usages. WDYT? Everything else is exposed (the default constants for gas pricing are just defaults/they're configurable through function arguments).

aquarat (changes requested):
Looking good, just some minor comments.

packages/airnode-utilities/src/gas-prices/gas-prices.ts
vponline: You're right, I had a VScode warning about this but it's gone now... must have been a bug with me adding the ethers dependency :smile:

packages/airnode-utilities/src/gas-prices/gas-prices.ts
vponline: Great idea, would it make sense to add the whole `go` [PromiseOptions](https://github.com/api3dao/airnode/blob/master/packages/airnode-utilities/src/promises/promise.ts#L12) as a field to `FetchOptions`
```js
export interface FetchOptions {
  readonly provider: ethers.providers.JsonRpcProvider | ethers.providers.Provider;
  readonly chainOptions: ChainOptions;
  readonly promiseOptions?: PromiseOptions;
}
```

and then for example:
```js
export const getLegacyGasPrice = async (options: FetchOptions): Promise<LogsData<GasTarget | null>> => {
  const { provider, promiseOptions } = options;

  const [err, gasPrice] = await go(() => provider.getGasPrice(), {
    retries: 1,
    timeoutMs: DEFAULT_RETRY_TIMEOUT_MS,
    ...promiseOptions,
  });
  if (err || !gasPrice) {
    const log = logger.pend('ERROR', 'All attempts to get legacy gasPrice from provider failed');
    return [[log], null];
  }

  return [[], { gasPrice }];
};
```

vponline: `FetchOptions` should probably be moved to `./types` as well :smile:

.changeset/fifty-months-laugh.md
```diff
@@ -0,0 +1,7 @@
+---
+'@api3/airnode-admin': patch
+'@api3/airnode-node': patch
+'@api3/airnode-utilities': patch
```
andreogle, line 4: I think this should be minor instead of patch?

packages/airnode-utilities/src/gas-prices/gas-prices.ts
```diff
@@ -1,15 +1,21 @@
 import { BigNumber, ethers } from 'ethers';
-import { go, logger, PendingLog } from '@api3/airnode-utilities';
-import { BASE_FEE_MULTIPLIER, DEFAULT_RETRY_TIMEOUT_MS, PRIORITY_FEE } from '../constants';
-import { GasTarget, LogsData } from '../types';
-import { ChainOptions, PriorityFee } from '../config/types';
+import { ChainOptions, PriorityFee, GasTarget } from './types';
+import { go, logger, PendingLog } from '../';
```
andreogle, line 3: I think we should try to avoid importing from index.ts as it creates a circular dependency - which can sometimes lead to weird bugs. Better to import from the module(s) directly

packages/airnode-utilities/src/gas-prices/gas-prices.ts
```diff
@@ -1,15 +1,21 @@
...
+import { LogsData } from '../logging';
+
+// The default amount of time before a "retryable" promise is timed out and retried
+export const DEFAULT_RETRY_TIMEOUT_MS = 5_000;
+// The Priority Fee in Wei
+export const PRIORITY_FEE = 3120000000;
```
andreogle, line 9: Maybe better to include the `_IN_WEI` in the name to help anyone implementing this?
```suggestion
export const PRIORITY_FEE_IN_WEI = 3120000000;
```

packages/airnode-utilities/src/gas-prices/gas-prices.ts
andreogle: Maybe it's worth merging `chainOptions` and `promiseOptions` into a single object?

packages/airnode-utilities/src/gas-prices/types.ts
```diff
@@ -0,0 +1,18 @@
+import { BigNumber } from 'ethers';
+
+export interface GasTarget {
+  readonly maxPriorityFeePerGas?: BigNumber;
```
andreogle, line 4: I'm not sure about all of these `readonly` modifiers in a "utilities" library. There might be use cases where the implementer needs mutation and this makes it quite restrictive. I know @Siegrift isn't a fan of `readonly` 😁

packages/airnode-utilities/src/gas-prices/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './gas-prices';
```
andreogle, line 1: I wonder if this should be in an "evm" folder? These functions/types are EVM specific

packages/airnode-utilities/src/gas-prices/gas-prices.ts
vponline: Thanks for pointing this out :+1:

packages/airnode-utilities/src/gas-prices/gas-prices.ts
vponline: @andreogle Did you mean as a field in chainOptions :detective: ?
```js
export interface ChainOptions {
  txType: 'legacy' | 'eip1559';
  baseFeeMultiplier?: number;
  priorityFee?: PriorityFee;
  promiseOptions?: PromiseOptions;
}
```
and then:
```js
const [err, gasPrice] = await go(() => provider.getGasPrice(), {
    retries: 1,
    timeoutMs: DEFAULT_RETRY_TIMEOUT_MS,
    ...chainOptions.promiseOptions,
  });
```

packages/airnode-utilities/src/gas-prices/types.ts
vponline: I removed `readonly` from the gas-prices types, but it looks like the other utility modules also have them :thinking:

packages/airnode-utilities/src/constants.ts
```diff
@@ -0,0 +1,6 @@
+// The default amount of time before a "retryable" promise is timed out and retried
+export const DEFAULT_RETRY_TIMEOUT_MS = 5_000;
+// The Priority Fee in Wei
+export const PRIORITY_FEE_IN_WEI = 3120000000;
```
Siegrift, line 4:
```suggestion
export const PRIORITY_FEE_IN_WEI = 3_120_000_000;
```

packages/airnode-utilities/src/evm/gas-prices/gas-prices.test.ts
```diff
@@ -1,8 +1,18 @@
-import { mockEthers } from '../../test/mock-utils';
+/**
+ * Mocks ethers library to return a mocked ethers provider with mocked gas prices and block data.
+ */
+function mockEthers({ ethersMocks = {} }) {
```
Siegrift, line 4: Nit: You can just inline the implementation, since this fn is used only here.

packages/airnode-utilities/src/gas-prices/types.ts
Siegrift: I dislike it because sometimes I want to write mutable code and this readonly modifier is in the way. This is especially true in tests, where you want to sometimes override a certain property. And lastly, it feels too verbose to me.

packages/airnode-utilities/src/constants.ts
```diff
@@ -0,0 +1,6 @@
+// The default amount of time before a "retryable" promise is timed out and retried
+export const DEFAULT_RETRY_TIMEOUT_MS = 5_000;
+// The Priority Fee in Wei
```
andreogle, line 3: I don't think this comment adds anything now 😄

packages/airnode-utilities/src/gas-prices/gas-prices.ts
vponline: I don't see why not :smile:

aquarat (approved):
LGTM :medal_sports:

PR #934, opened 2022-03-11
Fix gateway templates

amarthadan:
[Issue AN-603](https://api3dao.atlassian.net/browse/AN-603)


packages/airnode-deployer/terraform/airnode/aws/templates/httpSignedDataApiGateway.yaml.tpl
```diff
@@ -35,10 +35,7 @@ components:
 
     EndpointRequestExample:
       summary: Endpoint request example
-      value: { "parameters": { "from": "EUR", "amount": 5 } }
-    EndpointResponseExample:
```
andreogle, line 39: Why was the response example removed? Just curious

packages/airnode-deployer/terraform/airnode/aws/templates/httpSignedDataApiGateway.yaml.tpl
```diff
@@ -35,10 +35,7 @@ components:
...
-      summary: Endpoint response example
-      value: { "response": 20 }
+      value: { "encodedParameters": "0x31737300000000000000000000...0000000000000000000000" }
```
andreogle, line 38: It doesn't matter that this value has been truncated, right?

andreogle (approved):

packages/airnode-deployer/terraform/airnode/aws/templates/httpSignedDataApiGateway.yaml.tpl
amarthadan: Just out of convenience. We don't have these in the GCP version (they're using old swagger format so we can't), it's constantly outdated and it doesn't show anywhere in AWS.

packages/airnode-deployer/terraform/airnode/aws/templates/httpSignedDataApiGateway.yaml.tpl
amarthadan: No, it's just an example.

PR #940, opened 2022-03-14
AN-374: Creating an example that will fail

kolenic-martin:
This pull request creates a new Airnode example that is intended to fail. I only made a minor alteration to the coingecko example in all the necessary places and documented the behaviour in the readme.


.changeset/clean-plants-kick.md
```diff
@@ -0,0 +1,2 @@
+---
```
Siegrift, line 1: One changeset is enough (you can remove these files). Also feel free to add a minor changeset with:
```
Add example that will fail
```

.gitignore
```diff
@@ -15,6 +15,7 @@
 coverage/
 coverage.json
 gas_report
+gcp.json
```
Siegrift, line 18: 👍

package.json
```diff
@@ -83,7 +83,9 @@
     "test:utilities": "(cd packages/airnode-utilities && yarn run test)",
     "test:validator": "(cd packages/airnode-validator && yarn run test)"
   },
-  "dependencies": {},
+  "dependencies": {
```
Siegrift, line 86: I think this is a mistake?

Siegrift (changes requested):
You are missing the Requester contract in this PR, did you forget to commit it?

I am not sure about the changes in yarn.lock, what is your yarn version?

package.json
kolenic-martin: This is interesting as I hadn't touched the file; it might've been the prettifier. We should definitely revert that.

.changeset/clean-plants-kick.md
kolenic-martin: Will do.

kolenic-martin:
Hi @Siegrift, I don't really know why the requester contract is missing; I'll investigate and recommit.

.changeset/fast-tomatoes-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@api3/airnode-examples': minor
+---
+
+Added an Airnode example intended to fail
```
Siegrift, line 5:
```suggestion
Added an Airnode example demonstrating invalid API request
```

packages/airnode-examples/package.json
```diff
@@ -36,7 +36,8 @@
     "@api3/airnode-node": "^0.5.0",
     "@api3/airnode-protocol": "^0.5.0",
     "ethers": "^5.4.5",
-    "is-wsl": "^2.2.0"
+    "is-wsl": "^2.2.0",
+    "yarn": "^1.22.18"
```
Siegrift, line 40: This is needed  implicitely, please remove it from here.

Siegrift (commented):
Also, remove `package-lock.json` as it is not needed.

packages/airnode-examples/integrations/failing-example/request-utils.ts
```diff
@@ -0,0 +1,17 @@
...
+import { cliPrint, getDeployedContract, readIntegrationInfo } from '../../src';
+
+const coinLabel = 'Ethereum';
+const coinId = coinLabel.toLowerCase();
+
+export const getEncodedParameters = () => {
+  return encode([{ name: 'coinId', type: 'string32', value: coinId }]);
+};
+
+export const printResponse = async (requestId: string) => {
```
Siegrift, line 11: This function is not needed (will never be called since the request fails) but I am OK with it staying

Siegrift (approved):

PR #943, opened 2022-03-16
Remove Errored variant of RequestStatus enum

dcroote:
[AN-482](https://api3dao.atlassian.net/browse/AN-482), part 4.

Potentially controversial changes include functions that previously accepted a single `Request<ApiCall>`, but now take `apiCalls: Request<ApiCall>[],` in order to use `reduce` to filter errored requests.

### DIFF
--- a/packages/airnode-node/src/coordinator/calls/disaggregation.ts
+++ b/packages/airnode-node/src/coordinator/calls/disaggregation.ts
@@ -1,63 +1,59 @@
 import flatMap from 'lodash/flatMap';
-import { logger, PendingLog } from '@api3/airnode-utilities';
+import { logger } from '@api3/airnode-utilities';
 import {
   AggregatedApiCallsById,
   ApiCall,
   Request,
   CoordinatorState,
   EVMProviderState,
-  GroupedRequests,
   LogsData,
   ProviderState,
-  RequestErrorMessage,
   RequestStatus,
+  UpdatedRequests,
 } from '../../types';
 
-export interface RequestsWithLogs {
-  readonly logs: PendingLog[];
-  readonly requests: GroupedRequests;
-}
-
 function updateApiCallResponse(
-  apiCall: Request<ApiCall>,
+  apiCalls: Request<ApiCall>[],
   aggregatedApiCallsById: AggregatedApiCallsById
-): LogsData<Request<ApiCall>> {
-  if (apiCall.status !== RequestStatus.Pending) {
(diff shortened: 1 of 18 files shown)


packages/airnode-node/src/evm/verification/request-verification.ts
```diff
@@ -13,100 +12,85 @@ export const isValidSponsorWallet = (hdNode: ethers.utils.HDNode, sponsor: strin
 
 // TODO: Remove this once https://api3dao.atlassian.net/browse/AN-400 is done
 export function verifySponsorWallets<T>(
-  requests: Request<T>[],
+  unverifiedRequests: Request<T>[],
   masterHDNode: ethers.utils.HDNode
 ): LogsData<Request<T>[]> {
-  const logs: PendingLog[] = [];
+  const { logs, requests } = unverifiedRequests.reduce(
```
dcroote, line 18: @andreogle  - after the long `filter` vs. `reduce` discussion last time, I now agree `reduce` is the safer, more consistent approach  😅

andreogle: haha it grows on you over time - at least it did for me. It can definitely be less readable though

packages/airnode-node/src/types.ts
```diff
@@ -180,6 +173,11 @@ export interface CoordinatorState {
   readonly settings: CoordinatorSettings;
 }
 
+export interface UpdatedRequests<T> {
```
andreogle, line 176: Nice 👍🏻

andreogle (approved):
Nice work 👍🏻

packages/airnode-node/src/coordinator/calls/disaggregation.ts
```diff
@@ -1,63 +1,59 @@
...
   RequestStatus,
+  UpdatedRequests,
 } from '../../types';
 
-export interface RequestsWithLogs {
-  readonly logs: PendingLog[];
-  readonly requests: GroupedRequests;
-}
-
 function updateApiCallResponse(
```
Siegrift, line 15: Rename to `updateApiCallResponses`? (now that it receives multiple apiCalls)

Siegrift (approved):
LGTM, just found one naming suggestion.

packages/airnode-node/src/coordinator/calls/disaggregation.ts
dcroote: Done! Thanks.

PR #944, opened 2022-03-16
Refactor wallet utility functions and callApi (for use in Airkeeper)

vponline:
This adds `deriveSponsorWalletFromMnemonic` as a new function to be used in Airkeeper [here](https://github.com/api3dao/airkeeper/blob/main/src/wallet.ts#L13) and updates `deriveWalletPathFromSponsorAddress` to accept `protocolId` as an optional parameter to enable it for use in Airkeeper [here](https://github.com/api3dao/airkeeper/blob/main/src/wallet.ts#L3).


vponline:
I'll update this with the `callApi` changes needed in Airkeeper for easier review

packages/airnode-node/src/api/index.ts
```diff
@@ -227,6 +229,11 @@ async function processSuccessfulApiCall(
         );
         return [[], { success: true, value: JSON.stringify({ timestamp, value: response.encodedValue }), signature }];
       }
+      case 'beacon': {
```
acenolaza, line 232: What if we just use the `http-gateway` type? 🤔

vponline: > What if we just use the http-gateway type?

Good point, I wasn't sure if it would be weird to have that on the Airkeeper (or if there would be need for more Beacon specific types in Airnode later on)

I can make the change to `http-gateway` if that's preferred

acenolaza: I'm not sure tbh so I'd like @amarthadan opinion on this since he's more familiar with this code these days.

vponline: Actually, I think we may be able to let **airnode-node** do the `templateId` verification so this could warrant a separate type :thinking:

acenolaza: 🤔 Legacy template Ids will include the [airnode address](https://github.com/api3dao/airnode/blob/master/packages/airnode-protocol/contracts/rrp/TemplateUtils.sol#L9) in the hash and [psp beacons won't](https://github.com/api3dao/airnode/blob/v1-proto-tob9-fix/packages/airnode-protocol-v1/contracts/protocol/StorageUtils.sol#L20). I'm not sure we can rely on the node's template verification function for all cases.

acenolaza: > What is the plan here? Does airkeeper want to call airnode to perform the API call instead doing the API call directly?

Right, that's the end goal

vponline: > Legacy template Ids will include the [airnode address](https://github.com/api3dao/airnode/blob/master/packages/airnode-protocol/contracts/rrp/TemplateUtils.sol#L9) in the hash and [psp beacons won't](https://github.com/api3dao/airnode/blob/v1-proto-tob9-fix/packages/airnode-protocol-v1/contracts/protocol/StorageUtils.sol#L20). I'm not sure we can rely on the node's template verification function for all cases.

Ah I see, I may have seem a comment about this somewhere  :smile:

amarthadan: @Siegrift Just to be clear, Airkeeper won't call a running Airnode (like invoking a Lambda or anything). We just want to reuse the `callApi` code in Airkeeper. The request will still be actually done by Airkeeper.

@vponline @acenolaza I'd say let's reuse `http-gateway` type (we might consider renaming it in the future) for now, we can diverse later if need be. And let's keep the template verification for Airkeeper since that would require bigger refactoring on the Airnode side.

amarthadan (changes requested):
Codewise the wallet stuff looks ok to me but maybe someone more knowledgable in this area can check that. @Siegrift @acenolaza :pray: ?

packages/airnode-node/src/config/index.ts
```diff
@@ -23,6 +23,17 @@ const readConfig = (configPath: string): unknown => {
...
+  const rawConfig = readConfig(configPath);
+  const parsedConfigRes = parseConfigWithSecrets(rawConfig, secrets);
+  if (!parsedConfigRes.success) {
+    throw new Error(`Invalid Airnode configuration file: ${parsedConfigRes.error}`);
+  }
+
+  const config = parsedConfigRes.data;
+  return config;
+}
+
```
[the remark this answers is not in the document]
vponline: Yes will do :+1:

Siegrift (approved):
It LGTM 👍

amarthadan (approved):
So, as a result, we didn't need to do anything for `callApi` :smile: Nice!

packages/airnode-node/src/evm/wallet.ts
```diff
@@ -15,16 +15,17 @@ import { Config } from '../config/types';
  * 1/1st31bits/2nd31bits/3rd31bits/4th31bits/5th31bits/6th31bits.
  *
  * @param sponsorAddress A string representing a 20bytes hex address
+ * @param protocolId An optional string representing the protocol id. Defaults to '1'.
  * @returns The path derived from the address
  */
-export const deriveWalletPathFromSponsorAddress = (sponsorAddress: string): string => {
+export const deriveWalletPathFromSponsorAddress = (sponsorAddress: string, protocolId?: string) => {
```
acenolaza, line 21: Nit: What do you think about using a default value instead of making `protocolId` optional? That way you wouldn't need to add the `|| '1' `below on line 28.

acenolaza (approved):
Wallet derivation LGTM 👍🏻

packages/airnode-node/src/evm/wallet.ts
vponline: It's a good idea, I'll open another PR :+1:

PR #952, opened 2022-03-18
Require conditionPspDapiUpdate caller to be zero address

bbenligiray:
When we require `msg.sender==address(0)` this function can no longer change the state, which makes its name accurate


aquarat (approved):
LGTM :1st_place_medal: 
I've requested a re-run of the tests. The unit tests pass on my local machine.

aquarat:
I fixed the things.

Ashar2shahid (commented):
A bit late with the review but everything looks good 👍

PR #953, opened 2022-03-18
Remove unused export

Siegrift:
This line caused CI to fail for some reason (dunno why). This export is not used anywhere so we can remove it to unblock the CI (hopefully).


bbenligiray (approved):
Thanks for looking into it

PR #954, opened 2022-03-18
Update conditions always return true if the data point is not initialized

bbenligiray:
Both for Beacons and dAPIs, the condition functions return `true` if the data point to be updated hasn't been initialized. This addresses the issue (even though it's not usable due to being front-runnable) where people can extract 1 bit from Beacon values by utilizing unintialized dAPIs. 

Also implemented a `updateDapiWithBeaconsAndReturnCondition()` with nicer interface. Note that it doesn't require `msg.sender==address(0)` because that`s not needed.

### DIFF
--- a/packages/airnode-protocol-v1/contracts/dapis/DapiServer.sol
+++ b/packages/airnode-protocol-v1/contracts/dapis/DapiServer.sol
@@ -470,16 +470,19 @@ contract DapiServer is
     ) external override returns (bool) {
         require(msg.sender == address(0), "Sender not zero address");
         bytes32 dapiId = keccak256(data);
-        int224 currentDapiValue = dataPoints[dapiId].value;
+        DataPoint memory initialDataPoint = dataPoints[dapiId];
         require(
             dapiId == updateDapiWithBeacons(abi.decode(data, (bytes32[]))),
             "Data length not correct"
         );
+        DataPoint storage updatedDataPoint = dataPoints[dapiId];
         return
             calculateUpdateInPercentage(
-                currentDapiValue,
+                initialDataPoint.value,
                 dataPoints[dapiId].value
-            ) >= decodeConditionParameters(conditionParameters);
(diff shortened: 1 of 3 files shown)


aquarat (approved):
LGTM :rocket: 
TIL a bit more about void signer contract interactions.
The tests for this PR pass on my local machine with the changes from PR #952 included.

packages/airnode-protocol-v1/contracts/dapis/DapiServer.sol
```diff
@@ -451,6 +454,32 @@ contract DapiServer is
...
+        bytes32 dapiId = keccak256(abi.encode(beaconIds));
+        DataPoint memory initialDataPoint = dataPoints[dapiId];
+        updateDapiWithBeacons(beaconIds);
+        DataPoint storage updatedDataPoint = dataPoints[dapiId];
+        return
+            calculateUpdateInPercentage(
+                initialDataPoint.value,
+                dataPoints[dapiId].value
+            ) >=
+            updateThresholdInPercentage ||
```
Ashar2shahid, line 479: is there any difference between using `updateThresholdInPercentage` vs just using `conditionParameters` like in `conditionPspBeaconUpdate` ? You can derive the updateThreshold from the condition parameters.

Ashar2shahid (commented):
it looks like ToB didn't really give an answer (on slack) to whether the exploit of 1 bit detection only works if the dAPI with ID is not initialized. Assuming that it does, the fix LGTM

bbenligiray:
@Ashar2shahid waiting for your review since the changes are somewhat substantial

packages/airnode-protocol-v1/contracts/dapis/DapiServer.sol
Ashar2shahid: ok that's the only comment I had, everything else looks good 👍

PR #961, opened 2022-03-23
Loosen Solidity version requirement for RrpRequester

bbenligiray:
I tried implementing a requester contract with Solidity 0.8.12 and realized that I can't because RrpRequester and the AirnodeRrp interfaces it depended on are fixed to 0.8.9. I updated the RrpRequester Solidity version to be ^0.8.0 but I couldn't do the same for AirnodeRrp interfaces because it changes the bytecode and breaks verifications. So I had to copy paste the AirnodeRrp interfaces to be used by RrpRequester. It's hacky but we can fix it in the v1 protocol.

### DIFF
--- a/.changeset/three-schools-jam.md
+++ b/.changeset/three-schools-jam.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-protocol': patch
+---
+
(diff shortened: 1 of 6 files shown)


acenolaza (approved):
👍🏻

PR #963, opened 2022-03-23
Enable Multi-architecture docker builds in CI and push to Docker Hub

aquarat:
This PR builds both x86_64 and arm64 Docker images for our dev Docker Hub repositories in the CI system.
The CI system tags the images appropriately and pushes them to Docker Hub, meaning that x86_64 systems (ie. a common PC) will get an x86_64 image on pull/run and an arm64 system (like a Raspberry Pi or M1 Mac) will get an arm64 image.

All automatic, no user intervention required.
A bonus is that removing some of the caching speeds things up slightly (by about 11 seconds).
The only bad aspect is that I'm using a self-hosted runner on an arm64 EC2 VM to run these builds. The reason for this is that Github runners don't support arm64... which means the alternative is QEMU, but QEMU slows things down a lot and makes network connectivity flakey. The EC2 machine is on API3's account and doesn't cost much. The Github self-hosted runner auto-updates and has no inbound connections or static IP.

When it comes time to execute a release it'll be possible to retag these images. I'll document this separately.

To test:
On both an arm64 system and an x86_64 system, run as an example: `docker run -ti --rm api3/airnode-deployer-dev:2519915b6ec46240fbed8787d62cf3e154388b77`
You can also view the manifest: `docker manifest inspect api3/airnode-deployer-dev:2519915b6ec46240fbed8787d62cf3e154388b77`, which will indicate two architectures.


dcroote (commented):
I'm not so sure about removing the caching. Are you sure it's necessary? The latest Action run for this branch is [21m 43s](https://github.com/api3dao/airnode/actions/runs/2033116044) and latest master is [13m 51s](https://github.com/api3dao/airnode/actions/runs/2033116044). What about caching where the key is the `sha` and `os`?

aquarat:
> I'm not so sure about removing the caching. Are you sure it's necessary?

Hey, I've had a bit of a crash course in self hosted Github runners; while Github's self hosted runner agent cleans up the filesystem after itself it doesn't clean up the docker cache. I added docker pruning but that causes a large slow-down. I've since read that a better way of implementing self hosted runners is using ephemeral runners, which are cheaper and start fresh for every job and thus don't require docker pruning. 

The ephemeral runner approach scales to 0 when there are no jobs. There's an off-the-shelf Terraform solution for this. I'll try and deploy this tomorrow and revert some of the caching related changes. If it runs acceptably I'll re-request review :)

amarthadan:
@aquarat It feels like it's getting way too complicated and you're spending way too much time on it. I don't think this is that important, is it? Wouldn't it be better to put it back to the backlog and come back to it later?

aquarat:
> @aquarat It feels like it's getting way too complicated and you're spending way too much time on it. I don't think this is that important, is it? Wouldn't it be better to put it back to the backlog and come back to it later?

Hey, yeah, that had crossed my mind. I think that's a good idea. I'll make a note of my research so far in the Jira issue.

PR #984, opened 2022-03-30
Add type generation for v1 contracts

aquarat:
The factories and types would be useful in other applications.

I wasn't familiar with this system until a few minutes ago, so there may be deficiencies in this PR.


Siegrift (commented):
LGTM so far :)

aquarat:
@Siegrift 
> LGTM so far :)

ahahahaa, I came to this PR now to ask for your opinion :laughing: 
Is this good enough? I'm using this currently in Telemetry, but maybe I made a mistake somewhere. I currently use the types like this:

```
const dapiServer = DapiServer__factory.connect(chain.dapiServer, voidSigner);
const operation = async () => await dapiServer.readWithDataPointId(beaconId);
```

bbenligiray:
@aquarat FYI I've recently migrated out of the monorepo because I don't think that `packages/airnode-protocol-v1` was a good idea.
https://github.com/api3dao/airnode-protocol-v1

PR #985, opened 2022-03-31
Add GCP billing info to Airnode examples readme

kolenic-martin:


packages/airnode-examples/README.md
```diff
@@ -124,7 +124,9 @@ yarn create-aws-secrets
 ### 6. (Only if deploying to GCP) Create GCP credentials
 
 If you intend to deploy Airnode on GCP, you will need to create a service account for your project and create and
-download an access key for the new account. If you are not sure how to do this or how to create a GCP account, see
+download an access key for the new account. If you haven't already done so, you will also need to enable billing for
+your project by pairing it with your credit card (no charges will be incurred). If you are not sure how to do this or
```
Siegrift, line 128: I wonder how safe is this statement. Since you set the billing credentials, since it's theoretically possible that the user pays if the free quota is exceeded.

I think we should formulate id a bit differently, something along these lines:
```
If you haven't already done so, you will also need to enable billing for
your project by pairing it with your credit card. Airnode examples should comfortably fit within the free tier so there will be no charges.
```

WDYT? cc: @amarthadan

Siegrift (approved):

PR #986, opened 2022-04-01
Phase 2 deployments

bbenligiray:
Deploying the v0 protocol contracts on Arbitrum, Avalanche, BSC, Fantom, Gnosis Chain, Optimism, Polygon


bbenligiray:
The `fantom-testnet` deployment didn't get verified on https://testnet.ftmscan.com/
`NETWORK=fantom-testnet yarn run verify-deployment` succeeds so I'm pretty sure this is a problem with their Etherscan clone. It also didn't let me log in using Firefox because the captcha wasn't being displayed so it's probably not properly maintained.

bbenligiray:
@bdrhn9
- Run the `verify-deployment` script for all newly deployed networks and see that the verification is successful
- Check the block explorers to see that all 3 contracts are verified (`fantom-testnet` and `arbitrum` didn't work)
- Run the `generate-references` script to see that `deployments/references.json` content remains the same
- Approve the PR

bdrhn9 (approved):

packages/airnode-protocol/credentials.example.json
```diff
@@ -1,30 +1,116 @@
...
+      "polygon": "",
+      "polygonMumbai": "<polygon-testnet>"
+    }
+  },
   "networks": {
     "mainnet": {
       "accounts": { "mnemonic": "" },
       "chainId": 1,
-      "url": ""
+      "url": "https://mainnet.infura.io/v3/8997be2479d94f91a24df8b56a8e98df"
```
Siegrift, line 28: These URLs are now public and they are valid (not a huge deal though).

bdrhn9: Yep, it belongs to me and it's free account.

bbenligiray: For reference :sweat_smile: 
https://github.com/ethers-io/ethers.js/blob/50d712d6afc927aaee68b482157598554a45d403/packages/providers/src.ts/infura-provider.ts#L17

Siegrift: lul :D

aquarat: This makes me feel better about past mistakes :P

amarthadan: You're too optimistic. It just makes me feel worse about using ethers.js :sweat_smile:

PR #995, opened 2022-04-05
Fix misleading log in initiateTransactions

dcroote:
[AN-643](https://api3dao.atlassian.net/browse/AN-643)


Siegrift (approved):
LGTM 👍 I think the last two (comments) also make sense with sponsor wallet, but just "sponsor" is better.

PR #996, opened 2022-04-06
Refactor AggregatedApiCall interfaces

dcroote:
Moves / removes `callApi` payload properties from `BaseAggregatedApiCall` according to call type. For example, `http-gateway` does not need `id`, `airnodeAddress`, or `endpointId` and `http-signed-data-gateway` does not need `airnodeAddress`.

Relevant for airkeeper, [see here](https://github.com/api3dao/airkeeper/blob/a6090d06facc60c16d60b4cbd8f0613c762dc261/src/api/call-api.ts#L12)

[AN-635](https://api3dao.atlassian.net/browse/AN-635)


packages/airnode-node/src/handlers/process-http-request.ts
```diff
@@ -12,7 +11,6 @@ export async function processHttpRequest(
   parameters: Record<string, string>
 ): Promise<[Error, null] | [null, ApiCallSuccessResponse]> {
   const requestId = randomHexString(16);
-  const airnodeAddress = wallet.getAirnodeWallet(config).address;
   const logOptions = buildBaseOptions(config, { requestId });
```
acenolaza, line 14: Since requestId is optional in LogMetadata, does it make sense to add a random hex string to the log options in this case?

acenolaza (approved):
LGTM 👍🏻 Just minor comment

vponline (approved):
Nice work! Tested with Airkeeper and works nicely :+1:

packages/airnode-node/src/handlers/process-http-request.ts
dcroote: I think you're right, it doesn't appear strictly necessary in isolation, but `processHttpSignedDataRequest` generates the `requestId` in the same (random) way and uses that `id` elsewhere in an error message:
https://github.com/api3dao/airnode/blob/62a090eddf37db93ebc64ba10ec70f21199c4dbe/packages/airnode-node/src/api/index.ts#L138-L139

So I'd lean towards keeping it as it also feels right to have a unique Id within the logs for a request, but if you'd like I can remove it.

acenolaza: Yes it makes sense but this will force us to submit a random hex in the payload from Airkeeper to only be used in those logs where [we are not sure if we should even be calling that function for beacons](https://github.com/api3dao/airnode/blob/62a090eddf37db93ebc64ba10ec70f21199c4dbe/packages/airnode-node/src/api/index.ts#L129) 😁. 

I think you are right and we should keep it just to avoid any error but I also wanted to ask you about it because to me a `requestId` is a hash computed on chain for AirnodeRrp and I'm not sure if it applies to http-gateway, signed or Airkeeper requests. 

Thanks for investigating tho 👏🏻

dcroote: You mean calling from airkeeper [here](https://github.com/api3dao/airkeeper/blob/a6090d06facc60c16d60b4cbd8f0613c762dc261/src/api/call-api.ts#L12-L27)? In that case the payload won't need `id`, `airnodeAddress`, or `endpointId` anymore because [these properties are being removed from `BaseAggregatedApiCall`](https://github.com/api3dao/airnode/pull/996/files#diff-bafef992d17b51ddeeecf5c4f67a79342f80c0b28bd383c18310ce0da1a57278L226) and [`HttpGatewayAggregatedApiCall` doesn't have them](https://github.com/api3dao/airnode/blob/62a090eddf37db93ebc64ba10ec70f21199c4dbe/packages/airnode-node/src/types.ts#L258-L260). That is the point of the PR 😅

packages/airnode-node/test/e2e/request-status.feature.ts
```diff
@@ -33,7 +33,7 @@ it('sets the correct status code for both successful and failed requests', async
   const validFulfillment = logs.find(
     (log) => log.args.requestId === validRequest!.args.requestId && log.name === 'FulfilledRequest'
   );
-  // The API responds with 723.39202 which multipled by the _times parameter
+  // The API responds with 723.39202 which multiplied by the _times parameter
```
andreogle, line 36:
```suggestion
  // The API responds with 723.39202 which is multiplied by the _times parameter
```

andreogle (approved):

PR #997, opened 2022-04-07
Replace API_CALL_FULFILLMENT_GAS_LIMIT with config option

vponline:
This replace `API_CALL_FULFILLMENT_GAS_LIMIT` constant with `fulfillmentGasLimit` configuration option.

### DIFF
--- a/packages/airnode-node/src/handlers/start-coordinator.test.ts
+++ b/packages/airnode-node/src/handlers/start-coordinator.test.ts
@@ -28,7 +28,6 @@ import * as adapter from '@api3/airnode-adapter';
 import * as validator from '@api3/airnode-validator';
 import { startCoordinator } from './start-coordinator';
 import * as fixtures from '../../test/fixtures';
-import { API_CALL_FULFILLMENT_GAS_LIMIT } from '../constants';
 
 describe('startCoordinator', () => {
   test.each(['legacy', 'eip1559'] as const)(`fetches and processes requests - txType: %s`, async (txType) => {
@@ -40,6 +39,7 @@ describe('startCoordinator', () => {
         ...chain,
         options: {
           txType,
+          fulfillmentGasLimit: 500000,
         },
       })),
     };
@@ -90,7 +90,7 @@ describe('startCoordinator', () => {
       '0x7c1de7e1',
       '0x0000000000000000000000000000000000000000000000000000000002a5213d',
       '0x69567b16514c2b799597247462cc6c3d9ac9dce88c0bc97c17db45dfb572cacb0fc7b38b2a73cf1fd78279251e5ef75b5e6fb06f4b0f0d023c4b215609e2e38f1b',
-      { gasLimit: API_CALL_FULFILLMENT_GAS_LIMIT, ...gasTarget, nonce: 212 }
+      { ...gasTarget, nonce: 212 }
     );
   });
 
--- a/packages/airnode-utilities/src/evm/gas-prices/gas-prices.ts
+++ b/packages/airnode-utilities/src/evm/gas-prices/gas-prices.ts
@@ -4,6 +4,13 @@ import { go } from '../../promises';
 import { LogsData, logger, PendingLog } from '../../logging';
 import { DEFAULT_RETRY_TIMEOUT_MS, PRIORITY_FEE_IN_WEI, BASE_FEE_MULTIPLIER } from '../../constants';
 
+// This will return the gasLimit with the gasTarget if getGasPrice is called from a place where the fulfillmentGasLimit
+// is available in the config.json and otherwise no gasLimit will be returned for use with other transactions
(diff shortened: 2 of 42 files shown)


packages/airnode-utilities/src/evm/gas-prices/gas-prices.ts
```diff
@@ -4,6 +4,13 @@ import { go } from '../../promises';
 import { LogsData, logger, PendingLog } from '../../logging';
 import { DEFAULT_RETRY_TIMEOUT_MS, PRIORITY_FEE_IN_WEI, BASE_FEE_MULTIPLIER } from '../../constants';
 
+// This will return the gasLimit with the gasTarget if getGasPrice is called from a place where the fulfillmentGasLimit
+// is available in the config.json and otherwise no gasLimit will be returned for use with other transactions
+// (e.g. withdrawal transactions and airnode-admin CLI/SDK)
+export const getGasLimit = (fulfillmentGasLimit?: number) => {
+  return fulfillmentGasLimit ? { gasLimit: BigNumber.from(fulfillmentGasLimit) } : {};
+};
+
```
vponline, line 13: @aquarat  If you have the chance to check this, I'd appreciate your thoughts on this solution.

We need this mainly for when transactions are submitted from the CLI/SDK from airnode-admin since we don't have the airnode `config.json` there. This should be fine since fulfillment transactions are not submitted from there, right?

packages/airnode-validator/templates/0.4/config.json
```diff
@@ -133,6 +133,12 @@
             }
```
amarthadan, line 133: You really don't need to update these and definitely not old ones. You would need to add a new 0.6 one but don't, we'll be removing these soon.

amarthadan (approved):
I'm fine with this but @aquarat you should check the pricing part.

packages/airnode-validator/templates/0.4/config.json
vponline: There are still some unit tests that failed if I didn't update them :thinking:

amarthadan: @Siegrift You've added the new validator, do you know what's going on there?

Siegrift: This should only affect old validator, but I'll take a look

Siegrift: I see the problem now. The old validator tests are still in place, but we no longer have the 0.6 template and we actually use the new validator in deployer and node. Since you have added a new field to the config.json the tests broke because they are validating it with 0.5 template.

While investigating this I accidentally changed the node version to 0.4 for the 0.5 templates. I do not thing we need to release a fix for v0.5 though (because the new validator does not check the node version).

I think the easiest way to fix this is to copy the `0.5` templates to `0.6` and add the `fulfillmentGasLimit` property there and fix the incorrect versions. See: https://github.com/api3dao/airnode/tree/gas-limit-emo 

I think we should fix the validator properly in v0.7.

amarthadan: Can't we comment out those tests that are using the old validator and not update the templates? I agree with finishing this properly in v0.7.

Siegrift: Yeah, we can do that as well.

amarthadan: @vponline Let's do that then.

vponline: @Siegrift  This means we need to remove all tests from 
**airnode-validator/src/validator.test.ts** and
**airnode-validator/test/validator.test.ts**, right?

So I'll need to remove the files since empty test files are not allowed.

Siegrift: Yeah.

packages/airnode-node/src/handlers/start-coordinator.test.ts
```diff
@@ -40,6 +39,7 @@ describe('startCoordinator', () => {
         ...chain,
         options: {
           txType,
+          fulfillmentGasLimit: 500000,
```
aquarat, line 42: It's minor, but in other areas of our code it's standard practice to use underscores to delineate thousands, eg. `5_000`.
:)

packages/airnode-utilities/src/evm/gas-prices/gas-prices.ts
aquarat: This approach seems fine to me :)
It's quite elegant.

aquarat (approved):
LGTM :bullettrain_front: 
Very neat :smile:

packages/airnode-node/src/handlers/start-coordinator.test.ts
vponline: Thanks, fixed (I keep forgetting about this) :smile:

PR #1009, opened 2022-04-15
Add chain names to references

bbenligiray:
Adds supported chain names to references.json for docs (@wkande FYI)

Can later be used in https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/evm/networks.ts (AN-648)


wkande (approved):
Yes this works well for the docs in that the chain names in addition to IDs are both pulled from the `references.json` file.

PR #1011, opened 2022-04-17
Update v0 contracts

bbenligiray:
- Copy-pasted AccessControlRegistry (so that we won't need to re-deploy for DapiServer, etc.) and Whitelist (just because) contracts from https://github.com/api3dao/airnode-protocol-v1
- Copy-pasted RequesterAuthorizer contracts from https://github.com/api3dao/airnode-protocol-v1 but also implemented a `isAuthorizedV0()` function that is backwards compatible. This will be updated at https://github.com/api3dao/airnode-protocol-v1 too.
- Added V0 to v0-specific contract names (so no version in the name means V1-compatible). I wasn't sure how far into the node code I should propagate the V0 name. I went with "if it's specifically referring to the contract name, add V0" (so contract types and factories are modified, the rest isn't)
- A few small changes in their individual commits


packages/airnode-admin/src/implementation.ts
```diff
@@ -393,7 +397,7 @@ export async function isRequesterWhitelisted(
   endpointId: string,
   requesterAddress: string
 ) {
-  return requesterAuthorizerWithAirnode.requesterIsWhitelisted(airnodeAddress, endpointId, requesterAddress);
```
bbenligiray, line 396: This function doesn't exist in v1

packages/airnode-protocol/contracts/access-control-registry/AccessControlManagerProxy.sol
```diff
@@ -1,100 +0,0 @@
-// SPDX-License-Identifier: MIT
```
bbenligiray, line 1: This is redundant with OwnableCallForwarder

packages/airnode-protocol/contracts/access-control-registry/AccessControlRegistry.sol
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: MIT
```
bbenligiray, line 1: Exact match with https://github.com/api3dao/airnode-protocol-v1

packages/airnode-protocol/contracts/access-control-registry/AccessControlRegistryAdminned.sol
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: MIT
```
bbenligiray, line 1: Exact match with https://github.com/api3dao/airnode-protocol-v1

packages/airnode-protocol/contracts/access-control-registry/AccessControlRegistryAdminnedWithManager.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: MIT
```
bbenligiray, line 1: Exact match with https://github.com/api3dao/airnode-protocol-v1

packages/airnode-protocol/contracts/authorizers/RequesterAuthorizer.sol
```diff
@@ -1,16 +1,16 @@
 // SPDX-License-Identifier: MIT
```
bbenligiray, line 1: Exact match with https://github.com/api3dao/airnode-protocol-v1 except `isAuthorizedV0()`

packages/airnode-protocol/contracts/authorizers/RequesterAuthorizer.sol
```diff
@@ -153,14 +157,15 @@ abstract contract RequesterAuthorizer is Whitelist, IRequesterAuthorizer {
     /// @param sponsor Sponsor address
     /// @param requester Requester address
     /// @return Authorization status of the request
-    function isAuthorized(
+    function isAuthorizedV0(
```
bbenligiray, line 160: Compatible with AuthorizationUtilsV0. The node never calls this directly so it doesn't need to be updated.

packages/airnode-protocol/contracts/authorizers/RequesterAuthorizerWithAirnode.sol
```diff
@@ -5,7 +5,7 @@ import "../whitelist/WhitelistRolesWithAirnode.sol";
 import "./RequesterAuthorizer.sol";
```
bbenligiray, line 5: Exact match with https://github.com/api3dao/airnode-protocol-v1

packages/airnode-protocol/contracts/authorizers/RequesterAuthorizerWithManager.sol
```diff
@@ -7,8 +7,6 @@ import "./interfaces/IRequesterAuthorizerWithManager.sol";
```
bbenligiray, line 7: Exact match with https://github.com/api3dao/airnode-protocol-v1

packages/airnode-protocol/contracts/rrp/requesters/RrpBeaconServerV0.sol
```diff
@@ -1,10 +1,9 @@
 // SPDX-License-Identifier: MIT
```
bbenligiray, line 1: Replaced Whitelist + WhitelistRolesWithManager with WhitelistWithManager, similar to DapiServer

packages/airnode-protocol/contracts/utils/OwnableCallForwarder.sol
```diff
@@ -15,8 +15,8 @@ import "./interfaces/IOwnableCallForwarder.sol";
 /// The ownership of this contract is deliberately renouncable. If this does
```
bbenligiray, line 15: Exact match with https://github.com/api3dao/airnode-protocol-v1

packages/airnode-protocol/contracts/whitelist/Whitelist.sol
```diff
@@ -1,12 +1,12 @@
 // SPDX-License-Identifier: MIT
```
bbenligiray, line 1: Exact match with https://github.com/api3dao/airnode-protocol-v1

packages/airnode-protocol/contracts/whitelist/WhitelistRoles.sol
```diff
@@ -1,13 +1,11 @@
 // SPDX-License-Identifier: MIT
```
bbenligiray, line 1: Exact match with https://github.com/api3dao/airnode-protocol-v1

packages/airnode-protocol/contracts/whitelist/WhitelistRolesWithAirnode.sol
```diff
@@ -1,37 +1,43 @@
 // SPDX-License-Identifier: MIT
```
bbenligiray, line 1: Exact match with https://github.com/api3dao/airnode-protocol-v1

packages/airnode-protocol/contracts/whitelist/WhitelistRolesWithManager.sol
```diff
@@ -1,24 +1,27 @@
 // SPDX-License-Identifier: MIT
```
bbenligiray, line 1: Exact match with https://github.com/api3dao/airnode-protocol-v1

packages/airnode-protocol/contracts/whitelist/WhitelistWithManager.sol
```diff
@@ -0,0 +1,143 @@
+// SPDX-License-Identifier: MIT
```
bbenligiray, line 1: Exact match with https://github.com/api3dao/airnode-protocol-v1

packages/airnode-protocol/package.json
```diff
@@ -43,6 +42,7 @@
     "typescript": "^4.2.4"
   },
   "dependencies": {
+    "@openzeppelin/contracts": "4.4.2",
```
bbenligiray, line 45: Otherwise contracts imported from `@api3/airnode-protocol` don't compile if they have an openzeppelin dependency

packages/airnode-protocol/test/rrp/requesters/RrpBeaconServerV0.sol.js
```diff
@@ -29,50 +28,92 @@ describe('RrpBeaconServerWithManager', () => {
...
-    const rrpBeaconServerFactory = await hre.ethers.getContractFactory('RrpBeaconServer', roles.deployer);
+    const rrpBeaconServerFactory = await hre.ethers.getContractFactory('RrpBeaconServerV0', roles.deployer);
     rrpBeaconServer = await rrpBeaconServerFactory.deploy(
       accessControlRegistry.address,
       rrpBeaconServerAdminRoleDescription,
       roles.manager.address,
       airnodeRrp.address
     );
     const managerRootRole = await accessControlRegistry.deriveRootRole(roles.manager.address);
+    // Initialize the roles and grant them to respective accounts
```
bbenligiray, line 41: This stuff can be done with a single tx because AccessControlRegistry now inherits Multicall

packages/airnode-protocol/hardhat.config.js
```diff
@@ -23,5 +23,11 @@ module.exports = {
   },
   solidity: {
     version: '0.8.9',
```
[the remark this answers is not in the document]
acenolaza: I'm also making the assumption that 0.8.0 is the highest patch version that inheritance permits right? I'm guessing that versions form 0.8.1 to 0.8.8 won't work

acenolaza (approved):
Will we need to update Airkeeper RRP beacon updater handler to point to the v0 contract instead?

packages/airnode-protocol/hardhat.config.js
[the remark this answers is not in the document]
bbenligiray: I used `^0.8.0` in interfaces and contracts to be inherited and `0.8.9` in contracts that will be deployed. This is because we will be using `0.8.9`, but we don't want to lock the user into that if they will be importing `@api3/airnode-protocol` to use some of these contracts to build their own project. For example, the user is expected to inherit `RrpRequester` but if the version of that is `0.8.9` then the requester contract will also have to be exactly `0.8.9`, which is annoying for the user.

PR #1022, opened 2022-04-22
Fix GCP signed data gateway receipt, let deployer fail with non zero status code

Siegrift:
https://api3dao.atlassian.net/browse/AN-695

You can review commit by commit. There is one commit which is not necessarily part of this issue (https://github.com/api3dao/airnode/pull/1022/commits/3dfc03d02ba892691105cf1b5fb471c2ac1b52c6) but it was easy to implement so I've added it. If you disagree, we can drop it.


packages/airnode-deployer/src/cli/commands.ts
```diff
@@ -69,6 +69,11 @@ export async function deploy(configPath: string, secretsPath: string, receiptFil
   // AWS doesn't allow uppercase letters in S3 bucket and lambda function names
   const airnodeAddressShort = shortenAirnodeAddress(airnodeAddress);
 
+  // Deployment is not an atomic operation. It is possible that some resources are deployed even when there is a
```
andreogle, line 72: 👍🏻

packages/airnode-deployer/src/cli/commands.ts
```diff
@@ -91,6 +95,15 @@ export async function deploy(configPath: string, secretsPath: string, receiptFil
   fs.rmSync(tmpDir, { recursive: true });
 
   writeReceiptFile(receiptFile, mnemonic, config, output, deploymentTimestamp);
+
+  if (deploymentError) {
+    logger.fail(
+      `Airnode deployment failed due to errors.\n` +
```
andreogle, line 101:
```suggestion
      `Airnode deployment failed due to an unexpected error.\n` +
```

packages/airnode-deployer/src/cli/commands.ts
```diff
@@ -91,6 +95,15 @@ export async function deploy(configPath: string, secretsPath: string, receiptFil
...
+        `  It is possible that some resources have been deployed on cloud provider.\n` +
+        `  Please use the "remove" command from the deployer CLI to make sure all cloud resources are removed.`
```
andreogle, line 103:
```suggestion
        `  Please use the "remove" command from the deployer CLI to ensure all cloud resources are removed.`
```

andreogle (approved):

packages/airnode-deployer/src/cli/commands.ts
Siegrift: There might me multiple errors actually (e.g. yesterday I got 3 for GCP deployment). Would you still consider your suggestion more accurate?

andreogle: Ok I didn't realise that. Maybe just "due to unexpected errors". 

Just "due to errors" sounds a bit weird to me

packages/airnode-deployer/src/cli/index.ts
```diff
@@ -22,17 +22,21 @@ function drawHeader() {
...
     await command();
-  } catch (err) {
-    loggerUtils.error((err as Error).message);
+  } catch {
+    // We are not logging the error since all errors should be handled by the specific deployer command. An error here
+    // means the command was not successful and should fail with non zero status code.
+    //
     // eslint-disable-next-line functional/immutable-data
     process.exitCode = 1;
   }
```
amarthadan, line 42: So, you've removed the logging of the error from `commands.ts` and let the exception go higher (which is here) but here you're ignoring it? Is it logged somewhere? Maybe I'm just blind.

Siegrift: For Airnode `deploy` I've added the error logging and for `remove` the error is actually logged here and rethrown https://github.com/api3dao/airnode/blob/master/packages/airnode-deployer/src/infrastructure/index.ts#L315

amarthadan: Those only log that the error happened they don't log the error itself. Without that, we actually don't know what really happened, just that something failed. We should definitely either log the error itself or throw it.

Siegrift: 👍 You're right. We should log it.

Siegrift: Actually, deployer package uses a pattern where the error is logged using `logger.fail` which outputs the error nicely in the terminal. Then, the error is rethrown. This means that functions should log the error themselves. We do this for all TF commands (https://github.com/api3dao/airnode/blob/master/packages/airnode-deployer/src/infrastructure/index.ts#L64), but there are some places where this is not the case and error is not handled at all (e.g. checking cloud provider state). 

After some thinking I think logging the error here most of the times just prints what was already printed... However, there might be an unhandled error which would be printed only by this place. If we do not log here, we could risk silent errors which is not worth it.

packages/airnode-deployer/src/cli/index.ts
```diff
@@ -22,8 +23,10 @@ function drawHeader() {
       "|  _  | | '__| '_ \\ / _ \\ / _` |/ _ \\",
       '| | | | | |  | | | | (_) | (_| |  __/',
       '\\_| |_/_|_|  |_| |_|\\___/ \\__,_|\\___|',
+      '',
```
Siegrift, line 26: Added this to make the CLI output less cluttered. It looks nicer now.

packages/airnode-deployer/src/cli/commands.ts
```diff
@@ -91,6 +94,17 @@ export async function deploy(configPath: string, secretsPath: string, receiptFil
   fs.rmSync(tmpDir, { recursive: true });
 
   writeReceiptFile(receiptFile, mnemonic, config, output, deploymentTimestamp);
+
+  if (deploymentError) {
+    logger.fail(
+      bold(
```
Siegrift, line 100: I've added it to `bold` because this is important, but the message can be in the middle of the error log. See:

image

packages/airnode-deployer/src/cli/commands.ts
```diff
@@ -21,26 +23,23 @@ export async function deploy(configPath: string, secretsPath: string, receiptFil
...
-    logger.fail(message);
-    throw new Error(message);
+    throw logAndReturnError(`Deployer can't deploy to "local" cloud provider`);
   }
 
   const mnemonic = config.nodeSettings.airnodeWalletMnemonic;
   if (!validateMnemonic(mnemonic)) {
-    logger.fail('AIRNODE_WALLET_MNEMONIC in your secrets.env file is not valid');
-    throw new Error('Invalid mnemonic');
+    throw logAndReturnError('AIRNODE_WALLET_MNEMONIC in your secrets.env file is not valid');
```
Siegrift, line 31: I wanted to avoid doing:
```js
const message = "xyz"
logger.fail(message)
throw new Error(message)
```

packages/airnode-deployer/src/utils/files.ts
```diff
@@ -12,8 +13,7 @@ export function parseSecretsFile(secretsPath: string) {
   try {
     return dotenv.parse(fs.readFileSync(secretsPath));
   } catch (e) {
-    logger.fail('Failed to parse secrets file');
-    throw e;
+    throw logAndReturnError(`Failed to parse secrets file: ${e}`);
```
Siegrift, line 16: Added `${e}` for more datails.

amarthadan (approved):

PR #1025, opened 2022-04-22
Enhance processing and add caching

aquarat:
This PR extends https://github.com/api3dao/airnode/pull/1008

This PR adds asynchronous pre and post processing, which enables some fun use cases, like OAuth2.

Unlike the sync eval variant, code should follow this format:
```
({input, resolve}) => {console.log('hello'); resolve({...input, something: 1})};
```


packages/airnode-node/src/api/processing.ts
```diff
@@ -51,20 +54,19 @@ export const postProcessApiSpecifications = async (input: unknown, endpoint: End
...
-                return unsafeEvaluate(input, currentValue.value);
-              default:
-                throw new Error(`Environment ${currentValue.environment} is not supported`);
-            }
-          }, input)
-        );
-      }),
-    { retries: 0, timeoutMs: 500 }
+    async () =>
+      await postProcessingSpecifications.reduce(async (input: any, currentValue: ProcessingSpecification) => {
```
aquarat, line 58: `await` may not be necessary?

packages/airnode-node/src/constants.test.ts
```diff
@@ -23,5 +23,6 @@ describe('constants', () => {
     expect(constants.PRIORITY_FEE).toEqual(3120000000);
     expect(constants.BASE_FEE_MULTIPLIER).toEqual(2);
     expect(constants.MAXIMUM_ONCHAIN_ERROR_LENGTH).toEqual(100);
+    expect(constants.PROCESSING_TIMEOUT).toEqual(10_000);
```
aquarat, line 26: Processing timeout should really be configurable from `config.json` as it can be much longer than 500ms when relying on remote I/O.

Siegrift (commented):
Took an initial look (was curious :p ). I like the vm module from node 💪

packages/airnode-node/src/constants.ts
```diff
@@ -44,3 +44,6 @@ export const BASE_FEE_MULTIPLIER = 2;
 
 // The maximum character length of an error message sent on-chain
 export const MAXIMUM_ONCHAIN_ERROR_LENGTH = 100;
+
+// The maximum amount of time in milliseconds that pre and post processing is allowed to execute
+export const PROCESSING_TIMEOUT = 10_000;
```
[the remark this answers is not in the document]
aquarat: Yes, ideally imo this should be part of the `ProcessingSpecification` in `config.json` (configurable per snippet).

aquarat:
> But isn't 10s default too much? Also this timer affects the Airnode cycle and those are not configurable as of now (I think we should discuss this).

The current OAuth2 snippet I'm using (which I can post on Slack if you're curious), takes around 2s to resolve a token (when it needs to refresh the token, often an existing token will be available from the cache in which case it will be much faster, hence why caching is here).

aquarat:
I've kept the timeout on reduce at 10s (constant) and given the individual code snippets configurable timeouts.
I've switched out `eval` for `vm.runInNewContext`.
I need to run a few real-world tests but this should be pretty close to ideal.

aquarat:
>But isn't 10s default too much? Also this timer affects the Airnode cycle and those are not configurable as of now (I think we should discuss this).

This is now the total timeout and it shouldn't affect anything else? I think the total parent cycle has 30 seconds of execution time?

aquarat:
This is still pending testing on cloud.
Caching is included for OAuth2 operation.

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -0,0 +1,252 @@
...
+    logger.error((e as Error).stack!);
+  }
+};
+
+/**
+ * Get all keys from the cache that start with startingKey.
+ *
+ * @param startingKey should take the form of `blockedRequestId-` and the full key could be `blockedRequestId-0x0000000...`
+ */
+const getKeys = (startingKey?: string): string[] => {
```
Siegrift, line 64:
```suggestion
const getKeys = (prefix?: string): string[] => {
```

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -0,0 +1,252 @@
...
+  try {
+    const keys = readdirSync(CACHE_BASE_PATH);
+    if (startingKey) {
+      return keys.filter((key) => key.indexOf(startingKey) === 0);
```
Siegrift, line 68:
```suggestion
      return keys.filter((key) => key.startsWith(startingKey) === 0);
```

because the function is misleading currently.

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -0,0 +1,252 @@
...
+  } catch (e) {
+    logger.error(`Unable to retrieve keys from fs cache`);
+    logger.error((e as Error).stack!);
+  }
+
+  return [];
+};
+
+/**
+ * Get the value stored for a key (aka the contents of the file with the key as name).
```
Siegrift, line 80:
```suggestion
 * Get the value stored for a key (a.k.a the contents of the file with the key as name).
```

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -0,0 +1,252 @@
...
+    logger.error((e as Error).stack!);
+  }
+};
+
+/**
+ * Add a cache entry
+ *
+ * @param key the name of the key, ideally in the form of `something-{uniqueId}`, eg. `requestIdBlocked-0x0000...`
+ * @param data the data to store for the key, arbitrary
+ * @param force force overwriting of an existing key (discouraged)
```
Siegrift, line 119: Mhh, is this really discouraged? Won't this be a relatively common operation (refresh the data returned by the API)?

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -0,0 +1,252 @@
...
+
+/**
+ * Get all keys and their associated values
+ * This is a potentially dangerous operation as it will cause the entire cache to be loaded into memory... which could
+ * be more than 500 MB.
+ *
+ * On Lambda RAM and tmpfs are separate resources, but on GCP tmpfs *is* RAM, so reading 500 MB of cache into RAM will
+ * incur a >1GB memory usage cumulatively.
+ */
+const getAllKeysAndValues = (): Record<string, any> => {
```
Siegrift, line 155: Do we need to provide this? Especially since it has those disadvantages?

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -0,0 +1,252 @@
...
+  }
+};
+
+/**
+ * Get all keys and their associated values
+ * This is a potentially dangerous operation as it will cause the entire cache to be loaded into memory... which could
+ * be more than 500 MB.
+ *
+ * On Lambda RAM and tmpfs are separate resources, but on GCP tmpfs *is* RAM, so reading 500 MB of cache into RAM will
+ * incur a >1GB memory usage cumulatively.
```
Siegrift, line 153: Will it result only in increased pricing or can it crash the lambda?

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -0,0 +1,252 @@
...
+
+    const remainingFiles = stattedFiles.filter((file) => !deletedFiles.includes(file));
+
+    if (remainingFiles.length < CACHE_MAX_FILES) {
+      return;
+    }
+
+    // Delete oldest cache entries first
+    type statFile = { file: string; mtimeMs: number };
+    const sortFn = (a: statFile, b: statFile) => {
```
Siegrift, line 199: 👍 I like this is not written as `b.mtimeMs - a.mtimeMs`

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -0,0 +1,252 @@
+import { existsSync, mkdirSync, readdirSync, readFileSync, rmdirSync, rmSync, statSync, writeFileSync } from 'fs';
+import { spawn } from 'child_process';
+import { join } from 'path';
+import { logger } from '../logging';
+
+export const CACHE_BASE_PATH = `/tmp/airnode-cache`;
+export const CACHE_MAX_FILESYSTEM_AGE_MINUTES = 60;
+export const CACHE_MAX_FILES = 100_000; // (500 (cache size) * 1024 (KB) ) / 4 (KB, fs "cluster size" = 130k
```
Siegrift, line 8: I am not sure I understand this comment, also 100K seems to me like too many files. 

Why is the cache size 500 ? Why `* 1024 / 4` ?

Siegrift (changes requested):
Looks mostly good to me but I left some comments.

packages/airnode-node/src/api/processing.ts
```diff
@@ -16,15 +17,17 @@ export const preProcessApiSpecifications = async (payload: CallApiPayload): Prom
 
   const [err, processedParameters] = await go(
     async () =>
-      preProcessingSpecifications.reduce((input: any, currentValue: ProcessingSpecification) => {
+      await preProcessingSpecifications.reduce(async (input: any, currentValue: ProcessingSpecification) => {
```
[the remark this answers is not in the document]
aquarat: :laughing: good point :stuck_out_tongue:

packages/airnode-node/src/api/unsafe-evaluate.ts
```diff
@@ -1,13 +1,42 @@
+import { runInNewContext } from 'vm';
```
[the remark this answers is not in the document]
aquarat: This is all pending testing on cloud. I expect some caveats around webpack minification. Just been a bit busy :sweat_smile: 
We can leave this thread hanging until tested on cloud :muscle:

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -0,0 +1,252 @@
...
+
+/**
+ * Calls the OS sync command to forcefully sync the cache filesystem.
+ * This is the functional equivalent of "commit" from SQL.
+ */
+const syncFsSync = () => {
+  try {
+    // TODO find a way to mock out spawn...
+    if (!isJest()) {
+      spawn(`sync`, [`-f`, CACHE_BASE_PATH]);
```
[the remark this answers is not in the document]
aquarat: ahhhh, thank you!

packages/airnode-node/src/api/unsafe-evaluate.ts
```diff
@@ -1,13 +1,42 @@
...
-export const unsafeEvaluate = (_input: any, code: string) => {
-  return eval(`
-    const ethers = require('ethers');
-    const input = _input;
+export const unsafeEvaluate = (input: any, code: string, timeout: number) => {
+  const vmContext = {
+    require,
+    input,
+    deferredOutput: undefined,
+  } as { deferredOutput?: any };
```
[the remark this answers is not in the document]
aquarat: You're right, it isn't, good spot.

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -0,0 +1,252 @@
...
+  } catch (e) {
+    logger.error(`Unable to sync cache fs`);
+    logger.error((e as Error).stack!);
+  }
+};
+
+/**
+ * Syncs the temporary filesystem asynchronously.
+ * Lambda will "freeze" and "thaw" the execution context and with it any async promises in progress. We create and
+ * forget this promise here, but it will resolve at some point, either in this invocation or a later one.
```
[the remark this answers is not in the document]
aquarat: After some more thought calling `sync` is probably unnecessary;
file system access should be thread safe and the cache is intended to be used in an append-only way (we create files, we don't modify them).
Explicitly calling `sync` is often used by databases to commit transactions to disk so that the database can be recovered in the event of a system crash or power outage, but this is not a problem here because we can expect cloud to be reliable and we don't care about the longevity of the cache beyond a few minutes.
So I think I'll remove it entirely.

packages/airnode-utilities/src/caching/index.ts
aquarat: Nice :)

packages/airnode-utilities/src/caching/index.ts
aquarat: One of the big use cases of this is to cache an API response for a unique `requestId`, in these cases you won't overwrite the API response because if the key exists you'd want to return the value for that key to the requester.

This is for when API operations are non-idempotent and where a request doesn't get honoured on the first, second, etc. cycle.

It should be fine @ force, I was being paranoid!

packages/airnode-utilities/src/caching/index.ts
aquarat: It's removal doesn't impact me and the app should first do a key search, so you're right, I'll remove it :)

packages/airnode-utilities/src/caching/index.ts
aquarat: Maybe @ 100K :D 
The Lambda `/tmp` directory is 500 MB in size. It uses a 4 KB sector/record size (different file systems call it different things, but it's the minimum file size).
500 MB * 1024 = KB, KB /4 gives you the maximum number of minimum-sized files, which works out to around 128 000 files.
I figured that 100k files was more than enough and gives us some headroom, so that's what I made the number, but we can make it 20k.
You'll probably tell me that I got the maths wrong @ Kibi vs Kilo :P but I tested the above limits too by running some test code on Lambda.

packages/airnode-utilities/src/caching/index.ts
aquarat: It would not increase pricing. It also would not crash the Lambda because Lambda uses separate "storage" backing for `/tmp` vs RAM, whereas GCP uses RAM for `/tmp`, so on GCP if you use too much cache you'll run out of RAM. Node's program memory footprint gets allocated up-front until there's more pressure AFAIK, so chances are it would not crash the app on GCP either - file writes would probably fail first.

aquarat:
Hopefully I can test this on cloud today.

packages/airnode-utilities/src/caching/index.ts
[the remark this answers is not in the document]
Siegrift: Yeah, that makes even more sense to me.

packages/airnode-utilities/src/caching/index.ts
Siegrift: It feels a bit unsafe. Usually when you are running out of RAM, the whole system starts to slow down. I fear that it could slow down/crash the lambda. I think we should start with a smaller cache first and then go big when we see it works without problems.

packages/airnode-utilities/src/caching/index.ts
Siegrift: Where did you get the information that minimum file size is 4 KB? That seems way too much for a minimum file size.

aquarat:
Works on cloud with the most recent commit.

packages/airnode-utilities/src/caching/index.ts
aquarat: 4 KB is a common sector size for modern filesystems. I think found the size using `statSync` on a freshly created file, but I can test again some time.

It is also the standard minimum file size for Linux tmpfs based volumes.

packages/airnode-utilities/src/caching/index.ts
aquarat: It only uses RAM on GCP, not Lambda.
The reason computers often slow down when using more RAM is not because of RAM per se but because the OS starts swapping out to disk (swap space). That won't occur on Lambda or GCP - but I've reduced the file count down to 20k, which is around 80 MB maximum usage. Let me know if you think it should be lower. The above excludes garbage collection and enumerating files as part of `sweep` (that's also extremely fast).

packages/airnode-utilities/src/caching/index.ts
Siegrift: Thanks for the links! :)

packages/airnode-utilities/src/caching/index.ts
Siegrift: > but I've reduced the file count down to 20k, which is around 80 MB maximum usage. Let me know if you think it should be lower

I think this is good.

packages/airnode-node/src/api/unsafe-evaluate.ts
```diff
@@ -1,13 +1,117 @@
...
+import trace_events from 'trace_events';
+import tty from 'tty';
+import url from 'url';
+import util from 'util';
+import v8 from 'v8';
+import vm from 'vm';
+import worker_threads from 'worker_threads';
+import zlib from 'zlib';
+
+const buildInNodeModules = {
```
Siegrift, line 38: Do we need to pass all of these? Will it work OK when user overrides the name of the built in module (e.g. does `const crypto = ...`)? I think it will work, but asking just in case.

packages/airnode-node/src/api/unsafe-evaluate.ts
```diff
@@ -1,13 +1,117 @@
...
 /**
  * This function is dangerous. Make sure to use it only with Trusted code.
  */
-export const unsafeEvaluate = (_input: any, code: string) => {
-  return eval(`
-    const ethers = require('ethers');
-    const input = _input;
+export const unsafeEvaluate = (input: any, code: string, timeout: number) => {
+  const vmContext = {
+    input,
```
Siegrift, line 82: We stopped passing `ethers`, I think that one is rather necessary.

packages/airnode-utilities/src/caching/index.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  it(`sweeps the cache - no old files present`, () => {
+    const files = ['1', '2', '3', '4', '5'];
+    const filesStatData = files.map((file) => ({ file, mtimeMs: Date.now() }));
+
+    const readdirSyncSpy = jest.spyOn(fs, 'readdirSync');
+    readdirSyncSpy.mockReturnValueOnce(files as unknown as Dirent[]);
+
+    const statSyncSpy = jest.spyOn(fs, 'statSync');
+    // @ts-ignore
```
Siegrift, line 65: Do you need these @ts-ignore(s) ?

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -0,0 +1,190 @@
+import { existsSync, mkdirSync, readdirSync, readFileSync, rmdirSync, rmSync, statSync, writeFileSync } from 'fs';
+import { join } from 'path';
+import { logger } from '../logging';
+
+export const CACHE_BASE_PATH = `/tmp/airnode-cache`;
+export const CACHE_MAX_FILESYSTEM_AGE_MINUTES = 60;
+export const CACHE_MAX_FILES = 20_000; // (500 (cache size) * 1024 (KB) ) / 4 (KB, fs "cluster size" = 130k
```
Siegrift, line 7: Nit: It wuld be nice to add the links you found here.

Siegrift (commented):
Added a few more comments, but it LGTM overall. 

Good job 💪

packages/airnode-node/src/api/unsafe-evaluate.ts
aquarat: Many things don't work without this, I'm not entirely sure why, webpack seems to override the default require function and strips out everything else. `fs` definitely doesn't work. I just got a list of default modules for node and included most of the modules except the deprecated ones. I did this because I don't know what we might need in future and the cost of including these extra modules seems negligible.

packages/airnode-node/src/api/unsafe-evaluate.ts
aquarat: Yeah, I think we should also pass all our airnode-* packages.

packages/airnode-utilities/src/caching/index.test.ts
aquarat: It complains because the return of the mock implementation function is incorrect. I'll look into trying to remove them.

packages/airnode-utilities/src/caching/index.ts
aquarat: Sure thing :)

aquarat:
Done a few more updates, hopefully it is acceptable :P

aquarat:
Actually I will split up `nodeModules` into `nodeModules` and `extraModules`.

packages/airnode-node/src/api/unsafe-evaluate.ts
Siegrift: I fear it will be confusing and harder to document and I do not see an use for using Airnode packages. If you this this is improtant we can keep it, but otherwise I'd just pass `ethers` and `axios`.

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -0,0 +1,207 @@
...
+import { logger } from '../logging';
+
+export const CACHE_BASE_PATH = `/tmp/airnode-cache`;
+
+/**
+ * Lambda, when last tested circa November 2021, persists /tmp contents for 2.5 hours.
+ */
+export const CACHE_MAX_FILESYSTEM_AGE_MINUTES = 60;
+
+/**
```
Siegrift, line 12: 🙏

packages/airnode-node/src/constants.test.ts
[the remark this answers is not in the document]
Siegrift: Btw. your solution handles timeouts well, but I think it would be nice to add a test which tests that. Something like:

```js
const preProcessingSpecifications = [
        {
          environment: 'Node 14' as const,
          value: 'const output = {...input, from: `ETH`}; while(true);',
          timeoutMs: 5_000,
        },
        ...
```

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -0,0 +1,207 @@
...
+ * Get all keys from the cache that start with startingKey.
+ *
+ * @param prefix should take the form of `blockedRequestId-` and the full key could be `blockedRequestId-0x0000000...`
+ */
+const getKeys = (prefix?: string): string[] => {
+  try {
+    const keys = readdirSync(CACHE_BASE_PATH);
+    if (prefix) {
+      return keys.filter((key) => key.indexOf(prefix) === 0);
```
Siegrift, line 51: Why not `.startsWith`?

Siegrift (approved):
Code LGTM 👍 (but I haven't tested manually). 

I do not see any more issues just by looking at the code and I believe this was (is) tested by you on the issue you work on.

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -0,0 +1,207 @@
...
+        return fileDetails;
+      });
+
+    const remainingFiles = stattedFiles.filter((file) => !deletedFiles.includes(file));
+
+    if (remainingFiles.length < CACHE_MAX_FILES) {
+      return;
+    }
+
+    // Delete oldest cache entries first
```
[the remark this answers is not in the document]
aquarat: We would have to sweep on every key addition which would slow things down. We have a lot of overhead @ 20k vs 130k files. It's not "correct" but it is highly unlikely, probably impossible on Lambda, for the cache file count to jump by 110k files in one cycle (Airnode would not be able to process that many requests on Lambda).

packages/airnode-utilities/src/caching/index.ts
aquarat: uh, because I don't know TS/JS that well ;)
My use of `indexOf` comes from Java bad habits :D 
I'll swap it out shortly.

aquarat:
@Siegrift : Regarding
> Btw. your solution handles timeouts well, but I think it would be nice to add a test which tests that. Something like:
and
> Ignore this. We set 30 seconds timeout and postprocessing can only take 10 seconds. So in worst case the API call has only 20 seconds to finish, but that is still all right.

I assume you mean I should ignore the first comment? I think the timeouts are robust and they definitely work in practice.

packages/airnode-node/src/api/unsafe-evaluate.ts
aquarat: It's difficult to know what's useful for the future.
At minimum anything can be achieved given the built-in node modules (like I re-implemented the `axios` requirement using built-in `https` in the target code snippet). Everything outside of the built in modules is convenience.
I'll remove the `airnode-*` require/imports :)

Siegrift:
> I assume you mean I should ignore the first comment?

Yeah.

aquarat:
Thanks again for your input and patience @Siegrift :)

PR #1027, opened 2022-04-25
Save log index

vponline:
This adds request sorting by `logIndex`.


packages/airnode-node/src/requests/sorting.ts
```diff
@@ -5,8 +5,9 @@ export function sortRequests<T>(requests: Request<T>[]): Request<T>[] {
   // In order to keep consistency between runs, requests are sorted by the following criteria:
   //
   //   1. Block number (ascending)
-  //   2. Transaction hash (ascending)
-  return orderBy(requests, ['metadata.blockNumber', 'metadata.transactionHash']);
+  //   2. Log index (ascending)
+  //   3. Transaction hash (ascending)
```
Siegrift, line 9: We do not need transaction hash now, do we?

PR #1037, opened 2022-04-26
Remove airnode address from template verification

vponline:
This removes airnode address as an input for templateId derivation.


Siegrift (approved):
👍 LGTM.

packages/airnode-node/src/evm/templates/template-verification.ts
```diff
@@ -8,7 +8,6 @@ interface ValidatedField {
 
 function getTemplateIdValidationFields(template: ApiCallTemplateWithoutId): ValidatedField[] {
   return [
-    { type: 'address', value: template.airnodeAddress },
```
amarthadan, line 11: I think @acenolaza [is correct](https://api3dao.atlassian.net/browse/AN-703?focusedCommentId=15321). The `getExpectedTemplateId` function is used in `api/index.ts` (https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/api/index.ts#L149) both for signed data gateway requests but also regular Airnode RRP requests. That's why you had to change tests everywhere, not just the ones for `callApi`. I think this change breaks the Airnode RRP. We need two different template ID derivation functions, one for Airnode RRP (v0 contracts) and one for signed data gateway (v1 contracts).

amarthadan (changes requested):

packages/airnode-node/src/evm/templates/template-verification.ts
vponline: Do you prefer to have two separate functions, or we could pass the request `type` (regular/http-signed-gateway) into `getExpectedTemplateId` and return based on that?

amarthadan: I think two separate functions will result in a nicer code.

amarthadan (approved):
LGTM :+1:

PR #1038, opened 2022-04-26
Remove dependabot changelog URLs causing failed CI

dcroote:
Docs CI has been failing repeatedly because of these changelog links.


Siegrift (approved):
🙏 LGTM

PR #1043, opened 2022-04-27
Add gasPriceMultiplier for legacy gas prices

vponline:
This adds a `gasPriceMultiplier` option to config to improve transactions for legacy providers reporting slow numbers.


packages/airnode-utilities/src/evm/gas-prices/gas-prices.ts
```diff
@@ -14,6 +14,9 @@ export const getGasLimit = (fulfillmentGasLimit?: number) => {
 export const parsePriorityFee = ({ value, unit }: PriorityFee) =>
   ethers.utils.parseUnits(value.toString(), unit ?? 'wei');
 
+export const multiplyGasPrice = (gasPrice: BigNumber, gasPriceMultiplier: number) =>
+  gasPrice.mul(Math.round(gasPriceMultiplier * 100)).div(BigNumber.from(100));
```
vponline, line 18: Added **Math.round** here to handle weird JS floating point math.

Also, two decimal point accuracy should be enough, right?

aquarat: Sorry about the delay in reviewing this. Sounds good @ two decimal points.

aquarat (approved):
LGTM, neat and succint :tada:

packages/airnode-utilities/src/evm/gas-prices/gas-prices.ts
acenolaza: Nit: I wonder if there's any benefit from making `gasPriceMultiplier` a `BigNumber` before using it in the calculation 🤔 Maybe for consistency it would be a good idea because `baseFeeMultiplier` is turned into a `BigNumber` in line 65

vponline: @acenolaza Thanks for checking this out. That makes perfect sense, I'll make a fix PR for this.

PR #1044, opened 2022-04-27
Fix condition for enabling GCP API gateway service

amarthadan:
This bug explains why all of a sudden we needed to enable the API gateway service manually on GCP.
Related to [AN-698](https://api3dao.atlassian.net/browse/AN-698) and [AN-700](https://api3dao.atlassian.net/browse/AN-700). Not sure it's the only issue though.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "dev:invoke": "(cd packages/airnode-node && yarn run dev:invoke)",
     "dev:list": "(cd packages/airnode-operation && yarn run dev:list)",
     "dev:stop": "(cd packages/airnode-operation && yarn run dev:stop)",
-    "docker": "yarn docker:build:artifacts && yarn docker:build:deployer && yarn docker:build:client && yarn docker:build:admin",
+    "docker:build": "yarn docker:build:artifacts && yarn docker:build:deployer && yarn docker:build:client && yarn docker:build:admin",
     "docker:build:admin": "docker build --tag api3/airnode-admin:latest --file packages/airnode-admin/docker/Dockerfile .",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -44,7 +44,7 @@
     "dev:invoke": "(cd packages/airnode-node && yarn run dev:invoke)",
     "dev:list": "(cd packages/airnode-operation && yarn run dev:list)",
     "dev:stop": "(cd packages/airnode-operation && yarn run dev:stop)",
-    "docker": "yarn docker:build:artifacts && yarn docker:build:deployer && yarn docker:build:client && yarn docker:build:admin",
+    "docker:build": "yarn docker:build:artifacts && yarn docker:build:deployer && yarn docker:build:client && yarn docker:build:admin",
```
amarthadan, line 47: Not sure why we call it just `docker` and all the "subtasks" are called `docker:build:*` so I changed it.

packages/airnode-deployer/terraform/airnode/gcp/main.tf
```diff
@@ -71,7 +71,7 @@ module "startCoordinator" {
 }
 
 resource "google_project_service" "apigateway_api" {
-  count = var.http_api_key == null || var.http_signed_data_api_key == null ? 0 : 1
+  count = var.http_api_key == null && var.http_signed_data_api_key == null ? 0 : 1
```
Siegrift, line 74: 🙏

Siegrift (approved):
👍 LGTM (but I didn't test this).

We should also patch this to the other versions and also update docs (they now say it needs to be enabled).

PR #1047, opened 2022-04-28
Remove v1 validator

vponline:
This removes v1 validator.


.husky/pre-commit
```diff
@@ -1,3 +1,2 @@
-yarn --cwd packages/airnode-validator/ run docs
 yarn prettier --write "./packages/airnode-validator/**/*.{md,json}" --loglevel silent
```
Siegrift, line 1: You can also remove this one (it was needed only to format the generated docs).

packages/airnode-validator/README.md
```diff
@@ -15,45 +15,15 @@ Validator is currently an internal dependency of Airnode and not intended to be
 You can clone and build the Airnode monorepo then run the validator as a yarn script from inside the
 `packages/airnode-validator` directory.
 
+<!-- TODO: fix commands for v2 validator cli -->
+
 ```sh
```
Siegrift, line 20: Feel free to remove this section below as well. The CLI will not use specs file (the validation will be based on the zod schema).

Siegrift (changes requested):
👍 LGTM, but we should also remove the unnecessary files in `example.specs.json`.

(later we can move the used ones should be later moved to some `test/fixture` directory).

Siegrift (approved):
👍 LGTM. I tried to check if there is anything else missing but this seems to be all. (Even if there is something we can remove it along the way).

Please, check docs whether there is something that needs to be removed. I don't think so, because I've removed validator from 0.6 docs but maybe I forgot something 😄

PR #1049, opened 2022-04-29
Create validator CLI

Siegrift:
https://api3dao.atlassian.net/browse/AN-608

The first PR implements the validator CLI and writes a few unit and e2e tests.
There is also one unrelated commit which builds the admin package (I didn't want to create a new PR for that).


packages/airnode-validator/src/cli/cli.ts
```diff
@@ -0,0 +1,63 @@
...
+export const fail = (s: string) => ora(s).fail();
+
+const examples = [
+  '--config pathTo/config.json --secrets pathTo/secrets.env',
+  '-c pathTo/config.json -s pathTo/secrets.env',
+];
+
+export const validateConfiguration = (configPath: string, secretsPath: string) => {
+  const goRawConfig = goSync(() => readFileSync(path.resolve(configPath), 'utf-8'));
+  if (!goRawConfig.success) return fail(`Unable to read config file at "${configPath}". Reason: ${goRawConfig.error}`);
```
amarthadan, line 20: None of these return statements will set an exit code and the CLI will finish with a 0 exit code.
Screenshot_20220502_110631

packages/airnode-validator/test/validator.feature.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  it('validates valid configuration', () => {
+    const args = [
+      `--config ${join(__dirname, './fixtures/valid-config.json')}`,
+      `--secrets ${join(__dirname, './fixtures/valid-secrets.env')}`,
+    ];
+
+    const output = runValidator(args).stderr.toString();
+
+    expect(output).toEqual('✔ The configuration is valid\n');
```
amarthadan, line 29: I would prefer not to have weird unicode characters in the codebase. They are weirdly displayed if your font in the editor doesn't support them and I don't think it brings much value in this case.

amarthadan (changes requested):

packages/airnode-admin/test/e2e/cli.feature.ts
```diff
@@ -66,6 +66,8 @@ describe('CLI', () => {
   beforeAll(() => {
     jest.setTimeout(45_000);
 
+    execSync('yarn build');
```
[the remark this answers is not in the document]
Siegrift: But when you modify the source files and run the e2e tests it will use the outdated build. I think it's better to just build it as it doesn't take too long.

packages/airnode-validator/test/validator.feature.ts
Siegrift: I think I had problems with these tests so I think I've updated it in the later PR to just `expect.stringContaining`. Will port these changes here as well.

packages/airnode-admin/test/e2e/cli.feature.ts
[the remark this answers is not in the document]
amarthadan: I don't agree as this one E2E test would behave differently from the rest. I think expecting a re-build for E2E tests is ok.

andreogle (approved):
LGTM 🥇

packages/airnode-admin/test/e2e/cli.feature.ts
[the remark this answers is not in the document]
Siegrift: Reverted.

packages/airnode-validator/src/cli/cli.ts
Siegrift: Done.

packages/airnode-validator/test/validator.feature.ts
Siegrift: Done.

amarthadan (approved):

PR #1050, opened 2022-04-30
Document how to build validator from source

Siegrift:
https://api3dao.atlassian.net/browse/AN-609

I've also added inline documentation, because it is nice to see know to pass to the functions directly in the IDE (especially in packages using the validator as dependency).


packages/airnode-validator/README.md
```diff
@@ -1,16 +1,13 @@
...
-The validator is used by the Docker Images to validate the configuration files you provide when deploying an Airnode.
-You can also use the validator to check the configuration files for correct formatting and other issues while creating
-them.
+> API and CLI for validating Airnode configuration files
 
-## Usage
+## User documentation
 
-<!-- TODO: Link to validator docs when available -->
+<!-- TODO: Add correct link to validator -->
```
amarthadan, line 7: I don't understand. What is this TODO for?

amarthadan (approved):

packages/airnode-validator/README.md
Siegrift: For the text below:

```
You can find documentation on how to use the deployer in the [validator docs](https://docs.api3.org/airnode/v0.7/).
```

but I now realized it says "deployer" instead of validator.

amarthadan: But there is already a link. What am I missing? :smile:

Siegrift: It doesn't point to validator, but to the root documentation.

PR #1051, opened 2022-04-30
Improve validator bundling

Siegrift:
https://api3dao.atlassian.net/browse/AN-622

Added build targets for non cjs targets. This makes it more optimal to load the validator inside the browser environment (although modern bundles would probably handle cjs as well). 

I've tested this using a simple FE application and validator published locally (using verdaccio). I've also made sure the source maps works.

### DIFF
--- a/packages/airnode-ois/src/index.ts
+++ b/packages/airnode-ois/src/index.ts
@@ -1,12 +1,14 @@
-import { oisSchema } from '@api3/airnode-validator';
+import { ois as oisModule } from '@api3/airnode-validator';
 import { OIS } from './types';
 
-export * from '@api3/airnode-validator/dist/cjs/src/ois';
 export * from './types';
 export * from './constants';
 
 // NOTE: The main purpose of this function is to make sure that the validator schema inferred from Zod (used internally
 // in validator) is compatible with our manually defined OIS types.
 export function parseOIS(ois: unknown): OIS {
-  return oisSchema.parse(ois);
+  return oisModule.oisSchema.parse(ois);
 }
+
+// Exports OIS schemas
+export { ois } from '@api3/airnode-validator';
--- a/packages/airnode-ois/src/index.ts
+++ b/packages/airnode-ois/src/index.ts
@@ -1,7 +1,7 @@
 import { oisSchema } from '@api3/airnode-validator';
 import { OIS } from './types';
 
-export * from '@api3/airnode-validator/dist/src/ois';
+export * from '@api3/airnode-validator/dist/cjs/src/ois';
 export * from './types';
 export * from './constants';
 
--- a/packages/airnode-validator/package.json
+++ b/packages/airnode-validator/package.json
@@ -3,19 +3,24 @@
   "license": "MIT",
   "version": "0.6.0",
   "private": false,
-  "main": "./dist/src/index.js",
-  "types": "./dist/src/index",
+  "main": "./dist/cjs/src/index.js",
+  "module": "./dist/esm/src/index.js",
+  "browser": "./dist/es6/src/index.js",
   "bin": {
     "airnode-validator": "./dist/bin/validator.js"
   },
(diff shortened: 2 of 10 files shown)


packages/airnode-validator/package.json
```diff
@@ -3,19 +3,24 @@
   "license": "MIT",
   "version": "0.6.0",
   "private": false,
-  "main": "./dist/src/index.js",
-  "types": "./dist/src/index",
```
Siegrift, line 7: I had to remove the "types" because it wasn't working. I've checked that the typings are working in a test project.

packages/airnode-validator/package.json
```diff
@@ -3,19 +3,24 @@
...
+  "main": "./dist/cjs/src/index.js",
+  "module": "./dist/esm/src/index.js",
+  "browser": "./dist/es6/src/index.js",
   "bin": {
     "airnode-validator": "./dist/bin/validator.js"
   },
   "files": [
-    "dist"
+    "dist",
+    "src"
```
Siegrift, line 14: We need "src" as well in order for source maps to work.

packages/airnode-validator/package.json
```diff
@@ -3,19 +3,24 @@
...
   ],
   "scripts": {
-    "build": "yarn run clean && yarn run compile && copyfiles templates/**/** dist/ && copyfiles conversions/** dist/",
+    "build": "yarn run clean && yarn run compile",
     "clean": "rimraf -rf ./dist *.tgz",
     "cli": "ts-node bin/validator.ts",
-    "compile": "tsc -p tsconfig.build.json",
+    "compile": "yarn compile:cjs && yarn compile:es6 && yarn compile:esm",
```
Siegrift, line 20: These are "inspired" by https://github.com/api3dao/promise-utils/blob/main/package.json#L17

packages/airnode-validator/package.json
```diff
@@ -32,8 +37,6 @@
   },
   "devDependencies": {
     "@types/yargs": "^17.0.2",
-    "copyfiles": "^2.4.1",
```
Siegrift, line 35: I've removed a few deps not needed after v1 cleanup.

amarthadan (commented):
Just to make sure, we'll be still able to use
```ts
import { parseConfigWithSecrets } from '@api3/airnode-validator';
```
and not something like
```ts
import { parseConfigWithSecrets } from '@api3/airnode-validator/dist/cjs/...';
```

Siegrift:
Yes. We will for all things that are exported in via the main `index.js` file

amarthadan (approved):

andreogle (approved):

packages/airnode-ois/src/index.ts
```diff
@@ -1,7 +1,7 @@
 import { oisSchema } from '@api3/airnode-validator';
 import { OIS } from './types';
 
-export * from '@api3/airnode-validator/dist/src/ois';
+export * from '@api3/airnode-validator/dist/cjs/src/ois';
```
Siegrift, line 4: @amarthadan WDYT? Shall we export all schemas / types from the `index.ts` to avoid this import? (I think we should, currently, only the API part is exported).

amarthadan: Yes, I think we should.

Siegrift: See: https://github.com/api3dao/airnode/pull/1051/commits/4611cd5806be654db574f0d8fef8d025b12da19a

I've used named exports for the schemas, since the important validator functions are all implemented in the `api` file. This also means, that the consumers who import the schemas from validator have to use `import {ois} from '@api3/airnode-validator'` and use `ois.someScheme` but I don't think this is that big of a deal.

Siegrift: Discussed and agreed upon on slack.

PR #1052, opened 2022-04-30
Enforce node version and create script for bumping them

Siegrift:
https://api3dao.atlassian.net/browse/AN-623
and
https://api3dao.atlassian.net/browse/AN-611

As part of this PR I've:
* enforced the `nodeVersion` number to match the package.json version. This version must be matched to the patch. We discussed that [here](https://api3workspace.slack.com/archives/C02AYRX8D89/p1634306006050500?thread_ts=1634303815.050000&cid=C02AYRX8D89)
* added back the tests to make sure the example files are always up to date. See: https://github.com/api3dao/airnode/blob/532f4f58daaa96d36987488dc7c8958d975ba21d/packages/airnode-validator/test/validator.test.ts#L30 (we have removed these tests in https://github.com/api3dao/airnode/pull/997/files#diff-dfd5b4cc8bd363dd247647c23ef5f9b66011e65e679ec928ebef064bc008d931).
* created a script which updates the `nodeVersion` for all example and test configuration files. This is "best effort" heuristic, but I don't think there is a better solution.
* updated the `nodeVersion` in order for tests to pass, refactored some test files along the way.

See the command to create a new version and bump the examples:
image

TODO: Update the versioning document


packages/airnode-deployer/src/cli/commands.test.ts
```diff
@@ -0,0 +1,124 @@
...
+import { mockReadFileSync } from '../../test/mock-utils';
+import { readFileSync } from 'fs';
+import { deploy, remove, removeWithReceipt } from './commands';
+import { version as packageVersion } from '../../package.json';
+import { Receipt } from '../types';
+import * as logger from '../utils/logger';
+
+const readExampleConfig = () => JSON.parse(readFileSync(join(__dirname, '../../config/config.example.json'), 'utf-8'));
+
+jest.mock('../infrastructure', () => ({
```
Siegrift, line 11: FYI: There are some non trivial things happening during mocking. 

These mocks are hoisted before the imports when the tests are run. However, if you use:
```js
const mockVariable = jest.fn()
jest.mock('../module', () => ({
    val: mockVariable
}))
```

this won't work unless you explicitely move the imports below. The variables are hoisted as well (the `const mockVariable` part), but the initialization seems not. 

This way it works a bit better. All imports are written first, which feels natural and inside code you get the mocked functions to make assertions.

packages/airnode-validator/src/config/config.ts
```diff
@@ -106,8 +107,15 @@ export const nodeSettingsSchema = z.object({
   cloudProvider: localOrCloudProviderSchema,
   logFormat: logFormatSchema,
   logLevel: logLevelSchema,
-  // TODO: This must match validator version
-  nodeVersion: z.string(),
+  nodeVersion: z.string().superRefine((version, ctx) => {
```
Siegrift, line 110: FYI: We should be using `superRefines` in favor of `refine` because the error message from `refine` is too generic.

packages/airnode-validator/src/receipt/receipt.ts
```diff
@@ -11,8 +12,15 @@ export const airnodeWalletSchema = z.object({
 export type AirnodeWallet = SchemaType<typeof airnodeWalletSchema>;
 
 export const deploymentSchema = z.object({
-  // TODO: This must match the version of validator
-  nodeVersion: z.string(),
+  nodeVersion: z.string().superRefine((version, ctx) => {
```
Siegrift, line 15: This is the same refinement check as in the `config.ts` validation, but I didn't want to abstract this to a "common" file just yet...

scripts/update-config-fixtures.ts
```diff
@@ -0,0 +1,36 @@
+import { readFileSync, writeFileSync } from 'fs';
+import fg from 'fast-glob';
+import { version as packageVersion } from '../packages/airnode-node/package.json';
+
+const configPatterns = ['./packages/**/*config.example.json', './packages/**/*config.valid.json'];
```
Siegrift, line 5: I found it a bit awkward to use `config.example.json` names for fixtures inside test files. After a few minutes think I settled on using `config.valid.json` pattern. I am open to better alternatives.

packages/airnode-deployer/config/config.example.json
```diff
@@ -44,7 +44,7 @@
     },
     "logFormat": "plain",
     "logLevel": "INFO",
-    "nodeVersion": "0.6.0",
+    "nodeVersion": "0.7.0",
```
Siegrift: Ignore these version updates, they are reverted in one of the next commits.

package.json
```diff
@@ -31,6 +31,7 @@
     "changeset": "changeset",
     "changeset:check": "changeset status --since=origin/master",
     "changeset:empty": "changeset --empty",
+    "changeset:new-version": "yarn changeset version && yarn ts-node scripts/update-config-fixtures.ts",
```
amarthadan, line 34: So, this needs to be called before doing the release, right? Have you updated the release docs?

packages/airnode-deployer/src/utils/validation.test.ts
```diff
@@ -1,21 +1,34 @@
...
 import { loadConfig, loadTrustedConfig } from '@api3/airnode-node';
-import * as fixtures from '../../test/fixtures';
+import dotenv from 'dotenv';
+import { version as packageVersion } from '../../package.json';
 
 describe('config validation', () => {
-  it('loads the config without validation', () => {
-    const config = fixtures.buildConfig();
-    mockReadFileSync('config.json', JSON.stringify(config));
+  const exampleConfigPath = join(__dirname, '../../config/config.example.json');
```
amarthadan, line 9: I agree that testing `config.example.json` is a good idea but I think that testing whatever comes from `fixtures.buildConfig()` is a valid configuration is also a good idea. That fixture is used all over the place and might easily become outdated. This would prevent that.

packages/airnode-deployer/test/fixtures/index.ts
```diff
@@ -1 +0,0 @@
-export * from './config';
```
amarthadan, line 1: I didn't even know we had these in deployer. Where did they come from?

scripts/update-config-fixtures.ts
```diff
@@ -0,0 +1,36 @@
+import { readFileSync, writeFileSync } from 'fs';
```
amarthadan, line 1: My concern here is that we match and rewrite some files we shouldn't. TBH I imagined this script to contain a list of files we know need to be changed without any search or anything. I guess this is fine but I would at least add a warning saying to check git that all the modified files should really be modified. Or print a list of modified files.

amarthadan (changes requested):
Btw this PR is quite ugly with changes all over the place that are not really related to the task at hand...

package.json
Siegrift: It's a TODO

packages/airnode-deployer/src/utils/validation.test.ts
Siegrift: > This would prevent that.

What would prevent it?

packages/airnode-deployer/test/fixtures/index.ts
Siegrift: I think they were introduced when unit tests were added.

scripts/update-config-fixtures.ts
Siegrift: It prints the modified files (see the PR description)

packages/airnode-deployer/src/utils/validation.test.ts
amarthadan: Testing that the output of `fixtures.buildConfig()` is valid.

Siegrift: I think those should import and use the `version` from `package.json`. I'll try to update them.

packages/airnode-node/test/fixtures/config/node-settings.ts
```diff
@@ -1,4 +1,5 @@
 import { NodeSettings } from '../../../src/config/types';
+import { version as packageVersion } from '../../../package.json';
```
amarthadan, line 2: Nit: We have a function for returning the [package version](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/version.ts#L3). Maybe you can use it for `airnode-node` code. In case the implementation changes (even though I doubt it).

amarthadan (approved):

PR #1067, opened 2022-05-02
Release v0.6.2

aquarat:
This is a patch release bumping the package versions to v0.6.2
v0.6.1 was broken and has been scheduled for "unpublishing".


Siegrift:
> v0.6.1 was broken

What was wrong?

aquarat:
> > v0.6.1 was broken
> 
> What was wrong?

A combination of factors but the main one was that the name of the branch was ambiguous due to an existing tag with the same - so I thought I was working on branch v0.6 but actually I had checked out a tag. But now I know how to do the releases :D

Siegrift:
> so I thought I was working on branch v0.6 but actually I had checked out a tag

You likely had those tags locally, because I've removed them recently (https://api3workspace.slack.com/archives/C02AYRX8D89/p1651137018466549). Sorry, I forgot to mention to remove them locally as well. 

Not sure what happened now, but those branches (`v0.5` and `v0.6`) are now removed and there are `v0.5.1` and `v0.6.1` which is wrong.

aquarat:
> You likely had those tags locally, because I've removed them recently (https://api3workspace.slack.com/archives/C02AYRX8D89/p1651137018466549). Sorry, I forgot to mention to remove them locally as well.
> 
> Not sure what happened now, but those branches (`v0.5` and `v0.6`) are now removed and there are `v0.5.1` and `v0.6.1` which is wrong.

I renamed them... is that a problem? We can change them back but it'll close these PRs.

Siegrift:
We should have only "v0.x" branch and have tags named "v0.x.y" this is because we do not want to create a new branch for every release, but instead have a branch for each major version. However, for now we are treating the "0.x" versions as majors (but the point is that there should be one branch "0.x" which will contain multiple tagged patches).

amarthadan:
What @Siegrift is saying. Also, this PR is open against `master` which makes no sense.

aquarat:
> What @Siegrift is saying. Also, this PR is open against `master` which makes no sense.

Okay, thank you both, in retrospect that's all obvious. I'll close these PRs and rename the branches. Thanks again :)

PR #1073, opened 2022-05-05
Convert gasPriceMultiplier to BigNumber

vponline:
Small fix to change `gasPriceMultiplier` into a **BigNumber** before using in calculations.


packages/airnode-utilities/src/evm/gas-prices/gas-prices.ts
```diff
@@ -15,7 +15,7 @@ export const parsePriorityFee = ({ value, unit }: PriorityFee) =>
   ethers.utils.parseUnits(value.toString(), unit ?? 'wei');
 
 export const multiplyGasPrice = (gasPrice: BigNumber, gasPriceMultiplier: number) =>
-  gasPrice.mul(Math.round(gasPriceMultiplier * 100)).div(BigNumber.from(100));
+  gasPrice.mul(BigNumber.from(Math.round(gasPriceMultiplier * 100))).div(BigNumber.from(100));
```
Siegrift, line 18: Why is this needed? Afaik BigNumber accepts regular numbers as well.

vponline: This is mostly for consistency (pointed out in #1043) as all the other numbers are converted to BigNumber in gas-prices (and mostly elsewhere in Airnode I believe).

acenolaza (approved):
Thanks 😁

PR #1075, opened 2022-05-06
Fix type of submitted transactions

dcroote:
While the calculations of gas / fees for legacy and eip1559 transactions are correct, legacy transactions are actually being submitted as eip1559 because of the missing `type` property in [ethers transactions](https://docs.ethers.io/v5/api/providers/types/#providers-TransactionRequest).

I've confirmed this is true by deploying an airnode with `legacy` as the `txType` in `config.json` and while the gas (base, max, max priority) reflect legacy pricing, the tx is still `Txn Type: 2 (EIP-1559)` [as shown in this fulfill tx](https://rinkeby.etherscan.io/tx/0x9898d62aae25e624228453b15990365a2ac2a9c83d3a251aa6dfd1bb4a8c6e7a) (click "Click to see More" in the bottom left then look at Gas Fees and Others). Implementing `type` of `0` for legacy as in this PR results in the `Txn Type` being `0 (Legacy)` [as shown here](https://rinkeby.etherscan.io/tx/0x1c0b1807f8a4f9a45d439e03b282d994835253334ac23fa5784805ebdbf30adb).


packages/airnode-node/src/types.ts
```diff
@@ -201,13 +201,6 @@ export interface TransactionOptions {
   readonly provider: ethers.providers.JsonRpcProvider;
 }
 
-export interface GasTarget {
-  readonly maxPriorityFeePerGas?: BigNumber;
-  readonly maxFeePerGas?: BigNumber;
-  readonly gasPrice?: BigNumber;
-  readonly gasLimit?: BigNumber;
-}
-
```
vponline, line 210: Glad to see this gone (and imported from utilities) :smile:

vponline (approved):
Nice investigation, LGTM :medal_sports:

Siegrift (approved):
👍 LGTM.

PR #1089, opened 2022-05-15
Setup TS project references, make tsconfig consistent

Siegrift:
## What does this do?

This PR simplifies development by references the TS packages. This means when you navigate to a definition from a different monorepo package (ctrl+click), the IDE will take you to the source instead of showing the built `d.ts` file from the `node_modules`. Also, it allows you to see the cross package errors instantaneously - e.g. you remove a function from airnode-utilities and you will see an error in the airnode-node package.

While implementing the change I had to modify all of the tsconfig files when enabling the `composite` flag ([see this](https://www.typescriptlang.org/docs/handbook/project-references.html#composite)). Since, I was doing this I took the time and tried to make the tsconfigs more simple and consistent. It works like this:
1. There is a single `tsconfig.json`, which contains common settings and there is no tsconfig.build.json (because we don't need one).
2. Every package has it's own `tsconfig.json` which inherits the root one and includes all TS source files (including tests).
3. Every package has it's own `tsconfig.build.json` which includes the `tsconfig.json` and excludes the non production files.

## Reading material

I've mostly followed the [TS documentation](https://www.typescriptlang.org/docs/handbook/project-references.html) and [this blog post](https://www.baltuta.eu/posts/typescript-lerna-monorepo-the-setup).


packages/airnode-abi/package.json
```diff
@@ -10,8 +10,8 @@
   ],
   "scripts": {
     "build": "yarn run clean && yarn run compile",
-    "clean": "rimraf -rf ./dist *.tgz",
-    "compile": "tsc -p tsconfig.build.json",
+    "clean": "rm -f *.tsbuildinfo && rimraf -rf ./dist *.tgz",
```
andreogle, line 13: I think this will break on windows?

```suggestion
    "clean": "rimraf -rf ./dist *.tgz *.tsbuildinfo",
```

packages/airnode-abi/tsconfig.json
```diff
@@ -2,8 +2,8 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "baseUrl": "./",
-    "outDir": "dist"
+    "outDir": "./dist",
```
andreogle, line 5: Should we not move all of these options to the base file to make it consistent for all packages?

packages/airnode-adapter/package.json
```diff
@@ -3,15 +3,15 @@
...
+  "main": "dist/src/index",
+  "types": "dist/src/index",
   "files": [
     "dist"
   ],
   "scripts": {
     "build": "yarn run clean && yarn run compile",
-    "clean": "rimraf -rf ./dist *.tgz",
-    "compile": "hardhat compile && tsc -p tsconfig.build.json",
+    "clean": "rm -f *.tsbuildinfo && rimraf -rf ./dist *.tgz",
```
andreogle, line 13:
```suggestion
    "clean": "rimraf -rf ./dist *.tgz *.tsbuildinfo",
```

packages/airnode-adapter/tsconfig.build.json
```diff
@@ -1,10 +1,7 @@
...
-  "extends": "../../tsconfig.build.json",
-  "compilerOptions": {
-    "baseUrl": "./",
-    "outDir": "./dist",
-    "rootDir": "src"
-  },
-  "include": ["src/**/*"],
-  "exclude": ["src/**/*.test.ts"]
+  "extends": "./tsconfig.json",
+  // References are not inhereted and need to be listed explicitly. See:
```
andreogle, line 3:
```suggestion
  // References are not inherited and need to be listed explicitly. See:
```

packages/airnode-admin/package.json
```diff
@@ -12,9 +12,9 @@
   ],
   "scripts": {
     "build": "yarn run clean && yarn run compile",
-    "clean": "rimraf -rf ./dist *.tgz",
+    "clean": "rm -f *.tsbuildinfo && rimraf -rf ./dist *.tgz",
```
andreogle, line 15:
```suggestion
    "clean": "rimraf -rf ./dist *.tgz *.tsbuildinfo",
```

packages/airnode-admin/tsconfig.build.json
```diff
@@ -1,9 +1,7 @@
 {
-  "extends": "../../tsconfig.build.json",
-  "compilerOptions": {
-    "baseUrl": "./",
-    "outDir": "./dist"
-  },
-  "include": ["./bin/**/*", "./src/**/*"],
-  "exclude": ["src/**/*.test.ts"]
+  "extends": "./tsconfig.json",
+  // References are not inhereted and need to be listed explicitly. See:
```
andreogle, line 3:
```suggestion
  // References are not inherited and need to be listed explicitly. See:
```

packages/airnode-deployer/package.json
```diff
@@ -11,10 +11,10 @@
     "dist"
   ],
   "scripts": {
-    "build": "yarn clean && yarn compile && yarn webpack && yarn copy:webpack && yarn copy:terraform",
-    "clean": "rimraf -rf ./dist ./.webpack *.tgz",
+    "build": "yarn run clean && yarn compile && yarn webpack && yarn copy:webpack && yarn copy:terraform",
```
andreogle, line 14: Why add `run` here but not for the others? I don't think it's needed?

packages/airnode-adapter/package.json
andreogle: Applies to all `package.json` files

andreogle (changes requested):

packages/airnode-abi/package.json
Siegrift: Will do. Btw. I am not sure we even support windows development. We [recommend using WSL2](https://github.com/api3dao/airnode/blob/master/README.md?plain=1#L63) for a while now.

packages/airnode-abi/tsconfig.json
Siegrift: Unfortunately, the `outDir` is relative to the path of the tsconfig. This path is different for every monorepo package so there is no other way then listing it explicitely.

packages/airnode-adapter/tsconfig.build.json
Siegrift: Not sure why my grammar corrector didn't catch this...

packages/airnode-deployer/package.json
Siegrift: I removed all `yarn clean` command at some point, but later reverted it via global find replace. Majority of other commands use `yarn run` so that's why. I've reverted it for this change.

andreogle (approved):

amarthadan (approved):
:+1:

Siegrift:
I'll want to do a few more changes to the configuration according to this demo: https://github.com/RyanCavanaugh/project-references-demo but I'll do this as part of a separate PR.

PR #1100, opened 2022-05-17
Paths validation

Siegrift:
https://api3dao.atlassian.net/browse/AN-612

Ensures that interpolated path variables in config.json exist and also that every parameter is used (ensuring bijection).

### DIFF
--- a/packages/airnode-validator/src/ois/ois.ts
+++ b/packages/airnode-validator/src/ois/ois.ts
@@ -143,7 +143,7 @@ const ensurePathParametersExist: ValidatorRefinement<SchemaType<typeof pathsSche
         if (p.in === 'path' && !matches.includes(p.name)) {
           ctx.addIssue({
             code: z.ZodIssueCode.custom,
-            message: `Parameter "${p.name}" is not used in the URL path`,
+            message: `Parameter "${p.name}" is not found in the URL path`,
             path: [rawPath, httpMethod, 'parameters', index],
           });
         }
--- a/packages/airnode-validator/src/ois/ois.ts
+++ b/packages/airnode-validator/src/ois/ois.ts
@@ -1,6 +1,13 @@
 import { z } from 'zod';
 import intersection from 'lodash/intersection';
+import forEach from 'lodash/forEach';
+import trimEnd from 'lodash/trimEnd';
+import trimStart from 'lodash/trimStart';
 import { SchemaType, ValidatorRefinement } from '../types';
+
+function removeBraces(value: string) {
+  return trimEnd(trimStart(value, '{'), '}');
+}
 
 export const paremeterTargetSchema = z.union([
   z.literal('path'),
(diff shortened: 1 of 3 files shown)


packages/airnode-validator/src/ois/ois.ts
```diff
@@ -106,11 +113,50 @@ export const operationSchema = z.object({
...
+          });
+        }
+      }, rawPath);
+
+      // Check that all parameters are used
+      parameters.forEach((p, index) => {
+        if (p.in === 'path' && !matches.includes(p.name)) {
+          ctx.addIssue({
+            code: z.ZodIssueCode.custom,
+            message: `Parameter "${p.name}" is not used in the URL path`,
```
andreogle, line 146: "is not used in" sounds a bit weird to me. Maybe something like:

```suggestion
            message: `Parameter "${p.name}" was not found in the URL path`,
```

packages/airnode-validator/src/ois/ois.ts
```diff
@@ -106,11 +113,50 @@ export const operationSchema = z.object({
...
+      const regex = /\{([^}]+)\}/g;
+      const matches = rawPath.match(regex)?.map(removeBraces) ?? [];
+
+      // Check that all path parameters are defined
+      matches.forEach((match) => {
+        const parameter = parameters.find((p) => p.in === 'path' && p.name === match);
+        if (!parameter) {
+          ctx.addIssue({
+            code: z.ZodIssueCode.custom,
+            message: `Path parameter "${match}" is not found in "parameters"`,
```
andreogle, line 135: Nit: Do we want past or present tense for these kinds of errors? Maybe past since the validation has already completed when the user sees it?

```suggestion
            message: `Path parameter "${match}" was not found in "parameters"`,
```

packages/airnode-validator/src/ois/ois.ts
```diff
@@ -106,11 +113,50 @@ export const operationSchema = z.object({
...
+export const httpStatusCodes = z.union([z.literal('get'), z.literal('post')]);
+
+export const pathSchema = z.record(httpStatusCodes, operationSchema);
+
+const ensurePathParametersExist: ValidatorRefinement<SchemaType<typeof pathsSchema>> = (paths, ctx) => {
+  forEach(paths, (pathData, rawPath) => {
+    forEach(pathData, (paramData, httpMethod) => {
+      const parameters = paramData!.parameters;
+      // Match on anything in the path that is braces
+      // i.e. The path /users/{id}/{action} will match ['{id}', '{action}']
```
andreogle, line 125: These comments are very helpful 👍🏻

andreogle (approved):

packages/airnode-validator/src/ois/ois.ts
Siegrift: Hehe :) , I copied this over from https://github.com/api3dao/airnode/blob/master/packages/airnode-adapter/src/request-building/path-parser.ts#L12 so they are most likely yours :D

packages/airnode-validator/src/ois/ois.ts
Siegrift: In my experience present is more used and it also sounds better. The way I think about it is that the validation happened in the past, but your config is invalid now (in present).

packages/airnode-validator/src/ois/ois.ts
Siegrift: "is not found" sounds good to me.

packages/airnode-validator/src/ois/ois.ts
andreogle: Sure as long as we try to keep it consistent 👌🏻

PR #1101, opened 2022-05-17
Hardcoded config templates

vponline:
This adds support for hardcoded (v0/RRP) templates in `config.json`. TemplateIds are verified and valid templates are not fetched from the chain.


packages/airnode-node/src/evm/templates/template-fetching.ts
```diff
@@ -80,13 +83,43 @@ export async function fetch(
...
+  const validTemplatesInConfig = templateIds.reduce((acc: ApiCallTemplatesById, id) => {
+    const configTemplateMatch = configTemplates[id];
+
+    if (configTemplateMatch) {
+      // Verify templateIds
+      const expectedTemplateIdV0 = getExpectedTemplateIdV0({
+        airnodeAddress: fetchOptions.airnodeRrpAddress,
+        endpointId: configTemplateMatch.endpointId,
+        encodedParameters: configTemplateMatch.encodedParameters,
+      });
```
acenolaza, line 106: This seems like something we must verify during config validation with a `refine` or `superRefine` function (packages/airnode-validator/src/config/config.ts)

packages/airnode-node/src/evm/templates/template-fetching.test.ts
```diff
@@ -18,6 +18,74 @@ describe('fetch (templates)', () => {
...
+        encodedParameters: '0x6874656d706c6174656576616c7565',
+        id: 'templateId-0',
+      },
+      // Invalid templateId
+      '0x38ba0e80224f14d0c654c4ba6e3745fcb7f310fd4f2f80994fe802da013edaxx': {
+        airnodeAddress: '0xD5659F26A72A8D718d1955C42B3AE418edB001e0',
+        endpointId: '0x13dea3311fe0d6b84f4daeab831befbc49e19e6494c41e9e065a09c3c68f43b6',
+        encodedParameters: '0x6874656d706c6174656576616c7565',
+        id: '0x38ba0e80224f14d0c654c4ba6e3745fcb7f310fd4f2f80994fe802da013edaxx',
+      },
```
acenolaza, line 66: If we validate config templates then we might not even need to handle these invalid cases

packages/airnode-node/src/evm/templates/template-fetching.ts
```diff
@@ -102,5 +135,5 @@ export async function fetch(
     return { ...acc, ...result[1] };
   }, {});
 
-  return [templateResponseLogs, templatesById];
+  return [templateResponseLogs, { ...templatesById, ...validTemplatesInConfig }];
```
andreogle, line 138: Nit: these variable names don't really "match" with each other. How about something like:

```ts
{ ...onchainTemplatesById, ...configTemplatesById }
```

packages/airnode-validator/src/config/config.ts
```diff
@@ -14,6 +14,13 @@ export const triggersSchema = z.object({
   httpSignedData: z.array(triggerSchema),
 });
 
+export const templateSchema = z.object({
+  endpointId: z.string(),
```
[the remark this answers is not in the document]
andreogle: Could this be generalized and made re-usable?

packages/airnode-node/src/evm/handlers/initialize-provider.ts
```diff
@@ -75,6 +75,7 @@ export async function initializeProvider(
   const templateFetchOptions = {
     airnodeRrpAddress: state3.contracts.AirnodeRrp,
     provider: state3.provider,
+    configTemplates: state3.config!.templates,
```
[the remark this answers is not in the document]
vponline: Yep this looks good also (I had it inside the template-fetching function [here](https://github.com/api3dao/airnode/pull/1101/files#diff-4e4808a951d4a7768c56d5433b7dc9b5bcc456bf31eb15087f98f2cb074d6acfR96))

packages/airnode-node/src/evm/templates/template-fetching.ts
vponline: Good idea. I will need to add **airnodeAddress** as an additional field as it's needed  to validate V0 templates :+1:

packages/airnode-validator/src/config/config.ts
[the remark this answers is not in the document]
vponline: Nice, will do the same for templateIds and addresses as well

packages/airnode-node/src/config/types.ts
```diff
@@ -15,6 +15,13 @@ export interface Triggers {
   httpSignedData: Trigger[];
 }
 
+export interface Template {
+  endpointId: string;
```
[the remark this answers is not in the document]
vponline: Added :smile:

packages/airnode-node/src/evm/templates/template-fetching.test.ts
vponline: True, I'll add one to the validator instead

packages/airnode-node/src/config/types.ts
```diff
@@ -113,5 +121,6 @@ export interface Config {
   readonly nodeSettings: NodeSettings;
   readonly ois: OIS[];
   readonly triggers: Triggers;
+  readonly templates?: Templates;
```
Siegrift, line 124: I'd prefer this to be an array of templates and the `templateId` to be inside the object. This is more consistent with what we have in `config.json` now and also a bit more explicit, since it is not clear what the record keys should be (from the first look).

packages/airnode-node/src/evm/templates/template-fetching.ts
```diff
@@ -76,17 +79,50 @@ async function fetchTemplateGroup(
   return [[], templatesById];
 }
 
+export const verifyConfigTemplates = (templateIds: string[], fetchOptions: FetchOptions) => {
```
Siegrift, line 82: Why do this check here if we want to do it in validator?

packages/airnode-validator/src/config/config.ts
```diff
@@ -14,14 +15,47 @@ export const triggersSchema = z.object({
...
+        // Verify that config.beacons.<beaconId> is valid
+        // by deriving the hash of the airnode address and templateId
+        const derivedTemplateId = ethers.utils.solidityKeccak256(
+          ['address', 'bytes32', 'bytes'],
+          [template.airnodeAddress, template.endpointId, template.encodedParameters]
+        );
+        if (derivedTemplateId !== templateId) {
+          ctx.addIssue({
+            code: z.ZodIssueCode.custom,
+            message: `Template ID "${templateId}" is invalid`,
```
Siegrift, line 43: You can omit the `${templateId}` since it will be specified in `path`.

Siegrift (changes requested):
👍  Overall LGTM, but I'd change a few things.

packages/airnode-node/src/evm/templates/template-fetching.ts
vponline: The difference is that in validator we're checking against the `airnodeAddress` field from the same object, and here we're checking against the state.contracts.AirnodeRrp address which comes from config.chains[] so there could be a mismatch here?

Does this seem redundant? Otherwise how to identify which template is for which chain (to get the correct AirnodeRrp address for the config validator :thinking:

packages/airnode-node/src/evm/templates/template-fetching.ts
acenolaza: You make a good point. `airnodeAddress` is not currently in config.json 🤔 We do have access to the airnode wallet mnemonic tho

packages/airnode-node/src/config/types.ts
acenolaza: > it is not clear what the record keys should be (from the first look).

For this reason I'm starting to dislike using `Record<keyType, valueType>` type in favor of the old `[usefulKeyName: keyType]: valueType` syntax (can't remember the name 😁 )

packages/airnode-node/src/evm/templates/template-fetching.ts
Siegrift: Mhhh, but why do you check `airnodeRrpAddress`? I didn't notice that, but the rrp address is the address of the AirnodeRrp protocol, not Airnode.

packages/airnode-node/src/config/types.ts
Siegrift: Yeah, but that is still only a dev convenience. If you imagine an user who just looks at the config.json, they see a random keys and have to understand Airnode in depth to know that it's the templateId.

acenolaza: Yes, I definitely see your point but I'm guessing Vekil did it this way to be consistent with [Airkeeper](https://github.com/api3dao/airkeeper/blob/main/config/airkeeper.example.json#L46) and [Airseeker](https://github.com/api3dao/airseeker/blob/main/config/airseeker.example.json#L66) config files right?

packages/airnode-node/src/evm/templates/template-fetching.ts
vponline: :facepalm: Thanks for catching this, I need to instead be deriving the airnode address from **nodeSettings.airnodeWalletMnemonic** (and this can be done in validator so this verification can be removed entirely).

packages/airnode-node/src/config/types.ts
vponline: Yep, but it's a simple conversion to an array if we want (and also now that we can do the whole verification in [validator](https://github.com/api3dao/airnode/pull/1101#discussion_r875852788)) :+1:

packages/airnode-node/src/config/types.ts
```diff
@@ -113,5 +119,6 @@ export interface Config {
   readonly nodeSettings: NodeSettings;
   readonly ois: OIS[];
   readonly triggers: Triggers;
+  readonly templates?: Template[];
```
Siegrift, line 122: Let's make this required. The user can specify `[]` if he doesn't want to hardcode any template.

Siegrift (changes requested):
We dislike optional fields so let's make the `templates` required.

packages/airnode-node/src/config/types.ts
vponline: Nice, I'll keep this in mind :smile:

packages/airnode-node/src/evm/templates/template-fetching.test.ts
```diff
@@ -15,9 +15,61 @@ describe('fetch (templates)', () => {
...
+      '0x38ba0e80224f14d0c654c4ba6e3745fcb7f310fd4f2f80994fe802da013edafe': {
+        airnodeAddress: '0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266',
+        endpointId: '0x13dea3311fe0d6b84f4daeab831befbc49e19e6494c41e9e065a09c3c68f43b6',
+        encodedParameters: '0x6874656d706c6174656576616c7565',
+        id: '0x38ba0e80224f14d0c654c4ba6e3745fcb7f310fd4f2f80994fe802da013edafe',
+      },
+    });
+
+    // Not called with a valid config templateId
+    expect(templatesMock).not.toHaveBeenCalled();
```
Siegrift, line 65: Are these assertions correct? I'd expect:
```
expect(getTemplatesMock).not.toHaveBeenCalledWith('0x38ba0e80224f14d0c654c4ba6e3745fcb7f310fd4f2f80994fe802da013edafe');
```

and
 
```
    expect(getTemplatesMock).toHaveBeenCalledWith([
      'templateId-0',
    ]);
```

Siegrift (approved):
One last comment, but otherwise LGTM 💪

packages/airnode-node/src/evm/templates/template-fetching.test.ts
vponline: I believe they are correct because the `fetchTemplateGroup` function first calls the `getTemplates` contract method:

```ts
const fetchTemplateGroup = () => 
{...
const contractCall = () => airnodeRrp.getTemplates(templateIds);
  const [err, rawTemplates] = await go(contractCall, { retries: 1, timeoutMs: DEFAULT_RETRY_TIMEOUT_MS });
  // If we fail to fetch templates, the linked requests will be discarded and retried
  // on the next run
  if (err || !rawTemplates) {
    const groupLog = logger.pend('ERROR', 'Failed to fetch API call templates', err);

    // If the template group cannot be fetched, fallback to fetching templates individually
    const promises = templateIds.map((id) => fetchTemplate(airnodeRrp, id));
....
```

In the test we **mockResolvedValue** for `getTemplates`:

```ts
const rawTemplates = {
      airnodes: ['airnode-0', '0xD5659F26A72A8D718d1955C42B3AE418edB001e0'],
      endpointIds: ['endpointId-0', '0x13dea3311fe0d6b84f4daeab831befbc49e19e6494c41e9e065a09c3c68f43b6'],
      parameters: ['0x6874656d706c6174656576616c7565', '0x6874656d706c6174656576616c7565'],
    };
    getTemplatesMock.mockResolvedValueOnce(rawTemplates);
```

So as I understand, it should not be fetching templates individually.

Siegrift: I think the comments mostly confused me, but I think this looks better now.

packages/airnode-validator/src/config/config.ts
```diff
@@ -15,14 +16,24 @@ export const triggersSchema = z.object({
   httpSignedData: z.array(triggerSchema),
 });
 
+export const evmAddressSchema = z.string().regex(/^0x[a-fA-F0-9]{40}$/);
+export const evmTemplateIdSchema = z.string().regex(/^0x[a-fA-F0-9]{64}$/);
+export const evmEndpointIdSchema = z.string().regex(/^0x[a-fA-F0-9]{64}$/);
```
acenolaza, line 21: Really like these 3 but maybe the 2 in lines 20 and 21 can be a single hexIdSchema or something with a better name :)

acenolaza (approved):
LGTM 👍🏻

PR #1102, opened 2022-05-17
AN-614: Allowed security scheme must exist in config.json

acenolaza:
[AN-614](https://api3dao.atlassian.net/browse/AN-614): used [zod.superRefine()](https://github.com/colinhacks/zod#superRefine) in order to validate that the security scheme enabled under `ois.apiSpecifications.security` is defined in `ois.apiSpecifications.components`.

### DIFF
--- a/packages/airnode-validator/src/config/config.test.ts
+++ b/packages/airnode-validator/src/config/config.test.ts
@@ -133,14 +133,7 @@ it('fails if a securitySchemeName is enabled and it is of type "apiKey" or "http
       name: 'access_key',
     },
   };
-  const apiCredentials = [
-    {
-      oisTitle: 'Currency Converter API',
-      securitySchemeName: 'Currency Converter Security Scheme',
-      securitySchemeValue: '${SS_CURRENCY_CONVERTER_API_KEY}',
-    },
-  ];
-  const configWithSecuritySchemes = {
+  const invalidConfig = {
     ...config,
     ois: [
       ...config.ois,
@@ -153,19 +146,15 @@ it('fails if a securitySchemeName is enabled and it is of type "apiKey" or "http
         },
       },
     ],
-    apiCredentials,
-  };
-
-  const invalidConfig = {
-    ...configWithSecuritySchemes,
     apiCredentials: [],
   };
+
   expect(() => configSchema.parse(invalidConfig)).toThrow(
     new ZodError([
       {
         code: 'custom',
         message: 'The security scheme is enabled but no credentials are provided in "apiCredentials"',
-        path: ['ois', 1, 'apiSpecifications', 'security', 'Currency Converter Security Scheme'],
+        path: ['ois', 1, 'apiSpecifications', 'security', securitySchemeName],
       },
     ])
   );
--- a/packages/airnode-validator/src/config/config.test.ts
+++ b/packages/airnode-validator/src/config/config.test.ts
@@ -127,7 +127,7 @@ it('fails if ois.apiSpecifications.security.<securitySchemeName> is defined in o
 
   const securitySchemeName = 'Currency Converter Security Scheme';
   const securityScheme = {
-    securitySchemeName: {
+    [securitySchemeName]: {
       in: 'query',
       type: 'apiKey',
       name: 'access_key',
@@ -156,33 +156,19 @@ it('fails if ois.apiSpecifications.security.<securitySchemeName> is defined in o
     apiCredentials,
   };
 
-  const invalidConfigEmtpyApiCredentialValue = {
-    ...configWithSecuritySchemes,
-    apiCredentials: [...config.apiCredentials, { ...apiCredentials[0], securitySchemeValue: '' }],
-  };
-  expect(() => configSchema.parse(invalidConfigEmtpyApiCredentialValue)).toThrow(
-    new ZodError([
-      {
-        code: 'custom',
-        message: 'Security scheme "Currency Converter Security Scheme" is not defined in "components.securitySchemes"',
-        path: ['ois', 1, 'apiSpecifications', 'security', 0],
-      },
-    ])
-  );
-
-  const invalidConfigNoApiCredential = {
+  const invalidConfig = {
     ...configWithSecuritySchemes,
     apiCredentials: [
       ...config.apiCredentials,
       { ...apiCredentials[0], securitySchemeName: apiCredentials[0].securitySchemeName + Date.now().toString() },
     ],
   };
-  expect(() => configSchema.parse(invalidConfigNoApiCredential)).toThrow(
+  expect(() => configSchema.parse(invalidConfig)).toThrow(
     new ZodError([
       {
         code: 'custom',
-        message: 'Security scheme "Currency Converter Security Scheme" is not defined in "components.securitySchemes"',
-        path: ['ois', 1, 'apiSpecifications', 'security', 0],
+        message: 'The security scheme is enabled but no credentials are provided in "apiCredentials"',
+        path: ['ois', 1, 'apiSpecifications', 'security', 'Currency Converter Security Scheme'],
       },
     ])
   );
--- a/packages/airnode-validator/src/config/config.test.ts
+++ b/packages/airnode-validator/src/config/config.test.ts
@@ -120,7 +120,7 @@ describe('nodeSettingsSchema', () => {
   });
 });
 
-it('fails if ois.apiSpecifications.security.<securitySchemeName> is defined in ois.apiSpecifications.components.<securitySchemeName> but apiCredentials.securitySchemeValue is empty string', () => {
+it('fails if a securitySchemeName is enabled and it is of type "apiKey" or "http" but is missing credentials in "apiCredentials"', () => {
   const config: Config = JSON.parse(
     readFileSync(join(__dirname, '../../test/fixtures/interpolated-config.valid.json')).toString()
   );
@@ -158,10 +158,7 @@ it('fails if ois.apiSpecifications.security.<securitySchemeName> is defined in o
 
   const invalidConfig = {
     ...configWithSecuritySchemes,
-    apiCredentials: [
-      ...config.apiCredentials,
-      { ...apiCredentials[0], securitySchemeName: apiCredentials[0].securitySchemeName + Date.now().toString() },
-    ],
+    apiCredentials: [],
   };
   expect(() => configSchema.parse(invalidConfig)).toThrow(
     new ZodError([
--- a/packages/airnode-validator/src/config/config.test.ts
+++ b/packages/airnode-validator/src/config/config.test.ts
@@ -2,14 +2,14 @@ import { readFileSync } from 'fs';
 import { join } from 'path';
 import { ZodError } from 'zod';
 import { chainOptionsSchema, configSchema, nodeSettingsSchema } from './config';
+import { Config, SchemaType } from '../types';
 import { version as packageVersion } from '../../package.json';
-import { SchemaType } from '../types';
 
 it('successfully parses config.json', () => {
-  const ois = JSON.parse(
+  const config = JSON.parse(
     readFileSync(join(__dirname, '../../test/fixtures/interpolated-config.valid.json')).toString()
   );
-  expect(() => configSchema.parse(ois)).not.toThrow();
+  expect(() => configSchema.parse(config)).not.toThrow();
 });
 
 describe('chainOptionsSchema', () => {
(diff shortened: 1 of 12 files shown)


packages/airnode-validator/src/ois/ois.ts
```diff
@@ -154,12 +154,27 @@ const ensurePathParametersExist: ValidatorRefinement<SchemaType<typeof pathsSche
...
-  security: z.record(z.tuple([])),
-});
+export const apiSpecificationSchema = z
+  .object({
+    components: apiComponentsSchema,
+    paths: pathsSchema,
+    servers: z.array(serverSchema),
+    security: z.record(z.tuple([])),
+  })
+  .superRefine((apiSpecification, ctx) => {
```
Siegrift, line 164: I know the issue doesn't mention this, but can we also check it the other way around? (that every scheme from `components.securitySchemes` is enabled in the `security`)?

Siegrift (commented):
👍  LGTM. 

I think we should also verify that the the security scheme value is set via the `apiCredentials` https://docs.api3.org/airnode/v0.6/grp-providers/guides/build-an-airnode/api-security.html#step-3-specify-the-values-for-the-defined-security-schemes but we can do that as a separate issue? WDYT?

acenolaza:
> +1 LGTM.
> 
> I think we should also verify that the the security scheme value is set via the `apiCredentials` https://docs.api3.org/airnode/v0.6/grp-providers/guides/build-an-airnode/api-security.html#step-3-specify-the-values-for-the-defined-security-schemes but we can do that as a separate issue? WDYT?

Right, I thought about doing that at first but then I realized that `securityScheme` [is optional](https://github.com/api3dao/airnode/blob/an614-security-scheme-validation/packages/airnode-adapter/src/request-building/authentication.ts#L119) in `apiCredentials` but maybe I could still do a one-way check (if `apiCredentials.securitySchemeName` is defined then check if `ois.apiSpecifications.components.securitySchemeName` is also defined. WDYT?

Siegrift:
Mhh, actually maybe we can bring this up at the call. I can imagine an user taking full OAS of their API, making a few changes to convert it to OIS and enable only a subset of triggers and security schemes. That suggests we should not do this check and allow "unused" security schemes.

However, for enabled security schemes (via the `security` field) we need make sure that  the value is defined in `apiCredentials`.

acenolaza:
> Mhh, actually maybe we can bring this up at the call. I can imagine an user taking full OAS of their API, making a few changes to convert it to OIS and enable only a subset of triggers and security schemes. That suggests we should not do this check and allow "unused" security schemes. But for other users this will be an annoying error they'll need to debug shrug .
 
Yes, and this has been how I've been approaching "reference" validation in Airkeeper and Airkeeper. For those config files I'm only adding checks for things being referenced but unused things are not being validated for convenience since having those config values there don't really hurt and makes testing easier when switching between config values.

> However, for enabled security schemes (via the `security` field) we need make sure that the value is defined in `apiCredentials`.

The thing is that [we only care](https://github.com/api3dao/airnode/blob/an614-security-scheme-validation/packages/airnode-adapter/src/request-building/authentication.ts#L97-L100) about `apiCredentials.securitySchemeValue` being set when `securityScheme.type` is `apiKey` or `http` 🤔

Siegrift:
> The thing is that [we only care](https://github.com/api3dao/airnode/blob/an614-security-scheme-validation/packages/airnode-adapter/src/request-building/authentication.ts#L97-L100) about apiCredentials.securitySchemeValue being set when securityScheme.type is apiKey or http 🤔

No. I mean when you have a security scheme that is defined and also enabled. Such as [this](https://github.com/api3dao/airnode/blob/config-txtype/packages/airnode-examples/integrations/weather-multi-value/config.example.json#L97) then we should verify the value is also supplied in `apiCredentials` ([here](https://github.com/api3dao/airnode/blob/config-txtype/packages/airnode-examples/integrations/weather-multi-value/config.example.json#L152)). Essentially, making sure that user defined  the [step 3 from the docs](https://docs.api3.org/airnode/v0.4/grp-providers/guides/build-an-airnode/api-security.html#step-3-specify-the-values-for-the-defined-security-schemes).

acenolaza:
> > The thing is that [we only care](https://github.com/api3dao/airnode/blob/an614-security-scheme-validation/packages/airnode-adapter/src/request-building/authentication.ts#L97-L100) about apiCredentials.securitySchemeValue being set when securityScheme.type is apiKey or http thinking
> 
> No. I mean when you have a security scheme that is defined and also enabled. Such as [this](https://github.com/api3dao/airnode/blob/config-txtype/packages/airnode-examples/integrations/weather-multi-value/config.example.json#L97) then we should verify the value is also supplied in `apiCredentials` ([here](https://github.com/api3dao/airnode/blob/config-txtype/packages/airnode-examples/integrations/weather-multi-value/config.example.json#L152)). Essentially, making sure that user defined the [step 3 from the docs](https://docs.api3.org/airnode/v0.4/grp-providers/guides/build-an-airnode/api-security.html#step-3-specify-the-values-for-the-defined-security-schemes).

But what if the Airnode is only relaying metadata using securitySchemes? https://docs.api3.org/airnode/v0.4/grp-providers/guides/build-an-airnode/api-security.html#relayed-meta-data-security-schemes

Siegrift:
Oh, I see what you mean now. Yes, we only care about those two. But if they are defined and used, there should be a value defined for them in `apiCredentials `

packages/airnode-validator/src/config/config.test.ts
```diff
@@ -2,14 +2,14 @@ import { readFileSync } from 'fs';
 import { join } from 'path';
 import { ZodError } from 'zod';
 import { configSchema, nodeSettingsSchema } from './config';
+import { Config, SchemaType } from '../types';
 import { version as packageVersion } from '../../package.json';
-import { SchemaType } from '../types';
 
 it('successfully parses config.json', () => {
-  const ois = JSON.parse(
+  const config = JSON.parse(
```
Siegrift, line 9: 🙏

packages/airnode-validator/src/config/config.test.ts
```diff
@@ -69,3 +69,71 @@ describe('nodeSettingsSchema', () => {
...
+          ...config.ois[0].apiSpecifications,
+          components: { securitySchemes: securityScheme },
+          security: { [securitySchemeName]: [] },
+        },
+      },
+    ],
+    apiCredentials,
+  };
+
+  const invalidConfigEmtpyApiCredentialValue = {
```
Siegrift, line 109:
```suggestion
  const invalidConfigEmptyApiCredentialValue = {
```

packages/airnode-validator/src/config/config.test.ts
```diff
@@ -69,3 +69,71 @@ describe('nodeSettingsSchema', () => {
     );
   });
 });
+
+it('fails if ois.apiSpecifications.security.<securitySchemeName> is defined in ois.apiSpecifications.components.<securitySchemeName> but apiCredentials.securitySchemeValue is empty string', () => {
```
Siegrift, line 73: `fails if ois.apiSpecifications.security.<securitySchemeName> is defined in ois.apiSpecifications.components.<securitySchemeName> but missing value in apiCredentials`

I don't think we should enforce non empty string here.

packages/airnode-validator/src/config/config.test.ts
```diff
@@ -69,3 +69,71 @@ describe('nodeSettingsSchema', () => {
...
+  const config: Config = JSON.parse(
+    readFileSync(join(__dirname, '../../test/fixtures/interpolated-config.valid.json')).toString()
+  );
+
+  const securitySchemeName = 'Currency Converter Security Scheme';
+  const securityScheme = {
+    securitySchemeName: {
```
Siegrift, line 80:
```suggestion
    [securitySchemeName]: {
```

packages/airnode-validator/src/config/config.test.ts
Siegrift: But I'd just call this `invalidConfig` and test that the value is missing in `apiCredential`.

packages/airnode-validator/src/config/config.test.ts
```diff
@@ -69,3 +69,71 @@ describe('nodeSettingsSchema', () => {
...
+    apiCredentials: [
+      ...config.apiCredentials,
+      { ...apiCredentials[0], securitySchemeName: apiCredentials[0].securitySchemeName + Date.now().toString() },
+    ],
+  };
+  expect(() => configSchema.parse(invalidConfigNoApiCredential)).toThrow(
+    new ZodError([
+      {
+        code: 'custom',
+        message: 'Security scheme "Currency Converter Security Scheme" is not defined in "components.securitySchemes"',
```
Siegrift, line 134: This error message will change once you fix the `[securitySchemeName]`

packages/airnode-validator/src/config/config.ts
```diff
@@ -138,10 +139,35 @@ export const apiCredentialsSchema = baseApiCredentialsSchema.extend({
...
+      const enabledSecurityScheme = ois.apiSpecifications.components.securitySchemes[enabledSecuritySchemeName];
+      if (enabledSecurityScheme && ['apiKey', 'http'].includes(enabledSecurityScheme.type)) {
+        const securitySchemeApiCredentials = config.apiCredentials.find(
+          (apiCredentials) => apiCredentials.securitySchemeName === enabledSecuritySchemeName
+        );
+        if (!securitySchemeApiCredentials || !securitySchemeApiCredentials.securitySchemeValue) {
+          ctx.addIssue({
+            code: z.ZodIssueCode.custom,
+            message: `The security scheme is enabled but no credentials are provided`,
+            path: ['ois', index, 'apiSpecifications', 'security', enabledSecuritySchemeName],
```
Siegrift, line 157: The value in the path should be index not a name.

```suggestion
            path: ['ois', index, 'apiSpecifications', 'security', schemeIndex],
```

packages/airnode-validator/src/config/config.ts
```diff
@@ -138,10 +139,35 @@ export const apiCredentialsSchema = baseApiCredentialsSchema.extend({
...
-  ois: z.array(oisSchema),
-  triggers: triggersSchema,
-  apiCredentials: z.array(apiCredentialsSchema),
-});
+const validateSecuritySchemesReferences: SuperRefinement<{
+  ois: OIS[];
+  apiCredentials: ApiCredentials[];
+}> = (config, ctx) => {
+  config.ois.forEach((ois, index) => {
+    Object.keys(ois.apiSpecifications.security).forEach((enabledSecuritySchemeName) => {
```
Siegrift, line 147:
```suggestion
    Object.keys(ois.apiSpecifications.security).forEach((enabledSecuritySchemeName, schemeIndex) => {
```

packages/airnode-validator/src/config/config.ts
```diff
@@ -138,10 +139,35 @@ export const apiCredentialsSchema = baseApiCredentialsSchema.extend({
...
+      const enabledSecurityScheme = ois.apiSpecifications.components.securitySchemes[enabledSecuritySchemeName];
+      if (enabledSecurityScheme && ['apiKey', 'http'].includes(enabledSecurityScheme.type)) {
+        const securitySchemeApiCredentials = config.apiCredentials.find(
+          (apiCredentials) => apiCredentials.securitySchemeName === enabledSecuritySchemeName
+        );
+        if (!securitySchemeApiCredentials || !securitySchemeApiCredentials.securitySchemeValue) {
```
Siegrift, line 153:
```suggestion
        if (!securitySchemeApiCredentials) {
```

packages/airnode-validator/src/config/config.ts
```diff
@@ -138,10 +139,35 @@ export const apiCredentialsSchema = baseApiCredentialsSchema.extend({
...
+          ctx.addIssue({
+            code: z.ZodIssueCode.custom,
+            message: `The security scheme is enabled but no credentials are provided`,
```
Siegrift, line 156: Maybe: (so that user knows where to specify the credentials). wdyt?

```suggestion
            message: `The security scheme is enabled but no credentials are provided in "config.apiCredentials"`,
```

Siegrift (changes requested):

packages/airnode-validator/src/config/config.ts
acenolaza: 🤔 sure about this? ois.apiSpecifications.security is not an array but an object where securitySchemeNames are the keys of it's properties. I can easily add the index but I'm wondering if it would be easier to determine where the error is if I just use the property key in the path. WDYT?

Siegrift: 🤦 Sorry, the security is a an object. I got confused somewhere  and thought it was an array. We definitely want a name not index.

acenolaza:
Thanks for the thorough and detailed code review @Siegrift 🙏🏻

packages/airnode-validator/src/config/config.test.ts
```diff
@@ -69,3 +69,57 @@ describe('nodeSettingsSchema', () => {
     );
   });
 });
+
+it('fails if ois.apiSpecifications.security.<securitySchemeName> is defined in ois.apiSpecifications.components.<securitySchemeName> but apiCredentials.securitySchemeValue is empty string', () => {
```
Siegrift, line 73: The test title is outdated. I guess it should be:

```
fails if ois.apiSpecifications.security.<securitySchemeName> is defined in ois.apiSpecifications.components.<securitySchemeName> but missing value in apiCredentials
```

packages/airnode-validator/src/config/config.test.ts
```diff
@@ -69,3 +69,57 @@ describe('nodeSettingsSchema', () => {
...
+
+  const securitySchemeName = 'Currency Converter Security Scheme';
+  const securityScheme = {
+    [securitySchemeName]: {
+      in: 'query',
+      type: 'apiKey',
+      name: 'access_key',
+    },
+  };
+  const apiCredentials = [
```
Siegrift, line 86: Why not remove this and pass empty `apiCredentials` to the config? The config setup would be shorter.

acenolaza: No reason. I just thought it might be closer to a real use-case if there was some data in apiCredentials.

Siegrift: You can remove this then and rename `configWithSecuritySchemes` to `invalidConfig`.

Siegrift (approved):
👍 LGTM

PR #1103, opened 2022-05-18
Validate chain options by type of transaction

dcroote:
[AN-618](https://api3dao.atlassian.net/browse/AN-618). Validates `legacy` options aren't being supplied for `eip1559` transactions and vice versa.


packages/airnode-validator/src/config/config.ts
```diff
@@ -43,13 +43,29 @@ export const priorityFeeSchema = z.object({
...
 });
 
-export const chainOptionsSchema = z.object({
-  txType: z.union([z.literal('legacy'), z.literal('eip1559')]),
-  gasPriceMultiplier: z.number().optional(),
-  baseFeeMultiplier: z.number().int().optional(),
-  priorityFee: priorityFeeSchema.optional(),
-  fulfillmentGasLimit: z.number().int(),
-});
+export const chainOptionsSchema = z
```
Siegrift, line 46: You can simplify this using the https://github.com/colinhacks/zod#discriminated-unions

Also, the inferred TS types will be better.

Siegrift (changes requested):

packages/airnode-validator/src/config/config.ts
dcroote: "simplify" 😅

I implemented using `discriminatedUnion`, which then required `strict()` validation in order to error on unrecognized  fields e.g. `priorityFee` being present for `legacy` `txType`. By default this produces a somewhat confusing error "Unrecognized key(s) in object", which I modified though a custom error mapping so that the user would know that the key(s) were unrecognized / disallowed _**for that transaction type**_.

Siegrift: I think the error message from zod is generic because it uses the `path` to tell you what is wrong. I like your message better, but I wouldn't be so strict about it in general.

Siegrift (approved):
👍 LGTM.

PR #1104, opened 2022-05-18
Update config txType to use number type instead of string

vponline:
This changes the **config.json** `txType` to accept either `0` (for legacy) or `2` (for eip1559) as values.

### DIFF
--- a/.changeset/beige-rocks-pull.md
+++ b/.changeset/beige-rocks-pull.md
@@ -1,0 +1,10 @@
+---
+'@api3/airnode-admin': minor
+'@api3/airnode-deployer': minor
+'@api3/airnode-examples': minor
+'@api3/airnode-node': minor
+'@api3/airnode-utilities': minor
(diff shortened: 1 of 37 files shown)


acenolaza:
> +1 LGMT.
> 
> It feels a bit less readable, but this seems to be the satndard naming for these transaacation types (e.g. etherscan also shows the numbers) and we decided on this in Slack.

I feel the same way about it being less readable but I'm wondering if we can use a zod.enum [here](https://github.com/api3dao/airnode/blob/77542e6f5e0c6511ff28dbb912e4fdd4b94ef6ee/packages/airnode-validator/src/config/config.ts#L47) or something in order to keep using `legacy` and `eip1559` in code.

Siegrift:
We could still use the strings internally and numbers in the `config.json`. But me personally would be happy to use the strings everywhere (but this is not a hard preference).

PR #1105, opened 2022-05-19
Add config validation for local Airnode invocation including `airnode-client` container

amarthadan:
[AN-595](https://api3dao.atlassian.net/browse/AN-595)

Doing the config file validation separately (instead of running it as a part of `invoke.ts`) so that if it fails, the container fails as well instead of running and throwing an error every minute.

### DIFF
--- a/packages/airnode-node/package.json
+++ b/packages/airnode-node/package.json
@@ -12,7 +12,7 @@
     "build": "yarn run clean && yarn run compile",
     "clean": "rimraf -rf *.tsbuildinfo ./dist *.tgz",
     "compile": "tsc --build tsconfig.build.json",
-    "dev:invoke": "ts-node src/cli/invoke.ts",
+    "dev:invoke": "ts-node src/cli/validate-config.ts && ts-node src/cli/invoke.ts",
     "dev:testApi": "ts-node src/cli/test-api.ts",
     "pack": "yarn pack",
     "test:e2e": "SILENCE_LOGGER=true jest --selectProjects e2e",
--- a/.changeset/warm-knives-visit.md
+++ b/.changeset/warm-knives-visit.md
@@ -1,0 +1,6 @@
+---
+'@api3/airnode-examples': minor
+'@api3/airnode-node': minor
+---
+
+Add config validation for local Airnode invocation including airnode-client container
--- a/packages/airnode-examples/README.md
+++ b/packages/airnode-examples/README.md
@@ -213,7 +213,7 @@ yarn run-airnode-locally api3/airnode-client-dev:bb9b8118940ec852c4223b13eba5a6e
 
 ### 11. Deploy a requester
 
-At this point, you have an RRP contract deployed. You will also either have Airnode running as a Docker container  
+At this point, you have an RRP contract deployed. You will also either have Airnode running as a Docker container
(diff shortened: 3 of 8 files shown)


packages/airnode-examples/scripts/run-airnode-locally.ts
```diff
@@ -23,7 +23,7 @@ const main = async () => {
   }
 
   const integrationPath = join(__dirname, '../integrations', integrationInfo.integration);
-  runShellCommand(`docker run --rm -v ${integrationPath}:/app/config --network="host" --name airnode ${imageName}`);
+  runShellCommand(`docker run --rm -it -v ${integrationPath}:/app/config --network="host" --name airnode ${imageName}`);
```
amarthadan, line 26: Added this so the signals from OS are passed correctly. Without it, the container was running unless I called `docker stop`.

Siegrift: I think I didn't include `-it` for a specific reason (can't remember why exactly) - but the examples tests fail currently.

packages/airnode-node/package.json
```diff
@@ -12,7 +12,7 @@
     "build": "yarn run clean && yarn run compile",
     "clean": "rimraf -rf *.tsbuildinfo ./dist *.tgz",
     "compile": "tsc --build tsconfig.build.json",
-    "dev:invoke": "ts-node src/cli/invoke.ts",
+    "dev:invoke": "ts-node src/cli/validate-config.ts && ts-node src/cli/invoke.ts",
```
Siegrift, line 15: I think the validate-config script should be in a separate command and not included in `dev:invoke`. You can use invoke even with invalid config when you want to test something specific.

Siegrift (changes requested):
Currently the tests fail.

packages/airnode-node/package.json
Siegrift: I just don't like that `dev:testApi` calls `test-api.ts`, but `dev:invoke` calls both validate and invoke and we have no script for validation only.

packages/airnode-examples/scripts/run-airnode-locally.ts
amarthadan: Ok, I'm giving up, spawning in the e2e test doesn't work properly with this. I'm reverting it but I'll add a notice to the examples README to stop the container manually.

packages/airnode-node/package.json
amarthadan: I understand. I still think this would be actually a less confusing approach since it will behave the same for local invocation and the container. If you don't have a hard preference I would keep it this way.

Siegrift: No hard preference.

packages/airnode-node/package.json
```diff
@@ -28,6 +28,7 @@
     "@api3/airnode-protocol": "^0.6.0",
     "@api3/airnode-utilities": "^0.6.0",
     "@api3/airnode-validator": "^0.6.0",
+    "@api3/promise-utils": "^0.3.0",
```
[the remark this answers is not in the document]
amarthadan: I initially was but ended up not using it and forget to remove it. Will do.

Siegrift (approved):
👍 LGTM

PR #1108, opened 2022-05-20
Validate API specification parameters

Siegrift:
https://api3dao.atlassian.net/browse/AN-613

### DIFF
--- a/packages/airnode-validator/src/ois/ois.test.ts
+++ b/packages/airnode-validator/src/ois/ois.test.ts
@@ -164,3 +164,45 @@ it('fails if apiSpecifications.security.<securitySchemeName> is not defined in a
     ])
   );
 });
+
+describe('apiSpecification parameters validation', () => {
+  it('fails if "apiSpecification.paths" parameter is not defined in "endpoints"', () => {
+    const invalidOis = loadOisFixture();
+    invalidOis.apiSpecifications.paths['/convert'].get!.parameters.push({
+      in: 'query',
+      name: 'non-existing-parameter',
+    });
+
+    expect(() => oisSchema.parse(invalidOis)).toThrow(
+      new ZodError([
+        {
+          code: 'custom',
+          message: `asd`,
+          path: [],
+        },
+      ])
+    );
+  });
+
+  it('"endpoint" parameter must reference parameter from "apiSpecification.paths"', () => {
+    const invalidOis = loadOisFixture();
+    invalidOis.endpoints[0].parameters.push({
+      name: 'some-new-param',
+      default: 'EUR',
+      operationParameter: {
+        in: 'query',
+        name: 'non-existing-param',
+      },
+    });
+
+    expect(() => oisSchema.parse(invalidOis)).toThrow(
+      new ZodError([
+        {
+          code: 'custom',
+          message: `asd`,
+          path: [],
+        },
+      ])
+    );
+  });
+});
--- a/packages/airnode-validator/src/ois/ois.test.ts
+++ b/packages/airnode-validator/src/ois/ois.test.ts
@@ -61,35 +61,17 @@ describe('disallows reserved parameter name', () => {
 
 it('verifies that the same parameter cannot be in fixedOperationParameters and parameters', () => {
   const ois = loadOisFixture();
-  ois.endpoints.push(cloneDeep(ois.endpoints[0]));
-
-  // Create the same parameter in the first endpoint
-  const parameter1 = { name: 'api-param-name1', in: 'query' } as const;
-  ois.endpoints[0].parameters.push({ name: 'test-param', operationParameter: parameter1 });
   ois.endpoints[0].fixedOperationParameters.push({
+    operationParameter: ois.endpoints[0].parameters[0].operationParameter,
     value: '123',
-    operationParameter: parameter1,
-  });
-
-  // Create the same parameter in the second endpoint
-  const parameter2 = { name: 'api-param-name2', in: 'query' } as const;
-  ois.endpoints[1].parameters.push({ name: 'test-param2', operationParameter: parameter2 });
-  ois.endpoints[1].fixedOperationParameters.push({
-    value: '456',
-    operationParameter: parameter2,
   });
 
   expect(() => oisSchema.parse(ois)).toThrow(
     new ZodError([
       {
         code: 'custom',
-        message: 'Parameters "api-param-name1" are used in both "parameters" and "fixedOperationOperations"',
+        message: 'Parameters "from" are used in both "parameters" and "fixedOperationOperations"',
         path: ['ois', 'endpoints', 0],
-      },
-      {
-        code: 'custom',
-        message: 'Parameters "api-param-name2" are used in both "parameters" and "fixedOperationOperations"',
-        path: ['ois', 'endpoints', 1],
       },
     ])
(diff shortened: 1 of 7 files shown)


amarthadan:
If you can squash these commits into one commit that actually has a meaningful name that would be great :smile:

Siegrift:
> If you can squash these commits into one commit that actually has a meaningful name that would be great 

Sure, but I think having it separated this way makes it easier to review.

amarthadan:
> > If you can squash these commits into one commit that actually has a meaningful name that would be great
> 
> Sure, but I think having it separated this way makes it easier to review.

Fair enough, but please, do so before merging.

packages/airnode-validator/src/ois/ois.test.ts
```diff
@@ -61,36 +61,18 @@ describe('disallows reserved parameter name', () => {
...
-    value: '456',
-    operationParameter: parameter2,
   });
 
   expect(() => oisSchema.parse(ois)).toThrow(
     new ZodError([
       {
         code: 'custom',
-        message: 'Parameters "api-param-name1" are used in both "parameters" and "fixedOperationOperations"',
+        message: 'Parameters "from" are used in both "parameters" and "fixedOperationOperations"',
```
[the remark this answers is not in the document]
Siegrift: [Allow parameters with same name, but different http method](https://github.com/api3dao/airnode/commit/fb92d912eb141f64d8990c38046800028f2ccf3a)

packages/airnode-validator/src/ois/ois.ts
```diff
@@ -195,7 +196,7 @@ export const endpointSchema = z.object({
   summary: z.string().optional(),
 });
 
-const ensureSingleParameterUsagePerEndpoint: ValidatorRefinement<OIS> = (ois, ctx) => {
+const ensureSingleParameterUsagePerEndpoint: ValidatorRefinement<SchemaType<typeof baseOisSchema>> = (ois, ctx) => {
```
vponline, line 199: Nit: could  extract `SchemaType<typeof baseOisSchema>`  and import it from `types.ts`

packages/airnode-validator/src/ois/ois.ts
```diff
@@ -211,6 +212,92 @@ const ensureSingleParameterUsagePerEndpoint: ValidatorRefinement<OIS> = (ois, ct
   });
 };
 
+const ensureEndpointAndApiSpecificationParamsMatch: ValidatorRefinement<SchemaType<typeof baseOisSchema>> = (
```
vponline, line 215: Nit: reuse the `BaseOisType` if extracted from above.

packages/airnode-validator/src/ois/ois.ts
```diff
@@ -219,4 +306,6 @@ export const baseOisSchema = z.object({
   endpoints: z.array(endpointSchema),
 });
 
-export const oisSchema = baseOisSchema.superRefine(ensureSingleParameterUsagePerEndpoint);
+export const oisSchema = baseOisSchema
+  .superRefine(ensureSingleParameterUsagePerEndpoint)
+  .superRefine(ensureEndpointAndApiSpecificationParamsMatch);
```
vponline, line 311: Nit: could rename the refine functions from `ensure`... to `validate`... if we want to keep the function names similar everywhere

packages/airnode-validator/src/ois/ois.ts
```diff
@@ -211,6 +212,92 @@ const ensureSingleParameterUsagePerEndpoint: ValidatorRefinement<OIS> = (ois, ct
...
+        paramData!.parameters.forEach((apiParam) => {
+          const allEndpointParams = [...endpoint.parameters, ...endpoint.fixedOperationParameters];
+          const endpointParam = allEndpointParams.find(
+            ({ operationParameter }) =>
+              operationParameter.in === apiParam.in && operationParameter.name === apiParam.name
+          );
+          if (!endpointParam) {
+            ctx.addIssue({
+              code: z.ZodIssueCode.custom,
+              message: `Parameter "${apiParam.name}" not found in "fixedOperationParameters" nor "parameters"`,
```
vponline, line 239: :detective: 
Nit: I believe this is a double-negative :smile: 
```suggestion
              message: `Parameter "${apiParam.name}" not found in "fixedOperationParameters" or "parameters"`,
```

packages/airnode-validator/src/ois/ois.test.ts
```diff
@@ -164,3 +146,90 @@ it('fails if apiSpecifications.security.<securitySchemeName> is not defined in a
...
+    ois.apiSpecifications.paths['/convert'].get!.parameters.push({
+      in: 'query',
+      name: 'api-param-name',
+    });
+
+    expect(() => oisSchema.parse(ois)).toThrow(
+      new ZodError([
+        {
+          code: 'custom',
+          message: 'Parameter "api-param-name" not found in "fixedOperationParameters" nor "parameters"',
```
vponline, line 203: If you decide to change the spelling:
```suggestion
          message: 'Parameter "api-param-name" not found in "fixedOperationParameters" or "parameters"',
```

packages/airnode-validator/src/ois/ois.test.ts
```diff
@@ -164,3 +146,90 @@ it('fails if apiSpecifications.security.<securitySchemeName> is not defined in a
...
+          path: ['ois', 'endpoints', 0],
+        },
+        {
+          code: 'custom',
+          message: 'Parameter "api-param-name" not found in "fixedOperationParameters" nor "parameters"',
```
vponline, line 208: And here:
```suggestion
          message: 'Parameter "api-param-name" not found in "fixedOperationParameters" or "parameters"',
```

vponline (approved):
LGTM. just noticed a couple minor details, feel free to ignore them :1st_place_medal:

packages/airnode-validator/src/ois/ois.ts
Siegrift: Yeah, definitely, but I think it's better to do it all at once as part of this task: https://api3dao.atlassian.net/browse/AN-723

packages/airnode-validator/src/ois/ois.ts
Siegrift: That is a good point and left me thinking about this.

I'd not replace this, because they are not synonyms. For example, you can't say `ensureSecuritySchemesReferences` instead of `validateSecuritySchemesReferences`. I like the `ensure` a bit more since it forces you to be more specific which is nicer. But I understand it might not be easy (or worth it) to name the functions this way all the time.

packages/airnode-validator/src/ois/ois.ts
vponline: I stand corrected!  :smile:

packages/airnode-validator/src/ois/ois.ts
vponline: That's good reasoning and I agree, thanks for explaining. Let's keep them   :+1:

packages/airnode-validator/src/ois/ois.ts
Siegrift: Actually @kolenic-martin and @amarthadan disliked this as well, so let me take your suggestion :D

PR #1117, opened 2022-05-23
Add axios dependency in Airnode for Processing

aquarat:
@Ashar2shahid found a bug in Airnode-node when using it within Airkeeper, relating to axios not being included as a dependency in package.json.


Siegrift (approved):
Yeah, I found a few minor issues while implementing the example as well. See: https://api3dao.atlassian.net/browse/AN-750

PR #1126, opened 2022-05-25
Allow parameters with same name, but different location

Siegrift:
https://api3dao.atlassian.net/browse/AN-753

### DIFF
--- a/packages/airnode-validator/src/ois/ois.test.ts
+++ b/packages/airnode-validator/src/ois/ois.test.ts
@@ -59,22 +59,98 @@ describe('disallows reserved parameter name', () => {
   });
 });
 
-it('verifies that the same parameter cannot be in fixedOperationParameters and parameters', () => {
-  const ois = loadOisFixture();
-  ois.endpoints[0].fixedOperationParameters.push({
-    operationParameter: ois.endpoints[0].parameters[0].operationParameter,
-    value: '123',
+describe('parameter uniqueness', () => {
+  it('allows parameter with same name, but different http method', () => {
+    const paramName = 'some-id';
+    const ois = loadOisFixture();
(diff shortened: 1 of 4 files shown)


packages/airnode-validator/src/ois/ois.test.ts
```diff
@@ -59,22 +59,98 @@ describe('disallows reserved parameter name', () => {
...
-  expect(() => oisSchema.parse(ois)).toThrow(
-    new ZodError([
-      {
-        code: 'custom',
-        message: 'Parameters "from" are used in both "parameters" and "fixedOperationOperations"',
-        path: ['ois', 'endpoints', 0],
-      },
-    ])
-  );
+  it(`fails if the same parameter is used in "parameters"`, () => {
```
Siegrift, line 92: I wanted to use `forEach` for this test and the test below, but it was messy so I kept them separate.

packages/airnode-validator/src/ois/ois.test.ts
```diff
@@ -59,22 +59,98 @@ describe('disallows reserved parameter name', () => {
...
+    const ois = loadOisFixture();
+    ois.endpoints[0].parameters.push(ois.endpoints[0].parameters[0] as any);
+
+    expect(() => oisSchema.parse(ois)).toThrow(
+      new ZodError([
+        {
```
Siegrift, line 98: I'm deliberately including both errors because it's easier for the user to find the offending pair... I'd even consider adding more details to the message (e.g. `Parameter "param-name" in "query" is used multiple times`). WDYT?

Siegrift: Yeah. I've added it. Btw. feel free to review if you have time :)

PR #1127, opened 2022-05-26
Remove relocated tx constants from airnode-node

dcroote:
These constants are unused except for tests as gas price handling has moved to airnode-utilities and promise handling has moved to promise-utils


Siegrift (approved):
👍 LGTM.

PR #1128, opened 2022-05-26
Fix lerna version in Docker build

amarthadan:
Regarding https://api3workspace.slack.com/archives/C02AYRX8D89/p1653545236645239

PR #1129, opened 2022-05-26
[0.6] Fix cloud resource naming

amarthadan:
Backport to 0.6 from https://github.com/api3dao/airnode/pull/1099

I had to add a fix from https://github.com/api3dao/airnode/pull/1128 as well so our Docker containers get built.


Siegrift (approved):
👍 LGTM

PR #1131, opened 2022-05-29
AN-669: Add option to subtract from withdrawals

### LINKED ISSUE #1077: Sponsor wallet withdrawals don't work on Optimism
Withdrawals ignore the L1 fee and attempt to transfer more than possible, which fails

dcroote:
Introduces an optional `withdrawalRemainder` within `chainOptions` of `config.json` that gets subtracted from the funds returned to the sponsor when making a withdrawal. Such a remainder is necessary to cover L1 data fees that aren't included when calculating the withdrawal transaction cost on Optimism (more on from the [Optimism docs here](https://community.optimism.io/docs/developers/build/transaction-fees/#displaying-fees-to-users)).

Fixes #1077


packages/airnode-node/config/config.example.json
```diff
@@ -20,7 +20,8 @@
           "value": 3.12,
           "unit": "gwei"
         },
-        "fulfillmentGasLimit": 500000
+        "fulfillmentGasLimit": 500000,
+        "withdrawalRemainder": "0"
```
dcroote, line 24: Another option instead of strings throughout could be numbers with something like `withdrawalRemainderGwei`

bbenligiray: Isn't the standard way of specifying amounts
```
{
  "value": 3.12,
  "unit": "gwei"
}
```
?

packages/airnode-node/src/evm/fulfillments/withdrawals.test.ts
```diff
@@ -85,6 +85,50 @@ describe('submitWithdrawal', () => {
...
+      expect(fulfillWithdrawalMock).toHaveBeenCalledWith(
+        withdrawal.id,
+        withdrawal.airnodeAddress,
+        withdrawal.sponsorAddress,
+        {
+          ...gasTarget,
+          gasLimit: ethers.BigNumber.from(70_000),
+          nonce: 5,
+          // 250_000_000 - ((50_000 + 20_000) * 1000) - 10_000_000
+          value: ethers.BigNumber.from(170_000_000),
```
Siegrift, line 126: You can remove the comment and inline the expression

```suggestion
          value: ethers.BigNumber.from(250_000_000 - ((50_000 + 20_000) * 1000) - 10_000_000),
```

packages/airnode-node/test/setup/e2e/utils.ts
```diff
@@ -26,6 +26,8 @@ export function buildChainConfig(contracts: Contracts): ChainConfig {
         unit: 'gwei',
       },
       fulfillmentGasLimit: 500_000,
+      // 1 ETH
+      withdrawalRemainder: '1000000000000000000',
```
Siegrift, line 30:
```suggestion
      withdrawalRemainder: utils.parseEther("1").toString(),
```

packages/airnode-node/src/evm/fulfillments/withdrawals.ts
```diff
@@ -59,13 +59,13 @@ export const submitWithdrawal: SubmitRequest<Withdrawal> = async (airnodeRrp, re
...
-  // send all the rest along with the transaction. The contract will direct
-  // these funds back to the sponsor wallet.
+  // We set aside some ETH to pay for the gas of the following transaction and
+  // return the rest to the sponsor wallet less a remainder, if specified.
+  const remainder = options.withdrawalRemainder ? ethers.BigNumber.from(options.withdrawalRemainder) : 0;
   const txCost = paddedGasLimit.mul(
     options.gasTarget.gasPrice ? options.gasTarget.gasPrice : options.gasTarget.maxFeePerGas!
   );
-  const fundsToSend = currentBalance.sub(txCost);
+  const fundsToSend = currentBalance.sub(txCost).sub(remainder);
```
Siegrift, line 68: What happens when this becomes negative?

packages/airnode-node/config/config.example.json
Siegrift, line 24: Do we want to have the value as `string` despite the `fulfillmentGasLimit` being number? It feels inconsistent, but we actually want to migrate to strings at some point anyway (see https://api3dao.atlassian.net/browse/AN-711)

packages/airnode-node/src/evm/fulfillments/withdrawals.ts
dcroote: That is handled just a few lines below 😄 

https://github.com/api3dao/airnode/pull/1131/files#diff-5ad6e79a7234dfa1621a5894aab96dca60eaa1ee9cc9e3c86869aab83c66c999R71-R76

packages/airnode-node/src/evm/fulfillments/withdrawals.test.ts
dcroote: Good call, thanks. I'm also applying it to another test.

packages/airnode-node/src/evm/fulfillments/withdrawals.ts
Siegrift: Sorry, I only looked at the diffed changes.

packages/airnode-node/test/setup/e2e/utils.ts
dcroote: I'm not applying this suggestion as I went with an amount type like @bbenligiray recommended above.

packages/airnode-node/config/config.example.json
dcroote: Same as the other comment: I'm not applying this suggestion as I went with an amount type.

dcroote: Yes- I've updated to use this approach.

packages/airnode-node/config/config.example.json
```diff
@@ -20,7 +20,11 @@
           "value": 3.12,
           "unit": "gwei"
         },
-        "fulfillmentGasLimit": 500000
+        "fulfillmentGasLimit": 500000,
+        "withdrawalRemainder": {
+          "value": 0,
```
Siegrift, line 25: Nit: Maybe add a non null value, since if the user wants to have 0 remainder he can just omit this.

packages/airnode-node/src/evm/fulfillments/withdrawals.ts
```diff
@@ -59,13 +59,15 @@ export const submitWithdrawal: SubmitRequest<Withdrawal> = async (airnodeRrp, re
     `Withdrawal gas limit estimated at ${paddedGasLimit.toString()} for Request:${request.id}`
   );
 
-  // We set aside some ETH to pay for the gas of the following transaction,
-  // send all the rest along with the transaction. The contract will direct
-  // these funds back to the sponsor wallet.
+  // We set aside some ETH to pay for the gas of the following transaction and
+  // return the rest to the sponsor wallet less a remainder, if specified.
```
Siegrift, line 63: Can you also mentioned that this is because some chains require this (e.g. Optimism)

Siegrift (approved):
👍 LGTM

packages/airnode-node/src/evm/fulfillments/withdrawals.ts
dcroote: Good idea- implemented.

packages/airnode-node/config/config.example.json
dcroote: Given `withdrawalRemainder` is only relevant to Optimism at the moment and is covered in the tests, I opted for omitting it from this `config.example.json` rather than leaving it as zero (potentially confusing) or adding a non-zero value (which wouldn't be appropriate for most chains). I'll describe it in the docs and within the Chain Idiosyncrasies page ([AN-761](https://api3dao.atlassian.net/browse/AN-761)).

PR #1140, opened 2022-05-31
Remove manual typings, use types infered by Validator

amarthadan:
Issue [AN-723](https://api3dao.atlassian.net/browse/AN-723)


packages/airnode-node/src/config/types.ts
```diff
@@ -1,124 +1 @@
...
-// TODO: Test that the parse result from validator is compatible with this config
-export interface Config {
-  readonly chains: ChainConfig[];
-  readonly nodeSettings: NodeSettings;
-  readonly ois: OIS[];
-  readonly triggers: Triggers;
-  readonly templates: Template[];
-  readonly apiCredentials: ApiCredentials[];
-}
+export * from '@api3/airnode-validator/dist/src/config';
```
amarthadan, line 1: In the end, I've opted for this approach.

Doing `export { config } from '@api3/airnode-validator'` makes cumbersome imports and re-exporting everything one by one is tedious and needs to be updated every time we want to add something new. Let me know what you think or whether this approach might break something.

packages/airnode-validator/src/config/config.ts
```diff
@@ -234,3 +234,18 @@ export const configSchema = z
...
+export type CloudProvider = SchemaType<typeof cloudProviderSchema>;
+export type AwsCloudProvider = SchemaType<typeof awsCloudProviderSchema>;
+export type GcpCloudProvider = SchemaType<typeof gcpCloudProviderSchema>;
+export type LocalOrCloudProvider = SchemaType<typeof localOrCloudProviderSchema>;
+export type Gateway = SchemaType<typeof gatewaySchema>;
+export type ChainOptions = SchemaType<typeof chainOptionsSchema>;
+export type ChainType = SchemaType<typeof chainTypeSchema>;
+export type ChainConfig = SchemaType<typeof chainConfigSchema>;
+export type Trigger = SchemaType<typeof triggerSchema>;
+export type Heartbeat = SchemaType<typeof heartbeatSchema>;
```
amarthadan, line 251: I've exported only pieces that are used by other packages. We may want to export everything instead. LMKWYT.

Siegrift (changes requested):
Found a few more thing we could do:

1. There are a few places in test files (e.g. https://github.com/api3dao/airnode/blob/remove-manual-typings/packages/airnode-validator/src/api/api.test.ts#L8) where we could use the named types. Can you fix those?
2. Use validator types in this file https://github.com/api3dao/airnode/blob/remove-manual-typings/packages/airnode-utilities/src/logging/types.ts#L1 (at least the LogLevel and logFormat)

packages/airnode-validator/src/config/config.ts
Siegrift: I don't think it's needed. In the worst case one can use the `SchemaType<typeof someShema>`...

packages/airnode-ois/src/index.ts
```diff
@@ -1,14 +1,7 @@
-import { ois as oisModule } from '@api3/airnode-validator';
-import { OIS } from './types';
+import { OIS, oisSchema } from './types';
```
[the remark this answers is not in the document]
amarthadan: What do you mean? The `oisSchema`? I think that's more suitable for `types` than `constants`.

packages/airnode-node/src/config/types.ts
Siegrift: We package different targets of the validator. This only works, because of a bug that I've introduced when I've added the references (read details below). Anyway, what you should use is `@api3/airnode-validator/dist/cjs/src/config`.

While testing this I realized that the `dist` folder is shared for tsconfig.build.json and tsconfig.json, but they build different files. This is essentially https://api3dao.atlassian.net/browse/AN-762 and the fix is probably to reference the tsconfig.build.json files instead.

packages/airnode-ois/src/index.ts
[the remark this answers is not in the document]
Siegrift: Mhhh, maybe a different file name then, e.g. `schemas`?

packages/airnode-node/src/config/types.ts
Siegrift: Btw. can we remove this file and just use the `export * ...` in `index.ts` ? Since the `types.ts` file exports also the schemas from the validator/ois. Having the file named `types.ts` it is mislieading.

packages/airnode-ois/src/index.ts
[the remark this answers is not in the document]
amarthadan: I can remove `types.ts` altogether as you're suggesting for config and no naming necessary.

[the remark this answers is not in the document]
Siegrift: Yeah, got this idea only after I submitted the comment :D

packages/airnode-node/src/config/types.ts
amarthadan: > While testing this I realized that the `dist` folder is shared for tsconfig.build.json and tsconfig.json, but they build different files. This is essentially https://api3dao.atlassian.net/browse/AN-762 and the fix is probably to reference the tsconfig.build.json files instead.

Or, to get rid of `tsconfig.build.json` as you did on Airseeker. I would actually prefer that.

Siegrift:
> Is that ok? Is this the intended state?

Yeah, it looks better to me.

Siegrift (approved):
👍 LGTM

PR #1144, opened 2022-06-01
Enable zod strict mode

amarthadan:
[AN-619](https://api3dao.atlassian.net/browse/AN-619)

### DIFF
--- a/packages/airnode-validator/src/ois/ois.ts
+++ b/packages/airnode-validator/src/ois/ois.ts
@@ -2,8 +2,8 @@ import forEach from 'lodash/forEach';
 import trimEnd from 'lodash/trimEnd';
 import trimStart from 'lodash/trimStart';
 import find from 'lodash/find';
-import { z } from 'zod';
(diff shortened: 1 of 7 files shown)


packages/airnode-validator/src/ois/ois.ts
```diff
@@ -238,10 +277,10 @@ const ensureSingleParameterUsagePerEndpoint: ValidatorRefinement<SchemaType<type
   });
 };
 
-const ensureEndpointAndApiSpecificationParamsMatch: ValidatorRefinement<SchemaType<typeof baseOisSchema>> = (
-  ois,
-  ctx
-) => {
+const ensureEndpointAndApiSpecificationParamsMatch: SuperRefinement<{
```
Siegrift, line 280: Why this rewrite?

amarthadan: So I could get rid of the `baseOisSchema` object and so it's consistent with how it's done for config.

Siegrift (approved):
👍 LGTM

PR #1145, opened 2022-06-01
Fix check for AWS API key uniqueness

amarthadan:
[AN-764](https://api3dao.atlassian.net/browse/AN-764) plus other small refactors (see commit messages).

### DIFF
--- a/packages/airnode-validator/src/config/config.test.ts
+++ b/packages/airnode-validator/src/config/config.test.ts
@@ -124,6 +124,16 @@ describe('nodeSettingsSchema', () => {
         region: 'region',
         disableConcurrencyReservations: false,
       },
+      httpGateway: {
+        enabled: true,
+        apiKey: 'sameApiKey',
+        maxConcurrency: 10,
+      },
+      httpSignedDataGateway: {
+        enabled: true,
+        apiKey: 'sameApiKey',
+        maxConcurrency: 10,
+      },
     };
 
     expect(() => nodeSettingsSchema.parse(invalidNodeSettings)).toThrow(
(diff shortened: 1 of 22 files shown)


packages/airnode-validator/src/config/config.test.ts
```diff
@@ -124,6 +124,16 @@ describe('nodeSettingsSchema', () => {
         region: 'region',
         disableConcurrencyReservations: false,
       },
+      httpGateway: {
```
Siegrift, line 127: 👍 I completely missed that I was not using any keys when implementing the unit test .

packages/airnode-validator/src/config/config.ts
```diff
@@ -16,7 +16,7 @@ export const triggerSchema = z
 export const triggersSchema = z
   .object({
     rrp: z.array(triggerSchema),
-    http: z.array(triggerSchema).optional(),
```
Siegrift, line 19: Yeah, I am working on another validator issue and I think there are multiple fields which should not be optional... I'll be creating similar commits to what you are doing now.

packages/airnode-validator/src/config/config.ts
```diff
@@ -132,15 +132,23 @@ export const disabledGatewaySchema = z
 
 export const gatewaySchema = z.discriminatedUnion('enabled', [enabledGatewaySchema, disabledGatewaySchema]);
 
-export const heartbeatSchema = z
+export const enabledHeartbeatSchema = z
```
Siegrift, line 135: Do we need separate schemas for those? Or you prefer that for readability? (I personally don't care with either).

Siegrift (approved):
👍 LGTM

PR #1146, opened 2022-06-01
Fix GCP gateway template filenames

amarthadan:
[AN-766](https://api3dao.atlassian.net/browse/AN-766)


Siegrift (approved):
👍  LGTM (but didn't test).

PR #1148, opened 2022-06-02
Legacy validator rules

Siegrift:
https://api3dao.atlassian.net/browse/AN-616

Review commit by commit.

### DIFF
--- a/packages/airnode-validator/src/config/config.test.ts
+++ b/packages/airnode-validator/src/config/config.test.ts
@@ -159,12 +159,12 @@ describe('nodeSettingsSchema', () => {
       },
       httpGateway: {
         enabled: true,
-        apiKey: '0x13dea3311fe0d6b84f4daeab831befbc49e19e6494c41e9e065a09c3c68f43b6',
+        apiKey: 'e83856ed-36cd-4b5f-a559-c8291e96e17e',
         maxConcurrency: 10,
       },
       httpSignedDataGateway: {
         enabled: true,
-        apiKey: '0x13dea3311fe0d6b84f4daeab831befbc49e19e6494c41e9e065a09c3c68f43b6',
+        apiKey: 'e83856ed-36cd-4b5f-a559-c8291e96e17e',
         maxConcurrency: 10,
       },
     };
@@ -300,27 +300,6 @@ describe('triggers references', () => {
           code: 'custom',
           message: `No matching endpoint for trigger with endpoint name "nonExistingEndpointName"`,
           path: ['triggers', 'rrp', 0, 'endpointName'],
-        },
-      ])
-    );
-  });
-
-  it('fails if endpoint ID is not a valid keccak hash', () => {
-    const invalidConfig = {
-      ...config,
-      triggers: {
-        ...config.triggers,
-        rrp: [{ ...config.triggers.rrp[0], endpointId: 'endpoint-id' }],
-      },
-    };
-
-    expect(() => configSchema.parse(invalidConfig)).toThrow(
-      new ZodError([
-        {
-          validation: 'regex',
-          code: 'invalid_string',
-          message: `Invalid`,
-          path: ['triggers', 'rrp', 0, 'endpointId'],
         },
       ])
     );
--- a/packages/airnode-validator/src/config/config.test.ts
+++ b/packages/airnode-validator/src/config/config.test.ts
@@ -298,4 +298,25 @@ describe('triggers references', () => {
       ])
     );
   });
+
+  it('fails if endpoint ID is not a valid keccak hash', () => {
+    const invalidConfig = {
+      ...config,
+      triggers: {
+        ...config.triggers,
+        rrp: [{ ...config.triggers.rrp[0], endpointId: 'endpoint-id' }],
+      },
+    };
+
+    expect(() => configSchema.parse(invalidConfig)).toThrow(
+      new ZodError([
+        {
+          validation: 'regex',
+          code: 'invalid_string',
+          message: `Invalid`,
+          path: ['triggers', 'rrp', 0, 'endpointId'],
+        },
+      ])
+    );
+  });
 });
--- a/packages/airnode-validator/src/config/config.test.ts
+++ b/packages/airnode-validator/src/config/config.test.ts
@@ -1,6 +1,7 @@
 import { readFileSync } from 'fs';
 import { join } from 'path';
 import { ZodError } from 'zod';
+import zip from 'lodash/zip';
 import {
   Config,
   chainOptionsSchema,
@@ -10,8 +11,13 @@ import {
   NodeSettings,
   amountSchema,
   Amount,
+  enabledGatewaySchema,
+  enabledHeartbeatSchema,
+  gatewaySchema,
+  heartbeatSchema,
 } from './config';
 import { version as packageVersion } from '../../package.json';
+import { SchemaType } from '../types';
 
 it('successfully parses config.json', () => {
   const config = JSON.parse(
@@ -153,12 +159,12 @@ describe('nodeSettingsSchema', () => {
       },
       httpGateway: {
         enabled: true,
-        apiKey: 'sameApiKey',
+        apiKey: '0x13dea3311fe0d6b84f4daeab831befbc49e19e6494c41e9e065a09c3c68f43b6',
         maxConcurrency: 10,
       },
       httpSignedDataGateway: {
         enabled: true,
-        apiKey: 'sameApiKey',
+        apiKey: '0x13dea3311fe0d6b84f4daeab831befbc49e19e6494c41e9e065a09c3c68f43b6',
         maxConcurrency: 10,
       },
     };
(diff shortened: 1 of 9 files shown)


packages/airnode-validator/src/config/config.ts
```diff
@@ -5,9 +5,12 @@ import { version as packageVersion } from '../../package.json';
 import { OIS, oisSchema } from '../ois';
 import { SchemaType } from '../types';
 
+export const evmAddressSchema = z.string().regex(/^0x[a-fA-F0-9]{40}$/);
+export const evmIdSchema = z.string().regex(/^0x[a-fA-F0-9]{64}$/);
+
 export const triggerSchema = z
   .object({
-    endpointId: z.string(),
+    endpointId: evmIdSchema,
```
amarthadan, line 13: Does this actually make sense? We're not enforcing the endpoint ID to be constructed from the endpoint name and OIS title (even though we encourage it in the docs), why force the format at all? I think it can be a random string and it will work, right? At least that's what docs are saying.

`This convention of determining an endpointId is not enforced at the protocol-level. For example, one could choose to generate an endpointId randomly, and as long as requesters use the correct endpointId, this will not be an issue.`

packages/airnode-validator/src/config/config.test.ts
```diff
@@ -126,12 +139,12 @@ describe('nodeSettingsSchema', () => {
       },
       httpGateway: {
         enabled: true,
-        apiKey: 'sameApiKey',
+        apiKey: '0x13dea3311fe0d6b84f4daeab831befbc49e19e6494c41e9e065a09c3c68f43b6',
```
amarthadan, line 142: It doesn't really matter but an address as an API key seems weird.

packages/airnode-validator/src/config/config.ts
```diff
@@ -133,7 +135,7 @@ export const gatewaySchema = z.discriminatedUnion('enabled', [enabledGatewaySche
 export const enabledHeartbeatSchema = z
   .object({
     enabled: z.literal(true),
-    apiKey: z.string(),
+    apiKey: apiKeySchema,
```
amarthadan, line 138: Heartbeat doesn't technically have such a restriction but I guess it's fine. Make sure to check with the ChainAPI guys. I presume they're using UUIDs for API keys but just to make sure.

packages/airnode-validator/src/config/config.test.ts
```diff
@@ -293,3 +306,64 @@ describe('triggers references', () => {
...
+  const gateway: SchemaType<typeof enabledGatewaySchema> = {
+    enabled: true,
+    apiKey: 'e83856ed-36cd-4b5f-a559-c8291e96e17e',
+    maxConcurrency: 100,
+  };
+  const heartbeat: SchemaType<typeof enabledHeartbeatSchema> = {
+    enabled: true,
+    apiKey: 'e83856ed-36cd-4b5f-a559-c8291e96e17e',
+    id: 'some-id',
+    url: 'https://www.uuidgenerator.net/version4',
```
amarthadan, line 320: :joy:

packages/airnode-validator/src/ois/ois.ts
```diff
@@ -37,10 +37,13 @@ export const fixedParameterSchema = z
 
 export const methodSchema = z.union([z.literal('get'), z.literal('post')]);
 
+// Path name must start wih "/" and must not contain space character
+export const pathNameSchema = z.string().regex(/^\/[^\s]+$/);
```
amarthadan, line 41: The path can be just a `/`, right? So, the regex should be `/^\/[^\s]*$/`. Adding a test for this edge case might also be a good idea.

amarthadan: Can really anything but whitespaces be in the path? What about another `/`? And random characters? Is `/@#%/%&*"\` a correct path?

packages/airnode-validator/src/ois/ois.ts
```diff
@@ -367,12 +367,19 @@ const ensureEndpointAndApiSpecificationParamsMatch: SuperRefinement<{
   });
 };
 
+export const semverSchema = z.string().refine((value) => {
+  const semver = value.split('.');
+  if (semver.length !== 3) return false;
+
+  return !semver.find((part) => /^\d+$/.test(part) === false);
+}, 'Expected semantic versioning "x.y.z"');
+
```
amarthadan, line 385: Can't this be done with a simple regex?

packages/airnode-validator/src/ois/ois.ts
```diff
@@ -67,7 +67,16 @@ export const reservedParameterSchema = z
     default: z.string().optional(),
     fixed: z.string().optional(),
   })
-  .strict();
+  .strict()
+  .refine((value) => {
```
amarthadan, line 71: What's the difference between `refine` and `superRefine`? Why do we use `superRefine` everywhere else?

amarthadan (changes requested):

packages/airnode-validator/src/config/config.ts
Siegrift: It was part of the legacy validator rules, but yeah we might drop this. I personally do not care about this much and it can be a validator warning later.

packages/airnode-validator/src/config/config.ts
Siegrift: @andreogle does this work for you? I am enforcing our API keys (including heartbeat) to have length 30-120 characters.

packages/airnode-validator/src/ois/ois.ts
Siegrift: refine is a shorthand for super refine, but it's less flexible. One disadvantage is that it doesn't set the `path` to the offending code. For simple validations though, refine is less boilerplate.

packages/airnode-validator/src/ois/ois.ts
Siegrift: Not a simple one https://github.com/api3dao/airnode/blob/bb509ca451d9003e9113b2a5f99cad44bdb0b2d5/packages/airnode-validator/templates/1.0/ois.json#L13 (but this one is a bit more flexible)

Anyway, implementing this in JS is much cleaner and maintanable so it is a deliberate choice.

packages/airnode-validator/src/ois/ois.ts
Siegrift: Probably not, but I wasn't sure what exactly is allowed (found conflicting information on the web) so I only disallowed spaces. 

Good catch with only a path.

packages/airnode-validator/src/config/config.ts
bbenligiray: Yeah it would be nice to have this as a warning

packages/airnode-validator/src/config/config.test.ts
```diff
@@ -1,8 +1,21 @@
 import { readFileSync } from 'fs';
 import { join } from 'path';
 import { ZodError } from 'zod';
-import { Config, chainOptionsSchema, configSchema, nodeSettingsSchema, ChainOptions, NodeSettings } from './config';
+import { zip } from 'lodash';
```
andreogle, line 4:
```suggestion
import zip from 'lodash/zip';
```
although I guess this doesn't matter as it's a test file. I still haven't been able to find anything like an ESLint rule that could ban certain phrases like `from 'lodash';` 😞

packages/airnode-validator/src/config/config.ts
andreogle: I'm fine with that. ChainAPI does use UUIDs for API keys

andreogle (approved):

dcroote (commented):
Only a couple of minor comments

packages/airnode-validator/src/config/config.ts
```diff
@@ -104,22 +107,24 @@ export const chainOptionsSchema = z.discriminatedUnion('txType', [
 
 export const chainConfigSchema = z
   .object({
-    authorizers: z.array(z.string()),
+    authorizers: z.array(evmAddressSchema),
     blockHistoryLimit: z.number().optional(),
     contracts: chainContractsSchema,
     id: z.string(),
     minConfirmations: z.number().optional(),
```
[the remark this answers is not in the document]
Siegrift: I'll fix this in a separate PR, I made a note to myself to check the schema for optionals/ints/record keys.

packages/airnode-validator/src/ois/ois.ts
amarthadan: It doesn't need to be that complex, we don't use suffixes like `-beta`. It would be something like `^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$` (haven't tested it though). Anyway, don't really care much, up to you.

Siegrift: `^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$` will not accept `10.20.30` and this is exactly what I want to prevent. Reading the JS implementation is much easier to reason about and get right.

amarthadan: Why wouldn't it?
Screenshot_20220606_091044

packages/airnode-validator/src/ois/ois.ts
amarthadan: But wouldn't it be better to be more strict? We're using the `path` to build an HTTP request, right? Imagine a user passing something weird to the path that is not supported. I would prefer to catch that during the validation which will give us a meaningful message and we will immediately know what's the problem instead of letting it fail in Axios most likely without any meaningful error message.

packages/airnode-validator/src/ois/ois.ts
Siegrift: You're right, I missed that you have `\d*` there. I still prefer JS impl though.

packages/airnode-validator/src/ois/ois.ts
Siegrift: What regex do you suggest? For reference these are the links I found:
- https://stackoverflow.com/questions/11469882/validate-the-path-of-an-url
- https://stackoverflow.com/questions/12928781/regex-validate-a-url-path-with-no-query-params
- https://stackoverflow.com/questions/161738/what-is-the-best-regular-expression-to-check-if-a-string-is-a-valid-url

I'd like to prevent a scenario where we make an error out of a valid URL, since the API provider will not be able to deploy his Airnode. For that reason I am more inclined to less bulletproof validation.

amarthadan: Ok, fair point. Looking at those regexes it can get out of hand pretty quickly. Let's let it looser and we can change it if it becomes a problem.

amarthadan (approved):

PR #1149, opened 2022-06-02
Template request example

Siegrift:
https://api3dao.atlassian.net/browse/AN-769

Review commit by commit.


packages/airnode-examples/integrations/coingecko-template/config.example.json
```diff
@@ -54,7 +54,13 @@
...
     "httpSignedData": []
   },
-  "templates": [],
+  "templates": [
+    {
+      "templateId": "0x02834eb43d56133982b7d6e5aa8b466c7ea4ba0fadf697698c1fee0996bba0fc",
+      "endpointId": "0xd9e8c9bcc8960df5f954c0817757d2f7f9601bd638ea2f94e890ae5481681153",
+      "encodedParameters": "0x3173000000000000000000000000000000000000000000000000000000000000636f696e49640000000000000000000000000000000000000000000000000000657468657265756d000000000000000000000000000000000000000000000000"
+    }
+  ],
```
amarthadan, line 63: I'm a bit confused. I must have missed that, what is this `templates` field for? I presume that it's for defining templates outside the blockchain, right? So, why do we need `create-template-on-chain.ts` then?

packages/airnode-examples/scripts/make-request.ts
```diff
@@ -34,33 +31,12 @@ const waitForFulfillment = async (requestId: string) => {
 
 const makeRequest = async (): Promise<string> => {
```
amarthadan, line 32: TBH I don't like this change. It introduces a lot of duplicate code (which I get it's not such a big deal for examples but still) and you've removed imo important comments.
Wouldn't be better to create a separate `makeTemplateRequest` method and use that in that one specific example?

packages/airnode-examples/integrations/coingecko-template/config.example.json
Siegrift: The template must exist on chain in order for you to make a template request. When you hardcode a template in config.json, Airnode will not fetch it from chain (so we avoid a blockchain call).

packages/airnode-examples/scripts/make-request.ts
Siegrift: > and you've removed imo important comments.

I've removed them, because previously it was a generic `make-request.ts` script that decided that the first endpoint is going to be used. This has now been moved to the individual `request-utils.ts` file of the integration.

> Wouldn't be better to create a separate makeTemplateRequest method and use that in that one specific example?

No, because there is a generic `make-request.ts` which encapsulates the common parts of the examples and it has to call the same method of examples.

amarthadan (approved):

PR #1154, opened 2022-06-03
Reject configuration with both relay metadata security schemes and gateways

amarthadan:
[AN-405](https://api3dao.atlassian.net/browse/AN-405)

### DIFF
--- a/packages/airnode-validator/src/config/config.ts
+++ b/packages/airnode-validator/src/config/config.ts
@@ -1,8 +1,9 @@
 import { ethers } from 'ethers';
 import { SuperRefinement, z } from 'zod';
 import forEach from 'lodash/forEach';
+import includes from 'lodash/includes';
 import { version as packageVersion } from '../../package.json';
-import { OIS, oisSchema } from '../ois';
+import { OIS, oisSchema, RELAY_METADATA_TYPES } from '../ois';
 import { SchemaType } from '../types';
 
 export const evmAddressSchema = z.string().regex(/^0x[a-fA-F0-9]{40}$/);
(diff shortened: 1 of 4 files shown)


packages/airnode-validator/src/config/config.ts
```diff
@@ -319,7 +344,9 @@ export const configSchema = z
   .strict()
   .superRefine(validateSecuritySchemesReferences)
   .superRefine(validateTemplateSchemes)
-  .superRefine(validateTriggersReferences);
+  .superRefine(validateTriggersReferences)
+  // TODO: Add check that the security schemes referenced in security are present
```
amarthadan, line 348: @Siegrift Do we track this somewhere?

Siegrift (changes requested):
I'd suggest improving the naming, but otherwise LGTM.

packages/airnode-validator/src/config/config.ts
amarthadan: Not really. That one is checking that if the security scheme is specified you have API credentials specified as well. I'm talking about a check that if you have something defined in `security` it needs to be available in `components.securitySchemes`. This check will just ignore if the security scheme defined in `security` doesn't exist.

Siegrift: I see. This is checked in https://github.com/api3dao/airnode/blob/reject-relay-metadata-gateway/packages/airnode-validator/src/ois/ois.ts#L178

amarthadan: Oh, ok, didn't notice that. In that case, I will have to change the code a bit. I thought that if it's caught early on I don't need to check for the field existence. But apparently the check is there and I still need to :smile:

Siegrift (approved):
👍 LGTM

PR #1155, opened 2022-06-03
Adjust memory allocation for gateway serverless functions

amarthadan:
https://api3workspace.slack.com/archives/C02AYRX8D89/p1654257231432529?thread_ts=1654185413.354569&cid=C02AYRX8D89


Siegrift (approved):
I think both AWS and GCP should be updated (like implemented in this PR, but wait for Buraks answer in Slack before merging).

PR #1156, opened 2022-06-03
[v0.6] Adjust memory allocation for gateway serverless functions

amarthadan:
Backporting changes from https://github.com/api3dao/airnode/pull/1155 to v0.6

### DIFF
--- a/.changeset/wise-pots-wait.md
+++ b/.changeset/wise-pots-wait.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-deployer': patch
+---
+
+Adjust memory allocation for gateway serverless functions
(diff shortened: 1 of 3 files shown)


Siegrift (approved):
I think both AWS and GCP should be updated (like implemented in this PR, but wait for Buraks answer in Slack before merging).

Btw. can you fix the documentation link? It seems there is a broken validator link.

amarthadan:
> Btw. can you fix the documentation link? It seems there is a broken validator link.

I was ignoring that as I thought `0.6.2` will be the last `0.6.x` release :sweat_smile: I'll fix it if `0.6.5` will become a thing :see_no_evil:

Siegrift:
Why not fix it now? :D

PR #1157, opened 2022-06-05
Add upper bound to Amount numbers

dcroote:
Rationale explained well by the ethers `BigNumber` [documentation](https://docs.ethers.io/v5/api/utils/bignumber/#BigNumber--notes-safenumbers).

### DIFF
--- a/packages/airnode-validator/src/config/config.ts
+++ b/packages/airnode-validator/src/config/config.ts
@@ -52,7 +52,7 @@ export const providerSchema = z
 
 export const amountSchema = z
   .object({
-    value: z.number(),
+    value: z.number().lte(9007199254740991), // 2**53 - 1
     unit: z
       .union([
         z.literal('wei'),
--- a/.changeset/fast-kangaroos-bake.md
+++ b/.changeset/fast-kangaroos-bake.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-validator': patch
+---
+
+Add upper bound to Amount numbers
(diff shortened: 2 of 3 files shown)


packages/airnode-validator/src/config/config.ts
```diff
@@ -52,7 +52,7 @@ export const providerSchema = z
 
 export const amountSchema = z
   .object({
-    value: z.number(),
+    value: z.number().lte(9007199254740991), // 2**53 - 1
```
Siegrift, line 55:
```suggestion
    value: z.number().lte(2**53 - 1),
```

Siegrift, line 55: Do we have this problem only with the amount field?

Siegrift (changes requested):

packages/airnode-validator/src/config/config.ts
dcroote: I chose to hardcode it in order to avoid 2**53 since that cannot be accurately represented as an integer.

Siegrift: Yeah, that's a good point.

dcroote: This would be relevant theoretically anywhere we might anticipate the user hardcodes a very very large **number**. My motivation here (though it is still probably unlikely) is protecting against a user denominating in `wei` instead of e.g. `ether` for an amount. 

Looking through `config.json` I think we're safe otherwise; the only other instance where I initially thought this might be relevant was the `_times` reserved parameter, but we're safe because that is a [string and not a number](https://github.com/api3dao/airnode/blob/bd4becb68ba334958b598f5a56e0e31278b0a71d/packages/airnode-validator/src/ois/ois.ts#L64).

I also looked through our codebase and couldn't find any instances where we manually specify overly large numbers (by searching for `000_000`, `2**`, `10**`, `.pow` ).

Siegrift (approved):
Makes sense.

PR #1158, opened 2022-06-05
Improve TS project references structure, fix publish configuration for airnode-examples

Siegrift:
https://api3dao.atlassian.net/browse/AN-762

Review commit by commit. I am sorry that the first commit is quite a large one, but I first wanted to make sure the configuration is fully working before committing.


jest.config.base.js
```diff
@@ -83,7 +83,7 @@ module.exports = {
   moduleNameMapper: null,
 
   // An array of regexp pattern strings, matched against all module paths before considered 'visible' to the module loader
-  modulePathIgnorePatterns: ['<rootDir>/.build', '<rootDir>/dist/'],
+  modulePathIgnorePatterns: ['<rootDir>/.build', '<rootDir>/dist/', '<rootDir>/build/'],
```
Siegrift, line 86: Currently, only airnode-validator uses the build folder, but I've put it to the root jest file because:
- we should not ever have tests in "build" folder
- was the easy choice

Validator needs to us the "build" folder, because projects which are referenced need to be set `composite` to `true`. This setting also implies that `noEmit` cannot be set to `true` and TS must be allowed to output the build files somewhere.

packages/airnode-abi/package.json
```diff
@@ -6,12 +6,13 @@
   "main": "dist/index",
   "types": "dist/index",
   "files": [
-    "dist"
+    "dist",
+    "!dist/**/*.tsbuildinfo"
```
Siegrift, line 10: You can use `npm publish --dry-run` to see what gets published. I checked every package and found two problems:
- `tsbuildinfo` files were published (addressed here)
- examples do not have `files` property set (addressed in https://github.com/api3dao/airnode/pull/1158/commits/1b504dc420fc406eed2a35957ce385bab8ac1bf0)

packages/airnode-examples/src/scripts/choose-integration.ts
```diff
@@ -3,7 +3,7 @@ import { readdirSync, writeFileSync } from 'fs';
 import { join } from 'path';
 import { PromptObject } from 'prompts';
 import { goSync } from '@api3/airnode-utilities';
-import { cliPrint, IntegrationInfo, promptQuestions, runAndHandleErrors } from '../src';
+import { cliPrint, IntegrationInfo, promptQuestions, runAndHandleErrors } from '../';
```
Siegrift, line 6: I know importing from `../` is awkward, but it's fundamentally not wrong, so I left it like this (instead of importing from the individual files).

.changeset/calm-pots-serve.md
```diff
@@ -0,0 +1,15 @@
+---
+'@api3/airnode-abi': patch
```
Siegrift, line 2: I chose `patch`, since this is mostly fixing the previous PR which introduced the TS references.

amarthadan (commented):
The code looks pretty much fine but the issue described in the ticket is still there.

packages/airnode-operation/.gitignore
```diff
@@ -1,5 +1,5 @@
 node_modules
-dist
+/dist
```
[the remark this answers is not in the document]
Siegrift: While I was doing the changes I touched some of the `.gitignores` and made them consistent before commiting.

image

packages/airnode-adapter/package.json
```diff
@@ -3,15 +3,16 @@
   "license": "MIT",
   "version": "0.6.0",
   "private": false,
-  "main": "dist/src/index",
-  "types": "dist/src/index",
+  "main": "dist/index",
+  "types": "dist/index",
```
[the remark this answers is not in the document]
Siegrift: It worked before and works now. What changed was TS config. (If I remember correctly, the `includes` field affected the output folder structure).

packages/airnode-examples/package.json
```diff
@@ -5,7 +5,7 @@
   "private": false,
   "description": "A collection of examples showcasing the features of Airnode",
   "scripts": {
-    "build": "yarn run clean && hardhat compile",
+    "build": "yarn run clean && hardhat compile && yarn tsc --build tsconfig.json",
```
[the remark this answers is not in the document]
Siegrift: Yes, this enforces that we actually fix the TS errors. Otherwise, they would only show up in the IDE. Alternatively, we could run `yarn tsc --build tsconfig.json` from the root (because it references all monorepo projects). But this way feels more natural.

Siegrift (commented):
> The code looks pretty much fine but the issue described in the ticket is still there.

Yeah, resolved in https://github.com/api3dao/airnode/pull/1158/commits/2c97cd4e88097f647746c2bf8fc364717b9b6b27. I forgot to add the tests inside the source folder to the tsconfig. I remember testing the isue in the ticket, but I probably used a file inside one of the `test` folders.

amarthadan:
The issue I described is fixed. If I cause a TS error in a test in `airnode-adapter` both running `yarn build` and `yarn test` within the `airnode-adapter` will reveal the error. Running `yarn build` in for example `airnode-node` (which has `airnode-adapter` as a dependency) won't show the error but I don't consider it a problem.

Unfortunately, these changes break the packaging :confused: Try running `yarn run pack` and check the archives on this branch and on `master`. You'll see that on this branch basically everything in the directory is packed which is not what we want.

Siegrift:
> Running yarn build in for example airnode-node (which has airnode-adapter as a dependency) won't show the error but I don't consider it a problem

Just to clarify: This is intended. If you look at the tsconfigs of the monorepo packages there are no references to other packages. This means that the builds work like before, however the "navigate to source code" still works, since the packages have `declarationMap` enabled.

> Unfortunately, these changes break the packaging 😕 Try running yarn run pack and check the archives on this branch and on master. You'll see that on this branch basically everything in the directory is packed which is not what we want.

I used to run `npm publish --dry-run` which packed the correct files. I came across this issue bug in yarn, which caused this: https://github.com/yarnpkg/yarn/issues/8332 (looking into it)

Siegrift:
I've replaced "yarn pack" with "npm pack" in https://github.com/api3dao/airnode/pull/1158/commits/396e602b64175089b615cc959b1d696bb3e1d960 which parses the files section correctly...

PR #1159, opened 2022-06-05
Strenghten schema

Siegrift:
https://api3dao.atlassian.net/browse/AN-770

Review commit by commit.

While implementing this change I realized that the validation of the parameters could be more thorough and I've implemented it as the last commit.

### DIFF
--- a/packages/airnode-validator/src/config/config.ts
+++ b/packages/airnode-validator/src/config/config.ts
@@ -110,10 +110,10 @@ export const chainOptionsSchema = z.discriminatedUnion('txType', [
 export const chainConfigSchema = z
   .object({
     authorizers: z.array(evmAddressSchema),
-    blockHistoryLimit: z.number().optional(),
+    blockHistoryLimit: z.number().int().optional(),
     contracts: chainContractsSchema,
     id: z.string(),
-    minConfirmations: z.number().optional(),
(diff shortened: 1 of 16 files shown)


packages/airnode-validator/src/config/config.ts
```diff
@@ -86,8 +86,8 @@ export const chainOptionsSchema = z.discriminatedUnion('txType', [
     .object(
       {
         txType: z.literal('eip1559'),
-        baseFeeMultiplier: z.number().int().optional(),
-        priorityFee: priorityFeeSchema.optional(),
+        baseFeeMultiplier: z.number().int().optional(), // Defaults to BASE_FEE_MULTIPLIER defined in airnode-utilities
```
Siegrift, line 89: I actually wanted to import the constants from airnode-utilities and use https://github.com/colinhacks/zod#default to make this explicit (because this can get out of sync easily if we make the field required).

I tried implementing that in https://github.com/api3dao/airnode/pull/1159/commits/46aed308ab237896b39cce6083c292a4616d4061 but I ended up reverting those changes, because lot of unit tests started to fail due to TS errors (because `default()` from zod makes the the field required and our fixtures needed to be updated).

amarthadan (approved):

PR #1171, opened 2022-06-07
Reject invalid secrets

Siegrift:
https://api3dao.atlassian.net/browse/AN-471

Turns out we already handle this. At first, I thought I will add a more descriptive error message (saying make sure the secret is a valid identifier), but this makes the error message inaccurate when there is a missing secret which will be the most common error. I've simplified the CLI error message a bit though.

### DIFF
--- a/packages/airnode-validator/src/api/api.ts
+++ b/packages/airnode-validator/src/api/api.ts
@@ -36,12 +36,17 @@ export function parseConfig(config: unknown): ValidationResult<Config> {
   return configSchema.safeParse(config);
 }
 
+const secretNamePattern = /^[A-Z][A-Z0-9_]*$/;
+export const secretNameSchema = z
+  .string()
+  .regex(secretNamePattern, `Secret name is not a valid. Secret name must match ${secretNamePattern.toString()}`);
+export const secretsSchema = z.record(secretNameSchema, z.string());
+
 /**
  * @param secrets a key value object with the secrets
  * @returns `{success: true, data: <secrets>}` if successful, `{success: false, error: <error>}` otherwise
  */
 export function parseSecrets(secrets: unknown): ValidationResult<Secrets> {
-  const secretsSchema = z.record(z.string(), z.string());
   return secretsSchema.safeParse(secrets);
 }
 
--- a/packages/airnode-validator/src/api/api.ts
+++ b/packages/airnode-validator/src/api/api.ts
@@ -14,11 +14,16 @@ import { ValidationResult } from '../validation-result';
  * @returns `{success: true, data: <interpolated config>}` if successful, `{success: false, error: <error>}` otherwise
  */
 export function parseConfigWithSecrets(config: unknown, secrets: unknown): ValidationResult<Config> {
-  const parseSecretsRes = parseSecrets(secrets);
-  if (!parseSecretsRes.success) return parseSecretsRes;
+  const parsedSecrets = parseSecrets(secrets);
+  if (!parsedSecrets.success) return parsedSecrets;
 
-  const interpolateConfigRes = interpolateSecrets(config, parseSecretsRes.data);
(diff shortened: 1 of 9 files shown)


packages/airnode-validator/src/api/api.ts
```diff
@@ -14,11 +14,16 @@ import { ValidationResult } from '../validation-result';
  * @returns `{success: true, data: <interpolated config>}` if successful, `{success: false, error: <error>}` otherwise
  */
 export function parseConfigWithSecrets(config: unknown, secrets: unknown): ValidationResult<Config> {
-  const parseSecretsRes = parseSecrets(secrets);
-  if (!parseSecretsRes.success) return parseSecretsRes;
+  const parsedSecrets = parseSecrets(secrets);
+  if (!parsedSecrets.success) return parsedSecrets;
```
amarthadan, line 18: I kinda like the `*Res` naming for variables that contain the `go` result. I know what can I expect there just by reading the name. But that's just my preference, feel free to ignore it.

amarthadan (approved):
I'm fine with this even though some messages are a bit confusing. For example message `Secrets interpolation failed. Caused by: STAGE is not defined` for a secret named `STAGE-NAME`.

But I thought that you want to go through the secrets' names, check that they are valid and throw an error saying that such a name can't be used if they're not. That would imo make more sense from the user's point of view.

We can merge this and do that in a separate PR or decide this is good enough. I'm fine either way.

amarthadan:
> > But I thought that you want to go through the secrets' names, check that they are valid and throw an error saying that such a name can't be used if they're not. That would imo make more sense from the user's point of view.
> 
> We can do that, but at that point I'd restrict the secrets name behind a convention (e.g. capital english letters + "_" character) to make sure they are valid identifier. WDYT?

I'm fine with that. Maybe allowing numbers as well? For something like `PROVIDER_URL_01`, `PROVIDER_URL_02`? Not sure how ChainAPI and operations generate these.

Siegrift:
@amarthadan Please review the last commit, I've changed the implementation.

Siegrift:
@amarthadan I've realized why e2e tests fail. It's because when you run Airnode in a docker, the ENV vars are merged with environment variables and we check every variable of `process.env` and there can be secrets that don't follow our convention.

See https://github.com/api3dao/airnode/pull/1171/commits/588fd678ebdab8cc51e5202636d935a739bba7df where I reworked the validation to only use the secrets loaded from the env file.

amarthadan:
> @amarthadan I've realized why e2e tests fail. It's because when you run Airnode in a docker, the ENV vars are merged with environment variables and we check every variable of `process.env` and there can be secrets that don't follow our convention.
> 
> See [588fd67](https://github.com/api3dao/airnode/commit/588fd678ebdab8cc51e5202636d935a739bba7df) where I reworked the validation to only use the secrets loaded from the env file.

Yeah, that solves it for validator's CLI but it still doesn't work if you want to invoke (`yarn dev:invoke`) Airnode locally :confused:

amarthadan:
> > @amarthadan I've realized why e2e tests fail. It's because when you run Airnode in a docker, the ENV vars are merged with environment variables and we check every variable of `process.env` and there can be secrets that don't follow our convention.
> > See [588fd67](https://github.com/api3dao/airnode/commit/588fd678ebdab8cc51e5202636d935a739bba7df) where I reworked the validation to only use the secrets loaded from the env file.
> 
> Yeah, that solves it for validator's CLI but it still doesn't work if you want to invoke (`yarn dev:invoke`) Airnode locally confused

Sorry, this actually will work for `yarn dev:invoke`, that's the whole point :man_facepalming:
The issue I ran into is connected to https://github.com/api3dao/airnode/pull/1172 which is pretty much the same thing but instead of wrong variable names we're dealing with empty variables that are common in the system.

PR #1172, opened 2022-06-07
Error on empty secrets

Siegrift:
https://api3dao.atlassian.net/browse/AN-486


packages/airnode-validator/src/api/api.ts
```diff
@@ -61,6 +61,12 @@ const NO_MATCH_REGEXP = /($^)/;
 const ES_MATCH_REGEXP = /\$\{([^\\}]*(?:\\.[^\\}]*)*)\}/g;
 
 function interpolateSecrets(config: unknown, secrets: Secrets): ValidationResult<unknown> {
+  const emptySecret = Object.entries(secrets).find(([_key, value]) => !value);
```
amarthadan, line 64: Can't we just do
```
const secretsSchema = z.record(z.string(), z.string().min(1));
```
?
That would throw on an empty secret, wouldn't it?

packages/airnode-validator/src/api/api.ts
```diff
@@ -61,6 +61,12 @@ const NO_MATCH_REGEXP = /($^)/;
...
+  if (emptySecret) {
+    const [secretName] = emptySecret;
+    return { success: false, error: new Error(`Secret "${secretName}" has empty value`) };
```
andreogle, line 67:
```suggestion
    return { success: false, error: new Error(`Secret "${secretName}" has an empty value`) };
```

packages/airnode-validator/src/api/api.ts
Siegrift: That's elegant. Can't see why it wouldn't work. WIll take a look.

Siegrift: We can do it, but the error message looks worse imo:
```
    -   "error": [Error: Secrets interpolation failed. Caused by: Secret "AIRNODE_WALLET_MNEMONIC" has empty value],
    +   "error": [ZodError: [
    +   {
    +     "code": "too_small",
    +     "minimum": 1,
    +     "type": "string",
    +     "inclusive": true,
    +     "message": "String must contain at least 1 character(s)",
    +     "path": [
    +       "AIRNODE_WALLET_MNEMONIC"
    +     ]
    +   }
    + ]]
```

I'd keep it the way it is implemented now.

amarthadan: Can't you do `z.string().min(1, {message: "Secret can't be empty"})`? :sweat_smile:
But I'm fine with the current way.

Siegrift: That just affects the message, but I dislike the "boilerplatish" ZodError response. (I know I could just extract the message, but it all feels unnecessary since the current implementation does this already)

PR #1177, opened 2022-06-10
Fail validation with empty secret value

dcroote:
[AN-776](https://api3dao.atlassian.net/browse/AN-776)

Note `emptySecret` has been removed from `interpolateSecrets` as the secrets are parsed first within `parseConfigWithSecrets` then interpolated so the `.min(1)` catches empty secrets.

### DIFF
--- a/packages/airnode-validator/src/api/api.ts
+++ b/packages/airnode-validator/src/api/api.ts
@@ -40,7 +40,7 @@ const secretNamePattern = /^[A-Z][A-Z0-9_]*$/;
 export const secretNameSchema = z
   .string()
   .regex(secretNamePattern, `Secret name is not a valid. Secret name must match ${secretNamePattern.toString()}`);
-export const secretsSchema = z.record(secretNameSchema, z.string());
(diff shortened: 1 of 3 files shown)


amarthadan (approved):

packages/airnode-validator/src/api/api.ts
```diff
@@ -40,7 +40,7 @@ const secretNamePattern = /^[A-Z][A-Z0-9_]*$/;
 export const secretNameSchema = z
   .string()
   .regex(secretNamePattern, `Secret name is not a valid. Secret name must match ${secretNamePattern.toString()}`);
-export const secretsSchema = z.record(secretNameSchema, z.string());
+export const secretsSchema = z.record(secretNameSchema, z.string().min(1, { message: 'Secret cannot be empty' }));
```
Siegrift, line 43: As mentioned [here](https://github.com/api3dao/airnode/pull/1172#discussion_r892315600) I disliked the verbose error message, but I don't mind keeping this.

PR #1178, opened 2022-06-11
Update ESLint and related packages

andreogle:
## Overview

I noticed that we were quite far behind with some ESLint and related packages. Dependabot hasn't been updating these ones so I just decided to update them myself


packages/airnode-protocol/src/index.ts
```diff
@@ -56,5 +56,5 @@ export {
   FailedRequestEvent,
   RequestedWithdrawalEvent,
   FulfilledWithdrawalEvent,
-} from './contracts/AirnodeRrpV0';
+} from './contracts/AirnodeRrpV0'; // eslint-disable-line import/no-unresolved
```
andreogle, line 59: I don't really like this but `eslint-plugin-import` didn't seem to want to work with ESLint v8

Siegrift: This file is autogenerated. Are you sure you've generated it beforehand?

andreogle: Are you sure this file is generated? There's a custom comment at the top. Or do you mean the contents of `./contracts/AirnodeRrpV0` are generated? I tried running `yarn build` again but this line is still an issue.

It [has been a problem before](https://github.com/api3dao/airnode/pull/1004#issuecomment-1096152730). We ignored the version then, but now we can't upgrade to ESLint v8 without bumping this package too. I can add a bit more context to this comment though if that helps?

Siegrift (approved):
👍 Left one comment, LGTM otherwise.

PR #1228, opened 2022-06-14
Improve coingecko-signed-data examples integration

### LINKED ISSUE #1199: Improve coingecko-signed-data integration
From https://api3dao.atlassian.net/browse/AN-557 :
* We should prevent users from choosing the local airnode client flow as the docker client image does not have http gateway functionality
* We should print the decoded value as well as the raw JSON string

Also, from the perspective of the user, I would have the following questions:
* What is the API key and where does it come from?
* Where does the templateId value come from?

These could potentially be answered with link(s) to the docs from the integration README

dcroote:
Closes #1199

### DIFF
--- a/.changeset/late-pans-occur.md
+++ b/.changeset/late-pans-occur.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-examples': patch
+---
+
+Improve coingecko-signed-data examples integration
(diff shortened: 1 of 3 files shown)


Siegrift (approved):
👍 LGTM

PR #1238, opened 2022-06-15
Update max fulfillment size

Siegrift:
https://api3dao.atlassian.net/browse/AN-784

### DIFF
--- a/packages/airnode-adapter/src/response-processing/extraction.test.ts
+++ b/packages/airnode-adapter/src/response-processing/extraction.test.ts
@@ -282,7 +282,7 @@ describe('respects maximum encoded size', () => {
       expect(encodedSize('string', str)).toBe(MAX_ENCODED_RESPONSE_SIZE + 32);
 
       expect(() => extractAndEncodeResponse(str, { _type: 'string' })).toThrow(
-        'Encoded value exceeds the maximum allowed size (1024 bytes)'
+        'Encoded value exceeds the maximum allowed size (16384 bytes)'
       );
     });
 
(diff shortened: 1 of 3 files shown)


dcroote:
I saw the tests had failed so I updated them as my understanding was this is holding back `0.7`

packages/airnode-adapter/src/response-processing/extraction.test.ts
```diff
@@ -295,14 +295,14 @@ describe('respects maximum encoded size', () => {
   });
 
   describe('encoding int256 arrays', () => {
-    const maxInt256ArrayLength = 30;
+    const maxInt256ArrayLength = 510;
 
     it('is under the limit', () => {
```
amarthadan, line 300: Test case names are switched. It should say `exceeds the limit` and `exceeds the limit` one should say `is under the limit`.

amarthadan: Same goes for the string tests above.

packages/airnode-adapter/src/response-processing/extraction.test.ts
```diff
@@ -295,14 +295,14 @@ describe('respects maximum encoded size', () => {
   });
 
   describe('encoding int256 arrays', () => {
-    const maxInt256ArrayLength = 30;
+    const maxInt256ArrayLength = 510;
```
amarthadan, line 298: I presume this is correct but out of curiosity, can somebody explain the math behind it? :smile:

amarthadan (approved):

packages/airnode-adapter/src/response-processing/extraction.test.ts
dcroote: Previously: `32 * 30 = 1024 - 64`
Now: `32 * 510 = 16384 - 64`

And my understanding for the reason to subtract 64 is:

```ts
const arrLen = ethers.utils.arrayify(ethers.utils.defaultAbiCoder.encode(['int256[]'], [[]])).byteLength;
console.log(arrLen);
```
Prints
```sh
64
```
Same as does:
```ts
const arrLen = ethers.utils.arrayify(ethers.utils.defaultAbiCoder.encode(['string'], [''])).byteLength;
```
hence:
https://github.com/api3dao/airnode/blob/5dd3b4a7dd387c413bee154c73f33f40511d14a8/packages/airnode-adapter/src/response-processing/extraction.test.ts#L278

Issue #1182, opened 2022-06-13
Make Airnode use internal chain references instead of log chainIds


amarthadan:
Airnode trusts the `chainId` reported by the chain provider in retrieved log events. These `chainId`s are used to match up API requests with provider config. This breaks the security rule of "Don't trust the client".

Ideally, Airnode should use the config index of the chain when retrieving logs and carry that absolute index through so that API calls are matched up using a trusted reference.

This is an example of where a configured chain is retrieved based on a provider-sourced `chainId` (from AirnodeRrp’s event log): https://github.com/api3dao/airnode/commit/44de4f1045b7fb126e1effab48fdc54e17e50e5e#diff-fe629d3de1ed4c9f10a71f38e083d007938620d9f5c81fd6c9fda5bbfac62a97R45

kolenic-martin:
Tagging @vponline for some more instructions (sorry, Michal). Thanks in advance for any instructions.

vponline:
Sorry for not getting back to this earlier, but I am afraid that I also don't quite know how to try to test that this works (besides running the airnode-examples and confirming that they still work).

This change is supposed to protect against the blockchain provider reporting a different chainId when the transaction logs are fetched, compared to what the chainId is in the `config.json`. But I'm not quite sure how we could 'fake' the provider giving us a wrong chainId to compare that the code is actually using the chainId from the config 🤔

amarthadan:
Hmm, I don't see how to easily test this from the user perspective either. Any ideas @Siegrift? If not, I suggest moving it to `Done`.

PR #1239, opened 2022-06-17
Use config chainId instead of the chainId from event logs

vponline:
#1182


vponline:
> There are basically two sources of chain, one is what we have in the config, and one that is passed as an argument. But now we don't need that check because we're passing chain ID early on from the config. Or am I misreading it?

I think I am the one misunderstanding the task because I have only changed where the `chainId` is coming from when the logs are fetched, i.e. I set it from the `FetchOptions` (instead of the logs data) [here](https://github.com/api3dao/airnode/pull/1239/files#diff-81b820a01f51db14bd135e9e9d34a499f2dd6da6be6551bbbb09cd075c568c6bR33): 

```ts
 const logsWithBlocks = rawLogs.map((log) => ({
     ...
    chainId: options.chainId,
    ...
  })) as EVMEventLog[];
```

But we are still matching the (chain) config based on this `chainId` from the log object (the destructured `chainId`  from `aggregatedApiCall`):

```ts
const { airnodeAddress, requesterAddress, sponsorAddress, sponsorWalletAddress, endpointId, id, chainId } =
        aggregatedApiCall;
const chain = config.chains.find((c) => c.id === chainId)!;
```

So to remove this matching I should be passing in the array index of the chain config (I just realized the task specifically said index). Sorry about this, I'll make the change

Siegrift:
Mhh, and what is the chainId in config.json for?

amarthadan:
> Mhh, and what is the chainId in config.json for?

What do you mean?

Issue #1194, opened 2022-06-13
Automatically recognize integrated chains


amarthadan:
Currently, one needs to update [this](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/evm/networks.ts) when there is a new chain integration. However, airnode-protocol already knows everything about these chains, so it can export these to be used here.

* Update the [generate-references script](https://github.com/api3dao/airnode/blob/master/packages/airnode-protocol/package.json#L20) to also generate the required information about the chains and store in [references.json](https://github.com/api3dao/airnode/blob/master/packages/airnode-protocol/deployments/references.json) to be [exported](https://github.com/api3dao/airnode/blob/master/packages/airnode-protocol/src/index.ts)
* Import this information in airnode-node instead of hardcoding it

This needs to refer to https://github.com/api3dao/airnode/blob/master/packages/airnode-protocol/deployments/references.json

kolenic-martin:
Hi @amarthadan, is there any way to test this for me? I have not generated a chain integration like ever so I might need some guidance on this. I'll be happy to learn, though.

vponline:
@kolenic-martin  You can test this one by going to `packages/airnode-protocol`.

In there you can delete the `deployments/references.json` file and then run `yarn deploy:generate-references`.

This script should re-generate the `references.json` file and it should contain the `networks` field:
```ts
"networks": {
    "1": {
      "chainId": 1,
      "name": "homestead"
    },
...
```

kolenic-martin:
Hi @vponline, I have proceeded as you instructed, i.e.:
- cd packages/airnode-protocol
- rm deployments/references.json
- yarn deploy:generate-references
- cat deployments/references.json
- [press PrtSc]
![image](https://user-images.githubusercontent.com/884979/174792944-af83a176-ff6f-4054-a7c5-ea7fe7a666e0.png)

This works just fine then :)

PR #1240, opened 2022-06-17
Add networks mapping to airnode-protocol references.json

vponline:
#1194


packages/airnode-protocol/src/index.ts
```diff
@@ -19,10 +20,11 @@ import {
 } from './contracts';
 import references from '../deployments/references.json';
 
-const AirnodeRrpAddresses: { [chainId: number]: string } = references.AirnodeRrpV0;
-const AccessControlRegistryAddresses: { [chainId: number]: string } = references.AccessControlRegistry;
-const RequesterAuthorizerWithAirnodeAddresses: { [chainId: number]: string } =
+const AirnodeRrpAddresses: { [chainId: string]: string } = references.AirnodeRrpV0;
+const AccessControlRegistryAddresses: { [chainId: string]: string } = references.AccessControlRegistry;
+const RequesterAuthorizerWithAirnodeAddresses: { [chainId: string]: string } =
```
vponline, line 25: I also noticed the types for the keys were set to numbers, but they should actually be strings so changed them here. It doesn't really affect any functionality though.

amarthadan (approved):
LGTM

packages/airnode-protocol/src/index.ts
```diff
@@ -19,10 +20,11 @@ import {
...
   references.RequesterAuthorizerWithAirnode;
+const networks: { [chainId: string]: ethers.providers.Network } = references.networks;
```
andreogle, line 27: Do we not want to keep this as the uppercase `NETWORKS` as it's a constant?

PR #1262, opened 2022-06-20
Improve mnemonic visibility in `generate-mnemonic` command

### LINKED ISSUE #1192: Improve the visibility of mnemonic in `generate-mnemonic` command
[Slack discussion](https://api3workspace.slack.com/archives/C01N6001YCE/p1649877336726019)

Burak: I think the nicest way would be to print the mnemonic in red

amarthadan:
Close #1192


Siegrift (changes requested):
Can you add a test please?

amarthadan:
> Can you add a test please?

Of course, sorry, I haven't realized there are e2e tests for this :man_facepalming:

packages/airnode-admin/test/e2e/cli.feature.ts
```diff
@@ -442,16 +442,22 @@ Template data:
     ];
     expect(out.slice(0, 3)).toEqual(explanationInfo);
 
-    const mnemonic = out[3];
+    const titleRows = [out[3], out[7]];
```
Siegrift, line 445: Mhh, could we make this more readable? Something along thse lines:
```js
###########
#${menmonic.padStart(n).padEnd(n)}#
###########
```

packages/airnode-admin/src/cli.ts
```diff
@@ -514,12 +514,29 @@ yargs
       const airnodeAddress = await admin.deriveAirnodeAddress(mnemonic);
       const airnodeXpub = admin.deriveAirnodeXpub(mnemonic);
 
+      const mnemonicLength = mnemonic.length;
```
Siegrift, line 517: Could we abstract this into a function and unit test the ASCII art output? It would be nice to ensure it's properly displayed.

packages/airnode-admin/test/e2e/cli.feature.ts
amarthadan: I don't see how using `padStart` or `padEnd` will help with readability as it needs the whole length as an argument so you can't really chain them. You would have to calculate the proper length anyway, I'd say even in a more complex way than just doing `/2`.

Siegrift: Yeah, true. Maybe something like: 

```
const space = ' '.repeat(n)
const line = `#${space.concat(mnemonic).concat(space)}#`
```

amarthadan: That's pretty much the same thing, you're just using `concat` instead of variable interpolation.

packages/airnode-admin/src/utils.test.ts
```diff
@@ -0,0 +1,27 @@
+import { encaseMnemonic } from './utils';
+
+describe('encaseMnemonic', () => {
+  it('draws a border for mnemonic of an even length', () => {
```
Siegrift, line 4: Nit: It would be nice to assert the parity of the string length.

packages/airnode-admin/src/utils.ts
```diff
@@ -0,0 +1,15 @@
+export const encaseMnemonic = (mnemonic: string) => {
```
Siegrift, line 1: Nit: maybe change the file name to `mnemonic` or `mnemonic-utils`, since utils should hold generic functions.

Siegrift (approved):
👍 LGTM.

PR #1275, opened 2022-06-21
Remove gateway URLs from receipt.json

### LINKED ISSUE #1225: Remove HTTP gateway endpoint URL from receipt.json
We add the HTTP gateway endpoint URL in `receipt.json` even though this is practically a secret. Now that we’re receiving the URL through the heartbeat, we can remove it from the receipt, which makes the receipt pushable to public repos.

Note that we will need to display the URL to the deploying user in case they aren’t using the heartbeat. Printing it to the console should be fine.

amarthadan:
Close #1225

### DIFF
--- a/packages/airnode-deployer/src/cli/commands.ts
+++ b/packages/airnode-deployer/src/cli/commands.ts
@@ -4,7 +4,7 @@ import * as path from 'path';
 import { CloudProvider, loadConfig } from '@api3/airnode-node';
 import size from 'lodash/size';
 import { bold } from 'chalk';
-import { deployAirnode, removeAirnode } from '../infrastructure';
+import { deployAirnode, DeployAirnodeOutput, removeAirnode } from '../infrastructure';
 import {
   deriveAirnodeAddress,
   writeReceiptFile,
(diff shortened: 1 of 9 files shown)


packages/airnode-deployer/src/cli/commands.ts
```diff
@@ -92,7 +92,10 @@ export async function deploy(configPath: string, secretsPath: string, receiptFil
   logger.debug('Deleting a temporary secrets.json file');
   fs.rmSync(tmpDir, { recursive: true });
 
-  writeReceiptFile(receiptFile, mnemonic, config, output, deploymentTimestamp);
+  writeReceiptFile(receiptFile, mnemonic, config, deploymentTimestamp);
+
+  if (output.httpGatewayUrl) logger.info(`HTTP gateway URL: ${output.httpGatewayUrl}`);
```
Siegrift, line 97: Won't this end up lost between the error logs in case there is an error?

Siegrift: Fair enough.

Siegrift (approved):
👍 LGTM

PR #1276, opened 2022-06-21
Replace attempt library `go` with promise-utils

vponline:
This replaces the attempt library go-function implementation with api3/promise-utils.


packages/airnode-utilities/src/caching/index.ts
```diff
@@ -65,18 +65,18 @@ const getKeys = (prefix?: string): string[] => {
...
+  const goRes = goSync(() => readFileSync(join(CACHE_BASE_PATH, key)));
+  if (!goRes.success) {
     logger.error(`Unable to read key data from fs cache`);
-    logger.error((e as Error).stack!);
+    logger.error(goRes.error.stack!);
+    return undefined;
   }
 
-  return undefined;
+  return JSON.parse(goRes.data.toString());
```
Siegrift: JSON.parse can throw so this should be inside the `goSync` (similarly how it was previously in try-catch)

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -108,22 +107,33 @@ const addKey = (key: string, data: any, force = false) => {
...
-    const fileExists = existsSync(filePath);
-    if (fileExists && !force) {
-      logger.error(`Unable to overwrite key from fs cache as key exists and force is set to false`);
-      logger.error(`Key: ${key}`);
-      return;
-    }
 
-    if (fileExists) {
-      rmSync(filePath, { force: true });
+  const goFileExists = goSync(() => existsSync(filePath));
```
Siegrift, line 111: Can this actually fail? I think `existsSync` does not throw an error.

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -133,51 +143,55 @@ const addKey = (key: string, data: any, force = false) => {
  * Sweep (garbage collect) the cache.
  */
 const sweep = () => {
-  try {
-    const now = Date.now();
+  const now = Date.now();
 
-    const stattedFiles = readdirSync(CACHE_BASE_PATH).map((file) => ({
+  const goRes = goSync(() =>
```
Siegrift, line 148: This now changed the behaviour. Previously all of the code was inside the try-catch. 

EDIT: I see the rest of the code does not throw so it's probably fine.

packages/airnode-node/src/evm/utils.ts
```diff
@@ -19,13 +20,17 @@ export function sortBigNumbers(bigNumbers: ethers.BigNumber[]) {
...
-  } catch {
-    return 'No revert string';
-  }
+
+  // Skip the funciton selector from the returned encoded data
+  // and only decode the revert reason string.
+  // Function selector is 4 bytes long and that is why we skip
+  // the first 2 bytes (0x) and the rest 8 bytes is the function selector
+  // return ethers.utils.defaultAbiCoder.decode(['string'], `0x${callData.substring(2 + 4 * 2)}`)[0];
+
```
Siegrift, line 29: Nit: remove `\n` as the comment is bound to that decode logic.

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -207,7 +209,7 @@ async function terraformAirnodeManage(
 }
 
 // `maxConcurrency` field is required for deployment but missing for removal. This is the easiest way to type it.
-type CloudProviderExtended = CloudProvider & {
+export type CloudProviderExtended = CloudProvider & {
```
Siegrift, line 212: Do we need to export this?

vponline:
This can be reviewed now. I managed to replace most of the try/catch blocks, but a few are left (mainly inside test files) and one [here](https://github.com/api3dao/airnode/blob/master/packages/airnode-examples/src/cli.ts#L11).

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -45,24 +46,25 @@
 async function runCommand(command: string, options: CommandOptions) {
   const stringifiedOptions = JSON.stringify(options);
   const commandSpinner = logger.debugSpinner(`Running command '${command}' with options ${stringifiedOptions}`);
-  try {
-    const { stdout } = await exec(command, options);
-    commandSpinner.succeed(`Finished command '${command}' with options ${stringifiedOptions}`);
-    return stdout;
-  } catch (err) {
+
+  const goExec = await go(() => exec(command, options));
```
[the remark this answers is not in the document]
Siegrift: You can ignore this. I've spend some time investigating these alerts in the past and they were all false positive. I no longer consider these scanning alerts as that important.

packages/airnode-utilities/src/caching/index.ts
Siegrift: I see you reverted this in a separate commit. For the future, try to squash it with the previous commit.

vponline: I'll keep this in mind :+1:

packages/airnode-deployer/src/infrastructure/index.ts
vponline: I was getting a type error without including it [here](https://github.com/api3dao/airnode/pull/1276/files#diff-13b7c9bcdda766f79084f5c09740a68e14cda9410e225e5d9bc44d643f332562R80). It's due to the nodeSettings.cloudProvider having the option of 'local', but that is not allowed once deploying.

packages/airnode-utilities/src/caching/index.ts
vponline: I think you are right. I'll remove this.

packages/airnode-examples/test/example-files-are-up-to-date.test.ts
```diff
@@ -11,15 +12,12 @@ describe('Verifies that all config.example.json files are up to date', () => {
...
-        const createConfig = await import(join(folder, 'create-config.ts'));
-        await createConfig.default(true);
-      } catch (e) {
-        throw new Error((e as Error).message);
-      }
+      const goCreateConfig = await go(() =>
+        import(join(folder, 'create-config.ts')).then((createConfig) => createConfig.default(true))
+      );
+
+      if (!goCreateConfig.success) throw goCreateConfig.error.message;
```
Siegrift, line 19: We should be throwing only Errors not strings.

packages/airnode-adapter/e2e/index.ts
```diff
@@ -294,17 +291,16 @@ describe('Extraction, encoding and simple on chain decoding', () => {
...
             _path: 'big.tooLargeDecimal',
           })
-        );
-        expect.fail();
-      } catch (e: any) {
-        expect(e.message).to.contain('out-of-bounds');
-      }
+        )
+      );
+      expect(goExtractAndEncode.success).to.equal(false);
```
Siegrift, line 302: You can use `assertGoError` instead of this assertion and the typecast below. (same for others).

Siegrift (commented):
Minor comments, but otherwise LGTM

packages/airnode-adapter/e2e/index.ts
vponline: This is cool, thanks for pointing it out!

Siegrift (approved):
👍 LGTM

Issue #1227, opened 2022-06-14
Use TS references for linked monorepo packages


Siegrift:
## Description

In https://github.com/api3dao/airnode/pull/1158 we removed the TS reference links between packages because they are hard to maintain and declaration maps worked even without it. Unfortunately, the links are necessary to show realtime errors in the dependant packages.

We should link the packages and create a script which verifies these links are up to date (as maintaining them is error prone).

Siegrift:
@amarthadan I'd move this to DONE as it can't really be tested and it didn't affect the publishing configuration (so it's a dev only change).

PR #1277, opened 2022-06-21
Use TS references to link monorepo packages

Siegrift:
Closes https://github.com/api3dao/airnode/issues/1227

This feature enables real time errors and compilation cross-project. This means that when `airnode-admin` depends on function X in `airnode-abi` and this function is removed, you will receive instant error in `airnode-admin`.

We assume that every dependency is used in the monorepo project sources and we require that the `tsconfig.json` of the sources references all other monorepo dependencies. 

When another tsconfig A (e.g. the tests) references the tsconfig B of project sources, the referenced projects in B are already already referenced by A. This means that importing stuff from monorepo packages in tests will still provide the same features as in sources.


amarthadan (commented):
Changes look fine to me but I found an issue.

When I, for example, remove the log function from the logger from airnode-utilities package and run a command from airnode-admin package (`yarn cli generate-mnemonic`) it still works. I would expect it to throw an error since the function for printing was removed from the airnode-utilities package.

Siegrift:
Thanks for the review, I didn't think that ts-node would behave differently. As discussed on slack, `ts-node` does not support project references ([link](https://github.com/TypeStrong/ts-node/issues/897)) so we decided to ignore this issue. It is not a big deal, since the errors are visible in the IDE and are showcased during build (this means also on CI). I've also mentioned this side effect in README. 

While investigating this issue I realized there is a setting https://www.typescriptlang.org/tsconfig#tsBuildInfoFile which allows us to move the tsbuildinfo files from the `/dist` directory which gets published to npm, so I've done that as well.

amarthadan (approved):

PR #1283, opened 2022-06-23
Improve Airnode logs for callStatic errors

### LINKED ISSUE #1207: Report better logs when `callStatic` Reverts
`testFulfill()` in [api-calls.ts](https://github.com/api3dao/airnode/blob/2685f63673b36556315604fdc82246a6c7f369c0/packages/airnode-node/src/evm/fulfillments/api-calls.ts#L115) does not report logs when it fails. 

Airnode just reports:

`Submitting API call fail for Request:${request.id}...`

This does not indicate if it failed in the static call or somewhere else. At AirnodeRrp we get the `no revert` string emitted in the event, which at least gives us some information to help debug, Airnode should also log where exactly it failed eg `static call fullfilment failed with : $err`

dcroote:
Fixes #1207


Siegrift (approved):
👍 LGTM

Ashar2shahid (approved):
This should help with debugging, thanks 👍

packages/airnode-node/src/evm/fulfillments/api-calls.ts
```diff
@@ -64,7 +64,7 @@ async function testFulfill(
     );
   const [err, res] = await go(operation, { retries: 1, timeoutMs: DEFAULT_RETRY_TIMEOUT_MS });
   if (err) {
-    const errorLog = logger.pend('ERROR', `Error attempting API call fulfillment for Request:${request.id}`, err);
+    const errorLog = logger.pend('ERROR', `Static call fulfillment failed for Request:${request.id} with ${err}`, err);
```
Ashar2shahid, line 67: if err is an object would this throw [object] ? Should we do JSON.stringify(err) @Siegrift ?

Siegrift: Mhh, I assumed it will always be an error instance which stringifies to string. It might be better to check for this explicitely and json.stringify in case it isn't.

dcroote: I believe @Siegrift  is correct and `err` has to be an `Error` instance because the `go` [function](https://github.com/api3dao/airnode/blob/7da75b33b2a9faffdd84518a84861845c46e3309/packages/airnode-utilities/src/promises/promise.ts#L30) returns a `Promise<GoResult<T>>` where `GoResult` is:
https://github.com/api3dao/airnode/blob/7da75b33b2a9faffdd84518a84861845c46e3309/packages/airnode-utilities/src/promises/promise.ts#L10

Issue #1197, opened 2022-06-13
Add hardcoded whitelisting into the config file


amarthadan:
API providers don’t want to make their endpoints public, and they don’t want to have to make transactions to whitelist requesters. This means there is no way for them to use Airnode at the moment (which is why they have been directed towards relaying metadata).

Instead, we want the user to be able to hardcode whitelistings for endpointId-requester pairs (similar to how the current authorizers work) in the config file. This should probably go inside the respective `chains` object, but we should talk about it.

[Slack discussion](https://api3workspace.slack.com/archives/C02AYRX8D89/p1646929455160219)

```
{
  "chains": [
    "authorizations": {
      "requesterAuthorizations": {
        "$ENDPOINT_ID_1": [
          "$REQUESTER_1",
          "$REQUESTER_2",
          ...
        ],
        ...
      }
    },
    ...
  ],
  ...
}
```

Related to https://github.com/api3dao/airnode/issues/1196

bbenligiray:
We're increasingly talking about authorizing requesters to have access to all endpoints of an Airnode. With that in mind, maybe `requesterAuthorizations` should be called `requesterEndpointAuthorizations` for future-proofness.

vponline:
The other task #1196  specifies the field name as 'authorizers' and here we have `authorizations`, is the intention to rename the field or create a new field?

bbenligiray:
An authorization is when you allow someone to use your Airnode. An authorizer is a contract that authorizes. So `authorizers` and `authorizations` refer to two distinct concepts, the specs are correct (other than [this](https://github.com/api3dao/airnode/issues/1197#issuecomment-1153665996)).

vponline:
I see, thanks for explaining!

vponline:
And just to confirm that should both `$ENDPOINT_ID_N` and `$REQUESTER_N`  be secrets in `secrets.env` and interpolated from there into `config.json`?

bbenligiray:
Not necessarily, but I can see that happen. When that requester makes a request to that endpoint and the Airnode responds on a public chain, the fact that that requester is authorized for that endpoint will be visible anyway, so this is not really a secret. However, this authorization may have been defined for a private chain, and the Airnode operator may want to be able to publish their config without disclosing their exact authorizations there.
If you're asking this to decide how you should create test fixtures, you don't need to do interpolation.

amarthadan:
[...]

PR #1285, opened 2022-06-25
Add authorizations configuration to config

vponline:
This adds the `authorizations` configuration field to `config.json`, authorizer type to the `authorizers` field and `checkConfigAuthorizations` functionality which skips fetching the on-chain authorization status for the `endpointId`/`requesterAddress` combinations defined in the config.


amarthadan:
We've discussed this a bit with @vponline on Slack as there was some confusion about this issue (from both of us :sweat_smile:). I'm moving #1197 back to `In progress`. Feel free to keep this PR open and ping me once you want me to review it :slightly_smiling_face:

vponline:
@amarthadan  I have added the off-chain authorizations field check which checks for the presence of an `endpointId`/`requesterAddress` combination and skips the on-chain check for that specific `apiCall` if found and sets the authorization status as `true`.

This PR now has both the authorizer and authorization task contents.

packages/airnode-node/src/evm/authorization/authorization-fetching.ts
```diff
@@ -114,15 +142,27 @@ export async function fetch(
...
+  if (isEmpty(fetchOptions.authorizers.requesterEndpointAuthorizers)) {
     const authorizationByRequestIds = apiCalls.map((pendingApiCall) => ({
       [pendingApiCall.id]: true,
     }));
     return [[], Object.assign({}, ...authorizationByRequestIds) as AuthorizationByRequestId];
   }
 
+  // Skip fetching authorization statuses if found in config for a specific authorization type
+  // and requester address
+  const { configAuthorizationsByRequestId, configAuthorizationRequestIds } = checkConfigAuthorizations(
```
amarthadan, line 154: Do you actually need both `configAuthorizationsByRequestId` and `configAuthorizationRequestIds` as return values? Isn't `configAuthorizationRequestIds` just `Object.keys(configAuthorizationsByRequestId)`?

packages/airnode-node/src/evm/authorization/authorization-fetching.test.ts
```diff
@@ -166,10 +178,85 @@ describe('fetch (authorizations)', () => {
     ]);
     expect(res).toEqual({});
   });
+
+  it('does not fetch authorizations if found in config and returns true for valid authorizations', async () => {
```
amarthadan, line 182: Can you also add a test for a combination of these two? Meaning that there are both requests that are authorized in the config and those that need to do the authorization on-chain.

amarthadan (changes requested):
Looks good, just two small things :+1:

packages/airnode-operation/README.md
```diff
@@ -135,8 +136,17 @@ WALLET LINKED TO A TEST MNEMONIC**
 
 **Authorizers**
 
-`authorizers` - a list of `authorizer` contracts. The values must correspond to a value defined in the `authorizers`
-top-level field.
+`authorizers` - a key/value object where the key is the type of `authorizer` and the value is a list of `authorizer`
+contracts. The values must correspond to a value defined in the `authorizers` top-level field.
```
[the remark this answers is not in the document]
vponline: I am not sure, but it was in the original description so I haven't edited that part :thinking:.

packages/airnode-node/src/evm/authorization/authorization-fetching.ts
vponline: > Isn't configAuthorizationRequestIds just Object.keys(configAuthorizationsByRequestId)

Not exactly as `configAuthorizationsByRequestId` is an array of objects so I thought this was a neater solution than mapping through `configAuthorizationsByRequestId` and getting each object's key, but this can also be done inline, fixed :smile:

amarthadan: Oh, I see now. The reason why the `successfulResults` (at the end of the `fetch` function) is an array is that we chunk the authorization requests. You actually don't need to create an array (`AuthorizationByRequestId[]`) but just one object (`AuthorizationByRequestId`) in the `checkConfigAuthorizations` function. It will simplify `checkConfigAuthorizations` itself and the subsequent filtering as well.

packages/airnode-operation/README.md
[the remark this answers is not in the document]
amarthadan: That's regarding the config in the `airnode-operations` package which is a bit different and it confused me. Carry on :smile:

packages/airnode-node/src/evm/authorization/authorization-fetching.ts
vponline: Very elegant, fixed!

packages/airnode-node/src/evm/authorization/authorization-fetching.ts
```diff
@@ -114,15 +132,24 @@ export async function fetch(
...
     return [[], Object.assign({}, ...authorizationByRequestIds) as AuthorizationByRequestId];
   }
 
+  // Skip fetching authorization statuses if found in config for a specific authorization type
+  // and requester address
+  const configAuthorizationsByRequestId = checkConfigAuthorizations(apiCalls, fetchOptions);
+
+  // Filter apiCalls for which a valid authorization was found in config
+  const apiCallsToFetchAuthorizationStatus = apiCalls.filter(
+    (apiCall) => !Object.keys(configAuthorizationsByRequestId).includes(apiCall.id)
```
amarthadan, line 148: Nit: You can store `Object.keys(configAuthorizationsByRequestId)` in a variable before calling the `filter` function so it won't be called for each iteration (for each api call).

amarthadan (approved):

Issue #1274, opened 2022-06-21
Use OIS package from separate repository in Airnode


Siegrift:
Blocked by https://github.com/api3dao/ois/issues/2

amarthadan:
Can you assign priority, size, and iteration?

Siegrift:
Blocked by https://github.com/api3dao/ois/issues/6

kolenic-martin:
re:testing this: I'll most probably have to make sure the examples still work and that when the repo is bootstrapped and built by yarn, it pulls the package from another repo, correct?

kolenic-martin:
Hi @Siegrift and @amarthadan , as mentioned above, it seems this could be tested by only running an example, which I've done several times today without any particular issues regarding the OIS package. I assume this is now done but feel free to ping me if not.

PR #1287, opened 2022-06-26
Use @api3/ois package and remove @api3/airnode-ois

Siegrift:
Closes https://github.com/api3dao/airnode/issues/1274


amarthadan (approved):

Siegrift:
FYI: I've remembered to check whether changeset versioning will not have problems with the removed package and indeed it had. I had to remove a reference to the OIS package (which is now removed) from one of the changesets.

Siegrift:
@amarthadan I've made some more changes, please take another look (I've removed OIS from validator).

amarthadan (approved):
LGTM

PR #1291, opened 2022-06-27
Add authorizer type to config.json

vponline:
This updates the `authorizer` field in `config.json` to be an object specifying the type of `authorizer` as the key and a list of `authorizer contracts` as the value.

### DIFF
--- a/packages/airnode-node/src/evm/authorization/authorization-fetching.ts
+++ b/packages/airnode-node/src/evm/authorization/authorization-fetching.ts
@@ -115,7 +115,7 @@ export async function fetch(
   }
 
   // If there are no authorizer contracts then endpoint is public
-  if (isEmpty(Object.values(fetchOptions.authorizers).flatMap((authorizer) => authorizer))) {
+  if (isEmpty(fetchOptions.authorizers.requesterEndpointAuthorizers)) {
     const authorizationByRequestIds = apiCalls.map((pendingApiCall) => ({
       [pendingApiCall.id]: true,
     }));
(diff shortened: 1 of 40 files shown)


packages/airnode-node/src/evm/authorization/authorization-fetching.ts
```diff
@@ -8,23 +8,24 @@ import { go } from '@api3/promise-utils';
...
   airnodeRrp: AirnodeRrpV0,
-  authorizers: string[],
+  authorizers: ChainAuthorizers,
   airnodeAddress: string,
   apiCall: Request<ApiCall>
 ): Promise<LogsData<boolean | null>> {
   const contractCall = (): Promise<boolean> =>
     airnodeRrp.checkAuthorizationStatus(
-      authorizers,
+      authorizers.requesterEndpointAuthorizers,
```
vponline, line 28: I assumed here that this contract call is currently supposed to take in the `requesterEndpointAuthorizations`, so please let me know if I should have been mapping all authorizer contractts from the new authorizers object (values) and passing them in here.

packages/airnode-node/src/evm/authorization/authorization-fetching.ts
```diff
@@ -64,7 +65,7 @@ async function fetchAuthorizationStatuses(
 
   const contractCall = (): Promise<boolean[]> =>
     airnodeRrp.checkAuthorizationStatuses(
-      authorizers,
+      authorizers.requesterEndpointAuthorizers,
```
vponline, line 68: ...and here

packages/airnode-node/src/evm/authorization/authorization-fetching.ts
```diff
@@ -114,7 +115,7 @@ export async function fetch(
   }
 
   // If there are no authorizer contracts then endpoint is public
-  if (isEmpty(fetchOptions.authorizers)) {
+  if (isEmpty(Object.values(fetchOptions.authorizers).flatMap((authorizer) => authorizer))) {
```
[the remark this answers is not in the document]
amarthadan: Since we're passing directly only `requesterEndpointAuthorizers` for the check (comment above) I think we should check only that for emptiness for v0.

amarthadan (approved):
LGTM :+1:
I don't think we need to wait for https://github.com/api3dao/airnode/issues/1197 with merging. Worst case we rework it a bit later.

Issue #1183, opened 2022-06-13
Unite error handling for AWS and GCP HTTP gateways


amarthadan:
GCP API gateway service is missing some features that AWS offers, namely request body validation and support for API keys. Due to this, each gateway handles error states a bit differently. We should unite the behavior so that gateways will act the same no matter which cloud provider the user uses.

From a recent [PR](https://github.com/api3dao/airnode/pull/843#discussion_r811193992), also reconsider if status 400 is the correct status code or if there might be a more suitable code to use.

kolenic-martin:
Hi @Siegrift, I've pushed this back and prioritized other issues but since I've managed to blast the rest of the column, I'm working on it now. I'm collecting various error messages via postman and putting together a list so I hope this will be done by the call.

kolenic-martin:
Hi, @amarthadan and @Siegrift, I've used the coingecko-testable example to attempt various scenarios (I helped myself out by reading the code and designing the test cases accordingly). I'm happy to say all of them performed as intended in the code on both providers, but let me just sum it up here:

- incorrect API key - a 401 error
- missing endpoint ID - a 400 error
- request without a body - a 400 error
- incorrect encodedParameters value - a 400 error

This didn't take that long but then I realized the signed and relayed gateway files were also changed so I ran the tests with those two as well on both providers. Thankfully, nothing's broken :) I'm happy to move this to Done.

PR #1292, opened 2022-06-27
Unify error handling

Siegrift:
Closes https://github.com/api3dao/airnode/issues/1183


Siegrift:
Code should be done. I tested both gateways on AWS and all seems fine. I wanted to compare the error handling on GCP, but I've been stuck at making a successful deployment for a couple of hours... I'm reaching gateway API quota, but GCP doesn't allow me to remove the gateways (even though they are not used anymore since I've removed all lambdas and buckets)

amarthadan (approved):
Just one comment, otherwise LGTM.

Issue #1195, opened 2022-06-13
Keep protocol indices at one place


amarthadan:
We’re currently (or will in the near future) supporting 5 different protocols, each with a different index ([Slack discussion](https://api3workspace.slack.com/archives/C02AYRX8D89/p1645698379511399))

We should have one place where these are stored. For now, there are just hardcoded values in both Airnode and Airkeeper.

Siegrift:
It seems to me there is one more. We currently have:

1: RRP
2: PSP
3: Relayed RRP
4: Relayed PSP
5: [Airseeker](https://cs.github.com/API3-CTT/airseeker/blob/2faee5766e4af07c778714c66f2a413987643708/test/setup/deployment.ts?q=deriveSponsorWalletFromMnemonic#L127)
12345: [Airkeeper](https://cs.github.com/api3dao/airkeeper/blob/32a06f6c6fb201917a8e7b377dfa913822ebb554/src/handlers/rrp.ts?q=deriveSponsorWalletFromMnemonic#L180)

I think the most reasonable place to define these is inside [airnode protocol](https://github.com/api3dao/airnode/blob/master/packages/airnode-protocol/src/index.ts).

Useful links:
* https://cs.github.com/?scopeName=all+api3&scope=org%3AAPI3-CTT+OR+org%3Aapi3dao&q=PROTOCOL_ID_PSP
* https://cs.github.com/?scopeName=all+api3&scope=org%3AAPI3-CTT+OR+org%3Aapi3dao&q=PROTOCOL_ID
* https://cs.github.com/?scopeName=all+api3&scope=org%3AAPI3-CTT+OR+org%3Aapi3dao&q=deriveSponsorWalletFromMnemonic

Siegrift:
After discussion with @amarthadan I've created 5 more issues to use these constants in the other repositories which currently hardcode the constants. See: https://api3dao.atlassian.net/browse/BEC-544?jql=issueKey%20in%20(BEC-540%2CBEC-541%2CBEC-542%2CBEC-543%2CBEC-544)

There is nothing to test in this issue so I'm moving it to done.

PR #1297, opened 2022-06-29
Define protocol indices in airnode-protocol

Siegrift:
Closes https://github.com/api3dao/airnode/issues/1195

### DIFF
--- a/packages/airnode-protocol/src/index.ts
+++ b/packages/airnode-protocol/src/index.ts
@@ -26,12 +26,14 @@ const RequesterAuthorizerWithAirnodeAddresses: { [chainId: string]: string } =
   references.RequesterAuthorizerWithAirnode;
 const networks: { [chainId: string]: ethers.providers.Network } = references.networks;
 
-const RRP_PROTOCOL_ID = '1';
-const PSP_PROTOCOL_ID = '2';
-const RELAYED_RRP_PROTOCOL_ID = '3';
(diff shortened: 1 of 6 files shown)


packages/airnode-protocol/src/index.ts
```diff
@@ -26,6 +26,13 @@ const RequesterAuthorizerWithAirnodeAddresses: { [chainId: string]: string } =
   references.RequesterAuthorizerWithAirnode;
 const networks: { [chainId: string]: ethers.providers.Network } = references.networks;
 
+const RRP_PROTOCOL_ID = '1';
+const PSP_PROTOCOL_ID = '2';
+const RELAYED_RRP_PROTOCOL_ID = '3';
+const RELAYED_PSP_PROTOCOL_ID = '4';
+const AIRSEEKER_PROTOCOL_ID = '5';
```
Siegrift, line 33: Not sure about the naming of these two, because they are not protocols per se.

Siegrift:
> RRP protocol ID = Request-response protocol protocol ID but RRP ID doesn't work so meh. 

Yeah, I was thinking the same.

> Maybe call it PROTOCOL_ID_RRP for better autocomplete (especially since Airkeeper/Airseeker won't be obvious) but it's up to you.

Alternatively, we can do enum style:

```
const PROTOCOL_IDS = {
    RRP: '1'
    ...
}
```
wdyt?

PR #1300, opened 2022-06-29
Add Airnode private key to env and use it to get Airnode wallet

vponline:
This adds the Airnode private key as an environment variable and uses that to derive the Airnode wallet.

### DIFF
--- a/packages/airnode-node/src/api/index.ts
+++ b/packages/airnode-node/src/api/index.ts
@@ -4,7 +4,7 @@ import { ethers } from 'ethers';
 import { logger, removeKeys, removeKey } from '@api3/airnode-utilities';
 import { go, goSync } from '@api3/promise-utils';
 import { postProcessApiSpecifications, preProcessApiSpecifications } from './processing';
-import { getMasterHDNode } from '../evm';
+import { getMasterHDNode, getAirnodeWalletWithPrivateKey } from '../evm';
 import { getReservedParameters } from '../adapters/http/parameters';
 import { API_CALL_TIMEOUT, API_CALL_TOTAL_TIMEOUT } from '../constants';
 import { isValidSponsorWallet, isValidRequestId } from '../evm/verification';
@@ -17,7 +17,7 @@ import {
   ApiCallErrorResponse,
   ApiCallParameters,
 } from '../types';
-import { Config } from '../config';
+import { Config, getEnvValue } from '../config';
 
 function buildOptions(payload: CallApiPayload): adapter.BuildRequestOptions {
   const { config, aggregatedApiCall } = payload;
@@ -78,9 +78,8 @@ function buildOptions(payload: CallApiPayload): adapter.BuildRequestOptions {
   }
 }
 
-async function signWithRequestId(requestId: string, data: string, config: Config) {
-  const masterHDNode = getMasterHDNode(config);
-  const airnodeWallet = ethers.Wallet.fromMnemonic(masterHDNode.mnemonic!.phrase);
+async function signWithRequestId(requestId: string, data: string, airnodeWalletPrivateKey: string) {
+  const airnodeWallet = getAirnodeWalletWithPrivateKey(airnodeWalletPrivateKey);
 
   return await airnodeWallet.signMessage(
     ethers.utils.arrayify(
@@ -89,9 +88,13 @@ async function signWithRequestId(requestId: string, data: string, config: Config
   );
 }
 
-async function signWithTemplateId(templateId: string, timestamp: string, data: string, config: Config) {
-  const masterHDNode = getMasterHDNode(config);
-  const airnodeWallet = ethers.Wallet.fromMnemonic(masterHDNode.mnemonic!.phrase);
+async function signWithTemplateId(
+  templateId: string,
+  timestamp: string,
+  data: string,
+  airnodeWalletPrivateKey: string
+) {
+  const airnodeWallet = getAirnodeWalletWithPrivateKey(airnodeWalletPrivateKey);
 
   return await airnodeWallet.signMessage(
     ethers.utils.arrayify(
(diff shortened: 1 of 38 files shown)


packages/airnode-node/src/api/index.ts
```diff
@@ -244,12 +247,19 @@ async function processSuccessfulApiCall(
 
   const response = goExtractAndEncodeResponse.data;
 
+  const airnodeWalletPrivateKey = getEnvValue('AIRNODE_WALLET_PRIVATE_KEY');
+  if (!airnodeWalletPrivateKey) {
+    const errorMessage = 'Missing Airnode wallet private key in environment variables.';
+    const log = logger.pend('ERROR', errorMessage);
+    return [[log], { success: false, errorMessage }];
+  }
```
amarthadan, line 255: Two things come to mind looking at this:
* This means that the env variable is required (which is fine). In that case, make the Terraform variable required so you can't somehow deploy without it.
* There is no need for this env variable for HTTP gateway (non-signed) requests. So, I would make this check only where is needed and also add the variable (in Terraform) only to handlers that need it.

packages/airnode-node/src/evm/wallet.ts
```diff
@@ -37,6 +37,10 @@ export function getAirnodeWallet(config: Config): ethers.Wallet {
   return ethers.Wallet.fromMnemonic(mnemonic);
 }
 
+export function getAirnodeWalletWithPrivateKey(privateKey: string): ethers.Wallet {
```
amarthadan, line 40: Do we still need the above `getAirnodeWallet` function?

amarthadan, line 40: I would name it `getAirnodeWalletFromPrivateKey`. Calling it "with private key" suggests that it returns a wallet and a private key. At least it sounds like that to me, it may be just my bad English :smile:

packages/airnode-node/src/handlers/process-http-signed-data-request.ts
```diff
@@ -25,7 +25,11 @@ export async function processHttpSignedDataRequest(
 
   const requestId = randomHexString(16);
   const logOptions = buildBaseOptions(config, { requestId });
-  const airnodeAddress = wallet.getAirnodeWallet(config).address;
+  const airnodeWalletPrivateKey = getEnvValue('AIRNODE_WALLET_PRIVATE_KEY');
+  if (!airnodeWalletPrivateKey) {
+    return [new Error('Missing Airnode wallet private key in environment variables.'), null];
+  }
+  const airnodeAddress = wallet.getAirnodeWalletWithPrivateKey(airnodeWalletPrivateKey).address;
```
amarthadan, line 32: I don't like much that we're doing these checks at so many places but can't really see a better solution :confused:

packages/airnode-node/src/providers/state.ts
```diff
@@ -16,7 +16,11 @@ export function buildEVMState(
   const masterHDNode = evm.getMasterHDNode(config);
   const chainProviderUrl = chain.providers[chainProviderName].url;
   const provider = evm.buildEVMProvider(chainProviderUrl, chain.id);
-  const airnodeAddress = evm.getAirnodeWallet(config).address;
+  const airnodeWalletPrivateKey = getEnvValue('AIRNODE_WALLET_PRIVATE_KEY');
+  if (!airnodeWalletPrivateKey) {
+    throw new Error('Missing Airnode wallet private key in environment variables.');
+  }
+  const airnodeAddress = evm.getAirnodeWalletWithPrivateKey(airnodeWalletPrivateKey).address;
```
amarthadan, line 23: Can't we pass the `airnodeAddress` to the `buildEVMState` function. It should be already available level higher, isn't it?

packages/airnode-node/src/workers/local-handlers.ts
```diff
@@ -12,6 +13,11 @@ function loadConfig() {
 
 export async function startCoordinator(): Promise<WorkerResponse> {
   const config = loadConfig();
+  // Set the Airnode wallet private key into the environment when running locally
+  // eslint-disable-next-line functional/immutable-data
+  process.env['AIRNODE_WALLET_PRIVATE_KEY'] = ethers.Wallet.fromMnemonic(
+    config.nodeSettings.airnodeWalletMnemonic
+  ).privateKey;
```
amarthadan, line 20: Hmm, this works for the other handlers just because the `startCoordinator` is called first and the env variables are shared. That's kinda meh.  What about creating a function and calling it for each handler the same way we do with `loadConfig`?

amarthadan (changes requested):

packages/airnode-node/src/evm/wallet.ts
vponline: Yep, that's still needed in the `airnode-deployer` to get the wallet from which to get the `privateKey` from initially. It is also used in `airnode-examples` in a few places.

packages/airnode-node/src/providers/state.ts
vponline: You're right, its available in CoordinatorSettings. I'll update

packages/airnode-node/src/workers/local-handlers.ts
vponline: >  That's kinda meh.

:smile: Indeed it is, added.

bbenligiray:
This is a bit :snail: but what if we derived the private key in the coordinator and kept it in the state?

vponline:
> This is a bit snail but what if we derived the private key in the coordinator and kept it in the state?

Seems like a valid alternative as well. Should I make the change?

bbenligiray:
Never mind, it doesn't work in the signed HTTP gateway case, which is the main reason we're doing this now

packages/airnode-node/src/evm/wallet.ts
amarthadan: No, using `getAirnodeWallet` is fine, I'm just making sure there's no unsued function :slightly_smiling_face:

packages/airnode-node/src/workers/local-handlers.ts
```diff
@@ -2,22 +2,25 @@ import * as path from 'path';
...
-import { loadTrustedConfig } from '../config';
+import { loadTrustedConfig, setEnvValue, Config } from '../config';
 import * as handlers from '../handlers';
 import * as state from '../providers/state';
 import { WorkerResponse, InitializeProviderPayload, CallApiPayload, ProcessTransactionsPayload } from '../types';
 
 function loadConfig() {
   return loadTrustedConfig(path.resolve(`${__dirname}/../../config/config.json`), process.env);
 }
+function setAirnodePrivateKeyToEnv(config: Config) {
```
amarthadan, line 13: Nit: You can pass just the mnemonic, not the whole config.

packages/airnode-node/src/api/index.ts
```diff
@@ -78,7 +78,11 @@ function buildOptions(payload: CallApiPayload): adapter.BuildRequestOptions {
   }
 }
 
-async function signWithRequestId(requestId: string, data: string, airnodeWalletPrivateKey: string) {
+async function signWithRequestId(requestId: string, data: string) {
+  const airnodeWalletPrivateKey = getEnvValue('AIRNODE_WALLET_PRIVATE_KEY');
+  if (!airnodeWalletPrivateKey) {
+    throw new Error('Missing Airnode wallet private key in environment variables.');
+  }
   const airnodeWallet = getAirnodeWalletFromPrivateKey(airnodeWalletPrivateKey);
```
amarthadan, line 86: This whole thing works the same way for both `signWith*` functions. Can it be abstracted into a separate function?

packages/airnode-node/src/workers/local-handlers.ts
vponline: Fixed :+1:

Issue #1188, opened 2022-06-13
Automatically remove unsuccessful Airnode deployments


amarthadan:
[Slack discussion](https://api3workspace.slack.com/archives/C02AYRX8D89/p1652783089926649)

kolenic-martin:
Hi, I've tried several testing scenarios on both AWS and GCP, including:

- shutting down my network connection mid-deployment
- re-deploying to AWS after previous deployment was interrupting by Ctrl+C kill signal on the command line
- deploying with incorrect credentials
- deploying to GCP without the App Engine Admin API enabled

This either resulted in an attempt to autoremove or in a notice that the removal needs to be performed by the user, depending on the circumstances (e.g. I expect the autoremoval to fail if the connection is shut down). It seems this behaves as intended but it's pretty difficult to come up with other real-life testing scenarios as this is oftentimes tied to unexpected errors on the cloud provider's side or other "vis maior" factors, so I think if any errors appear, they will arise as I keep testing and deploying Airnode routinely. For now,I think it's safe to move this to Done.

kolenic-martin:
Hi, I've tried several testing scenarios on both AWS and GCP, including:

- shutting down my network connection mid-deployment
- re-deploying to AWS after previous deployment was interrupting by Ctrl+C kill signal on the command line
- deploying with incorrect credentials
- deploying to GCP without the App Engine Admin API enabled

This either resulted in an attempt to autoremove or in a notice that the removal needs to be performed by the user, depending on the circumstances (e.g. I expect the autoremoval to fail if the connection is shut down). It seems this behaves as intended but it's pretty difficult to come up with other real-life testing scenarios as this is oftentimes tied to unexpected errors on the cloud provider's side or other "vis maior" factors, so I think if any errors appear, they will arise as I keep testing and deploying Airnode routinely. For now,I think it's safe to move this to Done.

PR #1302, opened 2022-06-30
Auto remove deployment

Siegrift:
Closes https://github.com/api3dao/airnode/issues/1188

image

image
image
image
image

### DIFF
--- a/packages/airnode-deployer/src/cli/commands.ts
+++ b/packages/airnode-deployer/src/cli/commands.ts
@@ -105,7 +105,7 @@ export async function deploy(configPath: string, secretsPath: string, receiptFil
       bold(
         `Airnode deployment failed due to unexpected errors.\n` +
           `  It is possible that some resources have been deployed on cloud provider.\n` +
-          `  We are attempting to remove them...\n`
+          `  Attempting to remove them...\n`
       )
     );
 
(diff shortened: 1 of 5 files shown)


amarthadan (approved):
Other than the flag name LGTM :+1:

packages/airnode-deployer/src/cli/commands.ts
```diff
@@ -96,14 +100,39 @@ export async function deploy(configPath: string, secretsPath: string, receiptFil
...
   }
 
-  const output = goDeployAirnode.data;
-  const deploymentTimestamp = new Date().toISOString();
+  if (!goDeployAirnode.success) {
+    logger.fail(
+      bold(
+        `Airnode deployment failed due to unexpected errors.\n` +
+          `  It is possible that some resources have been deployed on cloud provider.\n` +
+          `  We are attempting to remove them...\n`
```
andreogle, line 108: I think we should strongly avoid "we" terminology in messages, or more broadly anything that implies the core team has some kind of admin ability to interfere with deployments. It's misleading and is likely going to cause confusion for users. 

I was looking at the screenshots and noticed this

Siegrift: Would:

```
It is possible that some resources have been deployed on cloud provider.
Attempting to remove them...
```

be OK?

andreogle: Yeah that's better 👍🏻

Issue #1187, opened 2022-06-13
Refactor `callApi` handler so it can be called without the whole configuration


amarthadan:
Based on the research done in https://api3dao.atlassian.net/browse/AN-749

acenolaza:
PR: https://github.com/api3dao/airnode/pull/1308

amarthadan:
Should this be in `In review`? The PR is still a draft and you're still working on it AFAIK.

acenolaza:
> Should this be in `In review`? The PR is still a draft and you're still working on it AFAIK.

Back to "In Progress"

acenolaza:
> Hi @acenolaza, this looks like an internals-related thing so there are no user-facing changes in terms of making requests. Is it enough if I run any of the examples to check whether that still works?

Yes, running the examples should be enough. I did run the `/operations` repo scripts to make sure requests are still being fullfilled.

kolenic-martin:
Hi, @acenolaza, I've run a test deployment earlier today (AWS + coingecko) and this worked, so I can confirm this did not break anything and we can move this to Done.

amarthadan:
I would definitely test also both gateways as the changes touch them as well.

kolenic-martin:
@amarthadan thanks for noting, I'm moving this back to testing and will do that right away.

kolenic-martin:
Hi, I've tested the coingecko-testable and coingecko-signed-data deployments locally and both on GCP and AWS. All of them perform requests and return values as they should, so I assume they are working normally. I'm moving this to Done.

PR #1308, opened 2022-07-01
AN-754: callApi handler argument refactor

acenolaza:
1. This is a draft PR because I've created a new branch out of `private-key-env` to be ahead of possible merge conflicts

2. Main change I did was limiting the required fields on the payload.config object by creating a [new type](https://github.com/api3dao/airnode/blob/f3c9957f8809629372009c97e758061c46ca41bb/packages/airnode-node/src/api/index.ts#L104-L108) using the typescript Pick utility type. I could have created a new type with only those required fields but I wanted to keep backwards compatibility with anyone calling this handler passing the whole config object.

### DIFF
--- a/packages/airnode-node/src/api/index.ts
+++ b/packages/airnode-node/src/api/index.ts
@@ -4,10 +4,10 @@ import { ethers } from 'ethers';
 import { logger, removeKeys, removeKey } from '@api3/airnode-utilities';
 import { go, goSync } from '@api3/promise-utils';
 import { postProcessApiSpecifications, preProcessApiSpecifications } from './processing';
-import { getMasterHDNode, getAirnodeWalletFromPrivateKey } from '../evm';
+import { getAirnodeWalletFromPrivateKey, deriveSponsorWalletFromMnemonic } from '../evm';
 import { getReservedParameters } from '../adapters/http/parameters';
 import { API_CALL_TIMEOUT, API_CALL_TOTAL_TIMEOUT } from '../constants';
-import { isValidSponsorWallet, isValidRequestId } from '../evm/verification';
+import { isValidRequestId } from '../evm/verification';
 import { getExpectedTemplateIdV0, getExpectedTemplateIdV1 } from '../evm/templates';
 import {
   AggregatedApiCall,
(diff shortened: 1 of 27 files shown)


packages/airnode-node/src/api/index.ts
```diff
@@ -105,46 +101,49 @@ async function signWithTemplateId(
...
+  Partial<Pick<Config, 'chains'>> &
+  Partial<{
+    nodeSettings: Pick<Config['nodeSettings'], 'airnodeWalletMnemonic'>;
+  }>;
+
 export interface CallApiPayload {
-  readonly config: Config;
+  readonly config: CallApiConfig;
   readonly aggregatedApiCall: AggregatedApiCall;
 }
```
amarthadan, line 113: I think we should distinguish the payload (and config) types for different call types (regular vs. http vs. http signed data). It makes no sense to keep the fields there as optional when we know that only some call type needs those fields (or rather will contain them for certain).Similar to how we use the `AggregatedApiCall` type.

packages/airnode-node/src/api/index.ts
```diff
@@ -105,46 +101,49 @@ async function signWithTemplateId(
...
   const { config, aggregatedApiCall } = payload;
+
   if (aggregatedApiCall.type !== 'regular') return null;
 
+  const { nodeSettings } = config;
+  if (!nodeSettings) {
+    const message = 'Missing config.nodeSettings object';
+    const log = logger.pend('ERROR', message);
+    return [[log], { success: false, errorMessage: message }];
+  }
```
amarthadan, line 125: I believe this check won't be necessary once the types are correct as we will be sure that the field is there for the regular request as the config is validated during the deployment.

amarthadan (changes requested):

packages/airnode-node/src/api/index.ts
acenolaza: Sounds good 👍🏻  Please let me know if [something like this](https://github.com/api3dao/airnode/pull/1308/commits/a4faf2aa8ac236c4ceffd0f06e0a8987aff2a98f) is what you were thinking of.

amarthadan: Almost. I think moving the `type` property from `*AggregatedApiCall` to `*ApiCallPayload` will make more sense now that we need to distinguish both and the payload is superior (contains the aggregated API call). You'll get rid of a lot of those unnecessary `as` statements. I think it shouldn't cause any problems and should be doable but can't really be certain without trying it. @Siegrift you've worked on this part of the code before, what do you think?

acenolaza: I even now wonder if the `type` property is even needed since we can always check using `instanceof` to determine the `AggregatedApiCall` type 🤔

Siegrift: > I think moving the type property from *AggregatedApiCall to *ApiCallPayload will make more sense now that we need to distinguish both and the payload is superior (contains the aggregated API call)

Yeah, now that the config is also differentiated based on the type it makes sense. That said, I am not sure if introducing a "config.json subset type" is worth the complexity and whether we should not just use the "whole" config.json (but I don't remember what was the decision when we discussed this for the Airkeeper).

> I even now wonder if the type property is even needed since we can always check using instanceof to determine the AggregatedApiCall type

Not sure what you mean. `instanceof` can be used only to check prototype inheritance (so you mostly use it for class/insance check). You can't use that to check if some value is of a certain type (there is no such feature in TS since the types are removed at runtime).

acenolaza: I just meant checking if an object has some shape or contains certain properties in order to determine they type but I'll just keep it as is.

acenolaza: I've just applied the changes but ended up making changes in places I was expecting not to touch 🤭

Please review it again @amarthadan when you have the chance

amarthadan:
Please, link https://github.com/api3dao/airnode/issues/1187 so it can be closed by merging this PR.

amarthadan (changes requested):

packages/airnode-node/src/api/index.ts
```diff
@@ -52,26 +53,23 @@ function buildOptions(payload: CallApiPayload): adapter.BuildRequestOptions {
...
-          airnodeAddress: airnodeAddress,
           requesterAddress: requesterAddress,
           sponsorAddress: sponsorAddress,
           sponsorWalletAddress: sponsorWalletAddress,
-          endpointId: endpointId,
           requestId: id,
           chainId: chainId,
-          chainType: chain.type,
-          airnodeRrpAddress: chain.contracts.AirnodeRrp,
+          chainType: chain?.type || 'evm',
```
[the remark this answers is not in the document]
dcroote: Currently the validator only allows `evm` for the non-optional `type`:
https://github.com/api3dao/airnode/blob/86f25ee98436fbd62a1b959099796040abf86344/packages/airnode-validator/src/config/config.ts#L45

packages/airnode-node/src/types.ts
```diff
@@ -311,6 +334,7 @@ export interface ProcessTransactionsPayload {
 
 export interface CallApiPayload {
   readonly functionName: 'callApi';
+  readonly type: 'regular' | 'http-gateway' | 'http-signed-data-gateway';
   readonly aggregatedApiCall: AggregatedApiCall;
```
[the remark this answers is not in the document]
acenolaza: Good catch!

packages/airnode-node/src/api/index.ts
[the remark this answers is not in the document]
acenolaza: Right, I've been seen that we have been defaulting to 'evm' and in this particular case we are just using the chain type for metadata which seems quite harmless as of today 🤷🏻

amarthadan (approved):
LGTM :+1:

PR #1315, opened 2022-07-05
Enable renovate automerge

Siegrift:
This will automatically merge the PR once there was approval from someone and the required status checks pass: https://docs.renovatebot.com/key-concepts/automerge/#required-pull-request-reviews

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,7 @@
 {
   "extends": ["config:base", "docker:disable"],
+  "automerge": true,
+  "automergeType": "pr",
   "packageRules": [
     {
       "matchPackageNames": ["chalk", "ora", "node", "@openzeppelin/contracts"],


amarthadan (approved):
From the docs:
```
When automerge is enabled on a PR, Renovate will not add assignees or reviewers at PR creation time, in order to decrease notifications noise a little. If tests subsequently fail, making automerge not possible, then Renovate will add the configured assignees and/or reviewers.
```

I'm fine with this, just making sure you're aware of this behavior. Maybe you would want to consider enabling this https://docs.renovatebot.com/configuration-options/#assignautomerge

Also, I haven't seen any example using these options at the top level, only for individual `packageRules`. But I didn't look for it so maybe it's fine.

Siegrift:
> I'm fine with this, just making sure you're aware of this behavior. 

Yeah, the most important thing is that nothing should be merged without review. From what I've read, the automerge is not possible when the status checks are not passed (including the required approvals). Also, bear in mind renovate bot currently does not use assignees as of now. I don't think it's needed since it [keeps sending emails (after every PR creation/rebase)](https://github.com/renovatebot/config-help/issues/88).

> Also, I haven't seen any example using these options at the top level, only for individual packageRules. But I didn't look for it so maybe it's fine.

Yep, it's fine. I've also validated the config locally (there is a docker container which can do the validation).

Issue #1312, opened 2022-07-05
Move logging utilities to a separate repo


Siegrift:
See: https://github.com/api3dao/airnode/issues/1181#issuecomment-1174671268

Siegrift:
Implemented in https://github.com/api3dao/commons/blob/main/src/logger/README.md

PR #1321, opened 2022-07-07
Add gas oracle with strategies

vponline:
This adds the gas oracle with 3 gas oracle strategies to Airnode.


packages/airnode-node/src/evm/handlers/process-transactions.ts
```diff
@@ -33,27 +33,17 @@ export async function processTransactions(
...
-  } else {
-    const gweiPrice = utils.weiToGwei(gasTarget.gasPrice!);
-    logger.info(`Gas price (legacy) set to ${gweiPrice} Gwei`, baseLogOptions);
-  }
+  const [logs, gasPrice] = await getGasPrice(state2.provider, chainOptions.gasPriceOracle);
+  logger.logPending(logs, baseLogOptions);
 
+  const gasPriceWithType =
+    chainOptions.txType === 'eip1559'
+      ? { type: 2, maxFeePerGas: gasPrice, maxPriorityFeePerGas: gasPrice }
```
amarthadan, line 41: Is it ok to use the same number for `maxFeePerGas` and `maxPriorityFeePerGas`?

packages/airnode-utilities/hardhat.config.ts
```diff
@@ -0,0 +1,13 @@
+import { HardhatUserConfig } from 'hardhat/types';
```
amarthadan, line 1: Hmm, this is necessary for the E2E tests, right? Can't we do it the same way as the `airnode-node` E2E tests? The blockchain preparation is implemented in `airnode-operation` and included. Can we do the same here?

packages/airnode-utilities/src/evm/gas-prices/gas-prices.ts
```diff
@@ -83,7 +83,7 @@ export const getEip1559GasPricing = async (
   ];
 };
 
-export const getGasPrice = async (options: FetchOptions): Promise<LogsData<GasTarget | null>> => {
+export const getProviderGasPrice = async (options: FetchOptions): Promise<LogsData<GasTarget | null>> => {
```
amarthadan, line 86: Do we even still need this file? I thought that the oracle will replace it.

packages/airnode-utilities/src/evm/gas-prices/types.ts
```diff
@@ -35,3 +35,27 @@ export interface FetchOptions {
...
+  gasPriceStrategy: 'constantGasPrice';
+  gasPrice: PriorityFee;
+}
+
+export type GasPriceOracleStrategy =
+  | LatestBlockPercentileGasPriceStrategy
+  | ProviderRecommendedGasPriceStrategy
+  | ConstantGasPriceStrategy;
+
+export type GasPriceOracleConfig = GasPriceOracleStrategy[];
```
amarthadan, line 61: Isn't this pretty much just a copy&paste of the config types? Can't we use those instead?

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts
```diff
@@ -0,0 +1,221 @@
...
+    logger.pend(
+      'INFO',
+      `All oracle strategies failed to return a gas price, returning constant gas price set to ${ethers.utils.formatUnits(
+        constantGasPrice,
+        'gwei'
+      )} gwei.`
+    )
+  );
+
+  return [logs, constantGasPrice];
```
amarthadan, line 220: Are we even going to get here? Isn't the `constantGasPrice` strategy mandatory and so will be processed in the code above among other strategies?

amarthadan (changes requested):
Since it's regarding gas pricing, would you @aquarat take a look as well?

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts
vponline: You're right, but I had to add that as without it typescript shows an error that '`Not all code paths return a value`'. 

Perhaps I need to refactor the logic a bit and handle the `constantGasPrice` strategy separately if you see this as an issue?

packages/airnode-utilities/src/evm/gas-prices/types.ts
vponline: I don't see why not (I saw the old gas prices had the types separately so I did it the same way), I'll make the change :smile:

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts
```diff
@@ -0,0 +1,198 @@
...
+// Returns the provider gas price and applies the recommended multiplier
+export const fetchProviderRecommendedGasPrice = async (
+  provider: ethers.providers.StaticJsonRpcProvider,
+  gasOracleOptions: config.ProviderRecommendedGasPriceStrategy
+) => {
+  const { recommendedGasPriceMultiplier } = gasOracleOptions;
+
+  const gasPrice = await go(() => provider.getGasPrice(), {
+    attemptTimeoutMs: GAS_ORACLE_STRATEGY_MAX_TIMEOUT_MS,
+    totalTimeoutMs: calculateTimeout(Date.now(), GAS_ORACLE_STRATEGY_MAX_TIMEOUT_MS),
```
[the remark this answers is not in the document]
vponline: :facepalm: Thanks for pointing this out. These were defined as separate variables in Airseeker and I have no idea now why I would have changed this

packages/airnode-utilities/package.json
```diff
@@ -23,7 +23,7 @@
     "ethers": "^5.6.9"
   },
   "devDependencies": {
-    "@nomiclabs/hardhat-ethers": "^2.0.2",
+    "@nomiclabs/hardhat-ethers": "^2.0.6",
```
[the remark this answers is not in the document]
vponline: Hmm, I think I only changed the version to match the `hardhat-ethers` dependency version in the other airnode packages.

packages/airnode-node/src/evm/handlers/process-transactions.ts
aquarat: It's common practice to set `maxFeePerGas` and `maxPriorityFeePerGas` to the gasPrice _if_ it's a txType 2 chain and the user has specified the older-style gas bid - but I don't know if that's what's going on here?

aquarat:
I will look at this properly tomorrow morning but the existing comments look thorough :)

packages/airnode-node/src/evm/handlers/process-transactions.ts
vponline: > It's common practice to set maxFeePerGas and maxPriorityFeePerGas to the gasPrice if it's a txType 2 chain and the user has specified the older-style gas bid - but I don't know if that's what's going on here?

This was the idea as `ethers` throws an error if we specify `type: 2` and try to set the `gasPrice` field.

packages/airnode-utilities/src/evm/gas-prices/types.ts
vponline: So as it turns out `airnode-validator` uses` airnode-utilities` and has it as a `tsconfig.json` reference and adding `airnode-validator` to `airnode-utilities` causes a circular dependency error so this may not be possible without changing our `tsconfig` settings. 

But I don't think we'd want to do that as circular dependencies are not recommended, right?

packages/airnode-utilities/hardhat.config.ts
vponline: Related to using the airnode-validator types in airnode-utilities, it turns out that airnode-operations also uses airnode-utilities as a reference so adding them to each other would also cause a circular dependency.

What do you think?

packages/airnode-utilities/src/evm/gas-prices/types.ts
amarthadan: We most certainly don't want a circular dependency :sweat_smile: It looks like `airnode-validator` has `airnode-utilities` as a dependency but doesn't use it anywhere. I think you can remove it from `airnode-validator` and that way you can use `airnode-validator` in `airnode-utilities`.

packages/airnode-utilities/src/evm/gas-prices/gas-prices.ts
amarthadan: Maybe more of a question for @aquarat. But I'd say remove it if we're not going to use it right away.

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts
amarthadan: I see. It's not an issue in terms that it will cause problems but the code is a bit misleading for whoever reads it.

What about changing the validation so that the `constantGasPrice` strategy must be the last one in the array (it doesn't make sense to have any other strategies behind this one, they won't be processed anyway). Then, looping through all the ones in front of it. And then, processing `constantGasPrice`.

packages/airnode-utilities/hardhat.config.ts
amarthadan: Hmm, that sucks. Ok, keep it as it is, for now, I've created an issue for this https://github.com/api3dao/airnode/issues/1327. Once https://github.com/api3dao/airnode/issues/1312 is done we will be able to rework it.

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts
```diff
@@ -0,0 +1,224 @@
...
+export const fetchProviderRecommendedGasPrice = async (
+  provider: ethers.providers.StaticJsonRpcProvider,
+  gasOracleOptions: config.ProviderRecommendedGasPriceStrategy
+) => {
+  const { recommendedGasPriceMultiplier } = gasOracleOptions;
+
+  const startTime = Date.now();
+  const gasPrice = await go(() => provider.getGasPrice(), {
+    attemptTimeoutMs: GAS_ORACLE_STRATEGY_ATTEMPT_TIMEOUT_MS,
+    totalTimeoutMs: calculateTimeout(startTime, GAS_ORACLE_STRATEGY_MAX_TIMEOUT_MS),
```
amarthadan, line 55: What is the `calculateTimeout` actually good for here? In Airseeker, we get the start time (current time) at the beginning of the processing and use it to calculate the timeout for all the chain calls down the line. Here, the `fetchProviderRecommendedGasPrice` function is called just once and you're getting the start time from within. `calculateTimeout` will just return pretty much the full `GAS_ORACLE_STRATEGY_MAX_TIMEOUT_MS` in this case.

You would have to get the start time before you start processing the various strategies and pass it to each of them that needs to make the chain calls so you will enforce one global timeout for gas price processing. If that's the intention.

packages/airnode-utilities/src/constants.ts
```diff
@@ -7,3 +7,7 @@ export const PRIORITY_FEE_IN_WEI = 3_120_000_000;
 export const BASE_FEE_MULTIPLIER = 2;
 // By default there is no gas price multiplier
 export const GAS_PRICE_MULTIPLIER = 1;
+export const RANDOM_BACKOFF_MIN_MS = 0;
+export const RANDOM_BACKOFF_MAX_MS = 2_500;
+export const GAS_ORACLE_STRATEGY_ATTEMPT_TIMEOUT_MS = 3000;
+export const GAS_ORACLE_STRATEGY_MAX_TIMEOUT_MS = 5000;
```
amarthadan, line 13: With these values, you most likely won't be able to make more than one attempt. Also, the question is, is the `GAS_ORACLE_STRATEGY_MAX_TIMEOUT_MS` supposed to be for one gas price strategy but multiple attempts or across all the strategies?

amarthadan (changes requested):

packages/airnode-utilities/src/constants.ts
vponline: Good point :thinking: 

I was thinking per strategy attempt, but now that you mention it, I think it would make more sense to have the `MAX_TIMEOUT` be for all strategies combined as we wouldn't want the gas oracle to take too long from a single cycle, right?

I would need to refactor the `attemptGasOracleStrategy` a bit to make the `MAX_TIMEOUT` apply to all attempts. What do you think, though? Should these values also be configurable in the strategy?

We'd likely want each `ATTEMPT_TIMEOUT` be at least 1.5-2 seconds as often it takes that long for the provider to respond, especially for getting the blocks for `latestBlockPercentileGasPrice` strategy.

amarthadan: I would say you're right and it would make sense to have a max timeout for all strategies combined.

Hmm, I'd say that we don't need these timeouts to be configurable at the moment. Just some sane defaults would imo suffice for the time being. Something like 10s for the global timeout across all the strategies (should be fine since the Airnode can run for up to 1 min anyway), something like 2s for an attempt, and 2 attempts for each call?

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts
vponline: Hmm I see what you mean, thanks for clarifying. :sweat_smile: 

> If that's the intention.

Should that be the intention, or we could just let the higher level function `attemptGasOracleStrategy`  handle the timeouts and retry-delays, and the `provider.getGasPrice` and `provider.getBlockWithTransactions` **Go-Options** should only include retries?

Or do you have something else in mind?

packages/airnode-utilities/src/evm/gas-prices/gas-prices.ts
aquarat: Adding them as an alternative strategy is probably a good idea (and wouldn't be much work it seems); none of the new strategies (as superior as they are) replicate the functionality of the above. The new gas pricing strategies are either very simple (constant pricing) or very advanced. I think this as an additional strategy offers a reasonable fall back for a lot of networks at low cost.

We're aiming to serve a lot of weird chains with Airnode (our business _edge_) and I'm scared of us restricting Airnode's functionality and then having do a hack job to support a specific chain. I realise it's a trade-off between too much _dead_ code vs support for weird chains.

I'm not too fussed though :shrug:

amarthadan: I mean if we turn it into a strategy and remove the `gas-prices.ts` I'm fine with it. If it's easy to add, it can be even done as part of this PR. @vponline WDYT?

aquarat (commented):
This is really great work, a lot of work too, and it's very polished, well done :)
I don't have any wonderfully insightful comments, it all looks reasonable.
The right questions are being asked @ timeouts.

packages/airnode-utilities/src/evm/gas-prices/gas-prices.ts
aquarat: (I'm not fussed where the code lives :smile: )

vponline: Ok sounds good.

> it can be even done as part of this PR. 

Yes, I'll work on adding these next :+1: .

vponline:
I've now added the EIP1559 strategy. I didn't add a `legacy` strategy separately as `providerRecommendedGasPrice` already covers it with the multiplier option as well.

So please feel free to review the last two commits for:
1. the updated timeout options and
2.  the new EIP1559 strategy 
when you get the chance.

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts
```diff
@@ -0,0 +1,296 @@
...
+
+// Returns the provider gas price and applies the recommended multiplier
+export const fetchProviderRecommendedGasPrice = async (
+  provider: Provider,
+  gasOracleOptions: config.ProviderRecommendedGasPriceStrategy
+): Promise<LegacyGasTarget> => {
+  const { recommendedGasPriceMultiplier } = gasOracleOptions;
+
+  const goGasPrice = await go(() => provider.getGasPrice(), {
+    onAttemptError: (goError) => logger.warn(`Failed attempt to get gas price. Error: ${goError.error}.`),
```
amarthadan, line 73: `onAttemptError` will have no effect (will never be called) if you have no retries. For just one function call the error is available as a return value.

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  const [logs, strategyGasTarget] = goProcessGasPriceOracleStrategies.data;
+
+  return [logs, getGasTargetWithGasLimit(strategyGasTarget, fulfillmentGasLimit)];
+};
+
+export const processGasPriceOracleStrategies = async (
+  provider: Provider,
+  gasPriceOracleConfig: config.GasPriceOracleConfig,
+  startTime: number
```
amarthadan, line 247: You actually don't need to pass `startTime` from `getGasPrice`. It's not used there anywhere, you might as well instantiate it here in `processGasPriceOracleStrategies`.

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts
```diff
@@ -0,0 +1,296 @@
...
+  );
+
+  // Attempt gas oracle strategies (excluding constantGasPrice) in order
+  for (const strategy of gasPriceOracleStrategies) {
+    const goAttemptGasOraclePriceStrategy = await go(() => attemptGasOracleStrategy(provider, strategy), {
+      attemptTimeoutMs: GAS_ORACLE_STRATEGY_ATTEMPT_TIMEOUT_MS,
+      totalTimeoutMs: calculateTimeout(startTime, GAS_ORACLE_STRATEGY_MAX_TIMEOUT_MS),
+      retries: 1,
+      delay: { type: 'random' as const, minDelayMs: RANDOM_BACKOFF_MIN_MS, maxDelayMs: RANDOM_BACKOFF_MAX_MS },
+    });
```
amarthadan, line 262: Hmm, this works because all the strategies contain only one blockchain call. In case of a strategy that would need to make more calls, this won't work (or it would but not ideally) because you would retry the whole strategy and not just a specific call you need to retry. I think it would be better to pass the `startTime` to each strategy letting them use it how many times they need it (calculating the timeout for each potential call), similar to how we do it in Airseeker. WDYT?

amarthadan (changes requested):

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts
vponline: I see that makes sense, I'll make the changes :+1:

packages/airnode-utilities/package.json
```diff
@@ -12,18 +12,22 @@
     "clean": "rimraf -rf *.tsbuildinfo ./dist ./build *.tgz",
     "compile": "tsc --build tsconfig.json",
     "test": "SILENCE_LOGGER=true jest --selectProjects unit",
+    "test:e2e": "SILENCE_LOGGER=true jest --selectProjects e2e --runInBand",
     "pack": "yarn pack"
   },
   "types": "dist/index",
   "main": "dist/index.js",
   "dependencies": {
+    "@api3/airnode-validator": "^0.7.1",
```
amarthadan, line 21: I think the latest version is 0.7.2, isn't it?

packages/airnode-utilities/src/constants.ts
```diff
@@ -7,3 +7,7 @@ export const PRIORITY_FEE_IN_WEI = 3_120_000_000;
 export const BASE_FEE_MULTIPLIER = 2;
 // By default there is no gas price multiplier
 export const GAS_PRICE_MULTIPLIER = 1;
+export const RANDOM_BACKOFF_MIN_MS = 0;
```
amarthadan, line 10: I think the min value should be always larger than 0, let's say 200 ms.

amarthadan (approved):
Just a few nitpicks, other than that LGTM :muscle:
Will need to be properly tested though, a lot of new stuff in this one :sweat_smile:

packages/airnode-utilities/package.json
vponline: hmm, all other airnode package versions are currently at  0.7.1 as well :thinking:

amarthadan: Oh, sorry, it should have been actually `0.7.0`, we're on `master` not on a `v0.7` branch. Never mind then, carry on :smile:

Issue #1205, opened 2022-06-13
Fix aws.env containing quotation marks causing problems


amarthadan:
[JIRA issue discussion](https://api3dao.atlassian.net/browse/AN-463)

kolenic-martin:
Hi @amarthadan, what is the expected behaviour when I do have quotation marks in aws.env? Thanks.

kolenic-martin:
Hi @vponline, thanks. I've just finished testing the deployment and removal with the quotation marks in aws.env and I can now confirm it works.

PR #1322, opened 2022-07-08
Improve deployer image interface

### LINKED ISSUE #1189: Improve Docker image interfaces
[Slack discussion](https://api3workspace.slack.com/archives/C02R79STBRA/p1650479082762419)
Currently, the Docker image interface is extremely demanding UX-wise. This should be much easier to use and not require you to know anything about the inner workings of the image. For example, `-r` can be an optional parameter that expects a particular directory structure (as in pre-alpha expected config.json and security.json to be in cd).

For reference API providers have no idea about what they are doing with the commands and even we, the internal users can’t easily tell if there is an error in a command or not.

This needs proper research and discussion.

dcroote:
Splits remove command in two and changes default receipt.json output folder to config/ rather than output/

For the two remove commands I went with `remove-with-receipt` and `remove-with-deployment-details`. I didn't choose `remove-without-receipt` for the latter as "without-receipt" doesn't tell the user what is expected.

Closes #1189.


dcroote:
CI docs failure being fixed in #1323

packages/airnode-deployer/src/cli/index.ts
```diff
@@ -91,14 +91,28 @@ yargs(hideBin(process.argv))
...
+    },
+    async (args) => {
+      logger.debugMode(args.debug as boolean);
+      logger.debug(`Running command ${args._[0]} with arguments ${longArguments(args)}`);
+
+      await runCommand(() => removeWithReceipt(args.receipt!));
+      return;
+    }
+  )
+  .command(
```
amarthadan, line 112: I think that the implementation of this command can be done simpler. I had to make all the arguments optional when there was an option to use just the receipt but that's not the case anymore. All of them except the `project-id` which is GCP specific can be made [required](http://yargs.js.org/docs/#demandOption) and the code will become simpler.

packages/airnode-examples/src/scripts/remove-airnode.ts
```diff
@@ -22,7 +22,7 @@ const main = async () => {
     integrationInfo.airnodeType === 'aws' && `--env-file ${secretsFilePath}`,
     integrationInfo.airnodeType === 'gcp' && `-v "${integrationPath}/gcp.json:/app/gcp.json"`,
     `-v ${integrationPath}:/app/output`,
```
amarthadan, line 24: There's no need to mount the `output` directory but you need to mount the `config` directory instead. The same goes for the deployment.

amarthadan (changes requested):

bbenligiray:
Does this only update the remove commands? What about the deployment commands?

amarthadan:
> Does this only update the remove commands? What about the deployment commands?

What about them? The default for the receipt placement is the config directory instead of the output directory but that's about it.

bbenligiray:
Don't you need to mount a bunch of volumes?

packages/airnode-deployer/src/cli/index.ts
```diff
@@ -91,14 +91,28 @@ yargs(hideBin(process.argv))
...
         description: 'Path to receipt file',
+        default: 'config/receipt.json',
         type: 'string',
       },
+    },
+    async (args) => {
+      logger.debugMode(args.debug as boolean);
+      logger.debug(`Running command ${args._[0]} with arguments ${longArguments(args)}`);
+
+      await runCommand(() => removeWithReceipt(args.receipt!));
```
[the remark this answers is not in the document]
dcroote: Fixed, thanks!

packages/airnode-examples/src/scripts/remove-airnode.ts
dcroote: I made the changes in the docs but missed that here. Fixed, thanks.

packages/airnode-deployer/src/cli/index.ts
dcroote: I've made the changes- significantly simpler!

packages/airnode-deployer/src/cli/index.ts
```diff
@@ -91,95 +86,76 @@ yargs(hideBin(process.argv))
...
-        gcp: [...descriptiveArgsCommon, 'project-id'],
-      };
-      const descriptiveArgsAll = uniq(
-        Object.values(descriptiveArgsCloud).reduce((result, array) => [...result, ...array])
-      );
-      const argsProvided = intersection([...descriptiveArgsAll, 'receipt'], keys(args));
-      const descriptiveArgsProvided = intersection(descriptiveArgsAll, keys(args));
 
-      if (isEmpty(argsProvided)) {
-        // Throwing strings to prevent yargs from showing error stack trace
```
amarthadan, line 144: Can you keep this comment on the throw statement?

amarthadan (approved):
LGTM :+1:

bbenligiray:
I think this issue is that https://github.com/api3dao/airnode/issues/1189#issuecomment-1169897970 doesn't include https://github.com/api3dao/airnode/issues/1189#issuecomment-1156292377, so this PR doesn't actually close https://github.com/api3dao/airnode/issues/1189

amarthadan:
> I think the issue is that [#1189 (comment)](https://github.com/api3dao/airnode/issues/1189#issuecomment-1169897970) doesn't include [#1189 (comment)](https://github.com/api3dao/airnode/issues/1189#issuecomment-1156292377), so this PR doesn't actually close #1189
> 
> What this was supposed to be was pre-alpha-like behavior, all files are read and written to the current directory.

It does include that.

There are two changes in this PR:
1. We will output the `receipt.json` file into `config` directory, removing the need for `output` directory. The only difference between this and the pre-alpha behavior is that it's not the current directory but the `config` directory which I think is better. The command has the `config` directory path as a default path for all the files so you don't need to actually provide any options. You still need to mount a volume when using Docker but you would have to do that anyway.
2. We will have two removal commands, `remove-with-receipt` for removing based on the `receipt.json` (again, the `config` directory is the default, no need to pass any options) and `remove-with-deployment-details` where you provide options instead of `receipt.json` (kind of a backup).

What do you think is missing?

bbenligiray:
My main goal was to get rid of `config/` because deployment files are already kept in directories with identifying names (for example dates in https://github.com/api3dao/operations/tree/main/data/apis/amberdata/deployments). I don't know what ChainAPI does but they should do something similar to prevent the user from deploying the first random configuration they find in `Downloads/`. Once you already keep all files in a timestamped/w.e. directory, requiring one more directory in that is redundant.
Also, it's oddly opinionated for the default config path to be `config/`, that's not intuitive and the only way to figure that out is by reading the docs or trial and error. 
So nothing seems to be missing, but I disagree with the default path for the configuration files.

amarthadan:
> My main goal was to get rid of config/ because deployment files are already kept in directories with identifying names (for example dates in https://github.com/api3dao/operations/tree/main/data/apis/amberdata/deployments).

And where would you put the config file then? In the root of the app directory? You won't be able to mount the whole directory then and you'll end up with more mount directives in the Docker command.
With `config` directory: 
```
docker run -it --rm -v $(pwd)/config:/app/config api3/airnode-deployer deploy
```
and without:
```
docker run -it --rm -v $(pwd)/config.json:/app/config.json -v $(pwd)/secrets.env:/app/secrets.env api3/airnode-deployer deploy
```
(not counting cloud auth files)

I don't know how exactly the operations work but I would assume that since they have these directories prepared like this they will just do for example:
```
docker run -it --rm -v $(pwd)/data/apis/amberdata/deployments/2022-01-31/airnode/aws:/app/config api3/airnode-deployer deploy
```
mounting the directory as a `config` directory within the container.

> Also, it's oddly opinionated for the default config path to be config/, that's not intuitive and the only way to figure that out is by reading the docs or trial and error.

I disagree. The directory is called `config` that's the first place where I would look for the configuration. Also, the docs are not the only way to figure it out, the CLI has help where the default is mentioned.

I'm fine with any other approach, maybe I don't get what's actually needed here but as far as I can tell the `config` directory is a better solution. Maybe @aquarat can provide some perspective from the operations standpoint.

bbenligiray:
You would do as in https://docs.api3.org/airnode/pre-alpha/guides/docker/deployer-image.html#deploy-first-time You can mount the whole `pwd` because `pwd` is not an app directory, it's a directory created and named to keep a specific deployment. This applies if the config files are in `deployments/2022-07-12/` or if they are downloaded from ChainAPI in a zipped directory called `blue-mackerel-2022-07-12`.
I'm saying that the user will struggle with requiring a specific config directory structure because I do each time I forget how everything works and have to figure it out again.

amarthadan:
> You would do as in https://docs.api3.org/airnode/pre-alpha/guides/docker/deployer-image.html#deploy-first-time You can mount the whole `pwd` because `pwd` is not an app directory, it's a directory created and named to keep a specific deployment. This applies if the config files are in `deployments/2022-07-12/` or if they are downloaded from ChainAPI in a zipped directory called `blue-mackerel-2022-07-12`. I'm saying that the user will struggle with requiring a specific config directory structure because I do each time I forget how everything works and have to figure it out again.

But that's exactly the same thing.

In the pre-alpha version you would do (assuming you're in the deployment directory, e.g. https://github.com/api3dao/operations/tree/main/data/apis/amberdata/deployments/2022-01-31/airnode/aws):
```
docker run -it --rm ... -v "$(pwd):/airnode/out" api3/airnode-deployer:pre-alpha
```
and in the current one:
```
docker run -it --rm ... -v "$(pwd):/app/config" api3/airnode-deployer
```

The only difference, on the container layer, is that before you had `/airnode/out` and now you have `/app/config`.

On the CLI layer, we by default (but still configurable) read the config from `config/config.json` and in pre-alpha it was read from `config.json` (at the root). The Docker container in pre-alpha was actually copying the files from `/airnode/out` to the app directory so they can be read (https://github.com/api3dao/airnode/blob/pre-alpha/Dockerfile#L19).

So, yes, we can do it "by a trick" like in the pre-alpha, mount to some other location and then copy the files within the container but I don't see a point. It makes more sense to use the same place for both CLI and Docker image for loading the configuration files. But if you insist we can change it.

dcroote:
> It makes more sense to use the same place for both CLI and Docker image for loading the configuration files.

This is a fair argument as it is a bit of a hack to mount `$(pwd)` to a random directory within the docker container and copy it to avoid having the `config/` directory.

That said, those details can be hidden from the user and I can see there is some friction added by `config/`. Further, having `aws.env` and `gcp.json` [default to being in the root directory](https://docs.api3.org/airnode/v0.7/grp-providers/guides/build-an-airnode/deploying-airnode.html#deployment) and not `config/` feels a bit strange- why not have everything in one place?

@bbenligiray if you'd like I'll work on the change.

amarthadan:
> Further, having aws.env and gcp.json [default to being in the root directory](https://docs.api3.org/airnode/v0.7/grp-providers/guides/build-an-airnode/deploying-airnode.html#deployment) and not config/ feels a bit strange- why not have everything in one place?

I initially didn't want to store `aws.env` in the `config` directory so it is not mounted into the container as the env variables were read by docker itself with `--env-file aws.env`. And I just followed the same placement for `gcp.json` even though that one is mounted. But that will change with https://github.com/api3dao/airnode/issues/1205 so we can move both `aws.env` and `gcp.json` to the `config` directory and have just one mount.

amarthadan:
> I'm confused. Tell me what the directory structure will be [here](https://docs.api3.org/airnode/v0.7/grp-providers/docker/deployer-image.html#configuration-files) and what the command will be [here](https://docs.api3.org/airnode/v0.7/grp-providers/docker/deployer-image.html#aws-2). What I was expecting is
> 
> ```
> my-airnode
> ├── aws.env     <- Used for AWS deployment
> ├── gcp.json    <- Used for GCP deployment
> ├── config.json
> └── secrets.env
> ```
> 
> and
> 
> ```
> docker run -it --rm \
>   --env-file aws.env \
>   -e USER_ID=$(id -u) -e GROUP_ID=$(id -g) \
>   -v "$(pwd):/app/config" \
>   api3/airnode-deployer:0.7.2 deploy
> ```
> 
> (credentials are mounted but we're getting a big UX improvement in return and it can be worked around by the advanced user)

Yes, exactly as you're expecting. Everything is mounted under the `config` directory, that's fine by me.

@vponline is already working on mounting the `aws.env` file in https://github.com/api3dao/airnode/issues/1205 and I'll create an issue for moving those cloud auth files into the `config` directory as well so the directory structure can be flat as you've described.

aquarat:
A bit late but this LGTM :+1: 
I like the `remove-with-deployment-details` functionality :smile:

packages/airnode-deployer/src/cli/index.ts
dcroote: Yes- comment restored.

PR #1328, opened 2022-07-12
Mount aws.env to deployer docker image and export with entrypoint

### LINKED ISSUE #1205: Fix aws.env containing quotation marks causing problems

vponline:
Closes #1205.

### DIFF
--- a/packages/airnode-deployer/docker/entrypoint.sh
+++ b/packages/airnode-deployer/docker/entrypoint.sh
@@ -4,6 +4,6 @@ DEPLOYER_BINARY="/usr/local/bin/airnode-deployer"
 USER_ID="${USER_ID:-0}"
 GROUP_ID="${GROUP_ID:-0}"
 
-test -f aws.env && export $(egrep -v '^#' aws.env | xargs)
+test -f aws.env && export $(egrep -v '^\s*#' aws.env | xargs)
 
 su-exec ${USER_ID}:${GROUP_ID} ${DEPLOYER_BINARY} "$@"
--- a/packages/airnode-deployer/docker/entrypoint.sh
+++ b/packages/airnode-deployer/docker/entrypoint.sh
@@ -4,4 +4,6 @@ DEPLOYER_BINARY="/usr/local/bin/airnode-deployer"
 USER_ID="${USER_ID:-0}"
 GROUP_ID="${GROUP_ID:-0}"
 
+test -f aws.env && export $(egrep -v '^#' aws.env | xargs)
+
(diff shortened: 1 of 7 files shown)


packages/airnode-deployer/docker/entrypoint.sh
```diff
@@ -4,4 +4,6 @@ DEPLOYER_BINARY="/usr/local/bin/airnode-deployer"
 USER_ID="${USER_ID:-0}"
 GROUP_ID="${GROUP_ID:-0}"
 
+test -f aws.env && export $(egrep -v '^#' aws.env | xargs)
```
amarthadan, line 7: Have you tested this?
Screenshot_20220713_101717

amarthadan (changes requested):

packages/airnode-deployer/docker/entrypoint.sh
vponline: Yes I tested deploying with both double quotes and without for the AWS credentials (on ubuntu). It doesn't work for you? 

I haven't done anything with the mnemonics since those are handled through the `secrets.env` file, right?

vponline: Also forgot to mention that I got that export line from Airkeeper (we use it to export credentials there)

amarthadan: Yeah, sorry, I haven't realized we're not talking about the secrets file but about the AWS credentials. Because this doesn't work with variables that contain spaces (as you can see in the screenshot). But the AWS credentials don't contain those so I guess it's fine.

Can you maybe change the regex to `'^\s*#'` to allow spaces in front of the `#` sign? I think that's an easy mistake that someone can make and it shouldn't cause problems.

vponline: Yeah definitely that makes sense, added! :smile:

amarthadan (approved):

packages/airnode-deployer/docker/entrypoint.sh
```diff
@@ -4,4 +4,6 @@ DEPLOYER_BINARY="/usr/local/bin/airnode-deployer"
 USER_ID="${USER_ID:-0}"
 GROUP_ID="${GROUP_ID:-0}"
 
+test -f aws.env && export $(egrep -v '^\s*#' aws.env | xargs)
```
Siegrift, line 7: (For the future): It's better to use the long argument versions, since it's totally not clear what `-v` does (compared to `--invert-match` which is self explanatory)

PR #1330, opened 2022-07-14
Move cloud auth files into the config directory

### LINKED ISSUE #1329: Move cloud auth files into the `config` directory
Currently, we encourage putting the cloud auth files (`aws.env` and `gcp.json`) into the root of the deployer app (e.g. `/app/aws.env` for the Docker image). We want to flatten the directory structure provided in our examples so we should encourage users to put all the files into one directory and mount it under the `/app/config` directory in the Docker image.

As far as I can tell, this will require only one change in the [airnode-deployer Dockerfile](https://github.com/api3dao/airnode/blob/master/packages/airnode-deployer/docker/Dockerfile#L33) and all the other changes will be only in the documentation.

dcroote:
Closes #1329.


packages/airnode-examples/src/scripts/deploy-airnode.ts
```diff
@@ -18,8 +18,8 @@ const main = async () => {
   const deployCommand = [
     `docker run -it --rm`,
     isWindows() ? '' : `-e USER_ID=$(id -u) -e GROUP_ID=$(id -g)`,
-    integrationInfo.airnodeType === 'aws' && `-v "${integrationPath}/aws.env:/app/aws.env"`,
-    integrationInfo.airnodeType === 'gcp' && `-v "${integrationPath}/gcp.json:/app/gcp.json"`,
+    integrationInfo.airnodeType === 'aws' && `-v "${integrationPath}/aws.env:/app/config/aws.env"`,
+    integrationInfo.airnodeType === 'gcp' && `-v "${integrationPath}/gcp.json:/app/config/gcp.json"`,
     `-v ${integrationPath}:/app/config`,
```
amarthadan, line 23: Is this actually necessary? The whole `integrationPath` is mounted as `/app/config` so the files will be there, right?

amarthadan (changes requested):

packages/airnode-examples/src/scripts/deploy-airnode.ts
dcroote: Yes the files will be there- good catch. I've removed the auth file mounts from the deploy and remove airnode-examples scripts.

amarthadan (approved):

PR #1349, opened 2022-07-26
Turn off GCP service disabling when removing deployment to enable multiple deployments per project

### LINKED ISSUE #1344: Removing one of multiple GCP deployments removes resources from a different deployment
I'm not 100% sure this is happening but it needs to be tested.

Steps to reproduce:
1. Deploy Airnode to GCP, stage = 'aaa'
2. Deploy another Airnode to GCP (first one still present), stage = 'bbb'
3. Remove the first one normally with `airnode-deployer` and receipt.json

Problems:
1. The removal will fail, saying that some of the services can't be disabled because there are still resources present (the second Airnode)
2. Some of the resources were removed anyway and some of the resources (namely cloud functions) from the second Airnode were removed as well

I suspect that this is because in the GCP recipes when using the `google_project_service` resource we have both `disable_dependent_services` and `disable_on_destroy` settings set to `true`. GCP is then trying to disable those services. My suspicion is that some of them can't be disabled because they contain resources (HTTP gateway), but some of them are disabled and their resources are removed (cloud functions). If this is the case, the fix should be setting those options to `false` for each `google_project_service` resource we have in the recipes.

vponline:
Closes #1344.

### DIFF
--- a/.changeset/yellow-emus-remain.md
+++ b/.changeset/yellow-emus-remain.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-deployer': patch
+---
+
+Set disable_dependent_services, disable_on_destroy in GCP recipes to false to allow removing one of multiple deployments from the same project
--- a/packages/airnode-deployer/terraform/airnode/gcp/main.tf
+++ b/packages/airnode-deployer/terraform/airnode/gcp/main.tf
@@ -1,8 +1,8 @@
 resource "google_project_service" "resourcemanager_api" {
   service = "cloudresourcemanager.googleapis.com"
 
-  disable_dependent_services = true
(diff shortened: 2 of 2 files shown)


Siegrift (commented):
@vponline did you test this?

vponline:
> @vponline did you test this?

Yep, I deployed two airnodes with different stages to the same GCP project, removed one with `deployer remove-with-receipt` (which worked) and then tested that the still remaining deployment worked with a request.

Siegrift (approved):
👍 Code LGTM, but I didn't test because I already spent too much time with reviews and I want to focus on my 0.8 tasks (when @kolenic-martin gets back I'd like him to test as well).

Issue #1184, opened 2022-06-13
Support HTTP gateways with airnode-client


amarthadan:
We need the containerized version to include an HTTP server that supports API test calls

Related to https://github.com/api3dao/airnode/issues/480

kolenic-martin:
The new functionality works as described in the docs; I've run the quick deploy tutorial and tried this out with various port numbers (3000 i.e. the default, 8000, 8080, 54321) as I did not come up with other variables that could potentially make this malfunction. Malformed requests are seen as such and thus are not fulfilled but that is expected.

Issue #1351, opened 2022-07-27
Add more extensive logging for gateway requests


Siegrift:
Currently, HTTP gateways (testing and signed) x (gcp, aws,local) do not log the incoming request before the arguments are validated. This means, we can end up with a rejected request but have no log entry about it (and what went wrong).

dcroote:
A [user on discord](https://discord.com/channels/758003776174030948/765618225144266793/1028888218822770688) could have been helped with more detailed logging:
> But it just return api call failed
I cannot figure out what is going wrong with this api call
Is there anyway to get more detail about failed calling?

martinkolenic:
Hi, I've tested this and reviewed it on the CloudWatch console.

Steps: Deploy Airnode on AWS, perform a HTTP request and check for log data on the CloudWatch console
Expected result: Request metadata logged in the console
Actual result: Matches expected result.

Screenshot:
![Image](https://user-images.githubusercontent.com/884979/203580752-664f77cc-2d8d-42c3-a10f-0dc3f8d31ff0.png)

dcroote:
Thanks @martinkolenic, but the logging changes were mostly added under the log level of `DEBUG`, so to see the updates you would need to deploy with `logLevel` set to `DEBUG` within `config.json` before making a gateway request.

Siegrift:
That's a good point. I think we the Airnode examples should use the DEBUG logger in config.json by default... WDYT?

martinkolenic:
Hi @dcroote, thanks for the information. I'll redeploy with that flag set to `DEBUG` and report back with results. @Siegrift, that might actually be a good idea in terms of investigating potential mishaps.

martinkolenic:
Hi @dcroote, I've rerun the test now.

Steps: In the coingecko-signed-data integration config.json, set the `logLevel` property to `DEBUG` and redeploy Airnode (if not deployed, deploy from scratch). Then, perform another POST request using `curl`. 

Expected result: Meaningful request response and a more extensive set of log entries compared to the `INFO` level that was set by default in the previous deployment (if any).
Actual result: Matches expected result

Screenshots:

AWS CloudWatch console:
![Image](https://user-images.githubusercontent.com/884979/203747371-251f54a4-72f3-4b75-a7a4-015cd0ce7a4a.png)

Airnode/`curl` output:
![Image](https://user-images.githubusercontent.com/884979/203747519-71d302bc-40f9-4497-8b9d-228e921af323.png)

[...]

PR #1352, opened 2022-07-27
Support HTTP gateways with airnode-client

Siegrift:
Closes https://github.com/api3dao/airnode/issues/1184

## How to review

Review commit by commit. Read https://github.com/api3dao/api3-docs/pull/935 and comment on the docs PR if there are things that should further clarified (or I simply forgot to document them).

## How to test

The documentation should be enough to have full understanding on how to configure the Airnode client. You can use https://docs.api3.org/airnode/v0.8/grp-providers/tutorial/quick-deploy-container/ or modify one of the Airnode examples. You can then use postman or curl.

## Bit related

While testing this I realized that we are not logging much information for the gateway requests. Created https://github.com/api3dao/airnode/issues/1351 for that.


packages/airnode-node/src/workers/local-gateways/server.ts
```diff
@@ -0,0 +1,141 @@
...
+      const [err, result] = await processHttpSignedDataRequest(config, endpointId, encodedParameters);
+      if (err) {
+        // Returning 500 because failure here means something went wrong internally with a valid request
+        res.status(500).send({ message: err.toString() });
+        return;
+      }
+
+      // We do not want the user to see {"success": true, "data": <actual_data>}, but the actual data itself
+      res.status(200).send(result!.data);
+    });
```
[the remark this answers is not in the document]
Siegrift: That is a good find, but out of scope this PR and likely never to be implemented. One should use cloud provider in production.

packages/airnode-node/src/workers/local-gateways/server.ts
```diff
@@ -0,0 +1,141 @@
...
+      const [err, result] = await processHttpRequest(config, endpointId, parameters);
+      if (err) {
+        // Returning 500 because failure here means something went wrong internally with a valid request
+        res.status(500).send({ message: err.toString() });
+        return;
+      }
+
+      // We do not want the user to see {"success": true, "data": <actual_data>}, but the actual data itself
+      res.status(200).send(result!.data);
+    });
```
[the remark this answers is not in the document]
Siegrift: That is a good find, but out of scope this PR and likely never to be implemented. One should use cloud provider in production.

packages/airnode-node/src/workers/local-gateways/gateways-api.ts
```diff
@@ -30,6 +30,7 @@ const httpRequestBodySchema = z.object({
 const httpSignedDataBodySchema = z.object({
   encodedParameters: z.string(),
 });
+const DEFAULT_PORT = 3000;
```
Siegrift: We can opt in to default ports, because the user can decide how to bind the port when running the Airnode container. E.g.:

```
docker run --rm -v /Users/emanuel/vacuumlabs/api3/airnode/packages/airnode-examples/integrations/coingecko:/app/config --publish 8000:3000 --name airnode api3/airnode-client
```

will expose the docker internal port (3000) to port 8000, so when making request you are doing it with port 8000.

Siegrift:
I am aware that the CI is failing and I'm investigating it. The PR is reviewable though.

packages/airnode-deployer/docker/entrypoint.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -e
```
vponline, line 3: I ran into some errors (first I had the wrong folder setup and then the wrong `nodeVersion` in `config.json`, 0.8 vs. 0.7.1) and I think this line causes the container to exit immediately if there is an error, so I had to do the `docker logs...` to see what happened.

I suspect this is the intentional flow, but thought I'd note it here in case anyone else is wondering why the container might be suddenly exiting.

vponline (approved):
This is a huge amount of work, very nice and clean implementation! The docs were also easy to follow. I think this is good to go after the tests and conflicts are resolved :+1: 

I tested this succesfully with the local quick deploy tutorial.

packages/airnode-deployer/docker/entrypoint.sh
Siegrift: Did you run docker in detached mode? We recommend that approach in https://docs.api3.org/airnode/v0.7/grp-providers/docker/client-image.html#checking-airnode-logs but I remember I disliked this.

The `set -e` is needed, because when the `node /app/src/cli/validate-config.js` fails, the bash script continues (with Airnode starting) which should not happen.

vponline: Yep I ran it with detached.

> The set -e is needed, because when the node /app/src/cli/validate-config.js fails, the bash script continues (with Airnode starting) which should not happen.

This makes perfect sense

packages/airnode-validator/src/config/config.ts
```diff
@@ -440,3 +446,4 @@ export type Trigger = SchemaType<typeof triggerSchema>;
 export type Triggers = SchemaType<typeof triggersSchema>;
 export type Heartbeat = SchemaType<typeof heartbeatSchema>;
 export type Amount = SchemaType<typeof amountSchema>;
+export type EnabledGatewaySchema = SchemaType<typeof enabledGatewaySchema>;
```
vponline, line 449:
```suggestion
export type EnabledGateway = SchemaType<typeof enabledGatewaySchema>;
```
This is a great idea (and I will make sure to add to cloud handlers as well), but perhaps rename to without the `Schema` part for the type name to keep consistent with the other types.

packages/airnode-node/src/workers/local-gateways/server.ts
```diff
@@ -0,0 +1,185 @@
...
+        if (!originVerification.success) {
+          res.status(400).send(originVerification.error);
+          return;
+        }
+
+        // Set headers for the responses
+        res.set(originVerification.headers).status(204).send('');
+        return;
+      }
+      res.set(originVerification.headers);
```
vponline, line 72:
```suggestion
        res.set(originVerification.headers).status(204).send('');
        return;
      }
       // Set headers for the responses
      res.set(originVerification.headers);
```
Nit: to move the comment down

vponline (approved):
Looks good, just two minor notes:

packages/airnode-node/src/workers/local-gateways/server.ts
Siegrift: I enriched the comment a bit. LMK if it makes sense.

vponline: Yep looks good :+1:

packages/airnode-validator/src/config/config.ts
```diff
@@ -173,7 +173,7 @@ export const apiKeySchema = z.string().min(30).max(120);
 
 export const corsOriginsSchema = z.array(z.string());
 
-export const enabledGatewaySchema = z
+export const EnabledGateway = z
```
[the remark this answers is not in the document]
Siegrift: Sorry, I've made an incorrect find/replace when renaming the type. Will fix.

I'm also looking into the unit and e2e tests, because I can't seem to get them passing for this PR.

vponline (approved):
Thanks for updates to the gas oracle tests. I hadn't seen jest timers used before, but will definitely come back to this PR for reference as they look quite useful.

Also thanks for adding the comments, very clear on what the steps are now. All the comments should be covered now.

Issue #1181, opened 2022-06-13
Make Airnode logger utility more usable for other projects


amarthadan:
The current [logging utility](https://github.com/api3dao/airnode/tree/master/packages/airnode-utilities/src/logging) in Airnode was built only with Airnode in mind. We’re starting using it in other tools (Airkeeper, Airseeker) and we’re running into some limitations.

I propose merging the [LogOptions](https://github.com/api3dao/airnode/blob/master/packages/airnode-utilities/src/logging/types.ts#L13) fields meta and additional into one field allowing to add any key/value pair instead of the prepared few in meta. That way it’s not bound to one project but reusable everywhere.

Siegrift:
It would be better if we moved the logging utilities into a separate repo.

amarthadan:
> It would be better if we moved the logging utilities into a separate repo.

What would be the difference?

Why is this blocked?

Siegrift:
> What would be the difference?

The logging utilities should be a generic library. It's awkward to depend on a generic purpose library of a specific service (Airnode) in other services we use. I am fine with first implementing this in the Airnode, but we should still discuss this and create a new task for it if necessary.

> Why is this blocked?

Until the discussion is over and we've decided where to implement this.

amarthadan:
> I am fine with first implementing this in the Airnode, but we should still discuss this and create a new task for it if necessary.

I agree, let's do that. Implement first under Airnode and potentially move it after.

Siegrift:
Created https://github.com/api3dao/airnode/issues/1312.

Siegrift:
I've started working on this and will resume after vacation. If this issue becomes a blocker and I am not back feel free to take over.

Siegrift:
@kolenic-martin I tested this on AWS, seeing the logs metadata (coordinator ID, chainId and provider name being set correctly). Can you test on GCP and locally please?

kolenic-martin:
Sure, no prob. I'll be off till Monday, is it okay to do it then? Or should
I squeeze this in sometime today?

Siegrift:
Monday is fine.

kolenic-martin:
Hi @Siegrift, sorry for the late reply but I actually tested this earlier in the week but I pushed back writing the report as I juggled other issues. The log metadata seems to be valid both in GCP and the local output. Sorry I didn't give an OK earlier.

PR #1353, opened 2022-07-29
Merge meta and additional fields of logger

Siegrift:
Closes https://github.com/api3dao/airnode/issues/1181

I also noticed that we stopped including some metadata in the logger in https://github.com/api3dao/airnode/pull/1341 so I added it back.

TODO:
- [x] - I want to test this on AWS to see if the logs work as expected.

### DIFF
--- a/packages/airnode-deployer/src/handlers/aws/index.ts
+++ b/packages/airnode-deployer/src/handlers/aws/index.ts
@@ -1,5 +1,5 @@
 import * as path from 'path';
-import { logger, DEFAULT_RETRY_DELAY_MS, randomHexString, setLogOptions } from '@api3/airnode-utilities';
+import { logger, DEFAULT_RETRY_DELAY_MS, randomHexString, setLogOptions, addMetadata } from '@api3/airnode-utilities';
 import { go } from '@api3/promise-utils';
 import {
   handlers,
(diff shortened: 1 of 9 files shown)


packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -44,7 +44,10 @@ export async function run(payload: WorkerPayload): Promise<AWSLambda.APIGatewayP
 // https://api3dao.atlassian.net/browse/AN-527
 
 async function initializeProvider(payload: InitializeProviderPayload) {
-  const stateWithConfig = { ...payload.state, config: parsedConfig };
+  const { state } = payload;
+  const { chainId, name: providerName } = state.settings;
+  addMetadata({ 'Chain-ID': chainId, Provider: providerName });
```
vponline, line 49: These could also be set in the `run` function but I guess it would have to be with some conditional checks because they are different for each of the handlers, so probably best to leave like this :thinking:

vponline (approved):
These look great, thanks for fixing the metadata  and adding tests :pray:

packages/airnode-deployer/src/handlers/aws/index.ts
Siegrift: Yeah, we could have the chainID set right away in the run handler, but I do not see much point in doing so since there need to be the individual `addMetadata` calls.

Siegrift:
image

image

See that in the first screenshot only the coordinator ID is set, in the second one all of coordinator, chain ID and provider name are set.

PR #1357, opened 2022-08-02
Add corsOrigins option

### LINKED ISSUE #1200: Research (and implement if possible) enabling CORS on the HTTP gateways
[Slack discussion](https://api3workspace.slack.com/archives/C02R79STBRA/p1646325797158989)
Should be configurable.

vponline:
Closes #1200


packages/airnode-deployer/terraform/airnode/aws/templates/httpGw.yaml.tpl
```diff
@@ -59,9 +59,6 @@ components:
       name: x-api-key
       in: header
 
-security:
-  - apiKey: []
-
```
vponline, line 64: I had to remove this at this level since the browser preflight checks do not include the apiKey

packages/airnode-deployer/terraform/airnode/aws/templates/httpGw.yaml.tpl
```diff
@@ -96,3 +103,34 @@ paths:
...
+              schema:
+                type: string
+            Access-Control-Allow-Origin:
+              schema:
+                type: string
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/EndpointResponse"
+      x-amazon-apigateway-request-validator: none
```
vponline, line 126: This is also to ensure that preflight checks are only passed through the the lambda without any validation of  the shape

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.test.ts
```diff
@@ -536,7 +536,8 @@ describe('Gas oracle', () => {
           new Promise((resolve) => {
             setTimeout(() => {
               return resolve({} as any);
-            }, GAS_ORACLE_STRATEGY_ATTEMPT_TIMEOUT_MS);
+              // Set timeout to exceed attempt maximum
+            }, GAS_ORACLE_STRATEGY_ATTEMPT_TIMEOUT_MS + 10);
```
vponline, line 540: This is to fix a flaky test where sometimes the timeout finished within the timeout limit and the tests were not passing.

packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -104,6 +104,17 @@ export async function processHttpRequest(
     format: parsedConfig.nodeSettings.logFormat,
     level: parsedConfig.nodeSettings.logLevel,
   });
+
+  // Check if the request origin header is allowed in the config
+  const originVerification = verifyRequestOrigin(
```
Siegrift, line 109: It's better to wrap the whole thing in:
```
if (event.httpMethod === 'OPTIONS')...
```

packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -104,6 +104,17 @@ export async function processHttpRequest(
...
+    parsedConfig.nodeSettings.httpGateway.enabled ? parsedConfig.nodeSettings.httpGateway.corsOrigins : [],
+    event.headers['origin']
+  );
+  // Respond to preflight requests if the origin is allowed
+  if (originVerification.success && event.httpMethod === 'OPTIONS') {
+    return { statusCode: 204, headers: originVerification.headers, body: '' };
```
Siegrift, line 115: I like 204 status code for success response.

However, we should also return early if there is an error. I suggest status code `400` with some message.

packages/airnode-deployer/src/handlers/common.ts
```diff
@@ -95,3 +95,22 @@ export function verifyHttpSignedDataRequest(
...
+  'Access-Control-Allow-Origin': origin,
+  'Access-Control-Allow-Methods': 'OPTIONS,POST',
+  'Access-Control-Allow-Headers': 'Content-Type,x-api-key',
+});
+
+export const verifyRequestOrigin = (allowedOrigins: string[], origin?: string) => {
+  const allowedOrigin = checkRequestOrigin(allowedOrigins, origin);
+
+  // Return CORS headers to be used by the response if the origin is allowed
+  if (allowedOrigin) return { success: true, headers: buildCorsHeaders(allowedOrigin) };
```
Siegrift, line 113: Usually, the service returns all allowed origins as a response to preflight request. Any reason why change this to only include the allowed one?

packages/airnode-deployer/terraform/airnode/aws/templates/httpGw.yaml.tpl
```diff
@@ -96,3 +103,34 @@ paths:
         responses:
           default:
             statusCode: 200
+    options:
+      parameters:
+        - $ref: "#/components/parameters/endpointId"
+      responses:
+        "200":
```
Siegrift, line 110: Not that it matters much, but in the service you actually return 204

packages/airnode-deployer/terraform/airnode/aws/templates/httpGw.yaml.tpl
Siegrift: Why is this needed?

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.test.ts
```diff
@@ -536,7 +536,8 @@ describe('Gas oracle', () => {
           new Promise((resolve) => {
             setTimeout(() => {
               return resolve({} as any);
-            }, GAS_ORACLE_STRATEGY_ATTEMPT_TIMEOUT_MS);
+              // Set timeout to exceed attempt maximum
```
Siegrift, line 539:
```suggestion
              // Set timeout to exceed attempt maximum to reduce this test flakiness
```

Siegrift (changes requested):

packages/airnode-deployer/src/handlers/aws/index.ts
vponline: hmm we need  the CORS headers returned for the `POST` method as well, right?

packages/airnode-deployer/src/handlers/aws/index.ts
vponline: By error do you mean that the origin is not allowed? The browser handles the CORS not allowed error if the correct headers are not returned, but do you mean that we should have a custom error for this case?

packages/airnode-deployer/src/handlers/common.ts
vponline: Do you mean all origin URLs as a comma separated list? I wasn't aware that this is possible and I can't find an example.

packages/airnode-deployer/terraform/airnode/aws/templates/httpGw.yaml.tpl
vponline: The preflight check doesn't pass the request body and parameters validation so this disabled it

packages/airnode-deployer/src/handlers/aws/index.ts
Siegrift: Your implementation currently does set it only for OPTIONS requests. But I think setting the headers for OPTIONS is enough (https://stackoverflow.com/a/71429784).

packages/airnode-deployer/src/handlers/aws/index.ts
Siegrift: If the request is OPTIONS we should not even attempt to call lambda, just return the headers to the browser.

packages/airnode-deployer/src/handlers/aws/index.ts
vponline: Hmm I do set them also for the POST method [here](https://github.com/api3dao/airnode/pull/1357/files#diff-6f3bd240ed3040ed5cbd29ed7ea1b83c7735fdcf318dddf814c7846f493bc830R137) (and they are included when testing with a react app)

vponline: The [GCP docs](https://cloud.google.com/functions/docs/samples/functions-http-cors#code-sample) do add the Origin header and [AWS docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-cors.html) seem to add all

:thinking:

Siegrift: Maybe it's a convention (or they simply did that out of convenience). My main point is not to call the gateway when you receive the OPTIONS request, but only return early the necessary headers. We can include the headers on the POST, I can't see an issue with that.

vponline: > My main point is not to call the gateway when you receive the OPTIONS request, but only return early the necessary headers.

I'm afraid I don't quite follow what you mean by this. 

For AWS it is possible to setup a mock response for the gateway to handle the options method (so that the lambda is not called), but this is not possible for GCP so I thought it would be better to handle both cloud providers the same way.

Siegrift: > I'm afraid I don't quite follow what you mean by this.

Sorry for the confusion, I realized I typed it wrong. I see that you return early when the OPTIONS request is successful, but we are missing an early return when the OPTIONS origin verification fails

packages/airnode-deployer/src/handlers/aws/index.ts
Siegrift: I meant the Airnode handlers, but I chose poor words to decscribe this. I mean basically the same thing as in the other comment, so we can resolve this one.

packages/airnode-deployer/src/handlers/aws/index.ts
vponline: > but we are missing an early return when the OPTIONS origin verification fails

Ah I see what you mean, I'll add a status 400 and an error message here :+1:

packages/airnode-deployer/terraform/airnode/aws/templates/httpGw.yaml.tpl
vponline: So it turns out that this is infact not needed (just tested a deployment), I've removed it.

packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -139,6 +154,21 @@ export async function processHttpSignedDataRequest(
     format: parsedConfig.nodeSettings.logFormat,
     level: parsedConfig.nodeSettings.logLevel,
   });
+
+  // Check if the request origin header is allowed in the config
+  const originVerification = verifyRequestOrigin(
+    parsedConfig.nodeSettings.httpGateway.enabled ? parsedConfig.nodeSettings.httpGateway.corsOrigins : [],
+    event.headers['origin']
```
Siegrift, line 161: Nit: Why not `.origin`? (We did not enable the TS setting for explicit property access so this syntax is not required).

Siegrift (approved):
👍 LGTM (but haven't deployed/tested on AWS).

After you merge I'll update the local gateways PR to support CORS as well.

packages/airnode-deployer/src/handlers/aws/index.ts
vponline: hmm no particular reason, I'll make the change

PR #1361, opened 2022-08-04
Cors config fix

vponline:
This fixes the CORS header verification to not affect non-CORS/non-browser `POST` requests.

Thanks @Siegrift for checking!


Siegrift (approved):
👍 LGTM.

PR #1364, opened 2022-08-05
Refactor `generate-mnemonic`

### LINKED ISSUE #1320: `generate-mnemonic` is too opinionated
The `generate-mnemonic` admin CLI command assumes the user is creating an Airnode wallet mnemonic and prints its messages accordingly. Then, we recommend the user to use this command for other things like creating a sponsor wallet, which potentially confuses users (https://discord.com/channels/758003776174030948/765618225144266793/994595049880944692).

The command should be renamed as `generate-airnode-mnemonic`. We should add another command `generate-mnemonic` that only displays the mnemonic and doesn't mention an Airnode address in the messages (instead refer to it as "Default wallet address (path:m/44'/60'/0'/0/0)")i

vponline:
Closes #1320


.changeset/olive-radios-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@api3/airnode-admin': patch
```
Siegrift, line 2: Nit: I'd say it's minor since it's more like a new feature not a bug.

packages/airnode-admin/src/cli.ts
```diff
@@ -509,7 +509,7 @@ yargs
     }
   )
   .command(
-    'generate-mnemonic',
+    'generate-airnode-mnemonic',
     'Generates a random mnemonic. Uses "ethers.Wallet.createRandom" under the hood',
```
Siegrift, line 513: We should enrich the description of this command, since now its not clear when to use which one.

packages/airnode-admin/src/cli.ts
```diff
@@ -530,6 +530,28 @@ yargs
...
+
+      logger.log(
+        [
+          'This mnemonic is created locally on your machine using "ethers.Wallet.createRandom" under the hood.',
+          'Make sure to back it up securely, e.g., by writing it down on a piece of paper:',
+          '',
+          ...encaseMnemonic(mnemonic),
+          '',
+          `The default wallet address (path:m/44'/60'/0'/0/0) for this mnemonic is: ${address}`,
+          `The default wallet xpub (path:m/44'/60'/0'/0/0) for this mnemonic is: ${xpub}`,
```
Siegrift, line 549: Do we need xpub mentioned here? It feels a bit extraneous.

Siegrift (changes requested):

packages/airnode-admin/src/cli.ts
vponline: I thought about leaving it in since it could be useful since it's also included for the other command, but I can remove if you'd prefer

Siegrift: Probably @bbenligiray is best to answer.

packages/airnode-admin/src/cli.ts
```diff
@@ -509,8 +509,8 @@ yargs
     }
   )
   .command(
-    'generate-mnemonic',
-    'Generates a random mnemonic. Uses "ethers.Wallet.createRandom" under the hood',
+    'generate-airnode-mnemonic',
+    'Generates a random mnemonic which can be used in the Airnode config and displays the corresponding address and extended public key. Uses "ethers.Wallet.createRandom" under the hood',
```
Siegrift, line 513: (subjective nit)

```suggestion
    'Generates a random mnemonic which can be used in the Airnode config and displays the extended public key. Uses "ethers.Wallet.createRandom" under the hood',
```

The correspodning address sounds wierd. We could replace "corresponding" with "default". Feel free to disregard.

vponline: We use `default wallet...` for the other `generate-mnemonic` command so perhaps just leave it out, although the current wording is how the api3-docs puts it :thinking:

packages/airnode-admin/src/cli.ts
bbenligiray: xpub is probably not needed here

Siegrift (approved):
👍 LGTM

PR #1372, opened 2022-08-08
Add per-endpoint rrp response caching

### LINKED ISSUE #1190: requestId-keyed per-endpoint caching
This leverages airnode-utilities caching to enable response caching. This is useful for RNG and non-idempotent API operations.

An example implementation for ANU QRNG can be found here: https://github.com/api3dao/airnode-qrng/commit/75c54a0f5dddbd2a29b7f303310356750d2fd1fc#diff-0586244b9496f86cf4c0bdd8ff5bcab4714ed16033fa4f6137e8dacdf732d124R93 

Ask @aquarat for more info if needed.

dcroote:
Closes #1190.


packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -27,6 +36,7 @@ export async function startCoordinator() {
   });
   await handlers.startCoordinator(parsedConfig, coordinatorId);
   const response = { ok: true, data: { message: 'Coordinator completed' } };
+  caching.syncFsSync();
```
Siegrift, line 39: I wonder how the caching will work for the gateways.

packages/airnode-deployer/src/handlers/gcp/index.ts
```diff
@@ -31,6 +40,9 @@ export async function startCoordinator(_req: Request, res: Response) {
   });
   await handlers.startCoordinator(parsedConfig, coordinatorId);
   const response = { ok: true, data: { message: 'Coordinator completed' } };
+  // TODO: line below (suspiciously?) not present in original PR:
+  // https://github.com/api3dao/airnode-qrng/commit/75c54a0f5dddbd2a29b7f303310356750d2fd1fc#diff-307ba728f020f0438491dbae15a76ec9421457225fcf4f4966f698b74a87263cR4-R8
+  caching.syncFsSync();
```
Siegrift, line 45: I wonder whether we need this and whether OS doesn't provide some guarantees for this. But it might just be forgotten in the original PR.

packages/airnode-node/src/handlers/start-coordinator.ts
```diff
@@ -88,12 +89,50 @@ function aggregateApiCalls(state: CoordinatorState) {
...
+  const aggregatedApiCalls = Object.values(filteredUncachedAggregatedApiCalls);
+
   const [logs, processedAggregatedApiCalls] = await calls.callApis(aggregatedApiCalls, getWorkerOptions(state));
   logger.logPending(logs);
 
+  processedAggregatedApiCalls
+    .filter((call) => call.success)
+    .forEach((call) => {
+      // "as any" from: https://github.com/api3dao/airnode/pull/871/files#diff-d9e817fb4873840d22bada5236869347e19f945d65b42eb439f5c9dc689404bfR38
+      const { encodedValue, signature } = (call as any).data;
```
Siegrift, line 126: as RegularApiCallSuccessResponse

packages/airnode-node/src/handlers/start-coordinator.ts
```diff
@@ -88,12 +89,50 @@ function aggregateApiCalls(state: CoordinatorState) {
...
+      caching.addKey(`requestId-${call.id}`, { encodedValue, signature });
```
Siegrift, line 127: You can just save `call.data`

packages/airnode-node/src/handlers/start-coordinator.ts
```diff
@@ -88,12 +89,50 @@ function aggregateApiCalls(state: CoordinatorState) {
...
+        ...value,
+        // errorMessage: undefined,
+        encodedValue,
+        signature,
+      };
+    });
+
+  const aggregatedApiCalls = Object.values(filteredUncachedAggregatedApiCalls);
+
   const [logs, processedAggregatedApiCalls] = await calls.callApis(aggregatedApiCalls, getWorkerOptions(state));
```
Siegrift, line 119: We do not want to make the API call for cached requests

packages/airnode-validator/test/fixtures/ois.json
```diff
@@ -52,6 +52,7 @@
         "method": "get",
         "path": "/convert"
       },
+      "cacheResponses": "false",
```
Siegrift, line 55: Should we make this more configurable? (E.g. how long to cache a value)

Siegrift: I disagree that this should be in OIS. OIS is an abstraction that shouldn't know anything about Airnode. This flag should probably go to `triggers.rrp[n].cacheResponses`

packages/airnode-node/src/handlers/start-coordinator.ts
dcroote: Right- that's why we now have (emphasis on "Uncached"): 
```ts
const aggregatedApiCalls = Object.values(filteredUncachedAggregatedApiCalls);
```

https://github.com/api3dao/airnode/pull/1372/files#diff-0586244b9496f86cf4c0bdd8ff5bcab4714ed16033fa4f6137e8dacdf732d124R117

packages/airnode-node/src/handlers/start-coordinator.ts
dcroote: Incorporated- thanks.

packages/airnode-node/src/handlers/start-coordinator.ts
dcroote: Incorporated- thanks.

packages/airnode-validator/test/fixtures/ois.json
dcroote: Under that suggestion wouldn't we also need `cacheResponses` in the gateways `http[n]` and `httpSignedData[n]`? It also seems like it would be easy to mistakenly configure caching inconsistently between triggers then.

packages/airnode-deployer/src/handlers/gcp/index.ts
dcroote: I think I found the reason: on GCP `/tmp` [is in RAM](https://cloud.google.com/appengine/docs/standard/go/using-temp-files)

packages/airnode-validator/test/fixtures/ois.json
Siegrift: > Under that suggestion wouldn't we also need cacheResponses in the gateways http[n] and httpSignedData[n]?

Yeah, it needs to go there, because it can't go to OIS and I don't see a better place.

> It also seems like it would be easy to mistakenly configure caching inconsistently between triggers then.

You mean having 2 triggers for the same endpoint and only one of them has caching enabled? Not sure if this is a feature or a bug though. We may open a slack discussion about this, because I see that this whole discussion can be opinionated... One might argue that caching is a generic concept and advocate to put it on OIS.

packages/airnode-node/src/handlers/start-coordinator.ts
```diff
@@ -88,12 +94,43 @@ function aggregateApiCalls(state: CoordinatorState) {
...
+  const cachedKeys = caching.getKeys('requestId-');
+
+  const filteredUncachedAggregatedApiCalls = pickBy(
+    aggregatedApiCallsById,
+    (_value, key) => !cachedKeys.includes(`requestId-${key}`)
+  );
+
+  const filteredCachedAggregatedApiCalls = Object.entries(aggregatedApiCallsById)
+    .filter(([key, _value]) => cachedKeys.includes(`requestId-${key}`))
+    .map(([key, value]) => {
```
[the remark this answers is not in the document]
dcroote: Thanks- updated within `filter` and `map`.

packages/airnode-validator/test/fixtures/ois.json
dcroote: Resolving per the agreement to have it in `triggers.rrp[n]` and not cache gateway responses.

packages/airnode-validator/src/config/config.ts
```diff
@@ -22,9 +22,17 @@ export const triggerSchema = z
   })
   .strict();
 
+export const rrpTriggerSchema = triggerSchema
+  .merge(
+    z.object({
+      cacheResponses: z.boolean().optional(),
```
Siegrift, line 28: Why not enforce this?

packages/airnode-utilities/src/caching/index.ts
```diff
@@ -202,6 +203,35 @@ const init = () => {
...
+    logger.error((e as Error).message);
+  }
+};
+
+/**
+ * Syncs the temporary filesystem asynchronously.
+ * Lambda will "freeze" and "thaw" the execution context and with it any async promises in progress. We create and
+ * forget this promise here, but it will resolve at some point, either in this invocation or a later one.
+ */
+export const syncFsASync = () => {
```
Siegrift, line 224: I am not sure I understand why would you want to call this asynchronously.

packages/airnode-node/src/handlers/start-coordinator.ts
```diff
@@ -88,12 +95,44 @@ function aggregateApiCalls(state: CoordinatorState) {
...
   const [logs, processedAggregatedApiCalls] = await calls.callApis(aggregatedApiCalls, getWorkerOptions(state));
   logger.logPending(logs);
 
+  processedAggregatedApiCalls
+    .filter((call) => call.success && call.cacheResponses)
+    .forEach((call) => {
+      caching.addKey(`requestId-${call.id}`, (call as RegularApiCallSuccessResponse).data);
+    });
+
+  caching.syncFsASync();
```
Siegrift, line 130: I think we should not ignore this promise and instead sync the FS synchronously. Also, do we need to call this at all? Isn't the call in the deployer enough? Is this for airnode-client?

Siegrift (commented):
The implementation looks good, but I am not sure about the storage syncing. It feels to me that:
1. It's important and we should do it more thoroughly/predictably
2. It's not needed and can be completely removed

If it's 1 (which I think it is because of the lambda timeout) then I'd suggest making the api more predictable. E.g.:
```
export const createCacheTransaction = (fn: (cache: Cache) => void) => {
  fn();
  syncFsSync();
};
```
where the `cache` object contains all cache functions. You only export the read-only functions from the caching module. This way it's guaranteed that the cache writes are persisted on the storage. (This mimics the DB transactional pattern).

WDYT?

packages/airnode-validator/src/config/config.ts
dcroote: Good point, I don't see why not

packages/airnode-utilities/src/caching/index.ts
dcroote: I see your point and would agree it seems sync is what we need to ensure predictability, but @aquarat may have had a reason for using it `start-coordinator` in his original implementation? https://github.com/api3dao/airnode-qrng/blob/75c54a0f5dddbd2a29b7f303310356750d2fd1fc/packages/airnode-node/src/handlers/start-coordinator.ts#L120

Siegrift: I'd say it was for performance. Basically, let Airnode continue processing the API responses as soon as possible and sync the FS sometime later. But I'd argue whether it helps. We may confirm this later when @aquarat comes back.

For now, I suggest using the sync version (and removing the async one completely). I don't want you blocked on this.

dcroote: > I suggest using the sync version (and removing the async one completely)

Done

packages/airnode-node/src/handlers/start-coordinator.ts
dcroote: > Is this for airnode-client?

Yes I believe so, but this raises the question of if it is here, why then do we need to initialize the cache and (for aws) call sync within the deployer? It seems like `caching.init()` and `caching.syncFsSync()` could be removed from there unless I am missing something.

Siegrift: Yeah, seems to me as well, but it doesn't hurt to keep them I guess.

packages/airnode-validator/src/config/config.ts
Siegrift: can we enforce it?

Siegrift (approved):
The implementation 👍 LGTM, but I haven't tested this.

packages/airnode-utilities/src/caching/index.ts
aquarat: Exactly as @Siegrift assumed.
Lambda's `tmpfs` is extremely fast, so `sync` vs `async` won't have a noticeable performance impact and `sync` is safer anyway. :)

dcroote: Great, thanks @aquarat!

packages/airnode-validator/src/config/config.ts
dcroote: 👍  done

dcroote:
To confirm function, I ran the coingecko airnode-examples integration locally with hardhat and airnode-client-dev (commit b382827397acb056c688ad9ee4a33f3e5ee3e022) after updating `config.json` with: `"cacheResponses": true`.

I made two requests:

```
0xbf65e684ecc2adb3e41212134fdd8c2999e6a125ba170b8624dc0938de86bad0
Ethereum price is 1685.89 USD

0x7fbbec776797f287f20fbd45095954939586f91e26e84bdd3cb4281c7da61236
Ethereum price is 1685.96 USD
```

I then restarted my local hardhat network and redeployed the contracts **without shutting down the airnode client container**.

If caching is working, making two new requests (which will have the same requestIds as before) should return the same (cached) response (ETH price) as before, which is indeed what happened:
```
0xbf65e684ecc2adb3e41212134fdd8c2999e6a125ba170b8624dc0938de86bad0
Ethereum price is 1685.89 USD

0x7fbbec776797f287f20fbd45095954939586f91e26e84bdd3cb4281c7da61236
Ethereum price is 1685.96 USD
```

And then a third request fetches a new response (price) as expected:
```
0x685bd4472679ce981d634fca32a10196784b8b24d5568c0a08a0ec8027865981
Ethereum price is 1686.25 USD
```

PR #1373, opened 2022-08-08
Improve validator cloudProvider error message

### LINKED ISSUE #1295: Provide better validator error message for invalid "cloudProvider" settings
You can reproduce this by creating an invalid `cloudProvider` section inside `config.json`. E.g. missing the `projectId` for GCP configuration:

```
    ...
    "cloudProvider": {
      "type": "gcp",
      "region": "us-east1",
      "disableConcurrencyReservations": false
    }
    ...
```

results in Image

The issue is in https://github.com/api3dao/airnode/blob/master/packages/airnode-validator/src/config/config.ts#L186 where we should use `discriminatedUnion` instead of `union`

dcroote:
Fixes #1295


Siegrift (approved):
👍 LGTM

Issue #1333, opened 2022-07-19
Automate releasing of proper packages (outside the CI)


amarthadan:
Based on https://api3dao.atlassian.net/l/cp/XEq99q90

amarthadan:
@martinkolenic I don't think there's anything for you to test here. Moving to Done.

Issue #1332, opened 2022-07-19
Automate releasing snapshot packages (outside the CI)


amarthadan:
Based on https://api3dao.atlassian.net/l/cp/XEq99q90

PR #1382, opened 2022-08-16
Rework building of Docker images

### LINKED ISSUE #1331: Building Docker images from published NPM packages
We can use published packages (either from a local NPM registry or from the official one) to build the Docker images. This will both make the Docker builds faster and the resulting docker images smaller.

amarthadan:
Close #1331

Should be pretty much the final version as far as the Docker containers go but it will be expanded on in #1332 and #1333 so we can publish NPM packages even without building the Docker images.


.github/workflows/build-test.yml
```diff
@@ -67,34 +67,12 @@ jobs:
...
-      - name: Push airnode artifacts to Docker Hub
-        run: |
-          docker push api3/airnode-artifacts-dev:${{ github.sha }}
-  docker-build-general:
-    name: Build docker image
-    runs-on: ubuntu-latest
-    needs: docker-artifacts
-    strategy:
-      matrix:
-        package: [deployer, admin, client]
```
dcroote, line 97: I was a bit concerned about build times when I saw the matrix strategy removed, but it doesn't appear to be too affected:

This run: 16m 19s

Some before this:
- 13m 44s
- 15m 26s
- 15m 3s
- 14m 32s
- 16m 42s

packages/airnode-node/docker/Dockerfile
```diff
@@ -1,30 +1,34 @@
...
-COPY --from=api3/airnode-artifacts /dependencies/airnode-node ${appDir}/node_modules
-COPY --from=api3/airnode-artifacts /packages ${appDir}/node_modules/@api3/
-COPY --from=api3/airnode-artifacts /build/packages/airnode-node/dist ${appDir}/
-COPY packages/airnode-node/docker/airnode-crontab ${cronjob}
-COPY packages/airnode-node/docker/entrypoint.sh /entrypoint.sh
+COPY airnode-crontab ${cronjob}
+COPY entrypoint.sh /entrypoint.sh
 
-# Install Tini to pass signals correctly
+    # Install Tini to correctlry pass signals
```
dcroote, line 17:
```suggestion
    # Install Tini to correctly pass signals
```

dcroote (commented):
I did a few quick tests of the dockerhub -dev images for this commit (`d0a04df8fdbbdca4b14548256422c0f7d075cf03`) and they appeared to be functional 👍

.github/workflows/build-test.yml
amarthadan: The matrix strategy was applied only to the builds for the final containers. Those builds are quite quick anyway so parallelizing them won't bring that much actually. The same goes for the new setup. And since there's no intermediate container (`airnode-artifacts`), I can't easily run these builds in a separate job. I'll figure out how to optimize it if necessary but I think it's fine for now.

docker/README.md
```diff
@@ -1,51 +1,66 @@
-# api3/airnode-artifacts
+# api3/airnode-packaging
 
-This is a Docker container for building all the packages in the monorepo and providing their build artifacts. Other
-package specific containers can use this one to obtain files necessary for their run without building or installing any
-additional dependencies.
+This is a Docker container for building all the packages in the monorepo, publishing the NPM packages (TODO) and
```
Siegrift, line 3:
```suggestion
This is a Docker container for building all the packages in the monorepo, publishing the NPM packages (TODO) and
```

These TODOs have a tendency to be forgotten, so I'd just avoid them and mention NPM publishing when implemented. Doesn't matter much to me though.

docker/README.md
```diff
@@ -1,51 +1,66 @@
...
+## Usage
 
-You can build the container image from your current workspace by providing a build type `local` (this is the default)
+In order for the container to work correctly you need to pass the Docker daemon socket from your system, so the
+containers can be used "within" the containers:
 
 ```bash
-docker build --build-arg build=local -f docker/Dockerfile -t api3/airnode-artifacts:latest .
-# or with no build argument
+... -v /var/run/docker.sock:/var/run/docker.sock ...
```
Siegrift, line 33: This is cross platform right?

scripts/packaging/prepare-containers.ts
```diff
@@ -0,0 +1,98 @@
+///////////////////////////////////////// WARNING /////////////////////////////////////////
+//                                                                                       //
+//                 Do NOT run these scripts outside a Docker container!                  //
```
Siegrift, line 3: Can we check this in this from with nodejs? I did a quick google and found https://www.npmjs.com/package/is-docker.

docker/README.md
amarthadan: Feel free to test it on Mac, it should work there. Not sure about Windows but I don't think being able to build the images there is as important as being able to run them there.

Siegrift: Yeah, will try to test this today.

scripts/packaging/prepare-containers.ts
amarthadan: I don't really think it's necessary. Nobody's going to run those scripts unless they're targeted in `package.json`.

Siegrift: But what is the disadvantage of doing it as I proposed?

amarthadan: I'm a bit worried about false positives but I can try it.

docker/Dockerfile
```diff
@@ -1,67 +1,18 @@
...
-
-# Source preparation - local
-FROM preparation as sourcelocal
-
-COPY . ${buildDir}
+ENV buildDir="/build"
 
-# Source preparation - git
-FROM preparation as sourcegit
+RUN apk add --update --no-cache git rsync docker && \
```
Siegrift, line 5: The comment:
```
# Download both solidity compilers as per: https://github.com/nomiclabs/hardhat/issues/1280#issuecomment-949822371
```

got lost.

docker/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+GIT_REF=${GIT_REF:-master}
+
+# Use Git respository as a source when there's no `/airnode` directory mounted
+if [ -d /airnode ]; then
+  # Efficient way to copy the whole directory with a lot of small files
+  (cd /airnode; tar cf - .) | (cd /build; tar xpf -)
```
Siegrift, line 8: 💭

docker/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
```
Siegrift, line 2: Can we add `set -e` for immediate exit when there is an error?

scripts/packaging/prepare-containers.ts
```diff
@@ -0,0 +1,98 @@
+///////////////////////////////////////// WARNING /////////////////////////////////////////
```
Siegrift, line 1: I'd move these scripts to the `docker` directory

scripts/packaging/prepare-containers.ts
amarthadan: I had to take the code from the library because it uses ESM modules which don't work with our setup. But it looks like it's working. At least for my setup :sweat_smile:

Siegrift: > I had to take the code from the library because it uses ESM modules which don't work with our setup

You can just use major version 2. See: https://github.com/sindresorhus/is-docker/releases/tag/v3.0.0

amarthadan: I'm really not going to spend more time on this, it's a useless check anyway :smile:

docker/README.md
Siegrift: Can we mention (or create an issue) that this is unusable on macOS due to docker copy being unusably slow?

packages/airnode-node/docker/README.md
```diff
@@ -1,4 +1,4 @@
-# Airnode client image
+# Airnode Docker image
```
Siegrift, line 1: Btw. why capital D?

Siegrift (changes requested):
There is https://github.com/api3dao/airnode/blob/docker-image-from-released-packages/packages/airnode-examples/src/scripts/rebuild-deployer-container.ts and similar ones for artifacts and client. These are no longer used in the examples and can be removed from package.json (and the test one).

packages/airnode-node/docker/README.md
amarthadan: Docker is a name :man_shrugging:

docker/README.md
amarthadan: I've changed the way the files are copied (as we've discussed on Slack). Can you test it again?

Siegrift: Sorry that it took so long. I can confirm this works, but I had to [add back some missing dependencies](https://github.com/api3dao/airnode/pull/1382/files#diff-f34da55ca08f1a30591d8b0b3e885bcc678537b2a9a4aadea4f190806b374ddcL15).

I have also a bit cleaner suggestion for copying the files. Essentially you can do a git clone from a local repo. See: https://github.com/api3dao/airnode/commit/47ddc74ad1be3a7248732a685d9e847fee14d991

Siegrift (approved):
Tested on macOS, left a minor suggestion, but 👍 LGTM.

One thing that threw me off a bit is that the images are tagged as `:local` by default (previously we used `:latest`) but I've checked the airnode repo and docs and there are no references to `:latest`.

docker/entrypoint.sh
```diff
@@ -0,0 +1,20 @@
...
+set -e
+
+GIT_REF=${GIT_REF:-master}
+
+# Use Git respository as a source when there's no `/airnode` directory mounted
+if [ -d /airnode ]; then
+  # Efficient way to copy the whole directory with a lot of small files
+  # Keeping any local code changes but excluding platform-specific files (dependencies, builds, etc.)
+  rsync -a --include packages/airnode-protocol/deployments --exclude node_modules --exclude build --exclude dist --exclude coverage --exclude artifacts --exclude cache --exclude .webpack --exclude deployments /airnode/ /build
```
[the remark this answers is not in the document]
amarthadan: I presume this will not include not-commited changes, will it? I wouldn't want to lose that as that's the point of a local build - you do some changes and you want to test whether it works in the container, not necessarily committing it first. I'll check and try to figure out something else if that's the case.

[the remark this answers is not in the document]
Siegrift: True. It will not include uncommitted changes. 

Btw. this command is fast for me as is. I just fear that we will have a new ignored folder and forget to add it here. But feel free to merge this as is.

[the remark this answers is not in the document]
amarthadan: I want to avoid that as well. I'll try to reuse the `.gitignore`.

amarthadan:
@Siegrift I've added the dependencies and changed the file copying to follow the `.gitignore`(s). Can you try it one more time? :pray:

PR #1383, opened 2022-08-17
Fix naming for some of the GCP regions

### LINKED ISSUE #1380: Fix behavior of some GCP regions
Based on @vponline's [report](https://api3workspace.slack.com/archives/C02AYRX8D89/p1660299786586649?thread_ts=1660225845.222229&cid=C02AYRX8D89) it looks like that [two GCP App Engine locations differ from what is used for other resources](https://cloud.google.com/appengine/docs/locations).
Namely `europe-west1` and `us-central1` are recognized only under `europe-west` and `us-central` in the App engine. We should add this "translation" to the deployer/Terraform recipes.

amarthadan:
Close #1380

I tested both the regions and it works now. You provide the name from the documentation (with the `1` at the end) and we "translate" those two region names for the App Engine resource.

### DIFF
--- a/packages/airnode-deployer/terraform/airnode/gcp/modules/function/variables.tf
+++ b/packages/airnode-deployer/terraform/airnode/gcp/modules/function/variables.tf
@@ -4,6 +4,10 @@ locals {
   tmp_configuration_dir = "${local.tmp_input_dir}/config-data"
   tmp_handlers_dir      = "${local.tmp_input_dir}/handlers"
   tmp_output_dir        = "${local.tmp_dir}/output"
+  # Two locations, which are called europe-west and us-central in App Engine commands and in the Google Cloud console,
+  # are called europe-west1 and us-central1, respectively, elsewhere in Google documentation.
+  # https://cloud.google.com/appengine/docs/locations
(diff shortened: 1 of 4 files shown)


packages/airnode-deployer/terraform/airnode/gcp/modules/function/variables.tf
```diff
@@ -4,6 +4,10 @@ locals {
   tmp_configuration_dir = "${local.tmp_input_dir}/config-data"
   tmp_handlers_dir      = "${local.tmp_input_dir}/handlers"
   tmp_output_dir        = "${local.tmp_dir}/output"
+  # Two locations, which are called europe-west and us-central in App Engine commands and in the Google Cloud console,
+  # are called europe-west1 and us-central1, respectively, elsewhere in Google documentation.
+  # https://cloud.google.com/appengine/docs/locations
+  app_engine_location_id = var.region == "europe-west1" ? "europe-west" : var.region == "us-central1" ? "us-central" : var.region
```
Siegrift, line 10: Nit: Can we add brackets on the inner ternary?

Siegrift (approved):
👍 LGTM (didn't test though)

PR #1391, opened 2022-08-22
Remove heartbeat id from receipt

### LINKED ISSUE #1296: Remove `heartbeatId` from `receipt.json`
When working on https://github.com/api3dao/api3-docs/issues/825 I've noticed that we store heartbeat ID in the `receipt.json` even though it's just a copy of the heartbeat ID from `config.json`. To add to the confusion, we send that ID with the heartbeat with the name `deployment_id`. I propose removing the `hearbeatId` field from the `receipt.json` (and with it the whole `api` section as it will become empty).

vponline:
Closes #1296


amarthadan (approved):
LGTM

Issue #217, opened 2021-02-20
Changing some node constants may have unexpected side effects


bbenligiray:
Keeping the node constants in a [separate file](https://github.com/api3dao/airnode/blob/2a214f0d188121e08c0a9a1362233a5cf03b0412/packages/node/src/constants.ts) implies that this is the single source of truth that can be updated to customize behavior.
However, there may be some unforeseen couplings that prevent that, for example:
https://github.com/api3dao/airnode/blob/2a214f0d188121e08c0a9a1362233a5cf03b0412/packages/node/src/constants.ts#L11
https://github.com/api3dao/airnode/blob/2a214f0d188121e08c0a9a1362233a5cf03b0412/packages/deployer/serverless.aws.yml#L34
Similarly, changing this 2->3 results in tests breaking because it is effectively hardcoded in them
https://github.com/api3dao/airnode/blob/2a214f0d188121e08c0a9a1362233a5cf03b0412/packages/node/src/constants.ts#L10

andreogle:
See discussion here: https://github.com/api3dao/airnode/pull/184#discussion_r567665416

One of the main problems with adjusting `OPERATION_RETRIES` is that many test mocks expect a specific number of calls. e.g. the first call fails, the second succeeds. This is done using `mockResolvedValueOnce` and `mockRejectedValueOnce`.

A potential solution for this (?):

```ts
for (var i = 0; i < OPERATION_RETRIES - 1; i++) {
  getTemplatesMock.mockRejectedValueOnce(new Error('Server says no'));
}
getTemplatesMock.mockResolvedValueOnce(rawTemplates);
```ts

**Comment by MrAngus666** [comment]
Probably it is worthy to have a delay between retries, what do you think about something like?:
```
function sleep(ms) {
  return new Promise((resolve) => {
    setTimeout(resolve, ms);
  });
}   
for (var i = 0; i < OPERATION_RETRIES - 1; i++) {
  await sleep(i * 200);
  getTemplatesMock.mockRejectedValueOnce(new Error('Server says no'));
}
getTemplatesMock.mockResolvedValueOnce(rawTemplates);
```

MrAngus666:
After checking #184 , I undertand that we need the retry because of the mocks behaviour, not because there is a failing API/Race Condition where al delay would give more changes of a success response.
Forget my previous comment!

bbenligiray:
An alternative solution would be to hard-code these non-variables

bbenligiray:
[...]

PR #1394, opened 2022-08-23
Avoid constants with side effects

Siegrift:
Relates to https://github.com/api3dao/airnode/issues/217

1. I went through all of the named constants in the codebase. The problem is that we have defined generic named constants (e.g. RANDOM_BACKOFF_MIN_MS, DEFAULT_RETRY_TIMEOUT_MS) but these could be used in different places and not necessarily bound to be the same. I renamed these generic names to a better name (more specific one).
2. I also removed some outdated and unused constants.
3. Currently, we define the API call timeout to 30s and processing timeouts to 5s each. We define the worker timeout also to 30s. However, API call + processing can last up to 40s, so it felt tempting to change these timeouts to gracefully account for this worst case. I was thinking about reducing the API call timeout to reserve time for processing is too limiting. It's better to let it timeout by the worker timeout.
4. It's also possible that the TF timeout will kill the running lambda, because it feels hard to guarantee what is the total execution of a handler, because blockchain calls are retried once when they fail. I think this is OK though, but I am not sure if our pending logs feature will handle this correctly (e.g. TF kills the lambda before logger.logPending is called and we end up with no logs).


packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -63,7 +56,7 @@ async function initializeProvider(payload: InitializeProviderPayload) {
   const stateWithConfig = { ...state, config: parsedConfig };
 
   const goInitializedState = await go(() => handlers.initializeProvider(stateWithConfig), {
-    delay: { type: 'static', delayMs: DEFAULT_RETRY_DELAY_MS },
+    delay: { type: 'static', delayMs: 50 },
```
amarthadan, line 59: Why haven't you used a constant for these?

packages/airnode-node/src/constants.test.ts
```diff
@@ -1,25 +1,12 @@
-import * as constants from './constants';
+import {
+  WORKER_CALL_API_TIMEOUT,
+  WORKER_PROCESS_TRANSACTIONS_TIMEOUT,
+  WORKER_PROVIDER_INITIALIZATION_TIMEOUT,
+} from './constants';
 
-describe('constants', () => {
```
amarthadan, line 3: Why did you remove all these tests?

packages/airnode-node/src/constants.test.ts
```diff
@@ -1,25 +1,12 @@
...
-    expect(constants.WORKER_CALL_API_TIMEOUT).toEqual(30_000);
-    expect(constants.WORKER_PROVIDER_INITIALIZATION_TIMEOUT).toEqual(20_000);
-    expect(constants.WORKER_PROCESS_TRANSACTIONS_TIMEOUT).toEqual(10_000);
-    expect(constants.MAXIMUM_SPONSOR_WALLET_REQUESTS).toBe(5);
-    expect(constants.MAXIMUM_ONCHAIN_ERROR_LENGTH).toEqual(100);
-    expect(constants.PROCESSING_TIMEOUT).toEqual(10_000);
-  });
+it('worker timeouts are smaller then terraform lambda timeout', () => {
+  // The 32 seconds timeout is currently hardcoded in TF
+  expect(WORKER_PROVIDER_INITIALIZATION_TIMEOUT).toBeLessThan(32_000);
```
amarthadan, line 9: This will be true even for `31_999` but that won't leave the desired 2 seconds for a graceful termination. What about something like `constant <= TF value - grace period (2 sec)`?

amarthadan (approved):
The changes make sense to me, just minor notes/questions. Feel free to merge when answered.

amarthadan:
I guess you shouldn't merge this to `master` as it's under the `0.10` milestone.

packages/airnode-deployer/src/handlers/aws/index.ts
Siegrift: It feels these could be different values for each handler, but they should be the same for AWS and GCP. I'll make constants for these.

packages/airnode-node/src/constants.test.ts
Siegrift: I think that should be done when we implement the 1. suggestion in https://github.com/api3dao/airnode/issues/217#issuecomment-1224141791.

Or do you want to add that grace period now? (I don't have a strong preference)

amarthadan: No, that sounds fine.

packages/airnode-deployer/src/handlers/aws/index.ts
Siegrift: Mhh, I just realized there are no retries specified for these functions, so the delay doesn't make sense... It feels to me we should not even retry those. WDYT?

amarthadan: Handlers? I agree.

Siegrift:
> I guess you shouldn't merge this to master as it's under the 0.10 milestone.

I've created `temp-v0.10.0` branch to merge this into.

aquarat:
Is the primary reason for holding this back risk mitigation?
(and this is a welcome change)

Siegrift:
> Is the primary reason for holding this back risk mitigation (too many changes at once)?

I think the primary reason should be to honor the things we decide and plan. It is a bit unfortunate that plan for 0.9 changed, but since we decided this should land in 0.10 it should not be released sooner.

PR #1395, opened 2022-08-24
Default to deployed AirnodeRrpV0 in airnode-examples

### LINKED ISSUE #1278: Default to deployed AirnodeRrpV0 contract addresses in `airnode-examples`
Rather than have users deploy the `AirnodeRrpV0` contract as part of the `airnode-examples` flow, the default should be be to use the address of the `AirnodeRrpV0` contract that API3 has already deployed on that chain. This would save the user a step and also better mimic the experience of Airnode deployment in which we do not ask providers to deploy this contract themselves. These addresses are available for all of the chains `airnode-examples` supports both in [the docs](https://docs.api3.org/airnode/latest/reference/airnode-addresses.html) and in [airnode-protocol deployments](https://github.com/api3dao/airnode/blob/4519831513db9e79caadafcd307cea23ee87d113/packages/airnode-protocol/deployments/references.json#L92-L121). The exception is for the local flow, which would still need the deployment step, hence deploying `AirnodeRrpV0` should still be an option.

dcroote:
Closes #1278.

### DIFF
--- a/packages/airnode-examples/README.md
+++ b/packages/airnode-examples/README.md
@@ -123,7 +123,7 @@ The [RRP contract](https://docs.api3.org/airnode/latest/concepts/#airnoderrp-sol
 [requester](https://docs.api3.org/airnode/latest/concepts/requester.html) triggers a request for Airnode. This contract
 is common for all Airnodes and requesters on a chain.
 
-If you are using the local blockchain, deploy the contact using the command below. Otherwise, the contract deployed by
+If you are using a local blockchain, deploy the contact using the command below. Otherwise, the contract deployed by
 API3 on the chosen chain will be used automatically.
 
 ```sh
--- a/packages/airnode-examples/README.md
+++ b/packages/airnode-examples/README.md
@@ -117,14 +117,14 @@ yarn print-account-details
 This script will show you the address of the account derived from the specified mnemonic. This account will be used to
 deploy the contracts and make transactions, so make sure it is funded. The recommended amount is at least 0.3 ETH.
 
-#### 4. Deploy the RRP contract
+#### 4. (Only if using local blockchain) Deploy the RRP contract
 
 The [RRP contract](https://docs.api3.org/airnode/latest/concepts/#airnoderrp-sol) is a contract through which the
 [requester](https://docs.api3.org/airnode/latest/concepts/requester.html) triggers a request for Airnode. This contract
 is common for all Airnodes and requesters on a chain.
 
-The API3 team will deploy this contract for the most popular chains at some point, but for now you have to deploy it
-yourself using:
+If you are using the local blockchain, deploy the contact using the command below. Otherwise, the contract deployed by
+API3 on the chosen chain will be used automatically.
 
 ```sh
 yarn deploy-rrp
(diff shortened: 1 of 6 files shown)


aTeoke:
Please do not push this to master, this will be part of 0.9. Please create 0.9 branch

packages/airnode-examples/README.md
```diff
@@ -117,14 +117,14 @@ yarn print-account-details
...
-#### 4. Deploy the RRP contract
+#### 4. (Only if using local blockchain) Deploy the RRP contract
 
 The [RRP contract](https://docs.api3.org/airnode/latest/concepts/#airnoderrp-sol) is a contract through which the
 [requester](https://docs.api3.org/airnode/latest/concepts/requester.html) triggers a request for Airnode. This contract
 is common for all Airnodes and requesters on a chain.
 
-The API3 team will deploy this contract for the most popular chains at some point, but for now you have to deploy it
-yourself using:
+If you are using the local blockchain, deploy the contact using the command below. Otherwise, the contract deployed by
```
Siegrift, line 126:
```suggestion
If you are using a local blockchain, deploy the contact using the command below. Otherwise, the contract deployed by
```

packages/airnode-examples/src/utils.test.ts
```diff
@@ -19,3 +19,15 @@ describe('readPackageVersion', () => {
     expect(res).toBe('0.5.0');
   });
 });
+
+describe('getExistingAirnodeRrpV0', () => {
+  test.each(supportedNetworks)('confirms all supported networks have AirnodeRrpV0 addresses', (name) => {
```
Siegrift, line 24: 👍 Nice.

packages/airnode-examples/src/utils.ts
```diff
@@ -3,6 +3,7 @@ import { join } from 'path';
 import { parse as parseEnvFile } from 'dotenv';
 import prompts, { PromptObject } from 'prompts';
 import isWsl from 'is-wsl';
+import { default as references } from '@api3/airnode-protocol/deployments/references.json';
```
Siegrift, line 6:
```suggestion
import references from '@api3/airnode-protocol/deployments/references.json';
```

Siegrift (approved):
👍 LGTM

packages/airnode-examples/src/utils.ts
dcroote: 😅  Updated, thanks.

packages/airnode-examples/README.md
dcroote: Updated, thanks.

dcroote:
Current [Docker build failure](https://github.com/api3dao/airnode/runs/8127741866?check_suite_focus=true#step:4:94) caused by https://github.com/npm/cli/issues/5453

aquarat:
I was curious and tested (as the linked issue suggests) pinning the installed npm version to 8.18.0 in the `Dockerfile` and it works. eg:

In docker/Dockerfile:
```
    yarn global add npm@8.18.0 && \
```
so that could be a solution until it's resolved upstream.

PR #1398, opened 2022-08-26
Heartbeat signature

### LINKED ISSUE #1337: Heartbeat payload signing
Based on research done in https://github.com/api3dao/airnode/issues/1336

From https://github.com/api3dao/airnode/issues/1299:
> The heartbeat payload will be cryptographically signed, allowing the recipient of the heartbeat payload to recover the public key and verify the public key+payload contents. The Airnode's "root" key should be used to sign the payload.

One thing to make sure is that the message that is being signed here can’t be manipulated to match a transaction hash or a Beacon update. So make sure to [sign a message](https://docs.ethers.io/v5/api/signer/#Signer-signMessage) that includes some salt (e.g. `"Airnode heartbeat authentication”`) so that it can’t be used for anything else.

vponline:
Closes #1337

### DIFF
--- a/.changeset/two-berries-beam.md
+++ b/.changeset/two-berries-beam.md
@@ -1,0 +1,7 @@
+---
+'@api3/airnode-deployer': minor
+'@api3/airnode-node': minor
+'@api3/airnode-validator': minor
+---
+
+Add heartbeat payload signing
--- a/packages/airnode-node/src/constants.test.ts
+++ b/packages/airnode-node/src/constants.test.ts
@@ -6,7 +6,7 @@ describe('constants', () => {
 
     // Ensure all constant values are checked
     expect.assertions(constCount + 1);
-    expect(Object.keys(constants).length).toEqual(14);
+    expect(Object.keys(constants).length).toEqual(13);
 
     expect(constants.API_CALL_TIMEOUT).toEqual(30_000);
     expect(constants.API_CALL_TOTAL_TIMEOUT).toEqual(30_000);
@@ -21,6 +21,5 @@ describe('constants', () => {
     expect(constants.MAXIMUM_SPONSOR_WALLET_REQUESTS).toBe(5);
     expect(constants.MAXIMUM_ONCHAIN_ERROR_LENGTH).toEqual(100);
     expect(constants.PROCESSING_TIMEOUT).toEqual(10_000);
-    expect(constants.HEARTBEAT_SALT).toEqual('Airnode heartbeat authentication');
   });
 });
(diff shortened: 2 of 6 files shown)


aquarat (commented):
I think this needs further discussion on Slack 😬 

This scheme was conceived assuming we weren't going to be using ChainAPI for data feed integrations (it wasn't considered intially)... but ChainAPI has proven that they're perfectly suited to handle this for us, therefore this functionality is probably no longer be required. This needs to be confirmed, especially with @andreogle . Sorry for the possible U-turn.

Separately, if we do proceed, I think there's a concern currently around the signature being replayable, because we don't have anything in the signature that changes per heartbeat event. If we embed the timestamp in the `heartbeatPayload`, the signature will change each time and it'll be tied to the wall-time/an absolute reference that can only move forward, so these won't be replayable.

Apart from that it can't "hurt" putting the timestamp in the payload vs outside of it.

But don't spend any more time on this until it's been explicitly confirmed we're proceeding with this.

aquarat (approved):
Looks great, neat as always 😊 
As a note to myself and others, this task needs https://github.com/api3dao/airnode/issues/1411 to be useful.

martinkolenic:
Hi @vponline, I've tested the latest commit in the `temp-v0.10.0` branch using the `coingecko-signed-data` Airnode example, enabling heartbeat and then listening to it using Beeceptor. This is the request my endpoint received.

image

Since having made this screenshot a couple minutes ago, I received several other request, meaning that the heartbeat is being sent in regular intervals as expected. If I understand correctly, we're interested in the signature here, which changes with every single heartbeat I receive. I am not sure how to check it for correctness but I'll be happy to learn how to do it if you have a moment to spare. If the signature is correct, we can move it to Done then.

vponline:
@martinkolenic  That's cool and thanks for testing! The signature is indeed supposed to be different for each request since it uses a timestamp in the signature and that (of course) changes for each request.

Here is how to verify the signature:

1. Go to https://playground.ethers.org/
2. Replace the TIMESTAMP, PAYLOAD and SIGNATURE into the following:
(note that the PAYLOAD will not include the timestamp and signature fields)
```ts
ethers.utils.verifyMessage(
  ethers.utils.arrayify(
    ethers.utils.solidityKeccak256(
      ['uint256', 'string'], [<TIMESTAMP>',JSON.stringify(<PAYLOAD>)])),
    <SIGNATURE>
)
```
4. Copy and paste the function into the Ethers playground and it should output the airnode address which signed it

martinkolenic:
Hi, Vekil, thanks very much for the detailed instructions. I checked the data and it was valid (a couple missing brackets there but I added those and the script worked). 

The GCP test went well too, the results of the request were valid as well. I'm moving this to Done now.

PR #1413, opened 2022-08-31
Exports functions under /api directory on airnode-node package

### LINKED ISSUE #1412: Export functions under /api directory on airnode-node package
This is required by Airkeeper in order to be able to re-use the callApi() functionality and pass only required config values as parameters.

acenolaza:
Closes #1412

### DIFF
--- a/.changeset/real-rice-train.md
+++ b/.changeset/real-rice-train.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-node': patch
+---
+
+Export functions in /api directory
--- a/packages/airnode-node/src/api/index.ts
+++ b/packages/airnode-node/src/api/index.ts
@@ -20,7 +20,7 @@ import {
   HttpSignedApiCallPayload,
 } from '../types';
 
-function buildOptions(payload: ApiCallPayload): adapter.BuildRequestOptions {
+export function buildOptions(payload: ApiCallPayload): adapter.BuildRequestOptions {
   const { type, config, aggregatedApiCall } = payload;
   const { endpointName, parameters, oisTitle } = aggregatedApiCall;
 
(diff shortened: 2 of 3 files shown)


amarthadan (approved):
As far as I can tell `callApi` is the only function that is exported. Is that enough?

acenolaza:
> As far as I can tell `callApi` is the only function that is exported. Is that enough?

That's all I'll need to call from Airkeeper but are you thinking that we should export all of the other functions just in case? I see [verifyTemplateId](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/api/index.ts#L127) is also being exported

amarthadan:
> > As far as I can tell `callApi` is the only function that is exported. Is that enough?
> 
> That's all I'll need to call from Airkeeper but are you thinking that we should export all of the other functions just in case? I see [verifyTemplateId](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/api/index.ts#L127) is also being exported

I don't mind either way, just making sure we don't miss something we need. Up to you :slightly_smiling_face:

acenolaza:
> > > As far as I can tell `callApi` is the only function that is exported. Is that enough?
> > 
> > 
> > That's all I'll need to call from Airkeeper but are you thinking that we should export all of the other functions just in case? I see [verifyTemplateId](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/api/index.ts#L127) is also being exported
> 
> I don't mind either way, just making sure we don't miss something we need. Up to you slightly_smiling_face

Good call 👍🏻 I wouldn't want to have to publish a new package because of forgetting to export something 🤭 so I'll probably export a few more functions that might be needed. Also [these types](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/types.ts#L303) would probably be used but they are already being exported.

PR #1418, opened 2022-09-05
Move from API keys to route-based authentication for HTTP gateways

### LINKED ISSUE #1374: Move from API keys to route-based authentication for HTTP gateways
Based on the [Slack discussion](https://api3workspace.slack.com/archives/C02AYRX8D89/p1659538932266949).

We want to get rid of the API keys for HTTP gateways and instead use a randomly generated route for the gateways' endpoints.
What needs to be done (roughly):
* Remove `apiKey` field from the config
* Remove API key initialization from Terraform recipes
* Turn off API key authentication in the OpenAPI templates
* Generate the random path section of the endpoint and update OpenAPI templates to use it
* Update the local HTTP gateways implementation accordingly

vponline:
Closes #1374

I have tested that this works with AWS, GCP and local gateways.


packages/airnode-node/src/workers/local-gateways/server.ts
```diff
@@ -51,7 +40,8 @@ export function startGatewayServer(config: Config, enabledGateways: GatewayName[
   const port = cloudProviderSettings.gatewayServerPort ?? DEFAULT_PORT;
 
   if (enabledGateways.includes('httpSignedDataGateway')) {
-    const httpSignedDataGatewayPath = `${HTTP_SIGNED_DATA_BASE_PATH}/:endpointId`;
+    const pathKey = randomUUID();
+    const httpSignedDataGatewayPath = `${HTTP_SIGNED_DATA_BASE_PATH}/${pathKey}/:endpointId`;
```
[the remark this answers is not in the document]
amarthadan: I agree, let's set it to constant.

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -186,14 +187,14 @@ async function terraformAirnodeManage(
   }
 
   if (httpGateway?.enabled) {
-    commonArguments.push(['var', 'http_api_key', httpGateway.apiKey!]);
+    commonArguments.push(['var', 'http_path_key', randomUUID()]);
```
amarthadan, line 190: Hmm. The problem with this approach is that the path (the UUID) will change every time the user will redeploy which is quite annoying. I think it would be better to use [`random_uuid`](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/uuid) Terraform resource. That will allow us to create a random path but it will be stored in the Terraform state and therefore won't change between redeployments. Also, we won't need the path variables to pass to the Terraform comamnds.

amarthadan (changes requested):

packages/airnode-deployer/src/infrastructure/index.ts
vponline: Updated to use this method and tested again that it works.

packages/airnode-node/src/workers/local-gateways/server.ts
```diff
@@ -18,6 +17,7 @@ const httpSignedDataBodySchema = z.object({
   encodedParameters: z.string(),
 });
 const DEFAULT_PORT = 3000;
+const DEFAULT_PATH_KEY = '05701bc4-4eb4-4f60-b4eb-075c80ea98c6';
```
amarthadan, line 20: Oh, you've chosen a true random UUID. I was thinking more like a "fake" UUID so nobody will wonder whether it should be configured somewhere. Something like `01234567-abcd-abcd-abcd-012345678abc` or `ffffffff-ffff-ffff-ffff-ffffffffffff`, you know what I mean.

amarthadan (approved):
Other than the default UUID LGTM :+1:

packages/airnode-node/src/workers/local-gateways/server.ts
vponline: Ah yeah I just switched the old `API_KEY` into the path :sweat_smile:. But you make a great point, I'll update :smile:

martinkolenic:
Hi @vponline, I noticed that API keys are still required for heartbeat; will this be changing as well? It seems to me from an outsider's perspective that it would be more consistent; that said, there might be more compelling reasons to keep it. I'm still testing this change at the moment; I'll let you know about the results and testcases as soon as I'm done.

vponline:
> Hi @vponline, I noticed that API keys are still required for heartbeat; will this be changing as well? It seems to me from an outsider's perspective that it would be more consistent; that said, there might be more compelling reasons to keep it. I'm still testing this change at the moment; I'll let you know about the results and testcases as soon as I'm done.

Good question @martinkolenic , the heartbeat will remain as is with api keys since the idea is that API providers can setup their own heartbeat-receiving endpoints using the more traditional 'api key in header' authorization approach) :smile:

Issue #1346, opened 2022-07-22
Update hardhat to ^2.10.0


Siegrift:
Hardhat version 2.10 consists of [considerable changes](https://github.com/NomicFoundation/hardhat/releases/tag/hardhat%402.10.0) and got slower so our e2e tests started failing (because it takes a long time until the node initializes).

We've excluded update of this dependency in https://github.com/api3dao/airnode/pull/1325 but we should look at this again at some point.

andreogle:
RIP semver 😭

I wonder how feasible it would be to try [Foundry](https://github.com/foundry-rs/foundry) instead of Hardhat. Foundry obviously isn't written in JS so integrating it with a JS/TS repo like Airnode may or may not be tricky, but it's _extremely_ fast (instant in most cases)

dcroote:
> I wonder how feasible it would be to try [Foundry](https://github.com/foundry-rs/foundry) instead of Hardhat. Foundry obviously isn't written in JS so integrating it with a JS/TS repo like Airnode may or may not be tricky, but it's _extremely_ fast (instant in most cases)

New title: Update hardhat to Rust 😂

dcroote:
The issue is the number of accounts we were initializing (1000!) - I reduced it to 100 in https://github.com/api3dao/airnode/pull/1647 and tests pass.

PR #1419, opened 2022-09-05
Updates Hardhat to v2.11.0

acenolaza:
Closes [#1346](https://github.com/api3dao/airnode/issues/1346)
 
Also removed ethereum-waffle in favor of the new hardhat matchers.

Note: I didn't add the new hardhat-toolbox plugin anywhere. https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-toolbox

### DIFF
--- a/.changeset/brave-cars-hope.md
+++ b/.changeset/brave-cars-hope.md
@@ -1,0 +1,9 @@
+---
+'@api3/airnode-adapter': patch
+'@api3/airnode-examples': patch
+'@api3/airnode-operation': patch
+'@api3/airnode-protocol': patch
+'@api3/airnode-utilities': patch
+---
+
+Updates Hardhat package to v2.11.0
--- a/packages/airnode-adapter/e2e/index.ts
+++ b/packages/airnode-adapter/e2e/index.ts
@@ -24,7 +24,7 @@ it('shows the need for assertArrayEquals', () => {
 
   const goBigNumber = goSync(() => expect(ethers.BigNumber.from([123])).to.equal(['123']));
   assertGoError(goBigNumber);
-  expect(goBigNumber.error.message).to.contain('invalid BigNumber value');
+  expect(goBigNumber.error.message).to.contain('The input value cannot be normalized to a BigInt');
 
   assertArrayEquals([[ethers.BigNumber.from(123)], [ethers.BigNumber.from(456)]], [['123'], [456]]);
 });
--- a/packages/airnode-adapter/e2e/tsconfig.json
+++ b/packages/airnode-adapter/e2e/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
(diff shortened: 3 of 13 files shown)


Siegrift (approved):
👍 LGTM

acenolaza:
> +1 LGTM

Thanks! I just need to update the Dockerfile because it is manually copying the solidity compiler and it seems that it is not compatible with latest version of Hardhat

aquarat:
For documentation: 
There was a [discussion on Slack about this](https://api3workspace.slack.com/archives/C040G0W7RH9/p1662497935694689)

The thread above explains things but in short we pre-populate the solc cache for HardHat to get around a corrupt downloads issue. Separately, in the past, due to a missing library in Alpine, HH would always download the amd64 binary, it would always fail to execute and then HH would fall back to the solc JS/WASM compiler, which would work across architectures. It seems HH now fails outright when the first executable fails to run instead of falling back to solc JS.

acenolaza:
[This](https://app.slack.com/client/T0104044FT5/C02AYRX8D89/thread/C02AYRX8D89-1637914176.266000) slack thread also adds more context.

acenolaza:
So I tried starting from scratch with a new tiny hardhat project that uses typescript and manages a single contract. Then dockerize it by using the same Alpine docker base image as Airnode. This worked just fine and hardhat was able to fetch the solc compiler just fine and compile the contract.

We've been saying that hardhat has issues to run the solc compiler in Alpine due to missing dependencies but wouldn't this test prove that theory wrong? Could it be something else causing the issue? Maybe the user docker uses to run the compilation or the location of the files, etc

I'm uploading the test project just in case you want to try by yourself.
To build the image: `docker build . -t hhdocker`
[hardhat-docker-test.zip](https://github.com/api3dao/airnode/files/9527820/hardhat-docker-test.zip)

To run the image (just starts a local eth node): `docker run -it --name myhd hhdocker`
To remove the image: docker rm -f myhd

aquarat:
> I'm uploading the test project just in case you want to try by yourself. To build the image: `docker build . -t hhdocker` To run the image (just starts a local eth node): `docker run -it --name myhd hhdocker` To remove the image: docker rm -f myhd [hardhat-docker-test.zip](https://github.com/api3dao/airnode/files/9527820/hardhat-docker-test.zip)

Interesting. From memory the *original* problem was that HH would inconsistently download the compiler(s). It affected some docker users, particularly Burak, but I saw it recently on my Mac (arm64) while investigating this problem (I removed our caching steps) - I didn't see the same issue on Linux x86. The download process was flakey.

aquarat:
I've run the airnode docker build process on an arm64 Mac without the pre-caching steps multiple times and it inconsistently builds; sometimes it downloads the JS compiler, sometimes it doesn't. So the problem is the amd64 compiler doesn't always work and when it doesn't work the fallback compiler isn't always downloaded correctly.

I haven't seen your zip archive fail to build on either platform so far (across 10 attempts).

dcroote:
I think https://github.com/api3dao/airnode/pull/1743 now allows us to close this, but @acenolaza I noticed you made additional changes beyond a version bump so will leave it to you

Issue #1405, opened 2022-08-29
Allow users to fetch deployment files from the cloud provider


amarthadan:
```bash
airnode-deployer fetch-files awsd1f89c33 # Stores a file awsd1f89c33-1662558010204.zip with the latest config.json and secrets.env to ./config directory
airnode-deployer fetch-files awsd1f89c33 --version 586f79c8 # Stores a file awsd1f89c33-1662557983568.zip (specific version) to ./config
airnode-deployer fetch-files awsd1f89c33 --output-dir /some/path/somewhere  # Stores a file awsd1f89c33-1662558010204.zip to /some/path/somewhere directory
airnode-deployer fetch-files awsd1f89c33 --output-dir /some/path/somewhere --version 586f79c8 # Combination of options
```
This command will download file(s) of the specified deployment based on the deployment ID from the `list` command (https://github.com/api3dao/airnode/issues/1403).
Users will be able to download both `config.json` and `secrets.env` in a zip archive.
They can provide a `--version` option to specify the deployment version from which the file(s) should be downloaded based on the version ID from the `info` command (https://github.com/api3dao/airnode/issues/1404). Without the option, the latest deployment version will be used.
They can also use the option `--output-dir` to specify where the files should be stored. Without the option, `./config` directory is used.

amarthadan:
Should be refined based on the [Slack discussion](https://api3workspace.slack.com/archives/C02AYRX8D89/p1663586385441409?thread_ts=1663228424.184189&cid=C02AYRX8D89).

amarthadan:
@bbenligiray @aquarat I've updated the description, let me know whether it makes sense this way.

bbenligiray:
- `--files` weirded me out a bit, the interface didn't feel elegant. I guess the reasoning is that you may want to not expose sensitive files, but I'm not sure if that's an actual requirement. I would wait for the users to ask for this.
- Could have "fetch all versions" and "fetch everything" commands. I see myself using "fetch everything" for convenience. This can also be a wishlist item.
- Is `--output-dir` an absolute path? I'd suggest [`--directory-prefix`](https://www.gnu.org/software/wget/manual/html_node/Directory-Options.html) instead.

amarthadan:
> --files weirded me out a bit, the interface didn't feel elegant. I guess the reasoning is that you may want to not expose sensitive files, but I'm not sure if that's an actual requirement. I would wait for the users to ask for this.

[...]

PR #1424, opened 2022-09-07
Store deployments in a common bucket

### LINKED ISSUE #1402: Rework deployments to use a common bucket
In order to avoid bucket name conflicts, we want to keep files necessary for the deployments under one common bucket (per AWS account/GCP account + project).

During the first deployment, we will create a bucket with the name `airnode-{unique ID}`. Under this bucket, we will create a "directory path" with a format `/{full Airnode address}/{stage}/{timestamp}`. This directory will be used to store all the necessary files for the deployment - `config.json`, `secrets.env` and Terraform state files.

If the user decides to deploy another Airnode instance with either a different Airnode address or a different stage, the corresponding path will be created for that deployment under the common bucket.

If the user decides to update their current deployment (e.g. changing the OIS part of the config) a new `{timestamp}` directory will be created under the same `/{full Airnode address}/{stage}/` path and the updated files will be stored there.

Deploying an Airnode that has the same Airnode address and stage name but a different region or Airnode version won't be possible. If the users want to change the region, they would first have to remove the old deployment and then make a new one. The same goes for changing the Airnode version. Airnode deployment will be possible to remove only with the airnode-deployer of the same version as it was deployed with.

When the deployment is removed and there are still other deployments with the same Airnode address, we remove just the specific stage directory (and its subdirectories). The same goes for multiple deployments with a different Airnode address. When removing the last deployment we also remove the common `airnode-{unique ID}` bucket.

To summarize:

Deployment steps:
1. Look for the common bucket - the name must be in a `airnode-AABBCCDDEEFF` format. If there is one, reuse it. If there is none, create one. If there is more than one, fail with an error.
2. Look for the corresponding `/{full Airnode address}/{stage}/` path. If there is none, create one.
3. If there already is a deployment with the same Airnode address and stage, check that the region and Airnode version match. Fail with an error if they don't.
4. Deploy the Airnode / update the old deployment.
5. Create a new `{timestamp}` directory and store `config.json`, `secrets.env` & Terraform state files.

Deployment removal steps:
1. Look for the common bucket - the name must be in a `airnode-AABBCCDDEEFF` format. If there is none or there is more than one, fail with an error.
2. Look for the `/{full Airnode address}/{stage}/` path. If there is no such path, fail with an error - there's no such deployment.
3. Check that the Airnode version match. Fail with an error if they don't.
4. Select the latest deployment (newest timestamp) and remove the Airnode deployment.
5. Remove the `{stage}` directory and its files.
6. If there's no other `{stage}` directory remove the {full Airnode address} directory
7. If there's no other {full Airnode address} directory remove the common bucket

amarthadan:
Close #1402

So sorry about the massive commit :disappointed: I was changing stuff all over the place and didn't really split it into some meaningful pieces.

I would suggest reviewing the code in `airnode-deployer/src/infrastructure` as a whole, not looking at the diff, as those files were pretty much rewritten.

I'm still missing tests (and the current ones are broken) but I wanted to open a draft PR for review while I finish them.


martinkolenic:
Hi @amarthadan, I've run multiple deployments with different stages both on GCP and AWS and it seems the changes are working fine. I was able to complete a request and the shared bucket now replaces the changes.

amarthadan:
> Hi @amarthadan, I've run multiple deployments with different stages both on GCP and AWS and it seems the changes are working fine. I was able to complete a request and the shared bucket now replaces the set of buckets we had previously.

Just to be clear, you tried multiple deployments at the same time, right? Have you also tried multiple deployments with a different Airnode address? That's important as well.
Also, you now shouldn't be able to redeploy if you change the region in the configuration file. Please, test that as well.

martinkolenic:
> > Hi @amarthadan, I've run multiple deployments with different stages both on GCP and AWS and it seems the changes are working fine. I was able to complete a request and the shared bucket now replaces the set of buckets we had previously.
> 
> Just to be clear, you tried multiple deployments at the same time, right? 

If running the deploy commands shortly one after another counts, then yes.

> Have you also tried multiple deployments with a different Airnode address? That's important as well. 
No, I haven't - will do.

> Also, you now shouldn't be able to redeploy if you change the region in the configuration file. Please, test that as well.

Will do.

amarthadan:
> If running the deploy commands shortly one after another counts, then yes.

No, I mean having actually 2 or more Airnodes deployed (with different stage names) at the same time, not being deployed at the same time. So they are running and have resources in the cloud at the same time. That's the main thing.

packages/airnode-deployer/src/infrastructure/aws.ts
```diff
@@ -1,74 +1,203 @@
...
+export const getAirnodeBucket = async (cloudProvider: AwsCloudProvider) => {
+  const s3 = initializeS3Service(cloudProvider);
+
+  logger.debug('Listing S3 buckets');
+  const goBuckets = await go(() => s3.listBuckets().promise());
+  if (!goBuckets.success) {
+    throw new Error(`Failed to list S3 buckets: ${goBuckets.error}`);
+  }
+
+  const airnodeBuckets = goBuckets.data.Buckets?.filter((bucket) => bucket.Name?.match(BUCKET_NAME_REGEX));
```
Siegrift, line 30: I'd add an 
```
if (!airnodeBuckets) throw new Error('...')
```

to avoid checking for undefined below.

packages/airnode-deployer/src/infrastructure/gcp.ts
```diff
@@ -1,22 +1,183 @@
...
+  translatePathsToDirectoryStructure,
+} from '../utils/infrastructure';
 import * as logger from '../utils/logger';
 
-export async function stateExists(bucketName: string, cloudProvider: GcpCloudProvider) {
-  logger.debug('Checking Terraform state existence in GCP');
-  const { projectId } = cloudProvider;
-  const storage = new Storage({ projectId });
-  const bucket = storage.bucket(bucketName);
+const initializeGCSService = (cloudProvider: GcpCloudProvider) => new Storage({ projectId: cloudProvider.projectId });
```
Siegrift, line 13: Our concention is to uppercase just the first letter even for all-uppercase words (e.g. HTTP is Http, GCP should be Gcp, so GCS should be Gcs).

Siegrift (changes requested):
Some minor comments, but I've only did a shallow review. Will try the feature properly later.

packages/airnode-deployer/src/utils/files.ts
```diff
@@ -32,11 +27,10 @@ export function writeReceiptFile(
       airnodeXpub: deriveAirnodeXpub(mnemonic),
     },
     deployment: {
-      airnodeAddressShort,
       cloudProvider: config.nodeSettings.cloudProvider as CloudProvider,
       stage: config.nodeSettings.stage,
       nodeVersion: config.nodeSettings.nodeVersion,
-      timestamp,
+      timestamp: new Date().toISOString(),
```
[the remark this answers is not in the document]
amarthadan: I was thinking about that and didn't know what approach would be better. These two can vary by up to 15 min (GCP so slooow). I think the timestamp in the receipt should be of the time the deployment ended. But I guess that's up to interpretation. It can be the time it started :man_shrugging: But that's odd, isn't it?

packages/airnode-deployer/src/utils/infrastructure.ts
```diff
@@ -34,3 +35,91 @@ export class MultiMessageError extends Error {
...
+
+  return directoryStructure;
+};
+
+export const getAddressDirectory = (directoryStructure: DirectoryStructure, airnodeAddress: string) => {
+  const addressDirectory = directoryStructure[airnodeAddress];
+  if (!addressDirectory) {
+    return null;
+  }
+  if (addressDirectory.type !== FileSystemType.Directory) {
```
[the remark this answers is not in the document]
amarthadan: They shouldn't but it doesn't hurt to check (whether somebody tempered with the bucket manually) and it also satisfies TS.

packages/airnode-deployer/src/utils/files.ts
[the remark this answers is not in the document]
Siegrift: Won't users identify a particular deployment by that timestamp?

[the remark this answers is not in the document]
amarthadan: You mean in the cloud? I don't think they will. First of all, they should not need to deal with the cloud resources manually. And second, the deployment is distinguished by Airnode address and stage. And you're currently deployed version is the one with the latest timestamp, no matter what's in your receipt. But I get what you're saying, I can sync those two timestamps, I don't have a strong preference.

packages/airnode-validator/src/config/config.ts
```diff
@@ -162,6 +166,21 @@ export const chainAuthorizersSchema = z.object({
   requesterEndpointAuthorizers: z.array(evmAddressSchema),
 });
 
+export const maxConcurrencySchema = z.number().int().positive();
+
+const validateMaxConcurrency: SuperRefinement<{ providers: Providers; maxConcurrency: MaxConcurrency }> = (
```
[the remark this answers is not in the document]
amarthadan: That's a different thing. This is a max concurrency per chain. The cumulation of this (across all the chains) is passed to the `run` handler. The `startCoordinator` handler has its own reservation of 1.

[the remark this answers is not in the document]
Siegrift: Fair, but it seems to me we don't check that at the moment.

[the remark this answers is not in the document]
amarthadan: But there's nothing to check. The reservation for the coordinator is hardcoded to 1. You would have to check how many reservations the user has left in the cloud.

packages/airnode-deployer/src/infrastructure/aws.ts
amarthadan: I'm pretty sure that's just bad typing on the AWS side. They type everything as optional even though it has to be in the response.

packages/airnode-deployer/src/utils/infrastructure.ts
```diff
@@ -34,3 +35,91 @@ export class MultiMessageError extends Error {
...
+};
+
+export type DirectoryStructure = Record<string, FileSystemItem>;
+
+export const translatePathsToDirectoryStructure = (paths: String[]) => {
+  const directoryStructure: DirectoryStructure = {};
+
+  paths.forEach((path) => {
+    // Keep '/' at the end (for identifying directories)
+    path.split(/\/(?!$)/).reduce((ds, name: string, i: number, a: string[]) => {
```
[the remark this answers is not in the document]
amarthadan: > Why /\/(?!$)/ regex

I added a better explanation why

> The reduce is pretty messy. I'd suggest just write a mutable for loop for this.

TBH, I don't think that would actually help the readability. It would have to be either as a recursion (which is basically what `reduce` is doing) or I would have to search for a correct spot in the directory structure for each path part which is also kinda counterintuitive.

I've added some more comments and changed the naming a bit. Let me know if it's better now.

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -260,91 +290,208 @@ function transformTerraformOutput(terraformOutput: string): DeployAirnodeOutput
...
-      ['var', 'airnode_address_short', airnodeAddressShort],
-      ['var', 'stage', stage],
-      ['input', 'false'],
-      'no-color',
-      'auto-approve',
-    ];
-    await execTerraform(execOptions, 'apply', prepareAirnodeManageArguments(cloudProvider, commonArguments));
+  const goDeploy = await go(async () => {
+    logger.debug('Fetching Airnode bucket');
+    let bucketName = await cloudProviderLib[type].getAirnodeBucket(cloudProvider as any);
```
aquarat, line 305: I'm just checking; a mutable with an if and set is preferred over a "IIFE"?
I don't mind relaxed mutability but we should probably make it a policy or something.

packages/airnode-validator/src/config/config.ts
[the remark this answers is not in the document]
Siegrift: Yeah, right.

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -260,91 +290,208 @@ function transformTerraformOutput(terraformOutput: string): DeployAirnodeOutput
...
+    if (!bucketName) {
+      logger.debug('No Airnode bucket found, creating');
+      bucketName = await cloudProviderLib[type].createAirnodeBucket(cloudProvider as any);
```
aquarat, line 308: > ``` as any```

I guess there's no better way of doing this 😅

packages/airnode-deployer/src/utils/infrastructure.ts
```diff
@@ -34,3 +35,101 @@ export class MultiMessageError extends Error {
...
+export type DirectoryStructure = Record<string, FileSystemItem>;
+
+export const translatePathsToDirectoryStructure = (paths: String[]) => {
+  const directoryStructure: DirectoryStructure = {};
+
+  paths.forEach((path) => {
+    // Split on each '/' except for the one on the end if there is such (for identifying directories).
+    // E.g. for path '0xA30CA71Ba54E83127214D3271aEA8F5D6bD4Dace/dev/1662730904/' it's ['0xA30CA71Ba54E83127214D3271aEA8F5D6bD4Dace', 'dev', '1662730904/']
+    //
+    // Reduce is used to traverse deeper and deeper through the directory structure, adding levels during the first pass (first path) and resuing them for the rest (paths with the common parts, path prefix).
```
aquarat, line 69:
```suggestion
    // Reduce is used to traverse deeper and deeper through the directory structure, adding levels during the first pass (first path) and reusing them for the rest (paths with the common parts, path prefix).
```

packages/airnode-deployer/src/utils/infrastructure.ts
```diff
@@ -34,3 +35,101 @@ export class MultiMessageError extends Error {
...
+    // The same `directoryStructure` variable is used as as starting point for each path so we end up with one fully populated directory structure.
```
aquarat, line 70:
```suggestion
    // The same `directoryStructure` variable is used as a starting point for each path so we end up with one fully populated directory structure.
```

packages/airnode-deployer/terraform/gcp/main.tf
```diff
@@ -104,22 +106,22 @@ module "httpReq" {
...
-  configuration_file = var.configuration_file
-  secrets_file       = var.secrets_file
-  region             = var.gcp_region
-  project            = var.gcp_project
-
+  name                  = "${local.name_prefix}-httpReq"
+  entry_point           = "httpReq"
+  source_dir            = var.handler_dir
+  memory_size           = 128
+  timeout               = 15
```
aquarat, line 113: I just noticed this, we give the handler 15 seconds for HTTP gateway requests but when run via the coordinator API calls can have 30 seconds? I can't remember why this (shouldn't they be the same)? #educate_me 😛

packages/airnode-deployer/test/fixtures/index.ts
```diff
@@ -0,0 +1,193 @@
+import { DirectoryStructure, FileSystemType } from '../../src/utils/infrastructure';
+
+export const mockBucketDirectoryStructureList = [
+  '0xd0624E6C2C8A1DaEdE9Fa7E9C409167ed5F256c6/',
+  '0xd0624E6C2C8A1DaEdE9Fa7E9C409167ed5F256c6/dev/',
+  '0xd0624E6C2C8A1DaEdE9Fa7E9C409167ed5F256c6/dev/1662557983/',
```
aquarat, line 6: Burak was benchmarking serverless for this and serverless stores the deployment timestamp as an ISO string. It makes it more readable if you're navigating the directory manually, not sure if we'd want to do that for this PR?

aquarat (approved):
I find this change exciting and it's a lot of work. A very interesting read 🔥.
I enjoyed the path traversal reduce and comments around it.
I don't see any major issues, just questions and perhaps enhancements that may save us pain in future.

packages/airnode-deployer/src/infrastructure/index.ts
amarthadan: I don't think there is. And I remember having the same discussion with @Siegrift the first time it came around.

packages/airnode-deployer/test/fixtures/index.ts
amarthadan: I don't really like the idea of an ISO time string in the directory name tbh. Users shouldn't access these files manually anyway. I'd suggest keeping it as it is for now. We'll add a way to fetch the files in a more user-friendly way in #1405 anyway.

packages/airnode-deployer/terraform/gcp/main.tf
amarthadan: > I can't remember why this (shouldn't they be the same)?

I have no clue :smile: I agree, they should be the same. Should I increase it?

> It looks like both GCP and AWS max-out at 30 seconds for HTTP gateway based requests.

That's strange if we have the limit set to 15 seconds, isn't it?

packages/airnode-deployer/test/fixtures/index.ts
aquarat: All good 👌

packages/airnode-deployer/src/cli/commands.ts
```diff
@@ -140,14 +71,15 @@ export async function deploy(configPath: string, secretsPath: string, receiptFil
   if (output.httpSignedDataGatewayUrl) logger.info(`HTTP signed data gateway URL: ${output.httpSignedDataGatewayUrl}`);
 }
 
-export async function remove(airnodeAddressShort: string, stage: string, cloudProvider: CloudProvider) {
-  await removeAirnode({ airnodeAddressShort, stage, cloudProvider });
+export async function remove(airnodeAddress: string, stage: string, cloudProvider: CloudProvider) {
```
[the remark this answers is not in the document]
amarthadan: Right, I haven't realized that. I'll adjust it :slightly_smiling_face:

packages/airnode-deployer/terraform/gcp/main.tf
aquarat: > I have no clue 😄 I agree, they should be the same. Should I increase it?
I think so, it should probably be 30 seconds.

> That's strange if we have the limit set to 15 seconds, isn't it?
I don't understand :P There's a hard-limit on signed data gateway requests of 30 seconds imposed by the cloud providers - so it's possible for the handler to be timed out by AWS/GCP because our underlying `go` based timeouts are also 30s I think.

Siegrift (changes requested):
I encountered an issue when running this which @amarthadan reproduced, will take another look when he debugs the issue.

martinkolenic:
Hi @amarthadan and @Siegrift, it seems that despite my best effort I did something wrong during testing and I cannot reproduce the functioning results now. I'm currently investigating it.

amarthadan:
@Siegrift It should be fixed now.

martinkolenic:
> @Siegrift It should be fixed now.

I managed to deploy and send a request now, it seems to be working correctly now.

Siegrift (commented):
## TLDR: 

All basic ops seems to be working, but there were some unexpected failures along the way. I wrote notes for the full journey, maybe it helps... Will take a look at the code once more tomorrow.

## Notes 

- Deployed Airnode with both gateways
    - success
- I forgot to specify HTTP trigger, so I redeployed with modified config (same stage)
    - failed (because it was already there) and autoremove removed the original Airnode
    - this is an awkward side effect of autoremove, but it’s the technically correct thing
    - BUG: This should not have happened (but later this exact thing works again)
Pasted Graphic 3

- Deployed again
    - success, both gateways work and RRP request works
- Changed stage, deployed again
    - failed, because gateway keys already exist, but autoremove was sucessfull
    - the error is kinda annoying, but we I don't think we can do anything about it
    - previous airnode works
- Changed secrets and redeployed
    - failed, got message `“Specified ReservedConcurrentExecutions for function decreases account's UnreservedConcurrentExecution below its minimum value of [100].”`
    - The error is correct. The account unreserved concurency is 189, the example reserves 100 for RRP and 20 per gateway = 140
    - TODO(new task): Decrease examples reserved concurrency to 1
- Changed “maxConcurrency” to 1
    - success, gateways + RRP works, aws folder structure is correct
    - I actually also changed the mnemonic so it deployed a completely different Airnode
Objects Properties

- Created new secrets again and changed stage, but preserved mnemonic, deployed again
    - success, everything works
0x6abEdc0A4d1A79eD62160396456¢95C5607369D3

- Changed mnemonic to the old Airnode, but preserved the same stage (and forgot to update gateway keys)
    - wanted to stop deployment by quickly pressing ctrl+z, deployer got stuck so I killed it the hard way (stopped the docker)
    - TODO(task): The docker (or the spinner) should not get frozen when pressing ctrl+z
    - the original airnode is still operational (both gateways, didn’t try RRP)
Pasted Graphic 2

- Created new gateway keys, trying to redeploy
    - failed, error: “Failed to copy file '0x5691ceD2b03593a8d67BdfF4Fe96107bfe603825/dev3/1663007611553/default.tfstate' to file '0x5691ceD2b03593a8d67BdfF4Fe96107bfe603825/dev3/1663007887640/default.tfstate' within S3 bucket 'airnode-6d46164e9ddd': NoSuchKey: The specified key does not exist.”
    - No idea what it means, autoremove worked
- Trying again the same thing
    - both gateways work, aws structure is correct
- removing the last airnode (and lost the receipts for others)
    - success
- deploying again
    - success
- deploying again (to verify re-deployment bug)
    - this time success
Properties

- trying to deploy again 2x
    - success, success
- change region, deploy
    - deploy fail, autoremove fail (which is actually a success)
Pasted Graphic 6

- changed config.json by disabling http gateway, deploy
    - deploy fail, autoremove success (so it removed the airnode)
image

- changed the config back, successfully redeployed, changed the config to disable one of the gateways, deployed again
    - worked

packages/airnode-deployer/src/utils/infrastructure.ts
```diff
@@ -34,3 +35,101 @@ export class MultiMessageError extends Error {
     super(messages.join('\n'));
   }
 }
+
+export const BUCKET_NAME_REGEX = /^airnode-[a-f0-9]{12}$/;
+export const generateBucketName = () => `airnode-${randomBytes(6).toString('hex')}`;
```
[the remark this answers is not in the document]
amarthadan: > 1. This is to address the GCP unique bucket names limitation right? (I'd prefer a comment for this, since this idea will get forgotten in time).

Sure.

>     2. The bucket name feels a bit awkward when user checks the buckets in AWS. I don't have a better idea though. Maybe I'd call it `airnodes` (plural) since it will hold multiple.
> 
>     3. The common bucket name needs to be documented in the docs. I'd also prefer if the deployer CLI told me that it's creating a shared bucket called `airnodes-xyz`. Same if it's reusing a bucket...

It does tell you that in the debug mode. I don't think users care about this much that's why I put it there. Serverless doesn't provide such info by default I think.

amarthadan:
@Siegrift Thanks for the thorough test.

As far as I can tell these are the issues that need to be investigated/fixed:
1. `The docker (or the spinner) should not get frozen when pressing ctrl+c`
2. `Failed to copy file` issue - This most likely happened because you forcefully stopped the deployment. But I agree that we can prevent at least some cases.
3. `changed config.json by disabling http gateway, deploy -> deploy fail` - I have no idea what this is, will investigate.

As for the topmost issue (screenshot), I don't think that happened because the Airnode was already there. From the screenshot, it looks like a network error. There's not much we can do about that. Or are they two different issues?

If you're ok with it I would address these in separate PRs. Also, do these need to be fixed for 0.9 to be released? That's more of a question for @aquarat I guess. I would like to address them but if we're waiting for 0.9 they are not critical and can be fixed later.

Siegrift:
> If you're ok with it I would address these in separate PRs.

Yeah. Btw. as you can see I only tested the AWS part. Leaving the GCP testing to @martinkolenic .
I'll take a one more look at code and give the final review...

packages/airnode-deployer/package.json
```diff
@@ -15,7 +15,7 @@
     "clean": "rimraf -rf *.tsbuildinfo ./dist ./.webpack ./build *.tgz",
     "cli": "ts-node bin/deployer.ts",
     "compile": "tsc --build tsconfig.json",
-    "copy:terraform": "copyfiles terraform/**/**/**/*.tf terraform/**/**/**/*.tpl  dist/",
+    "copy:terraform": "copyfiles terraform/**/**/*.tf terraform/**/**/*.tpl  dist/",
```
Siegrift, line 18: Mhh, is the pattern "/**/" (x2) necessary? The ** stands for any nested directory so **/**  feels redundant.

packages/airnode-deployer/src/infrastructure/aws.test.ts
```diff
@@ -0,0 +1,315 @@
...
+  putObject: jest.fn(),
+  getObject: jest.fn(),
+  copyObject: jest.fn(),
+  deleteObjects: jest.fn(),
+  deleteBucket: jest.fn(),
+};
+
+jest.mock('aws-sdk', () => ({
+  config: {
+    update: jest.fn().mockImplementation(() => {}),
```
Siegrift, line 32:
```suggestion
    update: jest.fn(),
```

packages/airnode-deployer/src/infrastructure/aws.test.ts
```diff
@@ -0,0 +1,315 @@
...
+  S3: jest.fn(() => mockS3),
+}));
+
+jest.mock('../utils/infrastructure', () => ({
+  ...jest.requireActual('../utils/infrastructure'),
+  generateBucketName: jest.fn(),
+}));
+
+const awsConfigUpdateSpy = jest.requireMock('aws-sdk').config.update;
+const awsListBucketsSpy: jest.SpyInstance = jest.requireMock('aws-sdk').S3().listBuckets;
```
Siegrift, line 43: Nit: You could just do `const abc = mockS3.xyz` or just use `mock.xyz` in the tests as well. (But I also like the explicitness). Leaving up to you.

packages/airnode-deployer/src/infrastructure/aws.test.ts
```diff
@@ -0,0 +1,315 @@
...
+  it(`throws an error if can't create a bucket`, async () => {
+    awsCreateBucketSpy.mockImplementation(mockPromise(jest.fn().mockRejectedValue(s3Error)));
+
+    await expect(createAirnodeBucket(cloudProvider)).rejects.toThrow(
+      new Error(`Failed to create an S3 bucket: Error: ${s3ErrorMessage}`)
+    );
+    expect(awsConfigUpdateSpy).toHaveBeenCalledWith({ region: cloudProvider.region });
+  });
+
+  it(`throws an error if can't set enctyprion`, async () => {
```
Siegrift, line 143:
```suggestion
  it(`throws an error if can't set encryption`, async () => {
```

packages/airnode-deployer/src/infrastructure/aws.test.ts
```diff
@@ -0,0 +1,315 @@
...
+
+    await expect(createAirnodeBucket(cloudProvider)).rejects.toThrow(
+      new Error(`Failed to setup a public access block for bucket '${bucketName}': Error: ${s3ErrorMessage}`)
+    );
+    expect(awsConfigUpdateSpy).toHaveBeenCalledWith({ region: cloudProvider.region });
+  });
+});
+
+describe('getBucketDirectoryStructure', () => {
+  it('returns bcuket directory structure', async () => {
```
Siegrift, line 163:
```suggestion
  it('returns bucket directory structure', async () => {
```

https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

packages/airnode-deployer/src/infrastructure/aws.ts
```diff
@@ -1,74 +1,203 @@
...
-  logger.debug('Removing Terraform state from AWS');
-  const { region } = cloudProvider;
-  AWS.config.update({ region });
-  const s3 = new AWS.S3();
+export const getBucketDirectoryStructure = async (cloudProvider: AwsCloudProvider, bucketName: string) => {
+  const s3 = initializeS3Service(cloudProvider);
 
-  await deleteBucket(s3, bucket);
-}
+  let paths: String[] = [];
```
Siegrift, line 89: Lowercase `string`,  never use the uppercase ones.

packages/airnode-deployer/src/infrastructure/gcp.test.ts
```diff
@@ -0,0 +1,328 @@
...
+
+    await expect(createAirnodeBucket(cloudProvider)).rejects.toThrow(
+      new Error(`Failed to setup IAM policy for bucket '${bucketName}': Error: ${gcsErrorMessage}`)
+    );
+    expect(gcsStorageSpy).toHaveBeenCalledWith({ projectId: cloudProvider.projectId });
+  });
+});
+
+describe('getBucketDirectoryStructure', () => {
+  it('returns bcuket directory structure', async () => {
```
Siegrift, line 178:
```suggestion
  it('returns bucket directory structure', async () => {
```

packages/airnode-deployer/src/infrastructure/gcp.ts
```diff
@@ -1,22 +1,181 @@
...
+    })
+  );
+  if (!goMetadata.success) {
+    throw new Error(`Failed to setup a uniform bucket-level access for bucket '${bucketName}': ${goMetadata.error}`);
+  }
+
+  const policy = {
+    bindings: [
+      {
+        role: 'roles/storage.legacyBucketReader',
```
Siegrift, line 61: Can we document why we need this? I am struggling to understand why many of the things like these are needed... (Same goes for the AWS logic).

Basically anything that has to be set in cloud is hard to reason about, whether it must be set and why. I fear this will be painful to modify in a few months advance.

packages/airnode-deployer/src/utils/infrastructure.ts
```diff
@@ -34,3 +35,101 @@ export class MultiMessageError extends Error {
...
+};
+
+export type File = {
+  type: FileSystemType.File;
+  bucketKey: string;
+};
+
+export type DirectoryStructure = Record<string, FileSystemItem>;
+
+export const translatePathsToDirectoryStructure = (paths: String[]) => {
```
Siegrift, line 62: I'd implement this recursively, but it would be a bit complex either way. 

I think documentation + unit tests are more then enough for this.

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -67,123 +71,175 @@ async function runCommand(command: string, options: CommandOptions) {
...
 
-async function terraformAirnodeManage(
-  command: string,
-  execOptions: child.ExecOptions,
-  cloudProvider: CloudProviderExtended,
-  bucket: string,
-  variables: AirnodeVariables
-) {
-  const terraformAirnodeCloudProviderDir = path.join(terraformAirnodeDir, cloudProvider.type);
+export function prepareAirnodeApplyDestroyArguments(variables: AirndoeApplyDestroyVariables): CommandArg[] {
```
Siegrift, line 177:
```suggestion
export function prepareAirnodeApplyDestroyArguments(variables: AirnodeApplyDestroyVariables): CommandArg[] {
```

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -260,91 +285,208 @@ function transformTerraformOutput(terraformOutput: string): DeployAirnodeOutput
...
+
+    // Delete the whole bucket if empty
+    if (Object.keys(directoryStructure).length === 0) {
+      logger.debug(`Deleting Airnode bucket '${bucketName}'`);
+      await cloudProviderLib[type].deleteBucket(cloudProvider as any, bucketName);
+    }
+  });
+
+  if (!goRemove.success) {
+    spinner.fail(`Failed removing Airnode ${airnodeAddress} ${stage} from ${type} ${region}`);
```
Siegrift, line 487: Nit:

```suggestion
    spinner.fail(`Failed to remove Airnode ${airnodeAddress} ${stage} from ${type} ${region}`);
```

Siegrift (changes requested):
Nice work. I've found a few comments and typos and I am not sure about updating using a different nodeVersion...

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -199,57 +255,26 @@ async function terraformAirnodeManage(
...
+  if (!isEmpty(importOptions)) {
+    await execTerraform(
+      { ...execOptions, ignoreError: true },
+      'import',
+      prepareCloudProviderAirnodeApplyDestoryArguments(cloudProvider, bucket, bucketPath, commonArguments),
+      cloudProviderAirnodeImportOptions[cloudProvider.type](cloudProvider as any)
+    );
   }
 
   commonArguments.push('auto-approve');
```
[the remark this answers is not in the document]
amarthadan: It does but that's what you want. Otherwise, Terraform's waiting for the user to type `yes` or something.

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -199,57 +255,26 @@ async function terraformAirnodeManage(
...
-    spinner.fail(`Failed deploying Airnode ${airnodeAddressShort} ${stage} to ${type} ${region}`);
-    throw goDeploy.error;
-  }
-
-  spinner.succeed(`Deployed Airnode ${airnodeAddressShort} ${stage} to ${type} ${region}`);
-  return goDeploy.data;
+  await execTerraform(
+    execOptions,
+    'apply',
+    prepareCloudProviderAirnodeApplyDestoryArguments(cloudProvider, bucket, bucketPath, commonArguments)
```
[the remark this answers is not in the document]
amarthadan: It is. The `apply` and `destroy` commands need the same arguments (mostly). That's why the abstraction, And I agree that the name's no good but I wasn't able to come up with a better one :sweat_smile:

packages/airnode-deployer/src/infrastructure/index.ts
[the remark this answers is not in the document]
Siegrift: But what if I want to disable the auto-remove?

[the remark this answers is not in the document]
Siegrift: Oh, I confused this briefly with `auto-remove` and found it strange what it does here. Ignore this.

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -260,91 +285,208 @@ function transformTerraformOutput(terraformOutput: string): DeployAirnodeOutput
...
+      logger.debug(`Fetching configuration file '${bucketConfigPath}'`);
+      const remoteConfig = JSON.parse(
+        await cloudProviderLib[type].getFileFromBucket(cloudProvider as any, bucketName, bucketConfigPath)
+      ) as Config;
+
+      const remoteNodeSettings = remoteConfig.nodeSettings;
+      const remoteCloudProvider = remoteNodeSettings.cloudProvider as CloudProvider;
+      if (remoteNodeSettings.nodeVersion !== nodeVersion) {
+        throw new Error(
+          `Can't update an Airnode deployment with airnode-deployer of a different version. Deployed version: ${remoteNodeSettings.nodeVersion}, airnode-deployer version: ${nodeVersion}`
```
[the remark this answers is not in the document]
amarthadan: Depends on what you consider updating. It won't affect changing a configuration of a deployed Airnode (redeployment). But you won't be able to update from let's say 0.8.0 to 0.9.0. That should be done by removing the old Airnode (with 0.8.0 deployer) and deploying the new one (with 0.9.0 deployer).

packages/airnode-deployer/src/infrastructure/index.ts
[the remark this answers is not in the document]
Siegrift: Maybe even `prepareCloudProviderTerraformArguments`?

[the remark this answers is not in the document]
amarthadan: But there's also a Terraform `init` command that needs different arguments :smile: Hence the distinction.

packages/airnode-deployer/src/infrastructure/index.ts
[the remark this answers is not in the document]
Siegrift: Oh, I see. And when you remove the Airnode it will remove the directory structure, so this check will not be reached and Airnode will be deployed sucessfully.

packages/airnode-deployer/src/infrastructure/index.ts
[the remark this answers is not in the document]
Siegrift: Ok, leave it as is :D

packages/airnode-deployer/package.json
amarthadan: You would think so but it doesn't work that way for `copyfiles`. I can doublecheck but I'm pretty sure I had to go as deep as the directory structure was in order to not miss any files.

packages/airnode-deployer/src/infrastructure/index.ts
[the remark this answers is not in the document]
amarthadan: The directory structure is common for all the Airnodes of all versions. You can have for example one Airnode with address A of version 0.8.0 and one with address B of version 0.9.0. The check checks only the specific Airnode (it's `config.json` to be precise) against the version of the deployer running the deployment/removal.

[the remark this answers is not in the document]
Siegrift: I mean it will remove the directory structure for the specific Airnode address (not the whole bucket folder)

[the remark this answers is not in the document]
amarthadan: Right, yes, that it will.

packages/airnode-deployer/package.json
amarthadan: Yes, it needs to be like this :man_shrugging:

martinkolenic:
Hi @amarthadan, with the newest commit `fa7340ca`, running `yarn build` in the root directory of the repo results in a failure to build `airnode-deployer`, see screenshot.

image

amarthadan:
> Hi @amarthadan, with the newest commit `fa7340ca`, running `yarn build` in the root directory of the repo results in a failure to build `airnode-deployer`, see screenshot.

That was just a temporary state as I accepted the suggested changes. Should be fixed now.

Siegrift (approved):
👍 LGTM, but haven't tested the GCP side.

btw. thx for not squashing the last commits.

martinkolenic:
Hi @amarthadan, I tried to follow the same steps as @Siegrift did but at the very first deployment attempt I encountered this error:
```
  ___  _                      _      
 / _ \(_)                    | |     
/ /_\ \_ _ __ _ __   ___   __| | ___ 
|  _  | | '__| '_ \ / _ \ / _` |/ _ \
| | | | | |  | | | | (_) | (_| |  __/
\_| |_/_|_|  |_| |_|\___/ \__,_|\___|

          Airnode v0.8.0

✖ Failed deploying Airnode 0x9AEd5F8b1B41100341acB5CE25b656C9b3bc5EAf dev to gcp us-east1
ℹ Outputted config/receipt.json
  This file does not contain any sensitive information.
✖ Airnode deployment failed due to unexpected errors.
  It is possible that some resources have been deployed on cloud provider.
  Attempting to remove them...

✖ Failed to remove Airnode 0x9AEd5F8b1B41100341acB5CE25b656C9b3bc5EAf dev from gcp us-east1
✖ Airnode removal failed due to unexpected errors.
  It is possible that some resources have been deployed on cloud provider.
  Please check the resources on the cloud provider dashboard and
  use the "remove" command from the deployer CLI to remove them.
  If the automatic removal via CLI fails, remove the resources manually.



Error details:
✖ Deployment error:
Failed to list GCS buckets: Error: invalid_grant: Invalid JWT: Token must be a short-lived token (60 minutes) and in a reasonable timeframe. Check your iat and exp values in the JWT claim.
✖ Removal error:
Failed to list GCS buckets: Error: invalid_grant: Invalid JWT: Token must be a short-lived token (60 minutes) and in a reasonable timeframe. Check your iat and exp values in the JWT claim.
Done in 3.69s.
```

amarthadan:
> Hi @amarthadan, I tried to follow the same steps as @Siegrift did but at the very first deployment attempt I encountered this error:

That sounds like an authentication error. Can you check that your `gcp.json` is correct and that you're passing it to the container (I presume you're using a Docker container)?

martinkolenic:
Hi @amarthadan, thanks for helping me troubleshoot as the message was rather cryptic. This was indeed caused by an incorrect gcp.json.

martinkolenic:
Hi @amarthadan, in the end, a newly-generated gcp.json did not help. The same message keeps appearing.

PR #1433, opened 2022-09-09
Fix wrong name in airnode-examples Requester contracts

Siegrift:
See the base RRP Requester constructor: https://github.com/api3dao/airnode/blob/master/packages/airnode-protocol/contracts/rrp/requesters/RrpRequesterV0.sol#L23


dcroote (approved):
Nice 👍

PR #1434, opened 2022-09-09
Remove unnecessary validator bundling

Siegrift:
Closes https://github.com/api3dao/airnode/issues/1431


amarthadan (approved):
LGTM but I haven't tested the package as NPM.

Siegrift:
I've tested it. It didn't work when I did `npx @api3/airnode-validator --config "config.json" --secrets "secrets.env"` inside the monorepo, but I believe that was because of some strange npx behaviour. When I ran the exact same command outside Airnode monorepo it worked correctly (It used the local verdaccio package).

Siegrift:
For @martinkolenic , double check that the validator packaging is not broken. This is just a refactor, there were no user facing changes.

martinkolenic:
Hi @Siegrift, I've repeatedly tried using the package, but when attempting to run the validator via npx (same command as you used above) within the repo, I keep receiving this error message:

```
sh: line 1: airnode-validator: command not found
```

Siegrift:
Yeah, had the same error. See https://github.com/api3dao/airnode/pull/1434#issuecomment-1246373814

martinkolenic:
I see, thanks. Is there any way to circumvent this or does it need to be fixed first?

Siegrift:
Just run it outside Airnode monorepo.

martinkolenic:
A quick recap of my testing process:

1. Installed verdaccio (had to reinstall as I recently distro-hopped and didn't have the package) and ran it in a console tab
2. Set the package registry to http://localhost:4873
3. Entered the airnode monorepo root dir and checked out the `temp-v0.10.0` branch
4. Ran `yarn bootstrap && yarn build` - this should produce the packages
5. Published the package to verdaccio
6. Installed the package from the local registry and ran it over a config+secrets combo

As I fed it a config with old OIS version, it complained (which was expected). Running the validator without any arguments led to the standard help output. If I did not mess up the process somewhere, this is working correctly.

Siegrift:
Yeah, that seems correct. Thanks for testing.

PR #1438, opened 2022-09-11
Correct test fixtures

Siegrift:
I noticed that `fulfillData` is encoded wrong. This doesn't matter for the unit tests, but would matter if the data was to be decoded on chain. This might confuse users poking around the tests to learn how stuff works.


bbenligiray (approved):
Very odd, good catch

PR #1439, opened 2022-09-11
Remove unnecessary async/await

Siegrift:
## Problem description

While working in an unrelated feature I installed @api3/airnode-admin and wanted to use `deriveSponsorWalletAddress` function which should be synchronous, but I had to use `await`... 

## Fix

I realized that it's because the function is has the `async` modifier in the implementation, but it doesn't await anything. I added two eslint rules (`require-await` to make sure async is used only when necessary and `no-return-await` to remove unnecessary await).

The implementation mostly removes the unnecessary async/await, but there are a few exception. Some functions need to be asynchronous even though in nature they could be synchronous. There are two workarounds:
- Use `Promise.resolve(value)` 
- Disable the rule for a particular line

### DIFF
--- a/.changeset/stupid-knives-lick.md
+++ b/.changeset/stupid-knives-lick.md
@@ -1,0 +1,11 @@
+---
+'@api3/airnode-admin': patch
+'@api3/airnode-deployer': patch
+'@api3/airnode-examples': patch
+'@api3/airnode-node': patch
+'@api3/airnode-operation': patch
+'@api3/airnode-protocol': patch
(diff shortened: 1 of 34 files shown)


amarthadan (approved):
LGTM but this could have easily broken something. It needs to be tested properly.

Siegrift:
>  this could have easily broken something. It needs to be tested properly.

TBH this is quite safe and even backwards compatible. E.g. if some code uses `await fn` and `fn` is not async anymore, the code will still work.

PR #1442, opened 2022-09-13
Fix broken README link

amarthadan:


amarthadan:
> +1 LGTM.
> 
> How did this error happened though? Shouldn't the PR fail?

Yes, it should have. No idea.

PR #1456, opened 2022-09-19
Fix Terraform dependency cycle error when removing HTTP gateways

### LINKED ISSUE #1450: Fix failing redeployment when the HTTP gatewayes become disabled
Based on the [discussion](https://github.com/api3dao/airnode/pull/1424#issuecomment-1244965821) in https://github.com/api3dao/airnode/pull/1424

During the PR review for https://github.com/api3dao/airnode/pull/1424 @Siegrift run into an issue where he got an error message noting a cyclic dependency in the Terraform recipes
```
Error: Cycle: module.httpSignedGw[0].aws_api_gateway_deployment.deployment (destroy), module.httpSignedGw[0].aws_api_gateway_rest_api.api_gateway (destroy), module.httpSignedReq[0].aws_lambda_function.lambda (destroy), module.httpSignedReq[0].null_resource.fetch_lambda_files (destroy), module.startCoordinator.aws_lambda_function.lambda
```

Steps to reproduce (not reliably though :smile:):
1. Deploy Airnode with both HTTP gateways enabled
2. Redeploy the same Airnode (running `deploy` again without a removal) with one of the gateways disabled
3. Redeploy the same Airnode (running `deploy` again without a removal) with both of the gateways disabled

amarthadan:
Close #1450

The first two commits are just some cleaning/minor refactor since I was going through the recipes anyway.

The last commit is the fix to #1450. The problem is in the [two lines](https://github.com/api3dao/airnode/blob/master/packages/airnode-deployer/terraform/aws/main.tf#L34-L35) that are passing the gateways' URLs to the coordinator's env variables. At this point, I'm almost certain that the dependency cycle error is a bug in Terraform and shouldn't happen. Even though the coordinator references the gateways (their output), the gateways don't reference the coordinator back, so there should not be any dependency cycle. I may create a minimal reproducible example and open it as an issue. The only solution that works is to condition the whole module instead of the individual environment variables. It's ugly, I don't like it, but, it's the only thing that works. That being said, I would still merge it. Even though the scenario where the error occurs is improbable I would keep the fix even though it complicates the code a bit.

The error doesn't occur with GCP (which is also odd as the structure of the modules and references between them are the same).

### DIFF
--- a/packages/airnode-deployer/terraform/aws/main.tf
+++ b/packages/airnode-deployer/terraform/aws/main.tf
@@ -19,8 +19,9 @@ module "run" {
   }
 }
 
-module "startCoordinator" {
+module "startCoordinatorNoGws" {
   source = "./modules/function"
+  count  = var.http_api_key == null && var.http_signed_data_api_key == null ? 1 : 0
 
   name               = "${local.name_prefix}-startCoordinator"
   handler            = "index.startCoordinator"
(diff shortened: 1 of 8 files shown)


packages/airnode-deployer/terraform/aws/main.tf
```diff
@@ -31,8 +32,79 @@ module "startCoordinator" {
...
+  invoke_targets                 = [module.run.lambda_arn]
+  schedule_interval              = 1
+  reserved_concurrent_executions = var.disable_concurrency_reservation ? null : 1
+
+  depends_on = [module.run]
+}
+
+module "startCoordinatorHttpGw" {
+  source = "./modules/function"
+  count  = var.http_api_key != null && var.http_signed_data_api_key == null ? 1 : 0
```
Siegrift, line 47: What is this count for? Also is it correct? (Currently, it's the same for all 4 options). Also, the `http_signed_data_api_key` seems strange here as this is for http gw.

amarthadan: > Currently, it's the same for all 4 options

It's not the same for all 4 options, it's different for each option, that's the point. I have to provide a different module for each possibility:
* no gateways
* only HTTP gateway
* only signed data HTTP gateways
* both gateways

The `count` argument is always set to either 1 (there should be exactly 1 instance of this module) or 0 (there should be no instance of this module). I'm basically switching on and off copies of the same module based on what gateway(s) are enabled. And passing their output (this time without a condition) to its environment variables.

Siegrift: Oh, sorry 🤦 . I see the difference now.

aquarat:
I don't have much insight to offer but I can't see any issues/problems.

Siegrift (approved):
I like the code as well. Thanks for splitting this into separate commits.

Giving approve since @aquarat also liked this, but I haven't tested whether this indeed fixes the issue. LMK if you want me to test this on cloud as well.

amarthadan:
@Siegrift Yes, if you can, please, do.

Siegrift (changes requested):
Tried to deploy an Airnode and got an issue due to missing source maps. I'll test again when @amarthadan fixes the issue.

amarthadan:
@Siegrift Should be fixed, you can try now.

Siegrift (approved):
👍 LGTM. I also tested

image

Issue #1460, opened 2022-09-20
Allow users to remove the deployment based on the deployment ID


amarthadan:
User will be able to remove the deployment based on the deployment ID, e.g.:

```bash
airnode-deployer remove aws1697238d
```
Based on the deployment ID from the `list` command (https://github.com/api3dao/airnode/issues/1403).

This will replace the current `remove-with-deployment-details` command.

The deployment ID will be also added to the `receipt.json` and will be printed on the console at the end of the deployment process.

amarthadan:
@bbenligiray @aquarat I've updated the description, let me know if it's ok.

bbenligiray:
l-g-t-m

martinkolenic:
Hi, I tested the following:

- removal of an AWS project (success)
- removal of a GCP project (success)

This looks well and from an end-user perspective, is also much easier to use than providing all the individual deployment details.

Issue #1404, opened 2022-08-29
Allow users to retrieve info about their deployments


amarthadan:
Allow users to see info and history of changes for a specific deployment.
```bash
airnode-deployer info 40207f25

Cloud provider: AWS (us-east-1)
Airnode address: 0xd0624E6C2C8A1DaEdE9Fa7E9C409167ed5F256c6
Stage: dev
Airnode version 0.9.0
┌────────────┬───────────────────────────┐
│ Version ID │ Deployment time           │
├────────────┼───────────────────────────┤
│ 5bbcd317   │ 2022-08-29 09:12:34 GMT+2 │
├────────────┼───────────────────────────┤
│ 21f1b492   │ 2022-09-29 07:34:33 GMT+2 │
├────────────┼───────────────────────────┤
│ 2c8b8bbe   │ 2022-08-12 10:17:09 GMT+2 │
└────────────┴───────────────────────────┘
```
Based on the deployment ID from the `list` command (https://github.com/api3dao/airnode/issues/1403).
The version ID will be constructed as the first 8 characters of sha([cloudProvider, region, airnodeAddress, stage, airnodeVersion, deploymentTime]) so it's unique for the deployment and its historical version.

Basically a preparation for https://github.com/api3dao/airnode/issues/1406

amarthadan:
@bbenligiray @aquarat I've updated the description, let me know if you're ok with this.

Siegrift:
Maybe:
```
Cloud provider: AWS (us-east-1)
Airnode address: 0xd0624E6C2C8A1DaEdE9Fa7E9C409167ed5F256c6
Stage: dev
Airnode version 0.9.0
┌────────────┬───────────────────────────┐
│ Version ID │ Deployment time           │
├────────────┼───────────────────────────┤
│ 5bbcd317   │ 2022-08-29 09:12:34 GMT+2 │ (current)
├────────────┼───────────────────────────┤
│ 21f1b492   │ 2022-09-29 07:34:33 GMT+2 │
├────────────┼───────────────────────────┤
│ 2c8b8bbe   │ 2022-08-12 10:17:09 GMT+2 │
└────────────┴───────────────────────────┘
```

Siegrift:
This info command will only show Airnodes deployed with the current Airnode version. Correct me if I am wrong, but this means that the user will first have to remove all of these (which will remove the files in AWS) and then redeploy with a new deployer version.

amarthadan:
[...]

PR #1462, opened 2022-09-21
Allow listing of active Airnode deployments

### LINKED ISSUE #1403: Allow listing of active Airnode deployments
Allow users to list the active Airnode deployment via airnode-deployer CLI in a similar fashion as `docker ps` command.

```bash
airnode-deployer list

Deployment ID   Cloud provider   Airnode address   Stage       Region        Node version   Last Update
1697238d        AWS              0x1234567...      dev         us-east-1     0.7.3          2022-08-29T09:12:34+0000
ada40890        AWS              0x1234567...      prod        us-east-1     0.6.5          2022-09-29T07:34:33+0000
1c4dbc3d        GCP              0xabcdef0...      mainnet-2   eu-central1   0.6.5          2022-08-12T10:17:09+0000
```
The output will be ordered by the cloud provider, Airnode address, region, stage, and Airnode version in this order. There's an argument to be made about ordering the list by the last update time but I would prefer if the output didn't move around every time I redeploy an Airnode. we can add options for multiple orderings later on.

The deployment ID will be constructed as the first 8 characters of `sha([cloudProvider, region, airnodeAddress, stage, airnodeVersion])`. The last update time is not included so the update to the given Airnode won't change it's deployment ID.

There will be a `--cloud-providers` option allowing the listing of Airnodes only from the given cloud provider(s). With no option provided, the deployer will try to list all (both) the cloud providers.

amarthadan:
Close #1403


packages/airnode-deployer/src/infrastructure/aws.test.ts
```diff
@@ -60,44 +59,38 @@ const cloudProvider = {
 const bucketName = 'airnode-aabbccdd0011';
 const fileContent = 'file content';
 const filePath = '/path/to/config.json';
-const bucketFilePath = '0xA30CA71Ba54E83127214D3271aEA8F5D6bD4Dace/dev/1662559204/config.json';
+const bucketFilePath = '0xA30CA71Ba54E83127214D3271aEA8F5D6bD4Dace/dev/1662559204554/config.json';
```
Siegrift, line 62: Mhh, is this needed? :D

packages/airnode-deployer/src/infrastructure/aws.ts
```diff
@@ -35,8 +34,8 @@ export const getAirnodeBucket = async (cloudProvider: AwsCloudProvider) => {
   return airnodeBuckets?.[0]?.Name ?? null;
 };
 
-export const createAirnodeBucket = async (cloudProvider: AwsCloudProvider) => {
-  const s3 = initializeS3Service(cloudProvider);
+export const createAirnodeBucket = async (_cloudProvider: AwsCloudProvider) => {
```
Siegrift, line 37: Why not remove the argument?

packages/airnode-deployer/src/infrastructure/index.test.ts
```diff
@@ -866,3 +853,165 @@ describe('removeAirnode', () => {
...
+    awsGetFileFromBucketSpy = jest
+      .spyOn(aws, 'getFileFromBucket')
+      .mockImplementation(() => Promise.resolve(fs.readFileSync(configPath).toString()));
+    gcpGetAirnodeBucketSpy = jest.spyOn(gcp, 'getAirnodeBucket').mockImplementation(() => Promise.resolve(bucket));
+    gcpGetBucketDirectoryStructureSpy = jest
+      .spyOn(gcp, 'getBucketDirectoryStructure')
+      .mockImplementation(() => Promise.resolve(directoryStructure));
+    gcpGetFileFromBucketSpy = jest
+      .spyOn(gcp, 'getFileFromBucket')
+      .mockImplementation(() => Promise.resolve(fs.readFileSync(configPath).toString()));
```
Siegrift, line 887: Nit: You can simplify this using https://jestjs.io/docs/mock-function-api#mockfnmockresolvedvaluevalue

packages/airnode-deployer/src/infrastructure/index.test.ts
```diff
@@ -866,3 +853,165 @@ describe('removeAirnode', () => {
...
+    consoleSpy = jest.spyOn(console, 'log');
+  });
+
+  it('lists Airnodes from multiple cloud providers', async () => {
+    const originalColorVariable = process.env.FORCE_COLOR;
+    // I have to disable table coloring so I can compare the output
+    process.env.FORCE_COLOR = '0';
+    const cloudProviders = ['aws', 'gcp'] as const;
+    const expectedOutput = `┌───────────────┬─────────────────┬───────────────────┬───────┬─────────────────┬───────────────────────────┐
```
Siegrift, line 896: Consider jest snapshots for these

packages/airnode-deployer/src/infrastructure/index.test.ts
```diff
@@ -441,22 +442,20 @@ describe('deployAirnode', () => {
   });
 
   it('deploys Airnode', async () => {
-    const terraformOutput = await infrastructure.deployAirnode(config, configPath, secretsPath);
+    const terraformOutput = await infrastructure.deployAirnode(config, configPath, secretsPath, Date.now());
```
Siegrift, line 445: Shouldn't the date be mocked somehow?

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -490,3 +487,103 @@ export async function removeAirnode(airnodeAddress: string, stage: string, cloud
...
+      cloudProvider,
+      airnodeAddress,
+      stage,
+      airnodeVersion,
+      lastUpdate,
+    ])
+  );
+
+  // eslint-disable-next-line no-console
+  console.log(table.toString());
```
Siegrift, line 588: You should use the `consoleLog` from utils. That one just console logs

packages/airnode-deployer/src/utils/cli.ts
```diff
@@ -1,3 +1,6 @@
+import * as crypto from 'crypto';
+import { format } from 'date-fns-tz';
```
Siegrift, line 2: Do we need a new library for this? Why not just use the format from date-fns ?

packages/airnode-deployer/src/infrastructure/aws.test.ts
amarthadan: What? The change from seconds to milliseconds? Not for these tests but I'm using the same mock dicrectory structure everywhere and it is needed for the new tests. Also, I want to be consistent with the reality, it should have been in milliseconds from the beginning.

packages/airnode-deployer/src/infrastructure/aws.ts
amarthadan: Because the argument is needed for the GCP counterpart of this function. It needs to have the same interface.

packages/airnode-deployer/src/infrastructure/index.test.ts
amarthadan: I don't think it needs to be for these tests, do you?

Siegrift: No, I was just wondering whether it doesn't cause some issues, but the date is probably irrelevant for that test.

packages/airnode-deployer/src/utils/cli.ts
Siegrift: Not sure if that is worth introducing a new library. But I don't have a strong preference.

amarthadan: So, I've checked and there is a difference based on the user's locale settings (e.g. difference between GMT+2 and CEST). I think it's worth keeping it to improve the UX but not a big deal though :man_shrugging:

packages/airnode-deployer/src/infrastructure/__snapshots__/index.test.ts.snap
```diff
@@ -0,0 +1,37 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`listAirnodes lists Airnodes from multiple cloud providers 1`] = `
+"┌───────────────┬─────────────────┬───────────────────┬───────┬─────────────────┬───────────────────────────┐
+│ Deployment ID │ Cloud provider  │ Airnode address   │ Stage │ Airnode version │ Last update               │
```
Siegrift, line 5: Shouldn't this be just "Version"? I think you discussed that with Andre somewhere...

packages/airnode-deployer/src/cli/index.ts
```diff
@@ -51,7 +51,6 @@ const cliExamples = [
   'remove-with-deployment-details --airnode-address 0x6abEdc0A4d1A79eD62160396456c95C5607369D3 --stage dev --cloud-provider aws --region us-east-1',
 ];
 
-drawHeader();
```
Siegrift, line 52: I am not so sure about this change. I'd prefer the header everywhere for consistency (but I know this is subjective).

packages/airnode-deployer/src/infrastructure/__snapshots__/index.test.ts.snap
amarthadan: @bbenligiray mentioned that he would prefer just a `Version` label but I opted for `Airnode version` to stress that this is the version of the Airnode deployed not the version of the deployment, or the API or something.
We might expand this in the future though, I think the ChainAPI has a concept of API versiosn which might make sense displaying here.

Siegrift: I see. There is a lot of room to bikeshed here :D 
I'd personally just call it `Airnode`.

packages/airnode-deployer/test/snapshots/index.ts
```diff
@@ -0,0 +1,33 @@
+// Manual snapshots
+
+// Ignoring the last column due to the length and content variability
+export const listAirnodes01 = `┌───────────────┬─────────────────┬───────────────────┬───────┬─────────────────┬
+│ Deployment ID │ Cloud provider  │ Airnode address   │ Stage │ Airnode version │ Last update
```
Siegrift, line 5: Why is the "Last update" column emopty? Shouldn't there be at least the creation date?

packages/airnode-deployer/src/infrastructure/__snapshots__/index.test.ts.snap
bbenligiray: I agree with "Airnode version"

packages/airnode-deployer/test/snapshots/index.ts
amarthadan: Read the comment:
`Ignoring the last column due to the length and content variability`
Can't really compare the date as its content (and therefore length) is dependent on the timezone the test is run in.

Siegrift (changes requested):
1. Consider adding `list` command to examples
2. The shortened `Airnode address` makes the thing unusable for removing Airnode with deployment details. (This should be changed back to full address imo)

Otherwise it seems to be working fine

amarthadan:
> 1. Consider adding `list` command to examples

Can be imo done as a different task.

>     2. The shortened `Airnode address` makes the thing unusable for removing Airnode with deployment details. (This should be changed back to full address imo)

I would agree but the plan is to provide an option to [delete the deployment based on the `Deployment ID`](https://github.com/api3dao/airnode/issues/1460). And the full Airnode address will be displayed in the output from the future [info](https://github.com/api3dao/airnode/issues/1404) command. But I'm not a UX expert and it's hard to say how would users actually use this.

Siegrift (approved):
👍 LGTM.

I had one more comment about drawing header, but it seems it got lost. I'd prefer having it in all commands for consistency 🤷

amarthadan:
> I had one more comment about drawing header, but it seems it got lost. I'd prefer having it in all commands for consistency shrug

I read it I just hoped that somebody else would chime in. I agree with @andreogle that it doesn't make sense/doesn't look good for the `list` command. But I see your point about consistency :man_shrugging:

PR #1464, opened 2022-09-22
Fix TS errors in generate-references script

Siegrift:
The protocol scripts were not TypeChecked by TS and the recent PR which removed all unnecessary async/await broke the script (you can't call `.then` on synchronous function).

This PR adds type checking to the protocol scripts and removes unnecessary wrapper for `generate-references` script.


bbenligiray (approved):
Confirmed that it works. I'm merging to continue working on `deploy-sepolia`.

PR #1467, opened 2022-09-23
Disable concurrency in airnode-examples

### LINKED ISSUE #1451: Disable concurrency in airnode-examples
Initially found during testing when hit maximum concurrency limit. See: https://github.com/api3dao/airnode/pull/1424#pullrequestreview-1104590686

As Burak pointed out we should disable concurrency reservations so the deployment also works with fresh AWS accounts.

dcroote:
Closes #1451.

### DIFF
--- a/.changeset/small-trains-hide.md
+++ b/.changeset/small-trains-hide.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-examples': minor
+---
+
+Disable concurrency in airnode-examples
--- a/packages/airnode-examples/integrations/config-utils.ts
+++ b/packages/airnode-examples/integrations/config-utils.ts
@@ -19,7 +19,7 @@ export const createCloudProviderConfiguration = (generateExampleFile: boolean): 
       return {
         type: airnodeType,
         region: 'us-east-1',
-        disableConcurrencyReservations: false,
+        disableConcurrencyReservations: true,
       };
     case 'local':
       return {
@@ -30,7 +30,7 @@ export const createCloudProviderConfiguration = (generateExampleFile: boolean): 
         type: airnodeType,
         region: 'us-east1',
         projectId: integrationInfo.gcpProjectId!,
-        disableConcurrencyReservations: false,
+        disableConcurrencyReservations: true,
       };
     }
   }


Siegrift (changes requested):
You will need to re generate the config.example.json files. Also you should remove maxConcurrency fields.

dcroote:
> You will need to re generate the config.example.json files.

All of the `config.example.json` files actually use `local` as the `cloudProvider.type`, for which `disableConcurrencyReservations` doesn't apply:
https://github.com/api3dao/airnode/blob/ce1598956438c124e6030ad96a8b6014877b1b04/packages/airnode-examples/integrations/coingecko/config.example.json#L47-L49



> Also you should remove maxConcurrency fields.

But this is required:

https://github.com/api3dao/airnode/blob/ce1598956438c124e6030ad96a8b6014877b1b04/packages/airnode-validator/src/config/config.ts#L169-L195

And I thought the point of `disableConcurrencyReservations` was to override it, for example:

https://github.com/api3dao/airnode/blob/ce1598956438c124e6030ad96a8b6014877b1b04/packages/airnode-deployer/terraform/aws/main.tf#L15

Siegrift:
Yeah, I am not very happy about how the `maxConcurrency` and `disableConcurrencyReservations` play together. For RRP it has some runtime implications even though it's disabled on cloud. For signed data gateway it doesn't have any runtime (in airnode-node) implications.

dcroote:
🤔 I'm not clear then- for the purposes of #1451, are the existing changes not sufficient?

Siegrift (approved):
> I'm not clear then- for the purposes of https://github.com/api3dao/airnode/issues/1451, are the existing changes not sufficient?

It's a limitation of the config structure, nothing can be done about it atm and I don't think it's a big deal.

dcroote:
> It's a limitation of the config structure, nothing can be done about it atm and I don't think it's a big deal.

Ah ok thanks.

@aTeoke - this is currently marked as milestone `0.10.0`, but since its an approved, minor, user-facing change, any reason to not merge and release sooner as part of `0.9.0`?

PR #1466, opened 2022-09-23
Bump @api3/ois to 1.2.0 and bump zod

dcroote:
Closes #1465.

Issue #1422, opened 2022-09-06
airnode-validator fails to build with the newly published @api3/ois package


aquarat:
## Description

airnode-validator fails to build with the newly published @api3/ois package.

## Version

Please provide your Node.js, Yarn and relevant Airnode versions. Are you building from a branch, which commit?
Node 14.19.3
Yarn 1.22.19
master branch with ois dependency bumped (5acdba4f88ffeb219fdd47d615c2c85ab725825e with patch included below applied)

## Present Behaviour

airnode-validator fails to build with the newly published @api3/ois package.

## Expected Behaviour

The build should succeed.

## Steps to reproduce

Bump @api3/ois dependency to the latest in packages in Airnode that use it and then attempt the build.

## Additional comments:
It probably relates to @api3dao/ois issue 

The build log:
```
In airnode, commit 5acdba4f88ffeb219fdd47d615c2c85ab725825e:
yarn bootstrap
yarn build
...
yarn run v1.22.19
$ lerna run build --stream && ts-node scripts/make-cli-executable.ts
lerna notice cli v5.5.0
lerna info Executing command in 10 packages: "yarn run build"
@api3/airnode-abi: $ yarn run clean && yarn run compile
@api3/airnode-adapter: $ yarn run clean && yarn run compile
@api3/airnode-validator: $ yarn run clean && yarn run compile
@api3/airnode-protocol: $ yarn run clean && yarn compile
@api3/airnode-abi: $ rimraf -rf *.tsbuildinfo ./dist ./build *.tgz
@api3/airnode-protocol: $ rimraf -rf ./src/contracts *.tsbuildinfo ./dist ./build *.tgz
@api3/airnode-validator: $ rimraf -rf *.tsbuildinfo ./dist ./build *.tgz
@api3/airnode-adapter: $ rimraf -rf *.tsbuildinfo ./dist ./build *.tgz
@api3/airnode-abi: $ tsc --build tsconfig.json
@api3/airnode-protocol: $ yarn run compile:contracts && yarn run compile:tsc
@api3/airnode-validator: $ yarn tsc --build tsconfig.json
@api3/airnode-adapter: $ hardhat compile && tsc --build tsconfig.json
@api3/airnode-protocol: $ hardhat compile
@api3/airnode-validator: $ /Users/aquarat/Projects/airnode/node_modules/.bin/tsc --build tsconfig.json
@api3/airnode-protocol: Nothing to compile
[...]

PR #1468, opened 2022-09-23
Allow patch-level changes to @api3/ois dependency

dcroote:
Follow-up to #1466 based on the suggestion by @amarthadan to relax the OIS dependency in the `package.json` files. 

`~1.2` is the same as `1.2.x` per [tilde ranges](https://github.com/npm/node-semver#tilde-ranges-123-12-1).

### DIFF
--- a/.changeset/happy-trees-act.md
+++ b/.changeset/happy-trees-act.md
@@ -1,0 +1,7 @@
+---
+'@api3/airnode-adapter': minor
+'@api3/airnode-node': minor
+'@api3/airnode-validator': minor
+---
+
+Allow patch-level changes to @api3/ois dependency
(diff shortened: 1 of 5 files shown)


Siegrift (approved):
👍 Yeah, I was thinking about that when we were discussing that in Slack, but forgot to notice it on the PR.

dcroote:
@Siegrift  - I realized we may not want to allow this (or at least wait until the underlying problem is understood): when there were two versions of `zod` between OIS and Airnode we encountered cryptic errors that caused build failures for `airnode-validator` ([comment from #1422](https://github.com/api3dao/airnode/issues/1422#issuecomment-1238900265)). By allowing different OIS patch versions, with potentially different `zod` versions, we could have the same issue.

Siegrift:
Yeah, let's hold it off. I'm not sure there is a simple solution. The biggest problem is that the OIS TS types are dependant on the zod version... Any update to that dependency creates a breaking change.

PR #1475, opened 2022-09-27
Add --transpileOnly flag to ts-node scripts

Siegrift:
Since we use TS references for typechecking TS files in the app (including scripts) we no longer need the ts-node to check the types. This makes the scripts run significantly faster (there is a noticable few seconds improvement).

The disadvantage is that it ts-node will not check types, so dev mistakes will be caught only once you run tsc --build (or on CI).


amarthadan:
TBH I personally don't like this. I would much rather see the mistake/error outright when running the script instead of just during the build/in the CI. Also, the improvement is like 2-3 seconds, not 20 seconds or something. I don't think it's worth and it's making development (arguably) more difficult :man_shrugging:

I let the rest decide.

Siegrift:
@dcroote and @aquarat If I understand your emoji right, you agree with @amarthadan right?

I personally always add the -T flag if I know I need to run the TS version repeatedly. But I can continue doing so and close this PR.

dcroote:
> @dcroote and @aquarat If I understand your emoji right, you agree with @amarthadan right?

While I like the focus on speed 😁, my vote would be with @amarthadan on this one for the reasons he mentioned.

aquarat:
> @dcroote and @aquarat If I understand your emoji right, you agree with @amarthadan right?
> 
> I personally always add the -T flag if I know I need to run the TS version repeatedly. But I can continue doing so and close this PR.

I agree with @amarthadan in that (1) I'd prefer to know the errors upfront and (2) it's not much of a speed improvement.
But that said I probably won't do much/any work on Airnode in future, so I really shouldn't have commented 😆. I'll probably have to set aside time soon to try and improve the API3 org permissions/teams structure 😅. So don't take my opinion seriously^

Issue #1449, opened 2022-09-14
Gracefully handle missing files of the deployed Airnode


amarthadan:
Based on the [discussion](https://github.com/api3dao/airnode/pull/1424#issuecomment-1244965821) in https://github.com/api3dao/airnode/pull/1424

Currently, we expect some files (`config.json`, `secrets.env`, and `default.tfstate`) to be part of the  deployment (stored in the common bucket). It may happen that these files are not present due to either some deployment error, premature deployment interruption, or manual interaction with the bucket content. We should add additional check(s) for the existence of these files and provide a more user-friendly error message with a suggestion of how to continue.

PR #1480, opened 2022-09-28
Add cloud provider prefix to deployment ID

### LINKED ISSUE #1473: Prefix deployment ID with the cloud provider identification
Prefixing deployment ID with the cloud provider identification will allow us to access only the cloud provider hosting the deployment for `info` and other commands. 
More information in the [Slack discussion](https://api3workspace.slack.com/archives/C02AYRX8D89/p1664179287904649)

Part of this is also a refactor of the `info` command mentioned in https://github.com/api3dao/airnode/pull/1477

amarthadan:
Close #1473


packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -511,6 +512,79 @@ export type Deployment = {
   versions?: DeploymentVersion[];
 };
 
+// If deploymentId is provided tries fetch only that one deployment and finishes early when found
```
Siegrift, line 515: If deploymentId is provided it tries to fetch only that one deployment and returns early when found

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -511,6 +512,79 @@ export type Deployment = {
...
+async function fetchDeployments(cloudProvider: CloudProvider['type'], deploymentId?: string) {
```
Siegrift, line 516: Nit: I think we will be using type `CloudProvider['type']` so consider defining it as `type CloudProviderType = CloudProvider['type']`

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -511,6 +512,79 @@ export type Deployment = {
...
+          `Invalid item in bucket '${bucketName}' (${cloudProvider.toUpperCase()}) with key '${
+            stageDirectory.bucketKey
+          }'. Skipping.`
+        );
+        continue;
+      }
+
+      const latestDeployment = Object.keys(stageDirectory.children).sort().reverse()[0];
+      const bucketConfigPath = `${airnodeAddress}/${stage}/${latestDeployment}/config.json`;
+      const config = JSON.parse(
```
Siegrift, line 548: I'd wrap this in `go` since this can fail in many ways (users manually edit aws config, we change the region and nodeVersion field). It would be nice to handle this gracefully.

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -511,6 +512,79 @@ export type Deployment = {
...
+        lastUpdate: latestDeployment,
+        versions: deploymentVersions,
+      };
+
+      // We're looking for just one deployment
+      if (deploymentId) {
+        if (deploymentId === id) {
+          // Return the deployment if found
+          return [deployment];
+        } else {
```
Siegrift, line 575: This is a bit awkward, why not just do it like this:

```
if (deploymentId === id) return [deployment]
deployments.push(deployment);
```

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -602,103 +629,33 @@ export async function listAirnodes(cloudProviders: readonly CloudProvider['type'
...
-    throw new Error(message);
+  if (goCloudDeploymentInfo.data.length === 0) {
+    spinner.stop();
+    throw new Error(`No deployment with id '${deploymentId}' found`);
   }
 
-  const { id, cloudProvider, region, airnodeAddress, stage, airnodeVersion, lastUpdate, versions } =
-    deployment as Deployment;
+  const deployment = goCloudDeploymentInfo.data[0];
+  const { id, region, airnodeAddress, stage, airnodeVersion, lastUpdate, versions } = deployment as Deployment;
```
Siegrift, line 658: Why is this not typed automatically? The go utils should provide inference for these things.

packages/airnode-deployer/src/infrastructure/index.ts
amarthadan: I tried to explain in the comments why this won't work. Imagine that you've specified a `deploymentId` you're searching for but such a deployment doesn't exist. With your snippet, you would still fill up the `deployments` array with all the deployments and you wouldn't be able to tell that such a deployment was not found.

packages/airnode-deployer/src/infrastructure/index.ts
amarthadan: I'm currently replacing it with a full `CloudProvider` type at some places so I'll see whether it makes sense.

packages/airnode-deployer/src/infrastructure/index.ts
Siegrift: But when you reach the return at the end, you know you haven't found it.

packages/airnode-deployer/src/infrastructure/index.ts
amarthadan: I'll do that as part of https://github.com/api3dao/airnode/issues/1449

packages/airnode-deployer/src/infrastructure/index.ts
amarthadan: Yes, and I would have to add another `if` statement there to see whether I'm looking for one specific deployment or not. And that's basically what I already have here.

Siegrift (approved):
👍 LGTM.

Issue #1318, opened 2022-07-06
Create an authorizer example


bbenligiray:
We can create an example in airnode-examples for implementing a custom authorizer and demonstrate its usage, similar to the relay security scheme example

dcroote:
For anyone coming across this issue, the `coingecko-cross-chain-authorizer` [airnode-examples integration](https://github.com/api3dao/airnode/tree/master/packages/airnode-examples/integrations/coingecko-cross-chain-authorizer), introduced in #1488 and planned to be released as part of Airnode v0.10, demonstrates how very basic same-chain and cross-chain authorizer contracts can be used together to authorize requests.

Leaving the task open as there is still an opportunity to demonstrate more complex authorizer contract logic.

dcroote:
Closing for the same reason as #1555: `RequesterAuthorizerWithAirnode` is unused and has been superceded. See https://github.com/api3dao/airnode/issues/1722 and https://api3workspace.slack.com/archives/C02AYRX8D89/p1680698758234279

dcroote:
Changing to "Closing as not planned"

PR #1488, opened 2022-10-04
Enable cross-chain authorizers

### LINKED ISSUE #1284: Cross-chain authorizers
Currently, the authorizer contracts specified in `chains.*.authorizers` are required to be located on the chain specified by `chains.*` (as in on Ethereum mainnet, only authorizer contracts that are deployed on mainnet can be used).

Especially for monetization functionality, we want to be able to specify authorizers on other chains. For example, say there is an authorizer on mainnet that authorizes requesters automatically when they pay USDC to the Airnode address. The Airnode is set up to have providers defined for both Rinkeby and mainnet. The authorizer on mainnet is only used on Rinkeby, while mainnet is closed to requests. This allows the Airnode operator to monetize their usage on Rinkeby.

This feature is required for ChainAPI, so I would have it in 0.9.0. Some issues with it:

- Currently, we have no way of specifying a chain with providers that doesn't accept any requests (as in the example above). In pre-alpha, when `authorizers` was defined to be `[]`, this meant everything is unauthorized on that chain. Currently, `[]` means everything is authorized, and one needs to remove the chain from `chains` to not accept any requests from it. However, in this case, we need to be able to specify providers without them being used to pick up requests, which is not supported with the current config format.
- We isolate interactions with providers (as in one Lambda invocation per provider). This flow requires two arbitrary providers to be used together. There are multiple ways of designing the Lambda invocation scheme for this and we need to decide on one (the solution below foresees no changes to the scheme, which is probably ideal since we want to wrap up v0).
- The config format needs to be updated to specify these cross-chain authorizers.

Here's a proposal (modifying https://github.com/api3dao/airnode/issues/1196 a bit):
```
"chains": [
    {
      "id": "3",
      "authorizers": [
        {
          "type": "crossChainRequesterAuthorizer",
          "chainType": "evm",
          "chainId": "1",
          "chainProvider": "${CP_SELF_HOSTED_MAINNET_URL}",
          "address": "0x1234..."
        }
      ]
    }
  ]
```

By embedding the entire chain information in the authorizer object, we don't need a chain object for Ethereum mainnet to exist. This means we don't need to worry about receiving requests on Ethereum mainnet. This provides flexibility in that we can come up with new authorizers that are not even on-chain conditions. So something like this:
```
"chains": [
    {
      "id": "3",
      "authorizers": [
        {
          "type": "http",
          "url": "https://some-authentication-api-that-is-practically-relay-metadata-but-much-nicer-because-it-lives-outside-of-ois.com"
        }
      ]
    }
  ]
```
Finally, this still allows provider-level redundancy to be specified, such as
```
"chains": [
    {
      "id": "3",
      "authorizers": [
        {
          "type": "crossChainRequesterAuthorizer",
          "chainType": "evm",
          "chainId": "1",
          "chainProvider": "${CP_SELF_HOSTED_MAINNET_URL}",
           "address": "0x1234..."
        },
        {
          "type": "crossChainRequesterAuthorizer",
          "chainType": "evm",
          "chainId": "1",
          "chainProvider": "${CP_INFURA_MAINNET_URL}",
           "address": "0x1234..."
        }
      ]
    }
  ]
```
We can also define same-chain authorizers as we did before
```
"chains": [
    {
      "id": "3",
      "authorizers": [
        {
          "type": "requesterAuthorizer",
          "address": "0x1234..."
        }
      ]
    }
  ]
```

dcroote:
Closes #1284.

The core logic is in [initialize-provider.ts](https://github.com/api3dao/airnode/pull/1488/files#diff-f6f66a98c3ef0c1afdb72171b67e11d8c3f410c314655c7e541284417532b035) and the coingecko-cross-chain-authorizer README has an explanation of how this user-runnable integration works.


packages/airnode-examples/contracts/coingecko-e2e/EverythingAuthorizer.sol
```diff
@@ -0,0 +1,20 @@
+//SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+import "@api3/airnode-protocol/contracts/authorizers/interfaces/IAuthorizerV0.sol";
+
+/* solhint-disable no-unused-vars */
```
Siegrift, line 6: You can do this:
```
contract EverythingAuthorizer is IAuthorizerV0 {
    function isAuthorizedV0(
        bytes32,
        address,
        bytes32,
        address,
        address
    ) external pure override returns (bool) {
        return true;
    }
}
```

packages/airnode-examples/integrations/coingecko-e2e/README.md
```diff
@@ -0,0 +1,7 @@
+# Coingecko-e2e example integration
```
Siegrift, line 1: I am fine with using the example in our e2e tests, but I would still want this to be runnable by users. People might want to use it to test this by using two different chains. This example would be available in via the examples CLI so it should be usable on its own.

packages/airnode-examples/package.json
```diff
@@ -15,6 +15,7 @@
     "create-airnode-config": "ts-node src/scripts/create-airnode-config.ts",
     "create-airnode-secrets": "ts-node src/scripts/create-airnode-secrets.ts",
     "create-aws-secrets": "ts-node src/scripts/create-aws-secrets.ts",
+    "deploy-authorizers": "ts-node src/scripts/deploy-authorizers.ts",
```
Siegrift, line 18: This is specific to the examples-e2e so it should be created inside the file of that integration. The `scripts` should be as generic as possible otherwise it will get messy.

packages/airnode-examples/test/example-files-are-up-to-date.test.ts
```diff
@@ -18,7 +18,10 @@ describe('Verifies that all config.example.json files are up to date', () => {
 
       if (!goCreateConfig.success) throw goCreateConfig.error;
 
-      const generatedConfigFile = readFileSync(path).toString();
+      const generatedConfigFile = readFileSync(path)
+        .toString()
+        // fix prettier newline disagreement with generated config for `coingecko-e2e`
```
Siegrift, line 23: I don't think you will need this.

packages/airnode-node/src/evm/handlers/initialize-provider.test.ts
```diff
@@ -228,4 +232,31 @@ describe('initializeProvider', () => {
...
+      '0x2': true,
+      '0x3': true,
+      '0x4': false,
+      '0x5': false,
+      '0x6': true,
+    };
+    const crossChainAuthorizations: AuthorizationByRequestId = {
+      '0x1': true,
+      '0x2': true,
+      '0x3': true,
```
Siegrift, line 248: I guess you want to have `false` here to cover all 4 cases.

packages/airnode-node/src/evm/handlers/initialize-provider.ts
```diff
@@ -109,6 +144,7 @@ export async function initializeProvider(
   const authAndTxCountPromises: readonly ParallelPromise[] = [
     fetchAuthorizations(state5),
     fetchTransactionCounts(state5),
+    fetchCrossChainAuthorizations(state5),
```
Siegrift, line 147: If `fetchCrossChainAuthorizations` is plural and fetches from multiple chains, I think `fetchAuthorizations` should be renamed to singular.

packages/airnode-operation/src/types.ts
```diff
@@ -111,7 +111,8 @@ export interface ConfigTemplate {
 }
 
 export interface ConfigAirnode {
-  readonly authorizers: { requesterEndpointAuthorizers: string[] };
+  // TODO
```
Siegrift, line 114: or not todo 😢

Siegrift (changes requested):
I like the idea of using the examples to test this, but I think the example should be a bit more polished and runnable by other users and real chains.

packages/airnode-examples/contracts/coingecko-e2e/EverythingAuthorizer.sol
dcroote: Great, thanks.

packages/airnode-examples/package.json
dcroote: Good call- I moved it.

packages/airnode-examples/test/example-files-are-up-to-date.test.ts
dcroote: Unfortunately, this is needed.

packages/airnode-node/src/evm/handlers/initialize-provider.test.ts
dcroote: Nice catch, thanks.

packages/airnode-examples/test/example-files-are-up-to-date.test.ts
Siegrift: Mhh, can you explain a bit more what is the issue?

packages/airnode-examples/integrations/coingecko-e2e/README.md
dcroote: I very much agree we should have an authorizer example integration (and in fact there is a task for this: #1318), but the contracts used in this test are contrived and unrelated to what an actual use case would involve. I think it is better to have this specifically for E2E testing of cross-chain authorizer functionality and build a separate integration that covers a real use case.

Siegrift: Actually, for me the big feature here is the cross chain authorization and not the Authorizer contracts themselves.

packages/airnode-examples/test/example-files-are-up-to-date.test.ts
dcroote: Without this (hack) you get the following test failure (notice the newline disagreement) when running this test for this new `coingecko-e2e` integration:

```ts
Verifies that all config.example.json files are up to date

    expect(received).toBe(expected) // Object.is equality

    - Expected  - 2
    + Received  + 6

    @@ -1,14 +1,18 @@
      {
        "chains": [
          {
            "maxConcurrency": 100,
            "authorizers": {
    -         "requesterEndpointAuthorizers": ["0xE2E0000000000000000000000000000000000000"],
    +         "requesterEndpointAuthorizers": [
    +           "0xE2E0000000000000000000000000000000000000"
    +         ],
              "crossChainRequesterAuthorizers": [
                {
    -             "requesterEndpointAuthorizers": ["0xE2E1111111111111111111111111111111111111"],
    +             "requesterEndpointAuthorizers": [
    +               "0xE2E1111111111111111111111111111111111111"
    +             ],
                  "chainType": "evm",
                  "chainId": "31337",
                  "contracts": {
                    "AirnodeRrp": "0x5FbDB2315678afecb367f032d93F642f64180aa3"
                  },

      26 |       writeFileSync(path, currentConfigFile);
      27 |
    > 28 |       expect(generatedConfigFile).toBe(currentConfigFile);
         |                                   ^
      29 |     });
      30 |   });
      31 | });

      at test/example-files-are-up-to-date.test.ts:28:35
      at fulfilled (test/example-files-are-up-to-date.test.ts:28:58)
```

Siegrift: Ahh, I see. Can you try fixing that by formatting the generated file? See: https://prettier.io/docs/en/api.html

packages/airnode-node/src/evm/handlers/initialize-provider.ts
dcroote: They're both plural as for each, authorizations are being fetched for an array of requests i.e. `Request<ApiCall>[]`

Siegrift: But `fetchAuthorizations` returns `data: AuthorizationByRequestId` and the `fetchCrossChainAuthorizations` returns `data: AuthorizationByRequestId[]`

packages/airnode-examples/test/example-files-are-up-to-date.test.ts
dcroote: Found the issue- the generated config didn't obey our prettier `printWidth` of 120. Your suggestion worked, thanks 👍

packages/airnode-node/src/evm/handlers/initialize-provider.ts
Siegrift: Yeah, that sounds a bit better to me. (I see your point though).

dcroote: 👍 Thanks for bringing it up and for the discussion, I think this will add clarity.

packages/airnode-operation/src/types.ts
dcroote: Done 😄

packages/airnode-examples/integrations/coingecko-e2e/README.md
dcroote: That's fair. This as a usable integration will be a decent departure from the other existing integrations (e.g. this will have multiple providers, multiple chains and different contracts deployed to each), but I'll work on it.

dcroote: Done!

packages/airnode-examples/integrations/coingecko-cross-chain-authorizer/create-config.ts
```diff
@@ -0,0 +1,236 @@
...
+  getAirnodeRrpAddress,
+  getChainId,
+} from '../config-utils';
+
+const createConfig = async (generateExampleFile: boolean): Promise<Config> => ({
+  chains: [
+    {
+      maxConcurrency: 100,
+      authorizers: {
+        requesterEndpointAuthorizers: ['0xE2E0000000000000000000000000000000000000'],
```
Siegrift, line 15: Nit: You can put them into secrets for easier implementation. But I know they are not secrets in practice so hardcoding makes sense as well.

Siegrift (approved):
👍 LGTM. Tested on localhost (same way as e2e tests), looks fine.

dcroote:
> Tested on localhost (same way as e2e tests), looks fine.

Thanks. I make some fixes to the cross-chain aspects of the integration and confirmed it's now working on Sepolia (same chain) with a polygon mumbai cross-chain authorizer (in addition to localhost E2E via CI).

packages/airnode-examples/integrations/coingecko-cross-chain-authorizer/create-config.ts
dcroote: Thanks- I considered this, but because canonically we do not store addresses like this as secrets I opted instead for the placeholder addresses that get replaced.

PR #1507, opened 2022-10-25
Add Docker image publishing to packaging scripts

amarthadan:
I added Docker image publishing (`docker push`) to our packaging scripts so it can be reused for publishing also official packages instead of using the command directly in CI.


Siegrift (approved):
👍 LGTM

PR #1508, opened 2022-10-25
Add CI job to enable/disable GH PR merging

amarthadan:
During the release flow, there's a time window when we need to "freeze" merging in order to release only what was tested during the testing of snapshot packages. I added a CI job whose sole purpose is just to check the GitHub secret and succeed or fail based on its value. I also added a command to the packaging container allowing us to set this secret to either `true` or `false`, effectively controlling whether the CI check will be successful and therefore whether the merge will be allowed.


package.json
```diff
@@ -56,6 +56,9 @@
     "docker:scripts:docker:build:local": "yarn docker:scripts:docker:build --npm-registry local --npm-tag local --docker-tag local",
     "docker:scripts:docker:publish": "docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -e DOCKERHUB_USERNAME -e DOCKERHUB_TOKEN api3/airnode-packaging:latest docker publish",
     "docker:scripts:docker:publish:latest": "yarn docker:scripts:docker:publish",
+    "docker:scripts:github": "docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -e GITHUB_TOKEN api3/airnode-packaging:latest github",
```
Siegrift, line 59: You need to set the GH token when running this command. It might be nice to mention this somewhere. Maybe root README?

amarthadan: There's a README in the `docker` directory that I'm forgetting to update. I'll do that.

Siegrift (approved):
👍 LGTM

PR #1513, opened 2022-10-27
Fix handling of quotation marks in secrets

amarthadan:
Fix #1510

PR #1514, opened 2022-10-28
[v0.9] Fix handling of quotation marks in secrets

amarthadan:
Backport of https://github.com/api3dao/airnode/pull/1513 to v0.9

### DIFF
--- a/.changeset/five-rivers-brush.md
+++ b/.changeset/five-rivers-brush.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-deployer': patch
+---
+
+Fix handling of quotation marks in secrets
--- a/packages/airnode-deployer/terraform/aws/modules/function/main.tf
+++ b/packages/airnode-deployer/terraform/aws/modules/function/main.tf
@@ -54,7 +54,7 @@ resource "aws_lambda_function" "lambda" {
   environment {
     variables = merge(
       var.environment_variables,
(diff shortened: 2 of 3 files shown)


Siegrift (approved):
👍 LGTM.

Maybe Prenaam could test this even before merging?

amarthadan:
> +1 LGTM.
> 
> Maybe Prenaam could test this even before merging?

@PrenaamD Would you be able to do that?

PrenaamD:
@amarthadan Thank you. I can use the opportunity to get familiar with running the deployer without using the image. But if it is urgent, I wouldn't hold anything while waiting for me to test this

PrenaamD:
@amarthadan I am testing this using the yarn cli deploy command and I am getting this error `Failed to list S3 buckets: CredentialsError: Missing credentials in config, if using AWS_CONFIG_FILE, set AWS_SDK_LOAD_CONFIG=1`
Not sure what I am missing

PrenaamD:
I got it to work by setting AWS credentials as environment variables instead of using the aws.env file as ChainAPI does.
The issue seems to be fixed. I used the quotes in the mnemonic field and the Airnode sent the heartbeats without any issue. No error in the `startCoordinator` lambda function logs.

Siegrift:
> Not sure what I am missing

We publish a dev-only version of all our docker images, they have the format `api3/airnode-deployer-<commit_hash>`, so you could just use this to deploy. The `yarn cli deploy` is only for development and we rarely use it.

> The issue seems to be fixed. I used the quotes in the mnemonic field and the Airnode sent the heartbeats without any issue.

Glad that it worked out.

Issue #1521, opened 2022-11-03
Fix incorrect merge of temp-v10 branch to master


Siegrift:
At the start of v0.10.0 we worked on a separate branch that we later merged to master. One of the issues we implemented was to [move from API-keys authentication to URL path based ones](https://github.com/api3dao/airnode/pull/1418). Once the temporary branch was merged to main, it was not merged correctly some of the previously removed code was reintroduced again https://github.com/api3dao/airnode/blob/master/packages/airnode-deployer/src/handlers/gcp/index.ts#L126.

We should check how the merge was performed, why it was incorrectly merged and check all of the features implemented in the temporary branch to see if they were merged correctly.

Siegrift:
Part of the reason why the broken merge occurred was my fault... I accidentally removed few commits while force pushing https://github.com/api3dao/airnode/pull/1434. I think I haven't pulled the latest changes from the origin before doing the rebase... The lesson for me is to use merge on shared branches and make the branch protected to avoid force pushes.

martinkolenic:
### Feature tests
**Heartbeat signature testing (and ID removal)**
Steps: Check for the presence of heartbeat signature using the coingecko-signed-data Airnode example, enabling heartbeat, deploying Airnode to AWS and then listening to it using Beeceptor. 

_Expected result: the Beeceptor endpoint should receive a request that includes a heartbeat signature; the heartbeat ID should not be present in the payload (nor the config)
Actual result: Matches expected result
Environment: Beeceptor, AWS cloud provider, Goerli testnet_

**HTTP Gateways not requiring API keys for authentication**
Steps: Deploy the coingecko-testable example integration, send request to the generated address via CURL without the API key parameter

Expected result: Receive a valid response
Actual result: Matches expected result
Environment: local OS Fedora 37 Beta, Linux 5.19.16, AWS cloud provider, CURL utility
0
### Standard tests
**Basic Airnode functionality test (Docker and npm package tests)**
Steps: Use airnode-deployer docker image to deploy Airnode to AWS (via Docker) and GCP (via the npm package), use the coingecko-signed-data example integration (altered to enable the basic gateway as well) and test making request via the provided script, and test the gateways via CURL.

_Expected result: Successful request fulfillment
Actual result: Matches expected result
Environment: AWS and GCP cloud providers, Görli testnet_

**Basic Airnode functionality test (Client Docker and image and npm package)**
Steps: Run Airnode locally on a local chain using the coingecko example and make a request.

[...]

PR #1497, opened 2022-10-13
v0.10 features so far

amarthadan:
Just merging the temporary v0.10 branch back to `master` so we can continue development there.

The last commit is newly added to fix broken deployer tests due to merge.

PR #1522, opened 2022-11-03
Fix broken temp-v0.10.0 branch merge

Siegrift:
Closes https://github.com/api3dao/airnode/issues/1521

## What happened

See https://github.com/api3dao/airnode/issues/1521#issuecomment-1301963469. Apart from the force push (which removed heartbeat signature, heartbeat data and shift from gateway API keys) two other features got lost (avoiding constants with side effects, heartbeat ID removal).

## The fix

I listed all PRs which were merged into `temp-v0.10.0` branch and sorted them. This can be seen [here](https://github.com/api3dao/airnode/pulls?q=type%3Apr+base%3Atemp-v0.10.0). The sorting is approximate, since GH doesn't have a sort function based on merged date.

I checked what was actually merged when we merged the `temp-v0.10.0` branch back to `master`. This can be seen [here](https://github.com/api3dao/airnode/pull/1497/commits). As mentioned everything merged after https://github.com/api3dao/airnode/pull/1434 was added to master.

To include all of the merged PRs before https://github.com/api3dao/airnode/pull/1434, I determined the order in which they were merged and cherry-picked them one by one. There were minimal merge conflicts since Airnode hasn't changed much.

Sorry, for the trouble I caused.


amarthadan (approved):
I would suggest testing all the "newly" added features again (after the merge), just to be on the safe side.

Siegrift:
> I would suggest testing all the "newly" added features again (after the merge), just to be on the safe side.

Yeah, we should do that. The heartbeat will be tested by @andreogle, but we should re-test the gateways.

PR #1526, opened 2022-11-10
Expand gateway request logging in AWS, GCP, and airnode-client

### LINKED ISSUE #1351: Add more extensive logging for gateway requests

dcroote:
Closes #1351. 

User story for the motivation and context of this PR:

As a [developer], I want to [access request processing logs on any of GCP, AWS, or the airnode-client without needing knowledge of Airnode request handling], so jthat [I can debug failing requests].


packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -117,6 +117,9 @@ export async function processHttpRequest(
     level: parsedConfig.nodeSettings.logLevel,
   });
 
+  // Unlike GCP, AWS automatically generates a unique request ID so no need to use one here
+  logger.info(`Received HTTP gateway request`);
+
```
amarthadan, line 122: Is this actually necessary? I think every request (lambda invocation) is logged in the AWS (and I think also in GCP) anyway. I don't think this adds anything. I would at least set it to `debug`.

packages/airnode-deployer/src/handlers/aws/index.ts
```diff
@@ -139,16 +144,20 @@ export async function processHttpRequest(
...
+  logger.debug(`HTTP gateway request passed request verification for endpoint ${endpointId}`);
 
   addMetadata({ 'Endpoint-ID': endpointId });
   const [err, result] = await handlers.processHttpRequest(parsedConfig, endpointId, parameters);
   if (err) {
     // Returning 500 because failure here means something went wrong internally with a valid request
+    logger.error(`HTTP gateway request processing error`);
     return { statusCode: 500, headers: originVerification.headers, body: JSON.stringify({ message: err.toString() }) };
   }
+  logger.info(`HTTP gateway request processed successfully`);
```
amarthadan, line 160: Again, I think this should be a `debug` log.

packages/airnode-deployer/src/handlers/gcp/index.ts
```diff
@@ -120,6 +121,10 @@ export async function processHttpRequest(req: Request, res: Response) {
     level: parsedConfig.nodeSettings.logLevel,
   });
 
+  // For logging, unlike blockchain requests, GCP gateway requests don't have requestIds
+  const requestId = randomUUID();
+  logger.info(`HTTP gateway request received, assigning it a request ID of: ${requestId}`);
```
amarthadan, line 126: I think you can use the logger's metadata to add the request ID instead of just adding it to the string.

amarthadan (commented):
I'm a bit skeptical about how much will this help with the debugging process. I'm not saying it's bad, it will definitely help but I don't know if it's enough. But it's a start :slightly_smiling_face:

amarthadan (commented):
I would use the log options with the `meta` field everywhere where you want to add some data to the log message. Similar to how it's done in [Airseeker](https://github.com/api3dao/airseeker/blob/main/src/update-data-feeds.ts#L285). Or do you feel like it's better this way?

dcroote:
> I would use the log options with the `meta` field everywhere where you want to add some data to the log message. Similar to how it's done in [Airseeker](https://github.com/api3dao/airseeker/blob/main/src/update-data-feeds.ts#L285). Or do you feel like it's better this way?

Ah sorry it wasn't ready for re-review just yet 😅 I pushed that commit in order for CI to build the images that I could then test. But I think the meta option will be the better route 👍

packages/airnode-deployer/src/handlers/aws/index.ts
dcroote: 👍 I've removed it for AWS given AWS automatically assigns a unique RequestID. For GCP, given the different layout of logs, I've changed it to `debug`.

packages/airnode-deployer/src/handlers/gcp/index.ts
dcroote: Yes this is much more elegant, thanks.

dcroote:
> I'm a bit skeptical about how much will this help with the debugging process.

Agreed, but it's a start and provides a lot more than what the user / Airnode operator gets access to currently. I actually found it useful in debugging encoded parameters for a signed gateway request 😅 

> I would use the log options with the meta field everywhere where you want to add some data to the log message

For the local client, a `requestId` gets generated and added as logging metadata (with `meta:` as you suggested) to distinguish requests, but I found this is not needed for AWS & GCP as each generates their own form of unique ID for an incoming request that groups logs nicely (`RequestId` and `executionId`, respectively).

amarthadan (approved):

PR #1537, opened 2022-11-15
Simplify heartbeat signature scheme

### LINKED ISSUE #1536: Simplify Heartbeat packing/signing
In Airnode v0.9 we introduced a new heartbeat schema, which involves a signature.

The signature mechanism itself is relatively well supported in languages other than JS/TS, but the binary packing method used as part of the signature process is not as well supported, particularly in Elixir, which is what ChainAPI uses.

Refer to [heartbeats.ts](https://github.com/api3dao/airnode/blob/55dd3c68bd075f998dae1148ee246298d536347e/packages/airnode-node/src/reporting/heartbeat.ts#L39), specifically:
```typescript
  return airnodeWallet.signMessage(
    ethers.utils.arrayify(
      ethers.utils.solidityKeccak256(['uint256', 'string'], [timestamp, JSON.stringify(heartbeatPayload)])
    )
  );
```
which should change to:
```typescript
  return airnodeWallet.signMessage(heartbeatPayload); // heartbeatPayload is a JSON string with heartbeatPayload as it's contents and a timestamp added
};
```

This removes the binary packing.
Ideally the JSON payload should be nested as a string to prevent encoding differences between languages causing verification to fail.

This is an urgent task.

Please check you're happy @andreogle

aquarat:
Closes #1536 

`ethers` packing is not well supported in the Elixir environment and even message signing has quirks associated with it that don't translate 1:1. This PR simplifies the heartbeat scheme by removing the packing component in favour of moving the timestamp into the JSON payload. This makes Airnode signatures easier to support in other languages.

The JSON is signed and sent as serialised JSON to prevent differences in JSON marshaling/unmarshaling across languages causing signatures to fail.


aquarat:
Currently waiting on feedback from @andreogle if he's happy with this scheme.

packages/airnode-node/src/reporting/heartbeat.ts
```diff
@@ -56,16 +42,18 @@ export async function reportHeartbeat(state: CoordinatorState): Promise<PendingL
   const httpGatewayUrl = getHttpGatewayUrl(config);
   const httpSignedDataGatewayUrl = getHttpSignedDataGatewayUrl(config);
 
-  const heartbeatPayload = {
+  const timestamp = Date.now();
+
+  const heartbeatPayload = JSON.stringify({
```
andreogle, line 47: I think it's worth adding a comment explaining why this is stringified. i.e. to avoid any minor differences that might occur when encoding/decoding. The payload/message needs to be exactly the same everywhere

packages/airnode-node/src/reporting/heartbeat.ts
```diff
@@ -56,16 +42,18 @@ export async function reportHeartbeat(state: CoordinatorState): Promise<PendingL
...
+    timestamp,
```
Siegrift, line 48: Nit: I'd suggest using the UNIX timestamp (in seconds) and not the JS timestamp which is in ms... But this is a breaking change so we might just not do it.

andreogle: Good point. Unix would be much better imo. And just to confirm, this in in UTC as well?

Why is it a breaking change?

Siegrift: Don't you rely on the heartbeat timestamp being in ms? (That's how it worked until now). Changing that to seconds would mean the old values would need to be migrated.

andreogle: ChainAPI hasn't been using the timestamp until now

packages/airnode-node/src/reporting/heartbeat.ts
amarthadan, line 47: Hmm. Maybe it would be wise to make sure that the payload content fields are always sorted. `JSON.stringify` doesn't do that. Maybe something like [`json-stable-stringify`](https://github.com/ljharb/json-stable-stringify] would be better. It's not like `JSON.stringify` will stringify it randomly, the fields will be in the order they are presented in when creating the object. But just to be on the safe side?

aquarat: If we serialise it like this we will always get a consistent result.
I have encountered a situation once before where the keys were serialised out of alphabetical order (I think they're serialised in the order they're set), but serialising a serialised payload gets rid of that issue and any other potential issues as you're comparing the signature with a byte array effectively. Adding an additional library to ensure serialisation ordering consistency seems unnecessary to me and introduces a new dependency.

aquarat: I've added the following comment:
```
  /*
  The heartbeat payload is serialised as JSON and then serialised again in JSON as the value of the payload field.
  The reason this is done is to avoid any inconsistencies between different JSON serialisation implementations.
   */
```

packages/airnode-node/src/reporting/heartbeat.ts
aquarat: I've made the change by using `Math.round(Date.now())` 👌

amarthadan (approved):

PR #1543, opened 2022-11-17
Remove outdated deployer CLI example

Siegrift:
Removes outdated example of a command that was removed as part of the 0.10.0 release.


packages/airnode-deployer/src/cli/index.ts
```diff
@@ -50,7 +50,6 @@ const cliExamples = [
   'list --cloud-providers gcp',
   'info 5bbcd317',
   'remove-with-receipt -r config/receipt.json',
-  'remove-with-deployment-details --airnode-address 0x6abEdc0A4d1A79eD62160396456c95C5607369D3 --stage dev --cloud-provider aws --region us-east-1',
```
dcroote, line 53: @amarthadan replaced `remove-with-deployment-details` with `remove <deployment-id>` as shown further down in the file. The v0.10 docs are also missing this change.
https://github.com/api3dao/airnode/blob/9cc4b493155a7091895285cdfb05eac1fa90e7a5/packages/airnode-deployer/src/cli/index.ts#L120-L129

dcroote (changes requested):

dcroote (approved):
LGTM 👍

PR #1548, opened 2022-11-23
Fix deployment ID generation

### LINKED ISSUE #1546: Deployment ID is incorrectly generated for removal and info commands when GCP project ID value is interpolated
The remove, list, and info commands generate the deployment IDs from config files stored in the cloud. They don't run the config files through the interpolation though. When one of the fields used for deployment ID generation (e.g. GCP project ID) requires the interpolation, the resulting deployment ID is incorrect.

We need to with each `config.json` file download also the `secret.env` file and run the interpolation before using the values for deployment ID generation.

Reported in #1545

amarthadan:
Fix #1546

Better error handling will be introduced in https://github.com/api3dao/airnode/issues/1449

### DIFF
--- a/.changeset/itchy-kings-pay.md
+++ b/.changeset/itchy-kings-pay.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-deployer': patch
+---
+
+Fix deployment ID generation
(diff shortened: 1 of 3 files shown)


Siegrift (approved):
👍 LGTM, but didn't test.

PR #1553, opened 2022-11-25
Fix e2e tests

Siegrift:
Fixes https://github.com/api3dao/airnode/issues/1551

See the issue thread for more details. For now, downgrade axios version until a new patch with a fix is released.

PR #1563, opened 2022-11-29
Bump axios

aquarat:


Siegrift:
Version 1.2.0 is broekn. See: https://github.com/api3dao/airnode/pull/1553

aquarat:
The latest axios supports data-URIs, hence this test. Thanks for the comment :)

PR #1564, opened 2022-11-30
fix: undefined error property in json logging

### LINKED ISSUE #1547: Unhandled error in logging when the API call fails
From #dev-support:

Screenshot from 2022-11-23 11-30-11

dcroote:
Closes #1547. Relatively simple fix as the bug appears to just have been a typo of copying the wrong line i.e. the code block several lines above this change has the correct conditional.

### DIFF
--- a/.changeset/strong-llamas-laugh.md
+++ b/.changeset/strong-llamas-laugh.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-utilities': patch
+---
+
+Fix logging undefined error property bug
--- a/packages/airnode-utilities/src/logging/logging.ts
+++ b/packages/airnode-utilities/src/logging/logging.ts
@@ -117,7 +117,7 @@ export function logFull(level: LogLevel, message: string, options: LogOptions | 
   }
 
   json(level, message, options);
-  if (level === 'ERROR' && (options as ErrorLogOptions).error!.stack!) {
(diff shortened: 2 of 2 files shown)


amarthadan (commented):
It does fix the unhandled error but since the error occurred, it means that the `logFull` function was at some point called with `level` set to `ERROR` and `options.error` being `null`. This way we will just ignore that call, won't we?

dcroote:
> It does fix the unhandled error but since the error occurred, it means that the `logFull` function was at some point called with `level` set to `ERROR` and `options.error` being `null`. This way we will just ignore that call, won't we?

There are some instances where we do log errors without an `error`, for example: https://github.com/api3dao/airnode/blob/74f33ea835d71f22152354ee8cff2e0747080c0b/packages/airnode-node/src/coordinator/calls/coordinated-execution.ts#L46

But that is somewhat beside the point here as the changed code block corresponds to providing an **additional** log when there is a `stack` available. If you look one line above the changed line, the error will still be logged in the case where there isn't an `error`: https://github.com/api3dao/airnode/blob/3034ad0b10e7cd9c464132bdcce50743029a31f5/packages/airnode-utilities/src/logging/logging.ts#L119-L121

amarthadan:
Oh, right, must have missed that, sorry. Carry on :slightly_smiling_face:

amarthadan (approved):

Issue #1581, opened 2022-12-01
Investigate coordinator failure for faulty provider


Siegrift:
https://api3workspace.slack.com/archives/C02AYRX8D89/p1669812538505089?thread_ts=1669715564.101439&cid=C02AYRX8D89

Siegrift:
The reason is the `JSON.parse` in https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/workers/cloud-platforms/aws.ts#L23

I've created PR's that fix the issue:
- https://github.com/api3dao/airnode/pull/1580
- https://github.com/api3dao/airnode/pull/1582

martinkolenic:
Hi @Siegrift, I'm not sure how to test this specific case, could you please advise? I understand I'd need a faulty provider but where to find/mock one?

amarthadan:
@metobom You can use the v0.9.2 snapshot packages/Docker images to test it.
https://github.com/api3dao/airnode/issues/1588

martinkolenic:
Hi @Siegrift, I've tested the issue.

Steps: Deploy Airnode with a faulty provider URL (a Beeceptor URL with a 1000s delay before a response)
Expected result: Coordinator keeps functioning but displays an error message
Actual result: Matches expected result

![Image](https://user-images.githubusercontent.com/884979/206211747-eae3de6a-3987-43cd-ade8-c87bddb23901.png)

Siegrift:
> @metobom is periodically trying to reproduce the issue using 0.9.1 and the problematic provider. When he does, he will check if 0.9.2 still causes the same issue. Maybe skip testing this one.

@bbenligiray Just to clarify, I asked Martin to test this on master and leave the 0.9.2 for @metobom. Thanks @martinkolenic, this can be moved to DONE.

PR #1582, opened 2022-12-01
Fix failed lambda execution crashing coordinator

Siegrift:
Closes https://github.com/api3dao/airnode/issues/1581

I wanted to write a test for this, but this is reproducible only on AWS...

### DIFF
--- a/packages/airnode-node/src/workers/cloud-platforms/aws.ts
+++ b/packages/airnode-node/src/workers/cloud-platforms/aws.ts
@@ -1,9 +1,11 @@
+import { logger } from '@api3/airnode-utilities';
+import { goSync } from '@api3/promise-utils';
 import AWS from 'aws-sdk';
 import { WorkerParameters, WorkerResponse } from '../../types';
 
 export function spawn(params: WorkerParameters): Promise<WorkerResponse> {
   // lambda.invoke is synchronous so we need to wrap this in a promise
-  return new Promise((resolve, reject) => {
+  return new Promise((spawnResolve, spawnReject) => {
     // Uses the current region by default
     const lambda = new AWS.Lambda();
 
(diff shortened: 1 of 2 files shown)


packages/airnode-node/src/workers/cloud-platforms/aws.ts
```diff
@@ -15,12 +17,32 @@ export function spawn(params: WorkerParameters): Promise<WorkerResponse> {
...
-      resolve(JSON.parse(JSON.parse(data.Payload as string).body) as WorkerResponse);
+      if (err) return reject(err);
+
+      if (data.FunctionError) return reject(data.Payload);
+
+      const goPayload = goSync(() => JSON.parse(data.Payload?.toString() ?? ''));
+      if (!goPayload.success) return reject(goPayload.error);
+
+      const goBody = goSync(() => JSON.parse(goPayload.data.body));
+      if (!goBody.success) return reject(goBody.error);
```
amarthadan, line 43: Does it make sense to provide a default for when the `Payload` is not present that will immediately fail during parsing? Isn't checking for `Payload` presence explicitly better?

amarthadan (approved):
Small comment, otherwise LGTM :+1:

packages/airnode-node/src/workers/cloud-platforms/aws.ts
Siegrift: I could do `JSON.parse(data.Payload!.toString())` but that suggests that Payload is defined, for which there is no reason. That's why I opted for `?` but `JSON.parse` requires a string argument, so I defaulted to `?? ''` empty string since it will fail on the next validation anyway.

amarthadan: I understand you're reasoning. I was just suggesting handling the `Payload` absence explicitly.
```suggestion
      if (!data.Payload) return reject(new Error('Missing payload');

      const goPayload = goSync(() => JSON.parse(data.Payload!.toString()));
      if (!goPayload.success) return reject(goPayload.error);

      const goBody = goSync(() => JSON.parse(goPayload.data.body));
      if (!goBody.success) return reject(goBody.error);
```

Siegrift: Sure, I can do that.

PR #1586, opened 2022-12-02
Skip api call

metobom:
Added a feature that lets skipping API calls. Instead of API call it uses `output`  of `preProcessingSpecifications`, `postProcessingSpecifications`, or both. To skip API call, `operation` must be undefined, `fixedOperationParameters` must be an empty array and one of `preProcessingSpecifications` or `postProcessingSpecifications` must be defined and not be an empty array.

### DIFF
--- a/packages/airnode-node/src/api/index.test.ts
+++ b/packages/airnode-node/src/api/index.test.ts
@@ -414,6 +414,146 @@ describe('callApi', () => {
       );
     });
   });
+  describe('skip api call', () => {
+    const createEncodedValue = (value: ethers.BigNumber, times = 100_000) =>
+      `0x${value.mul(times).toHexString().substring(2).padStart(64, '0')}`;
+
+    it('skip api call with preProcessingSpecifications', async () => {
+      const config = fixtures.buildConfig();
+      const parameters = { _type: 'int256', _path: 'result', parameter1: '25' };
+      const aggregatedApiCall = fixtures.buildAggregatedRegularApiCall({ parameters });
+      const preProcessingSpecifications = [
+        {
+          environment: 'Node 14' as const,
+          value: 'const output = { result: (parseInt(input.parameter1) * 2).toString() }',
+          timeoutMs: 5000,
+        },
+      ];
+
+      config.ois[0].endpoints[0].operation = undefined;
+      config.ois[0].endpoints[0].fixedOperationParameters = [];
+      config.ois[0].endpoints[0].preProcessingSpecifications = preProcessingSpecifications;
+
+      const [logs, res] = await callApi({
+        type: 'regular',
+        config,
+        aggregatedApiCall,
+      });
+
+      expect(logs).toEqual([]);
+      expect(res).toEqual({
+        success: true,
+        data: {
+          encodedValue: createEncodedValue(ethers.BigNumber.from(25 * 2)),
+          signature:
+            '0x148d65210b201c3ddd6f8e08cfc29032e9f3c781919eeb80b950dbb984ebbbee289105ea8b591ace166d04b65e334801ea0177bb431775db2b4268d55fa4e2a81c',
+        },
+      });
+    });
+
+    it('skip api call with postProcessingSpecifications', async () => {
+      const config = fixtures.buildConfig();
+      const parameters = { _type: 'int256', _path: 'result', parameter1: '25' };
+      const aggregatedApiCall = fixtures.buildAggregatedRegularApiCall({ parameters });
+      const postProcessingSpecifications = [
+        {
+          environment: 'Node 14' as const,
+          value: 'const output = { result: (parseInt(input.parameter1) * 2).toString() }',
+          timeoutMs: 5000,
+        },
+      ];
+
+      config.ois[0].endpoints[0].operation = undefined;
+      config.ois[0].endpoints[0].fixedOperationParameters = [];
+      config.ois[0].endpoints[0].postProcessingSpecifications = postProcessingSpecifications;
+
+      const [logs, res] = await callApi({
+        type: 'regular',
+        config,
+        aggregatedApiCall,
+      });
+
+      expect(logs).toEqual([]);
+      expect(res).toEqual({
+        success: true,
+        data: {
+          encodedValue: createEncodedValue(ethers.BigNumber.from(25 * 2)),
+          signature:
+            '0x148d65210b201c3ddd6f8e08cfc29032e9f3c781919eeb80b950dbb984ebbbee289105ea8b591ace166d04b65e334801ea0177bb431775db2b4268d55fa4e2a81c',
+        },
+      });
+    });
+
+    it('skip api call with both preProcessingSpecifications and postProcessingSpecifications', async () => {
+      const config = fixtures.buildConfig();
+      const parameters = { _type: 'int256', _path: 'result', parameter1: '25' };
+      const aggregatedApiCall = fixtures.buildAggregatedRegularApiCall({ parameters });
+      const preProcessingSpecifications = [
+        {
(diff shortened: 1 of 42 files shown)


packages/airnode-node/src/api/index.ts
```diff
@@ -287,11 +288,33 @@ export async function callApi(payload: ApiCallPayload): Promise<LogsData<ApiCall
   if (verificationResult) return verificationResult;
 
   const processedPayload = await preProcessApiSpecifications(payload);
-
-  const [logs, response] = await performApiCall(processedPayload);
-  if (isPerformApiCallFailure(response)) {
-    return [logs, response];
+  const ois = payload.config.ois.find((o: OIS) => o.title === payload.aggregatedApiCall.oisTitle) as OIS;
+  const endpoint = ois.endpoints.find((e: Endpoint) => e.name === payload.aggregatedApiCall.endpointName) as Endpoint;
```
andreogle, line 292: Doesn't this achieve the same thing without the explicit cast?

```suggestion
  const endpoint = ois.endpoints.find((e: Endpoint) => e.name === payload.aggregatedApiCall.endpointName)!;
```

packages/airnode-node/src/api/index.ts
```diff
@@ -287,11 +288,33 @@ export async function callApi(payload: ApiCallPayload): Promise<LogsData<ApiCall
...
+
+  if (!endpoint.operation && endpoint.fixedOperationParameters.length === 0) {
```
andreogle, line 294: This is pretty difficult to follow. Maybe it could be refactored/shortened a bit? Also adjusted the message grammer a little bit. Might also be worth adding some information about the specific invalid endpoint for easier debugging (e.g. the name)

```ts

if (!endpoint.operation && isEmpty(endpoint.fixedOperationParameters)) {
  if (isEmpty(endpoint.preProcessingSpecifications) && isEmpty(endpoint.postProcessingSpecifications) {
    const message = `Failed to skip API call. Ensure at least one of 'preProcessingSpecifications' or 'postProcessingSpecifications' is defined and is not an empty array.`;
    const log = logger.pend('ERROR', message);
    return [log, { success: false, errorMessage: message }];
  }
  return processSuccessfulApiCall(payload, { data: processedPayload.aggregatedApiCall.parameters });
}

const [logs, response] = await performApiCall(processedPayload);
if (isPerformApiCallFailure(response)) {
  return [logs, response];
}
return processSuccessfulApiCall(payload, response);
```

andreogle: It's probably also worth adding a comment explaining what's going on here too, since it's not obvious

packages/airnode-node/src/api/index.ts
```diff
@@ -287,11 +288,33 @@ export async function callApi(payload: ApiCallPayload): Promise<LogsData<ApiCall
...
+        [
+          logger.pend(
+            'ERROR',
+            `Failed to skip API call. Make sure one, or both 'preProcessingSpecifications', 'postProcessingSpecifications' exists and not empty array.`
+          ),
+        ],
+        {
+          success: false,
+          errorMessage: `Failed to skip API call. Make sure one, or both 'preProcessingSpecifications', 'postProcessingSpecifications' exists and not empty array.`,
+        } as SkipApiCallError,
```
andreogle, line 309: What's the reason for casting to this specific interface? The function signature specifies that this returns `Promise<LogsData<ApiCallResponse>>` so it's kinda confusing.

packages/airnode-node/src/api/index.ts
```diff
@@ -287,11 +288,33 @@ export async function callApi(payload: ApiCallPayload): Promise<LogsData<ApiCall
...
+      ];
+    }
+    return processSuccessfulApiCall(payload, { data: processedPayload.aggregatedApiCall.parameters });
+  } else {
```
andreogle, line 313: No need for the `else` since the `if` blocks above already return early in all of the paths

packages/airnode-node/src/types.ts
```diff
@@ -246,6 +246,11 @@ export interface ApiCallErrorResponse {
   errorMessage: string;
 }
 
+export interface SkipApiCallError {
```
andreogle, line 249: This is a duplicate interface of `ApiCallErrorResponse` above

packages/airnode-node/src/api/index.test.ts
```diff
@@ -414,6 +414,146 @@ describe('callApi', () => {
...
+        success: true,
+        data: {
+          encodedValue: createEncodedValue(ethers.BigNumber.from(25 * 2 + 5)),
+          signature:
+            '0x9796e8ba07c517a43b2ca1416b9c975541345c5b7bf73613738ed84216c4573d1656cbc385acb9111b7077fba65418b59cfe3f59ee81109523664bbc25efcfd71b',
+        },
+      });
+    });
+
+    it('fail when both preProcessingSpecifications and postProcessingSpecifications are empty or undefined', async () => {
```
andreogle, line 528:
```suggestion
    it('fails when both preProcessingSpecifications and postProcessingSpecifications are empty or undefined', async () => {
```

andreogle: (This test is also only testing empty arrays, while the description also says "undefined")

packages/airnode-node/src/api/index.ts
```diff
@@ -287,11 +288,33 @@ export async function callApi(payload: ApiCallPayload): Promise<LogsData<ApiCall
...
+            `Failed to skip API call. Make sure one, or both 'preProcessingSpecifications', 'postProcessingSpecifications' exists and not empty array.`
+          ),
+        ],
+        {
+          success: false,
+          errorMessage: `Failed to skip API call. Make sure one, or both 'preProcessingSpecifications', 'postProcessingSpecifications' exists and not empty array.`,
+        } as SkipApiCallError,
+      ];
+    }
+    return processSuccessfulApiCall(payload, { data: processedPayload.aggregatedApiCall.parameters });
```
andreogle, line 312: Could we add a comment here too explaining why this returns the aggregatedApiCall parameters as the result?

packages/airnode-node/src/api/index.test.ts
```diff
@@ -414,6 +414,146 @@ describe('callApi', () => {
...
+        success: true,
+        data: {
+          encodedValue: createEncodedValue(ethers.BigNumber.from(25 * 2)),
+          signature:
+            '0x148d65210b201c3ddd6f8e08cfc29032e9f3c781919eeb80b950dbb984ebbbee289105ea8b591ace166d04b65e334801ea0177bb431775db2b4268d55fa4e2a81c',
+        },
+      });
+    });
+
+    it('skip api call with postProcessingSpecifications', async () => {
```
andreogle, line 454:
```suggestion
    it('skips the API call with postProcessingSpecifications and no operation', async () => {
```

packages/airnode-node/src/api/index.test.ts
```diff
@@ -414,6 +414,146 @@ describe('callApi', () => {
       );
     });
   });
+  describe('skip api call', () => {
+    const createEncodedValue = (value: ethers.BigNumber, times = 100_000) =>
+      `0x${value.mul(times).toHexString().substring(2).padStart(64, '0')}`;
+
+    it('skip api call with preProcessingSpecifications', async () => {
```
andreogle, line 421:
```suggestion
    it('skips the API call with preProcessingSpecifications and no operation', async () => {
```

andreogle (changes requested):

andreogle (approved):
Looks great 👌🏻 Nothing more from my end

amarthadan (approved):
LGTM :+1:

Issue #1567, opened 2022-12-01
The deploy:verify-local script in airnode-protocol only works for undeterministic deployments


bbenligiray:
The `deploy:verify-local`  npm script in `airnode-protocol` only work for undeterministic deployments. This is actually fine, because deterministic deployments all have the same address, and the address being the same is the verification (that they have the correct bytecodes and constructor arguments). However, it would still be nice if it worked for all deployments so that when it's added to CI one doesn't need to make exceptions for deterministic deployments, which may be cumbersome to detect.

Ashar2shahid:
PR https://github.com/api3dao/airnode/pull/1590

martinkolenic:
Hi @bbenligiray, is this testable from an end-user perspective? Thanks.

PR #1590, opened 2022-12-07
verify deterministic deployment

Ashar2shahid:
addresses #1567

### DIFF
--- a/packages/airnode-protocol/scripts/verify-local.ts
+++ b/packages/airnode-protocol/scripts/verify-local.ts
@@ -18,6 +18,8 @@ const verifyBytecode = (deployedData: string, localData: string, contractName: s
     logger.log(`✅  ${contractName} deterministic deployment on ${hre.network.name} matches the local build!`);
     return true;
   }
+  // The compiler appends by default the IPFS hash of the metadata file to the end of the bytecode (the last 106 char)
+  // for reference: https://docs.soliditylang.org/en/v0.8.17/metadata.html#encoding-of-the-metadata-hash-in-the-bytecode
   if (deployedData.slice(0, -106) === localData.slice(0, -106)) {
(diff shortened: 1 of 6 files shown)


packages/airnode-protocol/scripts/verify-local.ts
```diff
@@ -9,14 +9,44 @@ import { logger } from '@api3/airnode-utilities';
...
     const creationTx = await hre.ethers.provider.getTransaction(deployment.transactionHash);
+    const deterministicDeploymentAddress = hre.ethers.utils.getCreate2Address(
+      '0x4e59b44847b379578588920ca78fbf26c0b4956c', // default create2 factory address in hardhat
+      hre.ethers.constants.HashZero,
+      hre.ethers.utils.keccak256(artifact.bytecode)
+    );
+
     // Verify that the creation tx hash belongs to the address
-    assert(creationTx.creates === deployment.address);
+    assert(creationTx.creates === deployment.address || deterministicDeploymentAddress);
```
bbenligiray, line 49: I think you intended `creationTx.creates === deployment.address || creationTx.creates === deterministicDeploymentAddress` here, yet this still accidentally creates the same outcome. `creationTx.creates` returns `null` when the deployment is deterministic, so `||` falls back to `deterministicDeploymentAddress`, which is treated as `true` because it's defined (though there is no case where it is not defined so this feels off).
I'll push a refactor.

bbenligiray (approved):

bbenligiray:
Confirmed that it perfectly verifies all current deployments. I created an issue for CI https://github.com/api3dao/airnode/issues/1594

packages/airnode-protocol/scripts/verify-local.ts
Ashar2shahid: right, the assertion will always be true in this case. I guess the correct assertion would be ```creationTx.creates === deployment.address || deterministicDeploymentAddress === deployment.address```

PR #1591, opened 2022-12-08
Unpin axios after v1.2.2 patch release

dcroote:
Reverts #1553 now that the patch release for the broken v1.2.0 is out


dcroote:
The `1.2.2` release works so this is ready for review & merge

package.json
```diff
@@ -113,7 +113,7 @@
     "@typescript-eslint/eslint-plugin": "^5.41.0",
     "@typescript-eslint/parser": "^5.41.0",
     "@vercel/ncc": "^0.34.0",
-    "axios": "1.1.3",
+    "axios": "^1.1.3",
```
Siegrift, line 116: Why not update it to latest version?

Siegrift (commented):
Why is yarn.lock not changed?

dcroote:
> Why is yarn.lock not changed?

My intention was to only revert #1553 and let Renovate take care of the update, but I went ahead and bumped it in the latest commit 👍

Siegrift (approved):
> My intention was to only revert https://github.com/api3dao/airnode/pull/1553 and let Renovate take care of the update, but I went ahead and bumped it in the latest commit 👍

I was just surprised, but it's better to update this to a working version explicitly.

👍 LGTM

PR #1593, opened 2022-12-09
Add missing GH token env variable

amarthadan:
When doing the v0.9.2 release I run into an issue that I forgot to pass the GitHub token environment variable to the Docker command running the build process.

The GitHub token is there to tag the commit after the publishing and push that tag to GitHub.


aquarat (approved):
👌 
Pretty cool that this is in production now 🚀 😇

PR #1599, opened 2022-12-21
Deployer missing files

### LINKED ISSUE #1449: Gracefully handle missing files of the deployed Airnode

vponline:
Closes #1449


packages/airnode-deployer/src/utils/logger.ts
```diff
@@ -36,15 +36,25 @@ export function fail(text: string) {
...
+  if (currentOra.isSpinning) {
+    currentOra.clear();
+    currentOra.frame();
+  }
   oraInstance().info(text);
 }
 
 export function debug(text: string) {
-  if (debugModeFlag) oraInstance().info(text);
+  if (debugModeFlag) info(text);
```
vponline, line 57: These changes are to allow printing a line to the console while the `ora` spinner is still spinning. WIthout these the spinner will overwrite any logs.

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -460,18 +460,43 @@ async function fetchDeployments(cloudProviderType: CloudProvider['type'], deploy
...
+        (requiredFileName) => !latestDepolymentFileNames.includes(requiredFileName)
+      );
+      if (missingRequiredFiles.length) {
+        logger.warn(
+          `Airnode '${airnodeAddress}' with stage '${stage}' is missing files: ${missingRequiredFiles.join(
+            ', '
+          )}. Deployer commands may fail and manual removal may be necessary.`
+        );
+      }
+
```
vponline, line 476: This will detect any of the three missing files, but note that if the missing file is either `secrets.env` or `config.json`, it is not possible to derive the deploymentId so the commands will fail with an error:

`No deployment with ID 'aws...' found`

image

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -460,18 +460,46 @@ async function fetchDeployments(cloudProviderType: CloudProvider['type'], deploy
       }
 
       const latestDeployment = Object.keys(stageDirectory.children).sort().reverse()[0];
+      // S
```
acenolaza, line 463: Should this be removed?

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -460,18 +460,46 @@ async function fetchDeployments(cloudProviderType: CloudProvider['type'], deploy
...
+      const requiredFileNames = ['config.json', 'secrets.env', 'default.tfstate'];
+      const missingRequiredFiles = requiredFileNames.filter(
+        (requiredFileName) => !latestDepolymentFileNames.includes(requiredFileName)
+      );
+      if (missingRequiredFiles.length) {
+        logger.warn(
+          `Airnode '${airnodeAddress}' with stage '${stage}' is missing files: ${missingRequiredFiles.join(
+            ', '
+          )}. Deployer commands may fail and manual removal may be necessary.`
+        );
```
acenolaza, line 477: Shouldn't we `continue` and skip this loop if this happens?

vponline: Good point :smile:

amarthadan (changes requested):
Nice start but I think we need to handle this more thoroughly.

There are currently 4 deployer commands (and more will come in the future):
* deploy - Deploy is not using `fetchDeployments` so it needs to be handled separately there. For deploy to work (assuming previous deployments) we need all three files (`config.json`, `secrets.env` and `default.tfstate`) to be present. The deployment won't work without them so I would just fail with an error message if one of them is missing.
* remove - Remove is using `fetchDeployments` because it needs to find the deployment based on the deployment ID. Again, we need all three files for deployment removal to work. And again, we can just fail if some of them are missing.
* list - List doesn't really care about the `default.tfstate` file, just the `config.json` and `secrets.env`. Here, it makes to skip the invalid ones (with a message) but still print those that are fine.
* info - Again, doesn't really care about the `default.tfstate`. It's concerned only with one deployment and its historical records. But without that deployment having all the files we can't really say whether it's there just broken or is not there at all :thinking:

My thinking here is that before running any actual functionality of these commands, we should do a sanity check on all the deployments in the bucket and print warnings accordingly. And then follow with whatever the command is supposed to do and react accordingly depending on the command (e.g. fail). WDYT?

Another thing is that maybe we should rethink making deployment ID from data that is in the files. Or rather moving that data into the directory structure of the bucket. That way we would be able to distinguish between a deployment missing and being broken. But that's a different topic.

vponline:
Thanks for reviewing. I am currently finishing up 2 tasks for operations and will get back to this one after.

vponline:
> deploy - Deploy is not using fetchDeployments so it needs to be handled separately there. 

This case has been handled now and is ok since we're only looking for a single deployment.

> remove - Remove is using fetchDeployments because it needs to find the deployment based on the deployment ID. Again, we need all three files for deployment removal to work. And again, we can just fail if some of them are missing.

This case is proving to be harder to implement since we can't be sure of the deploymentId when iterating over the bucket deployments. We can print the missing files in a warning (just like list and info), but we can't fail straight after finding missing files for any deployment. I think this would require implementing a differently created `deploymentId` which might be worth the effort, though. :thinking:

> Another thing is that maybe we should rethink making deployment ID from data that is in the files. Or rather moving that data into the directory structure of the bucket. That way we would be able to distinguish between a deployment missing and being broken. But that's a different topic.

Yes definitely this seems to be required to handle either missing or broken deployment checks properly.

packages/airnode-deployer/src/utils/infrastructure.ts
```diff
@@ -163,3 +167,38 @@ export const deploymentComparator = (a: Deployment, b: Deployment) => {
 
   return compareVersions(a.airnodeVersion, b.airnodeVersion);
 };
+
+export const checkBucketMissingFiles = (
```
amarthadan, line 171: Let's call it `getMissingBucketFiles` as it returns something rather than just checking and potentially throwing an exception.

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -461,18 +478,38 @@ async function fetchDeployments(cloudProviderType: CloudProvider['type'], deploy
...
-      const config = JSON.parse(
-        await cloudProviderLib[cloudProviderType].getFileFromBucket(bucket.name, bucketConfigPath)
+      const goGetConfigFileFromBucket = await go(() =>
+        cloudProviderLib[cloudProviderType].getFileFromBucket(bucket.name, bucketConfigPath)
       );
+      if (!goGetConfigFileFromBucket.success) {
+        logger.warn(`Failed to fetch configuration file. Error: ${goGetConfigFileFromBucket.error.message} Skipping.`);
+        continue;
+      }
+      const config = JSON.parse(goGetConfigFileFromBucket.data);
```
amarthadan, line 501: We should also wrap `JSON.parse` with `goSync` as it can throw as well.

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -333,6 +334,20 @@ export const deployAirnode = async (config: Config, configPath: string, secretsP
...
+        throw new Error(
+          `Can't update an Airnode with missing files: ${bucketMissingFiles[airnodeAddress][stage].join(
+            ', '
+          )}. Deployer commands may fail and manual removal may be necessary.`
+        );
+      }
+
       const latestDeployment = Object.keys(stageDirectory.children).sort().reverse()[0];
       const bucketConfigPath = `${bucketStagePath}/${latestDeployment}/config.json`;
       logger.debug(`Fetching configuration file '${bucketConfigPath}'`);
```
amarthadan, line 353: Can't select the correct line, it's meant to be the line below this one :smile: 

Even though the file will be present we should still wrap the call in `go` as it can throw for any other reason.
The same goes for JSON parsing.

amarthadan (changes requested):
I'm fine with this solution for now. Later we should rethink the deployment ID generation though. But I don't want to block the 0.10 release even more and it's good enough.

Just some minor issues.

vponline:
Thanks! The final changes should be covered now.

amarthadan (approved):

PR #1600, opened 2023-01-05
Add --cache property to prettier and eslint

### LINKED ISSUE #1595: Add "--cache" option to prettier
Prettier introduced a new option that greatly speeds up formatting. See: https://prettier.io/docs/en/cli.html#--cache...

This is not really an Airnode issue in particular and can be done for all our repos. I've tried to enable the option for the docs repo, where prettier is taking a long time to complete.

dcroote:
Closes #1595


Siegrift:
Btw. eslint also supports `--cache` so we might as well add it there (do not forget to add `.eslintcache` to gitignore)

Siegrift (approved):
👍 LGTM, but add it for ESlint as well.

dcroote:
> Btw. eslint also supports `--cache` so we might as well add it there (do not forget to add `.eslintcache` to gitignore)

Good call- `yarn lint:eslint` went from 17s down to 1.5s

PR #1603, opened 2023-01-08
Add fetch-files command to deployer

### LINKED ISSUE #1405: Allow users to fetch deployment files from the cloud provider

amarthadan:
Close #1405

### DIFF
--- a/packages/airnode-deployer/src/cli/index.ts
+++ b/packages/airnode-deployer/src/cli/index.ts
@@ -9,7 +9,7 @@ import { deploy, removeWithReceipt } from './commands';
 import * as logger from '../utils/logger';
 import { longArguments } from '../utils/cli';
 import { MultiMessageError } from '../utils/infrastructure';
-import { deploymentInfo, listAirnodes, removeAirnode } from '../infrastructure';
+import { deploymentInfo, fetchFiles, listAirnodes, removeAirnode } from '../infrastructure';
 
 function drawHeader() {
   loggerUtils.log(
@@ -49,6 +49,7 @@ const cliExamples = [
   'deploy -c config/config.json -s config/secrets.env -r config/receipt.json',
   'list --cloud-providers gcp',
   'info aws808e2a22',
+  'fetch-files aws808e2a22',
   'remove-with-receipt -r config/receipt.json',
   'remove aws808e2a22',
 ];
@@ -124,7 +125,6 @@ yargs(hideBin(process.argv))
       yargs.positional('deployment-id', {
         description: `ID of the deployment (from 'list' command)`,
         type: 'string',
-        demandOption: true,
       });
     },
     async (args) => {
@@ -167,7 +167,6 @@ yargs(hideBin(process.argv))
       yargs.positional('deployment-id', {
         description: `ID of the deployment (from 'list' command)`,
         type: 'string',
-        demandOption: true,
       });
     },
     async (args) => {
@@ -178,6 +177,40 @@ yargs(hideBin(process.argv))
       const goDeploymentInfo = await go(() => deploymentInfo(args.deploymentId as string));
       if (!goDeploymentInfo.success) {
         logger.fail(goDeploymentInfo.error.message);
+        // eslint-disable-next-line functional/immutable-data
+        process.exitCode = 1;
+      }
+    }
+  )
+  .command(
+    'fetch-files <deployment-id> [version-id]',
+    'Fetch deployment files for the deployed Airnode',
+    (yargs) => {
+      yargs.positional('deployment-id', {
+        description: `ID of the deployment (from 'list' command)`,
+        type: 'string',
+      });
+      yargs.positional('version-id', {
+        description: `ID of the deployment version (from 'info' command)`,
+        type: 'string',
+      });
+      yargs.option('output-dir', {
+        alias: 'o',
+        description: 'Where to store fetched files',
+        default: 'config/',
+        type: 'string',
+      });
+    },
+    async (args) => {
+      logger.debugMode(args.debug as boolean);
+      logger.debug(`Running command ${args._[0]} with arguments ${longArguments(args)}`);
+
+      // Looks like due to the bug in yargs (https://github.com/yargs/yargs/issues/1649) we need to specify the types explicitely
+      const goFetchFiles = await go(() =>
+        fetchFiles(args.deploymentId as string, args.outputDir as string, args.versionId as string | undefined)
+      );
+      if (!goFetchFiles.success) {
+        logger.fail(goFetchFiles.error.message);
         // eslint-disable-next-line functional/immutable-data
         process.exitCode = 1;
       }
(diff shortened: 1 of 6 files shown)


packages/airnode-deployer/src/cli/index.ts
```diff
@@ -183,6 +182,40 @@ yargs(hideBin(process.argv))
...
+        type: 'string',
+      });
+      yargs.positional('version-id', {
+        description: `ID of the deployment version (from 'info' command)`,
+        type: 'string',
+      });
+      yargs.option('output-dir', {
+        alias: 'o',
+        description: 'Where to store fetched files',
+        default: 'config/',
```
Siegrift, line 200: I prefer explicit setting of `output` locations when commands generate/download stuff, but we do provide default for `deploy` command.

packages/airnode-deployer/src/infrastructure/index.test.ts
```diff
@@ -1521,3 +1534,288 @@ describe('deploymentInfo', () => {
     expect(gcpGetFileFromBucketSpy).not.toHaveBeenCalled();
   });
 });
+
+describe('fetchFiles', () => {
+  const bucket = {
+    name: 'airnode-123456789',
+    region: 'us-east-1',
+  };
+  const configPath = path.join(__dirname, '..', '..', 'test', 'fixtures', 'config.aws.valid.json');
```
Siegrift, line 1543: Nit: You can just do `path.join(__dirname, '../../test/fixtures/config.aws.valid.json')`

packages/airnode-deployer/src/infrastructure/index.test.ts
```diff
@@ -1521,3 +1534,288 @@ describe('deploymentInfo', () => {
...
+    });
+  });
+
+  afterEach(() => {
+    fs.rmSync(tempOutputDirectory, { recursive: true });
+  });
+
+  it('fetch files for deployment', async () => {
+    const deploymentId = 'aws7195b548';
+    const fetchedFilename = `${deploymentId}-3580a278.zip`;
```
Siegrift, line 1579: I wonder why the filename is not just `${deploymentId}.zip`.

packages/airnode-deployer/src/infrastructure/index.test.ts
```diff
@@ -1521,3 +1534,288 @@ describe('deploymentInfo', () => {
...
+    );
+    expect(awsGetFileFromBucketSpy).toHaveBeenNthCalledWith(
+      2,
+      bucket.name,
+      '0xd0624E6C2C8A1DaEdE9Fa7E9C409167ed5F256c6/dev/1662558010204/secrets.env'
+    );
+    expect(awsGetFileFromBucketSpy).toHaveBeenNthCalledWith(
+      3,
+      bucket.name,
+      '0xd0624E6C2C8A1DaEdE9Fa7E9C409167ed5F256c6/dev/1662557983568/config.json'
```
Siegrift, line 1652: Why is it called with different timestamps?

packages/airnode-deployer/src/infrastructure/index.test.ts
```diff
@@ -1521,3 +1534,288 @@ describe('deploymentInfo', () => {
...
+    expect(awsGetFileFromBucketSpy).toHaveBeenNthCalledWith(
+      6,
+      bucket.name,
+      '0xA30CA71Ba54E83127214D3271aEA8F5D6bD4Dace/dev/1662559204554/secrets.env'
+    );
+  });
+
+  it(`fails if deployment version can't be found`, async () => {
+    const deploymentId = 'aws7195b548';
+    const nonexistingDeploymentVersion = '11223344';
```
Siegrift, line 1756: Nit: `nonExistingDeploymentVersion`

packages/airnode-deployer/src/infrastructure/index.test.ts
amarthadan: Because even though you don't fetch the specific version in this test a specific version is fetched (the latest one). And in case you would do `airnode-deployer fetch-files aws7195b548` and `airnode-deployer fetch-files aws7195b548 3580a278` I didn't want it to store the result in two different files because it's the same thing.

packages/airnode-deployer/src/infrastructure/index.test.ts
amarthadan: The first two calls are searching for the right deployment. We need to read those files in order to generate the deployment ID and find the correct one. In this case, it's the first one. And the second two calls are actually fetching those files we need, those with the specific version (older ones).

packages/airnode-deployer/src/infrastructure/index.test.ts
Siegrift: Maybe it could be called `${deploymentId}-latest.zip`, but what you did makes sense as well.

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -688,3 +689,75 @@ export async function deploymentInfo(deploymentId: string) {
   const tableStringWithCurrent = tableString.replace(new RegExp(`(?<=${currentVersionId}.*?)\n`), ' (current)\n');
   consoleLog(tableStringWithCurrent);
 }
+
+export async function fetchFiles(deploymentId: string, outputDir: string, versionId?: string) {
```
Siegrift, line 693: Nit: Maybe use `try finally` to stop the spinner at the end of the function execution instead of doing it manually?

Siegrift (approved):
:+1 LGTM

packages/airnode-deployer/src/infrastructure/index.ts
amarthadan: Can't really do that because I need to call `spinner.stop` in case of a failure or `spinner.succeed` if it's successful. I would need to have a `catch` block, catching the exceptions I'm throwing, stopping the spinner, and immediately rethrowing them. And the whole function would need to be in a `try catch finally` block, that seems weird. I'd rather keep it this way.

Siegrift: 👍 Makes sense. I didn't notice you need to call a different one if the operation was successful.

PR #1609, opened 2023-01-11
Add new _gasPrice reserved parameter

### LINKED ISSUE #1492: Add a new reserved parameter to allow gasPrice to be specified
See https://github.com/api3dao/airnode/issues/1393

If an endpoint has a `_gasPrice` reserved parameter defined, the requester can specify a gas price, which will be used in fulfillment. Unlike `_minConfirmations`, requests that will be fulfilled by the same sponsor wallets can have independent `_gasPrice` values. In other words, if there are three requests, with 10, 1 and 100 gwei gas prices specified, the Airnode should respond to each request with the respective gas price.

dcroote:
Allows gasPrice to be specified by a requester.

Closes #1492

### DIFF
--- a/packages/airnode-adapter/src/types.ts
+++ b/packages/airnode-adapter/src/types.ts
@@ -62,7 +62,8 @@ export type BaseResponseType = typeof baseResponseTypes[number];
 // Use might pass a complex type (e.g. int256[3][]) which we cannot type
 export type ResponseType = string;
 
-export interface ReservedParameters {
+// Reserved parameters specific for response processing i.e. excludes _gasPrice
+export interface ResponseReservedParameters {
   _path?: string;
   _times?: string;
   _type: ResponseType;
--- a/packages/airnode-node/src/api/index.ts
+++ b/packages/airnode-node/src/api/index.ts
@@ -226,7 +226,7 @@ export async function processSuccessfulApiCall(
   const { endpointName, oisTitle, parameters } = aggregatedApiCall;
   const ois = config.ois.find((o) => o.title === oisTitle)!;
   const endpoint = ois.endpoints.find((e) => e.name === endpointName)!;
-  const reservedParameters = getReservedParameters(endpoint, parameters);
+  const { _type, _path, _times, _gasPrice } = getReservedParameters(endpoint, parameters);
 
   const goPostProcessApiSpecifications = await go(() => postProcessApiSpecifications(rawResponse.data, endpoint));
   if (!goPostProcessApiSpecifications.success) {
@@ -235,10 +235,11 @@ export async function processSuccessfulApiCall(
   }
 
   const goExtractAndEncodeResponse = goSync(() =>
-    adapter.extractAndEncodeResponse(
-      goPostProcessApiSpecifications.data,
-      reservedParameters as adapter.ReservedParameters
-    )
+    adapter.extractAndEncodeResponse(goPostProcessApiSpecifications.data, {
+      _type,
+      _path,
+      _times,
+    } as adapter.ResponseReservedParameters)
   );
   if (!goExtractAndEncodeResponse.success) {
     const log = logger.pend('ERROR', goExtractAndEncodeResponse.error.message);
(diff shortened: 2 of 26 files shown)


packages/airnode-adapter/src/types.ts
```diff
@@ -62,7 +62,8 @@ export type BaseResponseType = typeof baseResponseTypes[number];
 // Use might pass a complex type (e.g. int256[3][]) which we cannot type
 export type ResponseType = string;
 
-export interface ReservedParameters {
+// Reserved parameters specific for response processing i.e. excludes _gasPrice
+export interface ResponseReservedParameters {
```
dcroote, line 66: I renamed `ReservedParameters` to `ResponseReservedParameters` to distinguish those that are relevant to response processing (`_type`, `_path`, and `_times`) from those that are not i.e. `_gasPrice` and (coming soon) `_minConfirmations`

packages/airnode-node/src/types.ts
```diff
@@ -229,6 +229,7 @@ export type ApiCallResponse = RegularApiCallResponse | HttpGatewayApiCallRespons
 export interface RegularApiCallSuccessResponse {
   success: true;
   data: { encodedValue: string; signature: string };
+  gasPriceOverride?: string;
```
dcroote, line 232: I spent a lot of time trying to find a way of retaining the requester-specified `_gasPrice` from the where API call response processing and reserved parameter parsing occurs for use where the `gasTarget` created and employed in submitting transactions. Obviously reserved parameters weren't exactly designed / implemented for more than response processing, but adding this property was the simplest solution I came up with.

packages/airnode-node/src/api/index.ts
```diff
@@ -235,10 +235,11 @@ export async function processSuccessfulApiCall(
...
   const goExtractAndEncodeResponse = goSync(() =>
-    adapter.extractAndEncodeResponse(
-      goPostProcessApiSpecifications.data,
-      reservedParameters as adapter.ReservedParameters
-    )
+    adapter.extractAndEncodeResponse(goPostProcessApiSpecifications.data, {
+      _type,
+      _path,
+      _times,
+    } as adapter.ResponseReservedParameters)
```
dcroote, line 242: Similarly here, trying to be explicit about which reserved parameters are relevant to response processing.

packages/airnode-node/src/types.ts
```diff
@@ -244,6 +245,7 @@ export interface HttpSignedDataApiCallSuccessResponse {
 export interface ApiCallErrorResponse {
   success: false;
   errorMessage: string;
+  gasPriceOverride?: string;
```
dcroote, line 248: Needed to add it here in order to match `RegularApiCallSuccessResponse` (satisfy type constraints)

amarthadan (commented):
LGTM, just a minor note. I want to test it first before approving.

packages/airnode-node/src/types.ts
dcroote: Do you mean something like
```ts
export interface RegularApiCallSuccessResponse {
  success: true;
  data: { encodedValue: string; signature: string };
  reservedParameterOverrides?: {
    _gasPrice?: string;
    _minConfirmations?: string;
  }
}
```

amarthadan: Yes.

amarthadan (changes requested):
Have you tested this? Because I did and I'm pretty sure it couldn't have worked :smile:
It took me quite some time but I found out what the problem is and why the value is ignored. The problem is here
https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/coordinator/calls/disaggregation.ts#L41
The response there is reconstructed but only the fields `data` and `success` are added. That was enough before but now we need to also add the `reservedParameterOverrides` field.
TS didn't complain because the field is optional anyway. But the set `_gasPrice` value was ignored.

Also, even though the value correctly overrides what's returned from the gas price oracle, the logs are quite confusing. They say how is the value chosen by the oracle but they don't mention anything about it being overridden later.

I'd also suggest writing tests using these reserved parameter overrides so that we won't run into something similar again.

amarthadan (approved):
LGTM

dcroote:
Oof 🤦 😞 well an apology is in order @amarthadan. I thought I had covered it through the tests I updated, but clearly you were right in this not working due to reconstruction. I added the E2E and confirmed it's working as expected locally. Sorry you ended up spending time debugging.

PR #1611, opened 2023-01-12
Deployer logs file

### LINKED ISSUE #1458: Improve deployer error messages
See: https://api3workspace.slack.com/archives/C02AYRX8D89/p1663092715835849

Currently, the deployer spits out terraform error messages which are too low level for end users. Also, due to an implementation limitation, the error is printed twice. 

What I propose is to:
* capture the terraform output and in case of an error provide just a short message
  * we can know contextually what operations is performed (update/redeploy/delete) and we can also parse the error message for common errors in the future (already existing gateway keys). For now use only a short generic error message.
* write full deployer debug information and error details into a file. The users can open this file to see more information about the problem and share the file to API3 to help debug. (The problematic part is that the file might contain sensitive information which should not be shared, e.g. stage, airnode address...).

This task could probably be split and could deserve a planning discussion on its own.

vponline:
Closes #1458


packages/airnode-deployer/src/utils/logger.ts
```diff
@@ -27,15 +33,22 @@ function oraInstance(text?: string) {
...
 export function succeed(text: string) {
+  writeLog(text);
   oraInstance().succeed(text);
 }
 
-export function fail(text: string) {
-  oraInstance().fail(text);
+export function fail(text: string, options: LoggerOptions = { bold: false }) {
+  writeLog(text);
+  oraInstance().fail(options.bold ? bold(text) : text);
```
vponline, line 47: I had to move the use of `bold` here because it created some weird characters when saved to a file

amarthadan (changes requested):
This issue could have been discussed/specified more. Sorry, that's on me and @aTeoke.

First of all, we need to allow users to change where the files should be stored. There should be a CLI option for that. Also, I'm not really a fan of the file names and the fact that we're storing everything in one (two) files.

I would propose having a `logs` directory. Users will be able to specify where it should be (the default being `config/logs`). In the directory, we will store the log files. Not just the two of them but two (or rather potentially two) for each deployer run, suffixing the name with a timestamp. Something like this:
```
├── logs
│   ├── deployer-2023-01-16_09:35:23.error.log
│   ├── deployer-2023-01-16_09:35:23.log
│   └── deployer-2023-01-16_09:50:15.log
```
I'm not 100% convinced that we actually need separate output and error log files though. Imagine debugging an issue. What good does it do if I have two different files and need to manually match timestamps from those logs to see what happened prior to the error? I would probably log it into the same file.

I'm of course open to discussing these design changes, this is just how I think it would work best.

I also run into a few issues while testing this.
As mentioned in the issue description, we should not log any sensitive information. Currently, you can find in the logs for example Airnode wallet private key or HTTP gateway.

Also, the debug mode is not working correctly. When the debug mode is enabled (`--debug` switch) most of the logs are written in the file twice. When the debug mode is not enabled, the logs written in the file are written as when the debug mode is enabled. This may be on purpose, I understand the intent, it's just a bit weird. I haven't seen such behavior anywhere else.

packages/airnode-deployer/src/utils/infrastructure.ts
```diff
@@ -31,11 +32,21 @@ export function formatTerraformArguments(args: CommandArg[]) {
...
 export const isCloudFunction = () => process.env.LAMBDA_TASK_ROOT || process.env.FUNCTION_TARGET;
 
-export const logAndReturnError = (message: string): Error => {
+export const logAndReturnError = (message: string) => {
   logger.fail(message);
   return new Error(message);
 };
 
+export const writeAndReturnError = (error: Error, message?: string) => {
+  fs.appendFileSync('config/deployer-error.log', `${new Date(Date.now()).toISOString()}: ${error.message}\n`);
```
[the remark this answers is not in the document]
vponline: Thanks for the spec, much appreciated! I'll work on fixing.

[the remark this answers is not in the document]
vponline: > I'm not 100% convinced that we actually need separate output and error log files though. 

TBH I was thinking the same. If that's ok, then I think we could just go with a single (timestamped) file per deployment attempt

packages/airnode-deployer/src/cli/commands.ts
```diff
@@ -68,8 +64,14 @@ export async function deploy(configPath: string, secretsPath: string, receiptFil
...
-  if (output.httpGatewayUrl) logger.info(`HTTP gateway URL: ${output.httpGatewayUrl}`);
-  if (output.httpSignedDataGatewayUrl) logger.info(`HTTP signed data gateway URL: ${output.httpSignedDataGatewayUrl}`);
+  if (output.httpGatewayUrl) {
+    logger.setSecret(output.httpGatewayUrl);
+    logger.info(`HTTP gateway URL: ${output.httpGatewayUrl}`);
+  }
+  if (output.httpSignedDataGatewayUrl) {
+    logger.setSecret(output.httpSignedDataGatewayUrl);
+    logger.info(`HTTP signed data gateway URL: ${output.httpSignedDataGatewayUrl}`);
+  }
```
vponline, line 74: I went with this option to set secret values to be sanitized from logs. The deployer command options are passed down multiple levels and I would have needed to pass the list of secret values along for them to reach the logger call. So this seemed simpler.

packages/airnode-deployer/src/infrastructure/index.ts
```diff
@@ -248,6 +259,7 @@ export async function terraformAirnodeApply(
   const cloudProvider = config.nodeSettings.cloudProvider as CloudProvider;
   const airnodeAddress = deriveAirnodeAddress(airnodeWalletMnemonic);
   const airnodeWalletPrivateKey = evm.getAirnodeWallet(config).privateKey;
+  logger.setSecret(airnodeWalletPrivateKey);
```
vponline, line 262: I went with this option to set secret values to be sanitized from logs. The deployer command options are passed down multiple levels and I would have needed to pass the list of secret values along for them to reach the logger call. So this seemed simpler.

packages/airnode-deployer/src/infrastructure/index.test.ts
```diff
@@ -18,6 +18,11 @@ jest.mock('../../package.json', () => ({
   version: '0.8.0',
 }));
 
+jest.mock('../utils/logger', () => ({
+  ...jest.requireActual('../utils/logger'),
+  writeLog: jest.fn(),
+}));
+
```
vponline: I tried this mock, but for some reason it resulted in logger having this structure:

{
  ...(regular logger functions),
  default: {
    writeLog: mock
      ...(regular logger functions),
  }
}

so `writeLog` was never actually mocked correctly. Mocking `fs.appendFileSync` as in the last commit works, though.

vponline:
I think that covers everything so please feel free to re-review :smile:

packages/airnode-deployer/src/utils/logger.ts
```diff
@@ -84,3 +90,21 @@ export function debugMode(mode: boolean) {
...
+export function replaceSecrets(input: string, secrets: string[]) {
+  let output = input;
+  secrets.forEach((secret) => (output = output.replace(secret, '*'.repeat(secret.length))));
+
+  return output;
+}
+
+export function setLogsDirectory(path: string) {
+  logsDirectory = path.endsWith('/') ? path.slice(0, -1) : path;
+  if (!fs.existsSync(logsDirectory)) fs.mkdirSync(logsDirectory, { recursive: true });
```
amarthadan, line 107: Hmm, I thought that since [`fs.exists` is being deprecated](https://nodejs.org/api/fs.html#fsexistspath-callback) `fs.existsSync` will be as well but it's not. Either way, I started using [`fs.accessSync`](https://nodejs.org/api/fs.html#fsaccesssyncpath-mode) instead. Can you do the same to be consistent?

packages/airnode-deployer/src/utils/logger.ts
```diff
@@ -84,3 +90,21 @@ export function debugMode(mode: boolean) {
...
+
+  logFileTimestamp = format(Date.now(), 'yyyy-MM-dd_HH:mm:ss');
```
amarthadan, line 109: If we're using the format at multiple places it should be extracted in a constant.

packages/airnode-deployer/src/utils/logger.ts
```diff
@@ -84,3 +90,21 @@ export function debugMode(mode: boolean) {
...
+
+export function replaceSecrets(input: string, secrets: string[]) {
+  let output = input;
+  secrets.forEach((secret) => (output = output.replace(secret, '*'.repeat(secret.length))));
+
+  return output;
+}
+
+export function setLogsDirectory(path: string) {
+  logsDirectory = path.endsWith('/') ? path.slice(0, -1) : path;
```
amarthadan, line 106: You don't need to "sanitize" the directory path if you then use `path.join` to create a file path (which we should be doing anyway).

```ts
fs.appendFileSync(path.join(logsDirectory, `deployer-${logFileTimestamp}.log`), `${timestamp}: ${sanitizedLogs}\n`);
```

amarthadan (changes requested):

amarthadan (changes requested):
Hmm, a few more issues (sorry :sweat_smile:)

There's a difference between what's written to the terminal and what's written to the log files. Not every output is using the logger and those are not in the logs (e.g. output of `list` or `info` commands). The question is whether we want those outputs there. Can they be considered sensitive information? I don't think so. I think we should include it in the logs.

Also, I would go with the debug mode as you had it before. Even though the debug mode is not enabled, we should write the logs to the file as if it is. That way, if something fails, the user already has the logs with all the info. Some issues may not be easily reproducible so asking them to enable the debug mode might not be a feasible option.

packages/airnode-deployer/src/utils/logger.ts
```diff
@@ -84,3 +90,21 @@ export function debugMode(mode: boolean) {
...
   return debugModeFlag;
 }
+
+export function setSecret(secret: string) {
+  secrets.push(secret);
+}
+
+export function replaceSecrets(input: string, secrets: string[]) {
+  let output = input;
+  secrets.forEach((secret) => (output = output.replace(secret, '*'.repeat(secret.length))));
```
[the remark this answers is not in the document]
vponline: > Hmm, this is happening for every secret and every line that is logged. But I guess we won't have that many secrets and that many logs so it will affect the performance. It is kinda ugly though

Yeah... I didn't have any other ideas, though. Did you happen to have an alternative in mind?

The other option I thought about is passing down the secrets values to the logger and using a conditional for `replaceSecrets`, but it just ended up being weird since we'd be passing them down so many levels.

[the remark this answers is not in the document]
amarthadan: > Did you happen to have an alternative in mind?

I was thinking about passing the secrets as an additional argument in the log functions and do the replace always only on that one passed line. But I'm not sure that's better. I mean it would perform better but not sure about usability.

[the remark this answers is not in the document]
vponline: How about we save the secrets into the `secrets`-array (as currently), but the logger functions will take in an options variable of whether to apply `replaceSecrets` to the text which would still check against the `secrets`-array? Best of both worlds :smile:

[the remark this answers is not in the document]
amarthadan: Yeah, let's go with that, that sounds like a nice compromise :slightly_smiling_face:

packages/airnode-deployer/src/utils/logger.ts
```diff
@@ -23,19 +41,39 @@ export function getSpinner() {
   return spinner;
 }
 
+export function useSpinner(method: OraMethod, text?: string, options?: LoggerOptions) {
+  if (text) writeLog(text, options);
+  getSpinner()[method](text);
+}
```
vponline, line 47: This wrapper is for using a particular instance of the spinner (mainly in `infrastructure/index.ts`) while allowing us to keep writing the logs in one place.

amarthadan (changes requested):

amarthadan (approved):

PR #1614, opened 2023-01-18
Add rollback command to deployer

### LINKED ISSUE #1406: Add rollback functionality for Airnode deployments
Allow users to rollback to their previous Airnode deployment version.

```bash
airnode-deployer rollback aws2c6ef2b3 5bbcd317
```
Based on the deployment ID from the `list` command (https://github.com/api3dao/airnode/issues/1403) and deployment version ID from the `info` command (https://github.com/api3dao/airnode/issues/1404).
The command will create a **new** deployment with the configuration from the selected previous one.

amarthadan:
Close #1406

At this point `src/infrastructure/index.ts` should be refactored a bit, some common code can be extracted into functions and the whole thing should be divided into separate files. I'm leaving that for a separate issue.

### DIFF
--- a/packages/airnode-deployer/src/cli/commands.test.ts
+++ b/packages/airnode-deployer/src/cli/commands.test.ts
@@ -1,24 +1,33 @@
-import { join } from 'path';
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
 import { mockReadFileSync } from '../../test/mock-utils';
-import { readFileSync } from 'fs';
 import { receipt } from '@api3/airnode-validator';
-import { deploy, removeWithReceipt } from './commands';
+import { Ora } from 'ora';
+import { deploy, removeWithReceipt, rollback } from './commands';
 import { version as packageVersion } from '../../package.json';
 import * as logger from '../utils/logger';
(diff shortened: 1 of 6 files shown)


packages/airnode-deployer/src/cli/commands.test.ts
```diff
@@ -1,25 +1,34 @@
-import { join } from 'path';
+import fs from 'fs';
```
Siegrift, line 1: Nit: Can we enforce it with eslint? I prefer the member import syntax, but I am OK with this if it's for the consistency sake.

Siegrift (changes requested):
I tested on AWS, looks good, but found a few edge cases which I think should not be present.
1. I can rollback even to current version
2. As a user I'd (or better said I could) expect the rollback to move the string "(current)" in the info output to the rollbacked version. However, what we do is we create a new version with config+secrets from the version to rollback to. I think this could be clarified in the output. E.g. `Creating a new version <version_id> with configuration files from version <version_to_rollback>`

Siegrift (approved):
👍 LGTM

PR #1616, opened 2023-01-22
fix _gasPrice e2e test and simplify getReservedParameterValue

dcroote:
`find` wasn't working as expected with the async callback. This PR corrects that deficiency and improves the robustness of the test by returning the transcation that used the expected `gasPrice`, confirming there was only one such transaction, and re-confirming the `gasPrice`.

The second commit simplifies the return logic of `getReservedParameterValue`

### DIFF
--- a/packages/airnode-node/test/e2e/requester-fulfill.feature.ts
+++ b/packages/airnode-node/test/e2e/requester-fulfill.feature.ts
@@ -1,3 +1,4 @@
+import compact from 'lodash/compact';
 import { startCoordinator } from '../../src/workers/local-handlers';
 import { operation } from '../fixtures';
 import {
@@ -65,7 +66,7 @@ it('should call fail function on AirnodeRrp contract and emit FailedRequest if r
 
 it('submits fulfillment with the gas price overridden', async () => {
   const requestedGasPrice = '99999999999'; // wei
-  const gasOverrideParameters = [
+  const overrideParameters = [
     { type: 'string32', name: 'from', value: 'ETH' },
     { type: 'string32', name: 'to', value: 'USD' },
     { type: 'string32', name: '_type', value: 'int256' },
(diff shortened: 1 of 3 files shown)


packages/airnode-node/test/e2e/requester-fulfill.feature.ts
```diff
@@ -65,24 +65,31 @@ it('should call fail function on AirnodeRrp contract and emit FailedRequest if r
...
-  expect(fulfillmentLog).not.toBeUndefined();
+  const filteredTxs = await Promise.all(
+    logs.map(async (log) => {
+      const tx = await provider.getTransaction(log.transactionHash);
+      if (tx.gasPrice && tx.gasPrice.toString() === requestedGasPrice) {
+        return tx;
+      }
+      return null;
+    })
+  ).then((logs) => logs.filter((log) => log));
```
amarthadan, line 91: Even though the code is working this way I would prefer avoiding mixing `await` and `then`.

```suggestion
  const filteredTxs = (await Promise.all(
    logs.map(async (log) => {
      const tx = await provider.getTransaction(log.transactionHash);
      if (tx.gasPrice && tx.gasPrice.toString() === requestedGasPrice) {
        return tx;
      }
      return null;
    })
  )).filter((log) => log));
```

amarthadan, line 91: You can also use lodash's [`compact`](https://lodash.com/docs/4.17.15#compact) instead of a custom filter.

amarthadan (approved):
Just a minor nits, otherwise LGTM.

packages/airnode-node/test/e2e/requester-fulfill.feature.ts
dcroote: Great- included in a fixup commit with you as co-author

dcroote: I always forget lodash has some convenience function for most everything like this. Included, thanks.

PR #1640, opened 2023-02-06
Release v0.10.0

amarthadan:


dcroote (approved):
Approving so that I don't delay merging, but I think the failing unit tests should be fixed first. The failures originate from the derivation of `deploymentId`, which takes the `airnodeVersion` as a parameter and is therefore different now than for v0.9.

https://github.com/api3dao/airnode/blob/dd64a54f8183ec740d443727a61384b2f1f384bf/packages/airnode-node/src/workers/index.ts#L42-L53

PR #1642, opened 2023-02-07
Disable number conversion for release CLI

amarthadan:
By default, yargs tries to parse everything that looks like a number as a number. This is not desired since some version tags (e.g. 0.10) can be parsed as a number but we need their string representation instead ('0.10' v 0.1).


dcroote (approved):
LGTM

Issue #1544, opened 2022-11-17
Relax engine specification for pre/post-processing


bbenligiray:
Currently pre/post-processing snippets require the node version to be specified. We're also constantly updating the Airnode node version. For a lack of a better alternative, I suggest dropping the node version (so `node` or `node.js` instead of `node14`). The alternative is to deploy multiple pre/post-processing environments for these to be run on, but even this is not a permanent solution because node versions fall out of support and cloud providers stop providing these even earlier than that.

dcroote:
> We're also constantly updating the Airnode node version.

You don't think we'll stay on 18 for a while after https://github.com/api3dao/airnode/issues/1631?

Siegrift:
These versions are supported roughly for ~2years, so they are not permanent. 

I support the idea of dropping the node version from processing name and use whatever node the Airnode was deployed on. Other alternatives don't seem feasible to me.

dcroote:
Blocked by https://github.com/api3dao/ois/issues/65 and a subsequent OIS release.

aTeoke:
Hey, I was trying to test this via `coingecko-pre-processing` example, but got this error. 
```
Error details:
2023-02-20T16:29:56.218Z ✖ Invalid Airnode configuration file: [
  {
    "code": "custom",
    "message": "oisFormat major.minor version must match major.minor version of \"1.4.0\"",
    "path": [
      "ois",
      0,
      "oisFormat"
    ]
  },
  {
    "code": "invalid_union",
    "unionErrors": [
      {
        "issues": [
          {
            "code": "invalid_literal",
            "expected": "Node 14",
            "path": [
              "ois",
              0,
              "endpoints",
              0,
              "preProcessingSpecifications",
              0,
              "environment"
            ],
            "message": "Invalid literal value, expected \"Node 14\""
          }
        ],
        "name": "ZodError"
      },
      {
        "issues": [
          {
            "code": "invalid_literal",
            "expected": "Node 14 async",
            "path": [
              "ois",
              0,
              "endpoints",
              0,
              "preProcessingSpecifications",
              0,
              "environment"
            ],
            "message": "Invalid literal value, expected \"Node 14 async\""
          }
        ],
        "name": "ZodError"
      }
    ],
    "path": [
      "ois",
      0,
      "endpoints",
      0,
      "preProcessingSpecifications",
      0,
      "environment"
    ],
    "message": "Invalid input"
  }
]
```

@dcroote could you please help?

[...]

PR #1647, opened 2023-02-10
Update Airnode to use Node 18

### LINKED ISSUE #1346: Update hardhat to ^2.10.0

### LINKED ISSUE #1631: Update Airnode to use Node 18
We will be skipping update to Node 16 and go straight to Node 18.
**Why?**
[Node.js 18 runtime for GCP App Engine](https://cloud.google.com/appengine/docs/standard/nodejs/release-notes#January_17_2023) is now generally available and Node.js 18 is recommended for GCP [Cloud Functions](https://cloud.google.com/functions/docs/concepts/nodejs-runtime). It has been supported on AWS Lambda [since Nov 2022](https://aws.amazon.com/fr/blogs/compute/node-js-18-x-runtime-now-available-in-aws-lambda/).

Node 16 is going to be EOL on 11 September 2023
Node 14 security support ends on 30 Apr 2023

**Dependencies:**
- blocks: [#1625](https://github.com/api3dao/airnode/pull/1625)
- We should address #1544 before or together with this

#1457 for reference

dcroote:
Closes #1631 and closes #1346. While this doesn't address https://github.com/api3dao/airnode/issues/1544, that can be done after (given it requires another OIS release) as the engine specification doesn't actually appear to enforce version 14 of `Node 14`.

~~Edit: The `airnode-node` unit tests all pass, yet the step fails...  Trying to isolate / reproduce.~~ Found the failing test and fixed it.

### DIFF
--- a/.changeset/short-donuts-pay.md
+++ b/.changeset/short-donuts-pay.md
@@ -1,0 +1,14 @@
+---
+'@api3/airnode-adapter': minor
+'@api3/airnode-admin': minor
+'@api3/airnode-deployer': minor
+'@api3/airnode-examples': minor
+'@api3/airnode-node': minor
+'@api3/airnode-operation': minor
+'@api3/airnode-protocol': minor
+'@api3/airnode-utilities': minor
+'@api3/airnode-abi': minor
+'@api3/airnode-validator': minor
+---
+
+Bump Node.js from 14 to 18
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -9,7 +9,7 @@ on:
 
 env:
   DOCKER_BUILDKIT: 1
-  TARGET_NODE_VERSION: '14.19.3'
+  TARGET_NODE_VERSION: '18.14.0'
 
 jobs:
   documentation:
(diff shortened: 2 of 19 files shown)


amarthadan (dismissed):

amarthadan:
I've approved this prematurely :smile: I've just realized that you're missing one crucial thing. You need to update the Node.js environment for serverless functions to Node 18 as well. [Here](https://github.com/api3dao/airnode/blob/master/packages/airnode-deployer/terraform/aws/modules/function/main.tf#L46) and [here](https://github.com/api3dao/airnode/blob/master/packages/airnode-deployer/terraform/gcp/modules/function/main.tf#L59).

amarthadan (approved):
Ok, I think that's all now :crossed_fingers: :smile:

dcroote:
Thanks @amarthadan and nice catch on the terraform functions.

I tested the airnode container, AWS deployment, and GCP deployment each with a blockchain and HTTP gateway request.

PR #1652, opened 2023-02-16
Removes Node version 14 from pre/post-processing specification.

### LINKED ISSUE #1544: Relax engine specification for pre/post-processing

dcroote:
Closes #1544

### DIFF
--- a/.changeset/rare-weeks-bathe.md
+++ b/.changeset/rare-weeks-bathe.md
@@ -1,0 +1,10 @@
+---
+'@api3/airnode-adapter': minor
+'@api3/airnode-deployer': minor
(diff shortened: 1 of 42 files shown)


amarthadan (approved):
LGTM :+1:

PR #1653, opened 2023-02-17
Replace hardcoded deploymentId values in airnode-node tests

### LINKED ISSUE #1641: Fix `airnode-node` tests breaking on package version update
This happened during [the v0.10.0 release](https://github.com/api3dao/airnode/actions/runs/4105569837/jobs/7092296880) and was temporarily fixed by [updating the fixtures](https://github.com/api3dao/airnode/pull/1640). This needs to be fixed permanently so it won't break for every new release.

dcroote:
Closes #1641.

`deriveDeploymentId` takes [as one argument](https://github.com/api3dao/airnode/blob/dd64a54f8183ec740d443727a61384b2f1f384bf/packages/airnode-node/src/workers/index.ts#L42-L53) the Airnode version. Consequently, when the Airnode version is bumped, the `deploymentId` changes, which breaks tests that have hardcoded values. This PR replaces these hardcoded values with a regex so bumping the version no longer causes the tests to fail. Tests on the output of `deriveDeploymentId` are already present elsewhere.


packages/airnode-node/src/handlers/start-coordinator.test.ts
```diff
@@ -172,7 +172,12 @@ describe('startCoordinator', () => {
     await startCoordinator(config, coordinatorId);
 
     // cached requests should not trigger an API call
-    expect(callApisSpy).toHaveBeenCalledWith([], expect.objectContaining({ deploymentId: 'localdd59d6d0' }));
+    expect(callApisSpy).toHaveBeenCalledWith(
+      [],
+      expect.objectContaining({
+        deploymentId: expect.stringMatching(new RegExp(`local[0-9a-z]{${DEPLOYMENT_ID_LENGTH}}`)),
```
amarthadan, line 178: Since the regular expression is always the same we can extract it as a constant and use that instead.
Also, it can be made more strict using `[0-9a-f]` (only hexadecimal) instead.

packages/airnode-node/src/constants.ts
```diff
@@ -35,3 +35,6 @@ export const MAXIMUM_ONCHAIN_ERROR_LENGTH = 100;
 
 // The maximum amount of time in milliseconds that pre and post processing is allowed to execute
 export const PROCESSING_TIMEOUT = 10_000;
+
+// The length of the hashed portion of a deployment ID
+export const DEPLOYMENT_ID_LENGTH = 8;
```
amarthadan, line 40: There is already such a constant https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/workers/index.ts#L8. I'm fine with using either of these places but we should just use the one.

amarthadan (changes requested):

packages/airnode-node/src/constants.ts
dcroote: I am using only one- I removed the one you link to, but it's probably cleaner to just export that. I'll update.

packages/airnode-node/src/handlers/start-coordinator.test.ts
dcroote: A constant across all tests? Without the `local` prefix? I may not be understanding how you envision this. 

Also `a-f`, right haha.

dcroote: I've updated it to `a-f` and to have a `const` regex variable per test file. I don't think it makes sense to have it as a "global" constant because the cloud provider type prefix of `local` applies to these tests and wouldn't apply to `aws` or `gcp` deployment ID regexes

amarthadan: I mean it as a global test constant since all the tests here use only the `local` prefix. But this is fine as well.

amarthadan (approved):

PR #1656, opened 2023-02-22
fix: apply prettier programmatically in update script

### LINKED ISSUE #1638: `update-config-fixtures.ts` script formats output differently than Prettier
I run into this during the v0.10.0 release
https://github.com/api3dao/airnode/pull/1637/files#diff-3317278434633fa1c9c58606616c717e25638b589a62d7ad79de68c2b98e2c77

dcroote:
Closes #1638. The fix is as @Siegrift suggested- to use `prettier` programmatically. It looked familiar because I ran into the same error in a slightly different context (though still generating config files) in #1488:

https://github.com/api3dao/airnode/blob/e4d3bee6e70bc733f4d95245fa320753d4204061/packages/airnode-examples/integrations/config-utils.ts#L70

I confirmed this solves the issue by running the script prior to the fix, seeing that multiple `config` files had formatting changes, then after this fix and noting that no files changed.


amarthadan (approved):
:+1:

PR #1659, opened 2023-02-24
Update AWS SDK

### LINKED ISSUE #1648: Update AWS SDK for JavaScript to v3
Within the Airnode logs after updating Node.js from 14 to 18 in #1647.
```sh
2023-02-13 23:49:28 (node:15) NOTE: The AWS SDK for JavaScript (v2) will be put into maintenance mode in 2023.
2023-02-13 23:49:28 
2023-02-13 23:49:28 Please migrate your code to use AWS SDK for JavaScript (v3).
2023-02-13 23:49:28 For more information, check the migration guide at https://a.co/7PzMCcy
```

amarthadan:
Close #1648

TBH I'm not sure why the commit 5917596 was needed. I presume a combination of a newer version of Docker (it should still work for the older ones though) and a new version of some dependencies (causing the minification not to work but we don't really need it in this place anyway).


packages/airnode-node/src/adapters/http/worker.test.ts
```diff
@@ -1,14 +1,12 @@
...
-}));
-
+import { mockClient } from 'aws-sdk-client-mock';
+import 'aws-sdk-client-mock-jest';
+import { LambdaClient, InvokeCommand } from '@aws-sdk/client-lambda';
 import { logger, LogOptions, setLogOptions } from '@api3/airnode-utilities';
 import * as worker from './worker';
 import * as fixtures from '../../../test/fixtures';
 
+const mockLambdaClient = mockClient(LambdaClient);
```
Siegrift, line 8: So this basically does the mocking for you internally, and returns you a spy that you can mock... Right?

packages/airnode-node/src/adapters/http/worker.test.ts
```diff
@@ -17,30 +15,35 @@ describe('spawnNewApiCall', () => {
...
+    mockLambdaClient.reset();
+  });
+
   it('handles remote AWS calls', async () => {
-    invokeMock.mockImplementationOnce((_params, callback) =>
-      callback(null, { Payload: JSON.stringify({ body: JSON.stringify({ ok: true, data: { value: '0x123' } }) }) })
-    );
-    const aggregatedApiCall = fixtures.buildAggregatedRegularApiCall();
+    mockLambdaClient.on(InvokeCommand, mockInvokeCommandInput).resolves({
+      Payload: new TextEncoder().encode(
```
Siegrift, line 30: Nit: Create `encode` (or stringifyAndEncode) function which will stringify and encode, to make it less boilerplate...

packages/airnode-node/src/evm/workers.test.ts
```diff
@@ -24,31 +22,33 @@ const providerErrorForWorker = {
...
   describe(`${workerType} worker`, () => {
     fixtures.setEnvVariables({ AIRNODE_WALLET_PRIVATE_KEY: fixtures.getAirnodeWalletPrivateKey() });
+    const state = fixtures.buildEVMProviderSponsorState();
+    const mockInvokeCommandInput = {
+      FunctionName: serverlessFunctionName,
+      Payload: new TextEncoder().encode(JSON.stringify({ state, functionName: functionNames[workerType] })),
+    };
+
+    beforeEach(() => {
+      mockLambdaClient.reset();
```
Siegrift, line 32: Doesn't it reset on it's own?

packages/airnode-node/src/workers/cloud-platforms/aws.ts
```diff
@@ -1,49 +1,43 @@
...
-
-      resolve(goBody.data as WorkerResponse);
-    });
-  });
+export async function spawn(params: WorkerParameters): Promise<WorkerResponse> {
+  // Uses the current region by default
+  const awsLambdaClient = new LambdaClient({});
+
+  // AWS doesn't allow uppercase letters in lambda function names
+  const resolvedName = `airnode-${params.deploymentId}-run`;
```
Siegrift, line 11: And we expect the deploymentId to be lowercased right? I guess previously we had .toLowerCase there just in case so that's what the comment is referring to.

packages/airnode-node/src/workers/cloud-platforms/aws.ts
```diff
@@ -1,49 +1,43 @@
...
+
+  const invokeCommand = new InvokeCommand(options);
+  const goInvoke = await go(() => awsLambdaClient.send(invokeCommand));
+  if (!goInvoke.success) {
+    logger.debug(`Lambda invocation or execution failed. ${goInvoke.error}}`);
+    throw goInvoke.error;
+  }
+  const invokeOutput = goInvoke.data;
+
+  if (invokeOutput.FunctionError) {
```
Siegrift, line 26: Where is this new API even documented? I tried quick googling but I only found https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/index.html which is closest to "documentation" I'd expect, but it's still pretty useless.

Siegrift: I know they probably haven't changed the lambda response, but still... Can you at least copy the comment that was there before?

packages/airnode-node/src/workers/cloud-platforms/aws.ts
```diff
@@ -1,49 +1,43 @@
...
+    throw invokeOutput.Payload;
+  }
+
+  if (!invokeOutput.Payload) {
+    const error = new Error('Missing payload in lambda response');
+    logger.debug(`Lambda invocation or execution failed. ${error}`);
+    throw error;
+  }
+
+  const payloadString = new TextDecoder('utf-8').decode(invokeOutput.Payload);
```
Siegrift, line 37: I noticed that you do not pass 'utf-8' during encode, but you pass it during encode... It feels wierd. Why not pass it to encode as well?

Siegrift (changes requested):
Small nits, otherwise LGTM

package.json
```diff
@@ -50,7 +50,7 @@
     "docker:build:latest": "yarn docker:build:packaging && yarn docker:build:images:latest",
     "docker:build:local": "yarn docker:build:packaging && yarn docker:build:images:local",
     "docker:build:packaging": "yarn docker:scripts:build && docker build --tag api3/airnode-packaging:latest --file docker/Dockerfile .",
-    "docker:scripts:build": "ncc build docker/scripts/cli.ts -o docker/scripts/dist --no-cache --minify --source-map --transpile-only",
+    "docker:scripts:build": "ncc build docker/scripts/cli.ts -o docker/scripts/dist --no-cache --source-map --transpile-only",
```
Siegrift, line 53: Btw. I am also not sure about why this should be needed... Maybe the new AWS is ESM and that causes problems??? (ESM is often to blame for stuff :D )

packages/airnode-node/src/adapters/http/worker.test.ts
amarthadan: Yes.

packages/airnode-node/src/evm/workers.test.ts
amarthadan: I'm not sure, I was following the documentation. It won't hurt to keep it.

package.json
amarthadan: I ran into this on another PR as well so I guess it was caused by some other dependency. It doesn't really matter though.

packages/airnode-node/src/workers/cloud-platforms/aws.ts
amarthadan: > Where is this new API even documented? I tried quick googling but I only found https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/index.html which is closest to "documentation" I'd expect, but it's still pretty useless.

It is quite useless but that's the correct documentation.

> Can you at least copy the comment that was there before?

Will do.

packages/airnode-node/src/workers/cloud-platforms/aws.ts
amarthadan: Because `TextEncoder` encodes **only** to UTF-8 (https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder). There is no way to set a different encoding. But `TextDecoder` can decode from multiple encodings (https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder).

Siegrift (approved):
👍 LGTM

PR #1670, opened 2023-03-13
Implement RequesterAuthorizerWithErc721 authorizers

### LINKED ISSUE #1632: Implement RequesterAuthorizerWithErc721 support
RequesterAuthorizerWithErc721 is a new authorizer type with the following interface
```
function isAuthorized(
        address airnode,
        uint256 chainId,
        address requester,
        address token
    ) external view returns (bool)
```
`airnode` is the address of the Airnode calling it, `requester` is the requester contract address, `chainId` is the chain ID that the requester contract is deployed on (and the respective RRP request is detected on), `token` is the address of the ERC721 token that is specified in the config of the Airnode.
RequesterAuthorizerWithErc721 will typically be used as a cross-chain authorizer, though this doesn't affect the implementation of the Airnode-side.
RequesterAuthorizerWithErc721 will inherit SelfMulticall (so `tryMulticall()` will be used for batch static calls). `checkAuthorizationStatuses()` can't be used for batch authorization checks because that expects a specific authorizer interface that RequesterAuthorizerWithErc721 doesn't support.
RequesterAuthorizerWithErc721 is currently under development under the ChainAPI Github organization but will be audited and made open source soon.
This blocks https://forum.api3.org/t/secondary-proposal-33-chainapi-july-2022-march-2023/1502 so it should be added to a milestone that is not too far away @aTeoke

dcroote:
Closes #1632 by adding two new authorizers (same chain and cross chain). In each, `RequesterAuthorizerWithErc721` corresponds to the address of the authorizer contract we deploy ([from here](https://github.com/api3dao/nft-authorizer/blob/f4bba58837f6d9d3b28254f3c2764cbe6e3bb88f/contracts/RequesterAuthorizerWithErc721.sol), incorporated completely herein along with other necessary contracts from `airnode-protocol-v1`), and `erc721s` is an array of NFTs that serve as tokens to authorize requesters.
```json
        "requesterAuthorizersWithErc721": [
          {
            "erc721s": ["0x00bDB2315678afecb367f032d93F642f64180a00"],
            "RequesterAuthorizerWithErc721": "0x999DB2315678afecb367f032d93F642f64180aa9"
          }
        ],
        "crossChainRequesterAuthorizersWithErc721": [
          {
            "erc721s": ["0x3FbDB2315678afecb367f032d93F642f64180aa6"],
            "chainType": "evm",
            "chainId": "4",
            "contracts": {
              "RequesterAuthorizerWithErc721": "0x6bbbb2315678afecb367f032d93F642f64180aa4"
            },
            "chainProvider": {
              "url": "http://127.0.0.2"
            }
          }
        ]
```

@bbenligiray  I'm still struggling a bit on the e2e test, see the second commit


Siegrift:
I've made the e2e test work (hacky way) in https://github.com/api3dao/airnode/commit/950f5ec9c7144472c0487247334d1aa8b8e9a7ba

The proper implementation requires a small refactor.

bbenligiray:
:+1: for the config format

dcroote:
@Siegrift  thanks for such a huge help, I really appreciate it.

> I've made the e2e test work

Unfortunately due to a legacy subtlety in `authorizer.requesterEndpointAuthorizers` empty array behavior, your request was being authorized by this authorizer instead of the new `requesterAuthorizersWithErc721` authorizer. See my last commit (in which I added a check for all authorizers being empty). I also changed dependencies, see my first commit after yours.

For [isAuthorized() within this new authorizer](https://github.com/api3dao/airnode/blob/beb4484d849e6296e170e5b5f190b20014323935/packages/airnode-protocol/contracts/dev/RequesterAuthorizerWithErc721.sol#L499-L520), it appears an NFT (token) has to be deposited... Still working through understanding that

Siegrift:
> For [isAuthorized() within this new authorizer](https://github.com/api3dao/airnode/blob/beb4484d849e6296e170e5b5f190b20014323935/packages/airnode-protocol/contracts/dev/RequesterAuthorizerWithErc721.sol#L499-L520), it appears an NFT (token) has to be deposited... Still working through understanding that

Yeah, sorry. I mainly focused on how to make the deployments (NFT and the respective authorizer) work. I thought that was the challenging part. Anyway, I think know how to make it work with token deposit. I'll try to fix the test today.

Siegrift:
@dcroote Take a look at https://github.com/api3dao/airnode/commit/b3db6aa8fcee91689eba1c4f93135674ab9265cc

packages/airnode-node/src/evm/authorization/authorization-application.test.ts
```diff
@@ -33,7 +33,12 @@ describe('mergeAuthorizations', () => {
     const apiCall = fixtures.requests.buildApiCall({ id: '0xapiCallId' });
     const authorizationByRequestId = { '0xapiCallId': true };
     const [logs, res] = authorization.mergeAuthorizations([apiCall], authorizationByRequestId);
-    expect(logs).toEqual([]);
+    expect(logs).toEqual([
+      {
+        level: 'DEBUG',
+        message: 'Requester:requesterAddress is authorized to access Endpoint ID:endpointId for Request ID:0xapiCallId',
```
Siegrift, line 39: It would also be nice debug level info which requester authorized the given call.

packages/airnode-node/src/evm/authorization/authorization-fetching.test.ts
```diff
@@ -39,26 +41,6 @@ describe('fetch (authorizations)', () => {
     expect(res).toEqual({});
   });
 
-  it('returns true for all pending requests if authorizers arrays are empty', async () => {
```
Siegrift, line 42: This is removed because the logic short-circuits the function call right?

packages/airnode-node/src/evm/handlers/initialize-provider.ts
```diff
@@ -10,30 +11,37 @@ import * as templates from '../templates';
...
-    authorizations: currentState.settings.authorizations,
-    airnodeAddress: currentState.settings.airnodeAddress,
-    airnodeRrpAddress: currentState.contracts.AirnodeRrp,
-    provider: currentState.provider,
-  };
-  const [logs, res] = await authorizations.fetch(currentState.requests.apiCalls, fetchOptions);
-  return { id: 'authorizations', data: res, logs };
+  if (isEmpty(currentState.settings.authorizers.requesterEndpointAuthorizers)) {
+    return { id: 'authorizations', data: {}, logs: [] };
+  } else {
```
Siegrift, line 21: rm else branch to avoid nesting.

packages/airnode-node/src/evm/handlers/initialize-provider.ts
```diff
@@ -142,32 +194,58 @@ export async function initializeProvider(
...
   // =================================================================
   // NOTE: None of these promises can fail otherwise Promise.all will reject
-  const authAndTxCountPromises: readonly ParallelPromise[] = [
-    fetchSameChainAuthorizations(state5),
-    fetchTransactionCounts(state5),
-    fetchCrossChainAuthorizations(state5),
-  ];
+
+  // If all authorizers arrays are empty then all requests are authorized
+  const allAuthorizersEmpty = Object.values(state5.settings.authorizers).every((arr) => isEmpty(arr));
```
Siegrift, line 199: This should deserve an INFO log I think.

packages/airnode-protocol/contracts/authorizers/mock/MockRequesterAuthorizerWithErc721.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import "../../dev/RequesterAuthorizerWithErc721.sol";
+
+contract MockRequesterAuthorizerWithErc721 is RequesterAuthorizerWithErc721 {
```
Siegrift, line 6: The `MockRequesterAuthorizerWithErc721` is not needed. You may just use `RequesterAuthorizerWithErc721` directly. I called it mock since it was only used for testing, but you may as well just use the real thing.

packages/airnode-protocol/contracts/authorizers/mock/MockRequesterAuthorizerWithErc721.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import "../../dev/RequesterAuthorizerWithErc721.sol";
```
Siegrift, line 4: I wonder if we want to copy the contents to the Airnode repo or use it as a dependency. Dependency soundsw better to me.

packages/airnode-validator/src/config/config.ts
```diff
@@ -153,23 +153,51 @@ export const chainOptionsSchema = z
   })
   .strict();
 
+// Authorizations
```
Siegrift, line 156: I am not sure what this comment want to say.

packages/airnode-validator/src/config/config.ts
```diff
@@ -153,23 +153,51 @@ export const chainOptionsSchema = z
...
 export const chainAuthorizationsSchema = z.object({
   requesterEndpointAuthorizations: z.record(endpointIdSchema, z.array(evmAddressSchema)),
 });
 
+// Authorizers: requesterEndpointAuthorizers
```
Siegrift, line 161: I am not sure what this comment want to say.

packages/airnode-validator/src/config/config.ts
```diff
@@ -153,23 +153,51 @@ export const chainOptionsSchema = z
...
 export const crossChainRequesterAuthorizerSchema = z.object({
-  requesterEndpointAuthorizers: requesterEndpointAuthorizersSchema,
+  requesterEndpointAuthorizers: requesterEndpointAuthorizersSchema.nonempty(),
+  chainType: chainTypeSchema,
+  chainId: z.string(),
+  contracts: airnodeRrpContractSchema,
+  chainProvider: providerSchema,
+});
+
+// Authorizers: requesterAuthorizersWithErc721
```
Siegrift, line 172: I am not sure what this comment want to say.

Siegrift (changes requested):

packages/airnode-protocol/contracts/authorizers/mock/MockRequesterAuthorizerWithErc721.sol
dcroote: Burak suggested copy the contents:https://api3workspace.slack.com/archives/C0436H2BJKE/p1677057747622019?thread_ts=1676612819.041809&cid=C0436H2BJKE

packages/airnode-validator/src/config/config.ts
dcroote: This comment (and the below similar ones) were meant as code dividers but I realize they're confusing so I'll remove.

packages/airnode-node/src/evm/authorization/authorization-fetching.test.ts
dcroote: Correct

packages/airnode-node/src/evm/handlers/initialize-provider.ts
dcroote: Good catch, thanks.

packages/airnode-node/src/evm/handlers/initialize-provider.ts
dcroote: Good idea, I added the log to the if/else block below

packages/airnode-protocol/contracts/authorizers/mock/MockRequesterAuthorizerWithErc721.sol
dcroote: Agreed- deleted.

packages/airnode-node/src/evm/authorization/authorization-application.test.ts
dcroote: We previously had no logging when a request was authorized, so even this is an upgrade 😅. I looked into adding the originating authorizer to the log, but that is not trivial to do based on how authorization fetching currently occurs. I think it's doable, but outside the scope of this PR. Happy to open a separate issue though.

packages/airnode-protocol/contracts/authorizers/mock/MockRequesterAuthorizerWithErc721.sol
bbenligiray: I can temporarily publish https://github.com/api3dao/nft-authorizer separately if you want, but to me these feel like equivalent solutions

dcroote: I was convinced by the argument that these will eventually be in this repo anyway.

Siegrift: Yeah, if it's going to be moved I am ok with it. An issue in the Airnode repo would be nice though.

bbenligiray: https://github.com/api3dao/airnode/issues/1675

Siegrift (approved):
👍 LGTM

packages/airnode-node/src/evm/authorization/authorization-application.test.ts
dcroote: https://github.com/api3dao/airnode/issues/1679

PR #1671, opened 2023-03-13
Simplify OEV GW logic

### LINKED ISSUE #1665: Simplify the OEV gateway behavior
The gateway will only receive beacons with signed data (with fields `timestamp`, `encodedValue`, and `signature`) that are considered by the OEV relay as a valid set to be signed. The gateway needs to make sure that:
* there is an absolute majority of them
* all of them are not older than 2 minutes
* all of them have correct `encodedValue` and `encodedParameters`
* all of them have a correct signature
* all of them are within the 1% deviation from their average value
Only if all these conditions are met the data are signed by the gateway. But the gateway is not trying to come up with the subset of beacons that would meet these conditions.

amarthadan:
Close #1665


patriksimurka (approved):
LGTM 👍

Siegrift (approved):
👍 LGTM

PR #1676, opened 2023-03-16
OEV GW timestamp calculation fix

amarthadan:


packages/airnode-node/src/workers/local-gateways/validation.ts
```diff
@@ -250,7 +250,11 @@ export function verifySignOevDataRequest(
     };
   }
 
-  return { success: true, validUpdateValues: map(validDecodedBeacons, 'decodedValue') };
+  return {
+    success: true,
+    validUpdateValues: map(validDecodedBeacons, 'decodedValue'),
```
Siegrift, line 255: Wouldn't it be better to return `validDecodedBeacons` ?

Siegrift (approved):
👍 LGTM

patriksimurka (approved):
LGTM 👍

packages/airnode-node/src/handlers/sign-oev-data.ts
```diff
@@ -80,7 +86,9 @@ export async function signOevData(
     null,
     {
       success: true,
-      data: signatures.map((signature) => ({ signature, timestamp, encodedValue: encodedUpdateValue })),
+      data: signatures.map((signature) => ({
```
Siegrift, line 89: I believe it would be better to return an array of `string | null` for every beacon. (No need to return object with one field). This makes makes in general more sense to me and will also simplify the implementation on the oev-relay.

packages/airnode-node/src/workers/local-gateways/validation.ts
amarthadan: I was thinking about it and decided against it since we only really need two things. If we end up needing/returning more I'll refactor it. It's a detail.

packages/airnode-node/src/handlers/sign-oev-data.ts
amarthadan: I agree that returning a one-field object doesn't make much sense but I don't agree with returning `null`s (as commented https://github.com/api3dao/oev-relay/pull/404/files#r1140277606).

Siegrift: I think it is more explicit which beacons were actually signed (that why I said I believe it's better in general).

For example, say the gateway receives beacons data A1, B1, A2, A3, B2 where A and B are Airnodes. If we later change the algorithm in Airnode to support not signing a particular beacon (e.g. only sign A1 and A3). But I agree with your comment, that it's mostly a cosmetic change so I am OK with leaving this as is.

Siegrift: A bit related to this... Currently, the gateway accepts an array of beacons, but that means you can call it with duplicates of a single beacon. This won't cause Airnode to misreport (or mis-sign) but it's wierd. I think it would make the most sense to accept a record `{[beaconId]: BeaconData}` and make the response be `{[beaconId]: signature}`.

amarthadan: Hmm, that's a fair point. Ok, let's do it this way then. I'll merge this PR and you can make changes in https://github.com/api3dao/airnode/pull/1681.

PR #1664, opened 2023-03-05
Combine airnode-examples http gateway integrations into one

dcroote:
I was doing some testing and again wanted to have a single integration with both HTTP gateways enabled. 

Closes #1635.

The failing documentation job will pass once merged.

Tested both gateways and a local blockchain request with `api3/airnode-client-dev:3a6a5bae75873c64d6a6648a81e852156738e833`.

PR #1680, opened 2023-03-17
Fix airnode-examples integration title

dcroote:
I missed this in #1664. This also reorders integrations alphabetically


Siegrift (approved):
👍

PR #1685, opened 2023-03-20
fix: chainId used in erc721CrossChainAuthorizations

dcroote:
In adding descriptions of the new `RequesterAuthorizerWithErc721 ` authorizers in https://github.com/api3dao/api3-docs/pull/1221, I realized I mistakenly used the wrong `chainId` for checking authorization within `crossChainRequesterAuthorizersWithErc721`. 

With cross-chain authorization, we are asking whether the request on chain A is authorized on chain A, not whether the request is authorized on the cross-chain chain B where we are checking the authorization.


Siegrift (approved):
👍 LGTM

Thanks for noticing. It slipped my 👀

PR #1686, opened 2023-03-21
feat: script to bump OIS version references

### LINKED ISSUE #1441: Bump OIS version references every time OIS package is updated
We already have a [script](https://github.com/api3dao/airnode/blob/master/scripts/update-config-fixtures.ts) for bumping node version in tests, example config files, etc. We need something similar for bumping OIS version references because otherwise, some tests will fail as the validator will incorrectly catch an OIS version mismatch.

dcroote:
Closes #1441. Using the script I found a config file that was quite outdated and updated it. I tested this by running the script with an arbitrary version different than the current version (`2.0.0`) and confirming that all files that include `oisFormat` were modified.


scripts/update-ois-version.ts
```diff
@@ -0,0 +1,55 @@
+import { readFileSync, writeFileSync } from 'fs';
+import fg from 'fast-glob';
+import { format } from 'prettier';
+
+if (process.argv.length < 3) {
```
Siegrift, line 5: Maybe it could parse the version from package.json?

scripts/update-ois-version.ts
Siegrift, line 1: Why not add it to `update-config-fixtures.ts`?

scripts/update-ois-version.ts
dcroote: You mean add a new field for OIS version within package.json and pull from there?

scripts/update-ois-version.ts
dcroote: That script is run when we bump the Airnode version, which happens at the end of a release cycle:

https://github.com/api3dao/airnode/blob/960bed4ca0c0723e986ed891611d38b2dbc12e8f/package.json#L33

OIS, on the other hand, is always bumped as part of some PR e.g. new feature within the release cycle.

scripts/update-ois-version.ts
Siegrift: I meant that OIS is already a dependency in the Airnode packages, so we might use that (since the package version dictates the OIS version).

scripts/update-ois-version.ts
Siegrift: Sure, but if the script is executed "as part of some PR e.g. new feature within the release cycle." it will be noop for configs.

scripts/update-ois-version.ts
dcroote: I think we have different ideas for the flow of bumping the OIS version. The way the script is implemented currently:
1. New OIS version gets released e.g. `2.1.0` from the ois repo
2. Run this script in the Airnode repo: `yarn update-ois-version 2.1.0`. This updates the 3 `package.json` files that OIS appears in, as well as all `.json` and `.ts` files where `oisFormat appears`.
3. `yarn run bootstrap && yarn run build`

What it sounds like you're suggesting:
1. Same as above
2. Bump ois package using a conventional approach (building / updating `yarn.lock` happens within here)
3. Run this script minus needing to enter the version and instead pull the version from a `package.json` where ois is a dependency e.g. `airnode-validator/package.json`

I can see it either way.

dcroote: Perhaps @amarthadan you had a flow in mind when creating the issue?

amarthadan: I would probably go with the second one. My reasoning is that the renovate bot will open a PR for the ois package version bump anyway. In the second flow, we just run the script on that branch and update the PR.
If we would go with the first flow I suspect we would end up doing pretty much the same thing because the person releasing ois package is not usually the same bumping Airnode dependencies to do it immediately after :smile:

dcroote: Sounds good, I'll revise.

dcroote:
Failing codeql CI appears unrelated as it is suddenly also affecting [this aws dependency update](https://github.com/api3dao/airnode/actions/runs/4507771382/jobs/7935862540).

Siegrift (approved):
👍 LGTM

PR #1691, opened 2023-03-22
Retry api call

### LINKED ISSUE #1658: Add retries to performApiCall
We want to add 2 retries for the [Airnode API call](https://github.com/api3dao/airnode/blob/e4d3bee6e70bc733f4d95245fa320753d4204061/packages/airnode-node/src/api/index.ts#L200-L219). First retry should take 10s, second should take 20s. Since the [function](https://www.npmjs.com/package/@api3/promise-utils) is not configurable to have differently timed timeouts, we must have 2 different retries.

Requirements

- 1st retry - 10s timeout
- 2nd retry - 20s timeout

Ashar2shahid:
closes #1658 

There are two ways to have 2 attempts one of them is this:
```
  const goRes = await go(() => adapter.buildAndExecuteRequest(options, { timeout: (timeout * 2) / 3 }), {
    retries: 1,
    attemptTimeoutMs: [timeout / 3, (timeout * 2) / 3],
    totalTimeoutMs: timeout,
  });
```

The `go` function will take 10s and 20s for each attempt before it timeouts, however the timeout that is passed to adapter cannot be adjusted dynamically for each attempt which means we have to supply the max timeout which will block the axios socket for an extra 10 seconds in the first attempt if the api call fails

The other way is:
```
  const goRes = await go(() => adapter.buildAndExecuteRequest(options, { timeout: timeout / 3 }), {
    totalTimeoutMs: timeout / 3,
  });
  if (!goRes.success) {
    const goRes = await go(() => adapter.buildAndExecuteRequest(options, { timeout: (timeout * 2) / 3 }), {
      totalTimeoutMs: (timeout * 2) / 3,
    });
    if (!goRes.success) {
      const { aggregatedApiCall } = payload;
      const log = logger.pend('ERROR', `Failed to call Endpoint:${aggregatedApiCall.endpointName}`, goRes.error);
      // eslint-disable-next-line import/no-named-as-default-member
      const axiosErrorMsg = axios.isAxiosError(goRes.error) ? errorMsgFromAxiosError(goRes.error) : '';
      const errorMessage = compact([RequestErrorMessage.ApiCallFailed, axiosErrorMsg]).join(' ');
      return [[log], { success: false, errorMessage: errorMessage }];
    }
    return [[], { ...goRes.data }];
  }
  ```
  
 This doesn't block but has extra lines of code

### CODE REVIEW

**Comment by Ashar2shahid** [comment]
@bdrhn9 @amarthadan thoughts on which is better ?

**Comment by amarthadan** [comment]
I see. Well, if we want to avoid Axios blocking file descriptors we should go with option 2. I don't think having those few extra lines is a big deal.

File: packages/airnode-node/src/api/index.ts [hunk 1]
```diff
@@ -203,16 +203,24 @@ export async function performApiCall(
   const options = buildOptions(payload);
   const timeout = API_CALL_TIMEOUT;
   // We also pass the timeout to adapter to gracefully abort the request after the timeout
-  const goRes = await go(() => adapter.buildAndExecuteRequest(options, { timeout }), {
-    totalTimeoutMs: timeout,
+  // timeout passed to adapter will cause axios socket to hang until the timeout is reached
```
Siegrift, line 206: Missing line stop.
```suggestion
  // . Timeout passed to adapter will cause axios socket to hang until the timeout is reached
```

packages/airnode-node/src/api/index.ts
```diff
@@ -203,16 +203,24 @@ export async function performApiCall(
...
+  // even if the attemptTimeoutMs is reached and the 2nd attempt is made
+  const goRes = await go(() => adapter.buildAndExecuteRequest(options, { timeout: timeout / 3 }), {
+    totalTimeoutMs: timeout / 3,
   });
   if (!goRes.success) {
```
Siegrift, line 211: nit: prefer early return and avoid nesting.

Siegrift (changes requested):
The extra complexity with `TIMEOUT/3` and `TIMEOUT*2/3` is not worth it. I think it's better to define two constants FIRST_API_CALL_TIMEOUT and SECOND_API_CALL_TIMEOUT.

packages/airnode-node/src/api/index.ts
```diff
@@ -201,21 +201,29 @@ export async function performApiCall(
   payload: ApiCallPayload
 ): Promise<LogsData<ApiCallErrorResponse | PerformApiCallSuccess>> {
   const options = buildOptions(payload);
-  const timeout = API_CALL_TIMEOUT;
-  // We also pass the timeout to adapter to gracefully abort the request after the timeout
-  const goRes = await go(() => adapter.buildAndExecuteRequest(options, { timeout }), {
-    totalTimeoutMs: timeout,
+  const firstTimeout = FIRST_API_CALL_TIMEOUT;
```
Siegrift, line 204: Just inline the constants. No need to redefine these variables.

packages/airnode-node/src/api/index.ts
```diff
@@ -201,21 +201,29 @@ export async function performApiCall(
...
+  const secondTimeout = SECOND_API_CALL_TIMEOUT;
+  // We also pass the timeout to adapter to gracefully abort the request after the timeout.
+  // timeout passed to adapter will cause axios socket to hang until the timeout is reached
```
Siegrift, line 207:
```suggestion
  // Timeout passed to adapter will cause axios socket to hang until the timeout is reached
```

packages/airnode-node/src/api/index.ts
```diff
@@ -201,21 +201,29 @@ export async function performApiCall(
...
+  // even if the totalTimeoutMs is reached and the 2nd attempt is made
+  const goRes = await go(() => adapter.buildAndExecuteRequest(options, { timeout: firstTimeout }), {
```
Siegrift, line 209:
```suggestion
  const goAttempt1 = await go(() => adapter.buildAndExecuteRequest(options, { timeout: firstTimeout }), {
```

packages/airnode-node/src/api/index.ts
```diff
@@ -201,21 +201,29 @@ export async function performApiCall(
...
-    // eslint-disable-next-line import/no-named-as-default-member
-    const axiosErrorMsg = axios.isAxiosError(goRes.error) ? errorMsgFromAxiosError(goRes.error) : '';
-    const errorMessage = compact([RequestErrorMessage.ApiCallFailed, axiosErrorMsg]).join(' ');
-    return [[log], { success: false, errorMessage: errorMessage }];
+  if (goRes.success) {
+    return [[], { ...goRes.data }];
   }
-
-  return [[], { ...goRes.data }];
+  const goRes2 = await go(() => adapter.buildAndExecuteRequest(options, { timeout: secondTimeout }), {
```
Siegrift, line 215:
```suggestion
  const goAttempt2 = await go(() => adapter.buildAndExecuteRequest(options, { timeout: secondTimeout }), {
```

Siegrift (approved):
👍 LGTM after the naming improvements.

PR #1693, opened 2023-03-23
Estimate gas for RRP fulfillments

### LINKED ISSUE #1589: Have RRP fulfillments estimate gas
Currently we have a [`fulfillmentGasLimit` field in `config.chains`](https://docs.api3.org/airnode/v0.9/reference/deployment-files/config-json.html#options-fulfillmentgaslimit) that is by default 500,000. This means you need a minimum of `500,000  *gasPrice` of the native currency in your sponsor wallet for it to be able to respond to a request. Considering that a fulfillment typically costs less than 100,000 gas, having to send the extra 400,000 to the sponsor wallet becomes very annoying.

The reasoning behind hardcoding the gas limit was to minimize the number of blockchain provider calls. This is also why we hardcode the chain ID in the config and only call `getTransactionCount()` once per sponsor wallet per coordinator invocation. However, we still follow this up with [a static call](https://github.com/api3dao/airnode/blob/34f94258e48f8f1dfe7e7e836539effceb4dd576/packages/airnode-node/src/evm/fulfillments/api-calls.ts#L122) to see if the fulfillment will revert. I propose replacing this static test call with a gas limit estimation call for the happy path. [Gas estimation calls are typically more expensive for the blockchain provider than regular static calls](https://stackoverflow.com/a/55431746/14558682), but they are by no means exotic (any transaction where the gas limit is not specified is preceded with a gas estimation) and assuming that the main issue is I/O, replacing a blockchain provider call with another should be fine.

So I'm proposing that in the case that the gas limit is not specified, a gas estimation call to be done first. If the estimation is successful, skip the static test call because it's guaranteed to not revert (otherwise gas estimation would have failed). If the estimation errors, follow up with the static test call to get the revert string, which will be submitted along with the failure. As a result, the fulfillment will be done with a gas limit that is just enough.

Finally, let me give the context about oracles and forced gas limits. There is this idea that relayers can trick their client to believe that their meta-tx reverts (which still results in the relayer being paid in most implementations), even though [the reason it reverts is that the relayer didn't provide enough gas](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/98967feeb294f26e8d99dcac57f340669ad7c09e/contracts/metatx/MinimalForwarder.sol#L59). The same issue is addressed in a less leet way by [not accepting fulfillments that don't come with a large enough gas limit](https://github.com/smartcontractkit/chainlink/blob/0518d8cf4c1bb1cb0d6b31b53a4c3c3915eb7ea1/contracts/src/v0.4/Oracle.sol#L171). In the Airnode protocol-context this is an irrelevant issue because: (1) A requester of a specific Airnode already trusts it with the delivery of the data, which makes a lot of sense in the first-party context because they already trust the API provider with the correctness of the data, (2) Fulfillments are already paid for by the requester so there is no incentive for the Airnode to minimize the gas burn, which is what the relayers perform this attack for. The Airnode operator can just claim the funds in the sponsor wallet, which is already accepted as a tradeoff.

bdrhn9:
Closes #1589. 

Since the issue was described well, I have implemented what it says. With unit-tests, covered all scenarios that can occur. Whole workflow is following:

* If fulfillmentGasLimit is defined in config.json,
  * It works like this PR wasn't added
* If fulfillmentGasLimit isn't defined in config.json, try to estimate gas,
  * If estimation was successful, submit fulfillment.
  * If estimation wasn’t successful, try to call statically to get revert string, if static call was successful unexpectedly, submit fail along with the gas estimation error. This scenario is very unlikely and probably it’ll occur because of RPC issues.
  * If estimation wasn’t successful, try to call statically to get revert string, if static call wasn’t successful as expected, submit fail along with revert string

### DIFF
--- a/packages/airnode-node/src/evm/fulfillments/api-calls.test.ts
+++ b/packages/airnode-node/src/evm/fulfillments/api-calls.test.ts
@@ -3,10 +3,14 @@ import { mockEthers } from '../../../test/mock-utils';
 const failMock = jest.fn();
 const fulfillMock = jest.fn();
 const staticFulfillMock = jest.fn();
+const estimateFulfillMock = jest.fn();
 mockEthers({
   airnodeRrpMocks: {
     callStatic: {
       fulfill: staticFulfillMock,
+    },
+    estimateGas: {
+      fulfill: estimateFulfillMock,
     },
     fail: failMock,
     fulfill: fulfillMock,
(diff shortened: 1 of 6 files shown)


packages/airnode-node/src/evm/fulfillments/api-calls.ts
```diff
@@ -151,6 +183,65 @@ async function testAndSubmitFulfill(
...
+      airnodeRrp,
+      updatedRequest,
+      estimateGasErr?.message ?? RequestErrorMessage.GasEstimationFailed,
+      options
+    );
+    return [[...estimateGasLogs, ...testLogs, ...submitLogs], submitErr, submittedRequest];
+  }
+
+  // If gas estimation is success, submit fulfillment without making static test call
+  const gasLimitNoticeLog = logger.pend('INFO', `Gas limit is set to ${estimateGasData} for Request:${request.id}...`);
```
dcroote, line 237: Minor nit, but the three trailing periods should be removed because this log reports an outcome (the gas limit), not that an action is occurring (e.g. that the gas limit is being estimated like in the DEBUG log).

dcroote (commented):
I tested locally with hardhat and the `api3/airnode-client-dev:b78233ac011b696ae8b9a84942f69b6bd79bd6a9` image and it worked 👍

packages/airnode-node/src/evm/fulfillments/api-calls.test.ts
```diff
@@ -526,8 +240,854 @@ describe('submitApiCall', () => {
...
+            { level: 'DEBUG', message: `Attempting to fulfill API call for Request:${apiCall.id}...` },
+            { level: 'INFO', message: `Submitting API call fulfillment for Request:${apiCall.id}...` },
+          ]);
+          expect(err).toEqual(null);
+          expect(data).toEqual({
+            ...apiCall,
+            fulfillment: { hash: '0xtransactionId' },
+          });
+          expect(staticFulfillMock).toHaveBeenCalledTimes(1);
+          expect(staticFulfillMock).toHaveBeenCalledWith(
```
Siegrift, line 284: @bdrhn9 These are the `toHaveBeenCalledWith` I meant to get rid of. This is basically repeated everytime and I'd argue it doesn't bring much value.

Issue #1690, opened 2023-03-22
Ensure beacon set ID is derived from ordered beacons in OEV gateway


Siegrift:
Beacon order is important when deriving a beacon set ID. The gateway receives beacons record and needs to order the beacons before deriving the beacon set ID.

Either sort the beacons, or make the order be provided via the request. If this ends up changing the request body/response do not forget to update relevant documentations.

bbenligiray:
I wouldn't enforce this to be consistent with the contract implementation

PR #1694, opened 2023-03-23
Revert back to using beacon array for OEV gateway request/response

Siegrift:
Closes https://github.com/api3dao/airnode/issues/1690

Beacons order is important when deriving beacon set ID. This was discussed https://api3workspace.slack.com/archives/C043YH02PGQ/p1679575311162519 and also a bit with @amarthadan privately. 

The decision was to NOT enforce this and be aligned with the (loose) contract implementation. We agreed to revert back to using arrays for request/response body. It is the responsibility of the OEV gateway caller to pass the beacons in correct order.

### DIFF
--- a/packages/airnode-deployer/terraform/aws/templates/oevGw.yaml.tpl
+++ b/packages/airnode-deployer/terraform/aws/templates/oevGw.yaml.tpl
@@ -74,7 +74,7 @@ components:
       type: array
       minItems: 1
       items:
-        type: integer
+        type: string
 
   examples:
     EndpointRequestExample:
--- a/packages/airnode-deployer/terraform/aws/templates/oevGw.yaml.tpl
+++ b/packages/airnode-deployer/terraform/aws/templates/oevGw.yaml.tpl
@@ -40,8 +40,10 @@ components:
         bidAmount:
           type: string
         beacons:
-          type: object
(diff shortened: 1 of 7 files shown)


Siegrift:
I tested on AWS and it works.

packages/airnode-deployer/terraform/aws/templates/oevGw.yaml.tpl
```diff
@@ -69,9 +71,10 @@ components:
                     type: string
 
     EndpointResponse:
-      type: object
-      additionalProperties:
-        type: string
+      type: array
+      minItems: 1
+      items:
+        type: integer
```
amarthadan, line 77: This should be `string` shouldn't it? It's an array of signatures.

amarthadan (approved):

packages/airnode-deployer/terraform/aws/templates/oevGw.yaml.tpl
Siegrift: Thanks.

Siegrift:
The QL check is broken for all PRs as of now and has already been raised https://github.com/api3dao/airnode/pull/1686#issuecomment-1482260623

PR #1702, opened 2023-03-28
Fix Terraform secrets parsing & URLs propagations

amarthadan:
Close #1698
Also solves issues raised by Mertcan (https://api3workspace.slack.com/archives/C02AYRX8D89/p1679929209604659)

Needs to be backported to v0.10 as well.

PR #1703, opened 2023-03-28
v0.10 - Fix Terraform secrets parsing & URLs propagations

amarthadan:
Backport of https://github.com/api3dao/airnode/pull/1702

### DIFF
--- a/.changeset/fifty-otters-heal.md
+++ b/.changeset/fifty-otters-heal.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-deployer': patch
+---
+
+Fix propagating of gateway URLs to heartbeat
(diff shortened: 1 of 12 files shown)


amarthadan:
The failed documentation check is caused by two URLs that have been removed:
* https://github.com/api3dao/airnode/tree/master/packages/airnode-examples/integrations/coingecko-signed-data
* https://github.com/api3dao/airnode/tree/master/packages/airnode-examples/integrations/coingecko-testable

The problem is caused by removing the URLs from the README of Airnode examples. To fix this properly, each version would have to point to its source tree on GitHub which can get complicated. I would just ignore it for the v0.10 branch as it's not causing any problems.

Siegrift (approved):
👍 LGTM

PR #1705, opened 2023-03-29
Fix Terraform secrets parsing for older TF version

amarthadan:
I needed to use `regex` function instead of `startswith` because the older Terraform 0.12 used for Airnode v0.10 doesn't support it.

Also includes fixes for Docker image builds.

The new regex explenation:
Screenshot_20230329_175437

### DIFF
--- a/packages/airnode-deployer/terraform/aws/modules/function/variables.tf
+++ b/packages/airnode-deployer/terraform/aws/modules/function/variables.tf
@@ -10,7 +10,7 @@ locals {
   # Trim whitespaces from the line
   secrets_lines_trimmed = [for line in local.secrets_lines : trimspace(line)]
   # Discard commented lines (starting with '#')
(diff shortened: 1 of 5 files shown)


packages/airnode-deployer/terraform/aws/modules/function/variables.tf
```diff
@@ -10,7 +10,7 @@ locals {
   # Trim whitespaces from the line
   secrets_lines_trimmed = [for line in local.secrets_lines : trimspace(line)]
   # Discard commented lines (starting with '#')
-  secrets_lines_uncommented = [for line in local.secrets_lines_trimmed : line if !startswith(line, "#")]
+  secrets_lines_uncommented = [for line in local.secrets_lines_trimmed : line if can(regex("^[^#]+.*$", line))]
```
Siegrift, line 13: Nit: The `+` is not necessary in that regex.

Siegrift (approved):
👍 LGTM, but didn't test.

PR #1720, opened 2023-04-06
Release v0.11.0

amarthadan:

PR #1725, opened 2023-04-10
Add vi(m) swap files to .gitignore

dcroote:
The motivation originates from: https://github.com/api3dao/airnode/pull/1720#pullrequestreview-1377054704


amarthadan (approved):
:+1: Please, don't merge it until v0.11.0 is released so I don't need to recreate #1720. I'll remove the `.swp` file in that PR though.

PR #1726, opened 2023-04-10
Fix order of commands in packaging script

amarthadan:
The problem was that the dependencies were installed before switching the branch causing potential incompatibility.

I'll merge this once 0.11.0 is released (so I don't need to recreate #1720) but I will build the releasing container with the fix in place.


dcroote (approved):
LGTM 👍

PR #1719, opened 2023-04-06
Release v0.10.1

amarthadan:

PR #1727, opened 2023-04-10
Fix hardcoded `deploymentId` in tests & prettier behavior

amarthadan:
Backporting https://github.com/api3dao/airnode/pull/1653 and https://github.com/api3dao/airnode/pull/1656 to fix failing [v0.10.1 release](https://github.com/api3dao/airnode/pull/1719)


dcroote (approved):
Backport looks good. 

...though at the risk of driving you even crazier on this release 😅  the failing documentation CI is real. Two `airnode-examples` integration links are broken for v0.10 because the URLs have `master` in them and the integrations were combined for v0.11 and therefore no longer exist. Specifically:

https://github.com/api3dao/airnode/blob/4502238823993b43cdd16e13ea1cd3bcbca361bf/packages/airnode-examples/README.md?plain=1#L45

and

https://github.com/api3dao/airnode/blob/4502238823993b43cdd16e13ea1cd3bcbca361bf/packages/airnode-examples/README.md?plain=1#L49

could be fixed by replacing `master` in the URL with `v0.10`, or I suppose with a local link e.g. 

```markdown
 - [coingecko-signed-data](integrations/coingecko-signed-data/README.md)
```

Which maybe we should do as part of a separate PR as well?

Issue #1716, opened 2023-04-05
Implement AirnodeRrpV0DryRun.sol


bbenligiray:
See https://api3workspace.slack.com/archives/C02AYRX8D89/p1680531711817959

We need this contract
```solidity
contract AirnodeRrpV0DryRun {
    function fulfill(
        bytes32 requestId,
        address airnode,
        address fulfillAddress,
        bytes4 fulfillFunctionId,
        bytes calldata data,
        bytes calldata signature
    ) external override returns (bool callSuccess, bytes memory callData) {
        require(
            keccak256(
                abi.encodePacked(
                    airnode,
                    msg.sender,
                    fulfillAddress,
                    fulfillFunctionId
                )
            ) == requestIdToFulfillmentParameters[requestId],
            "Invalid request fulfillment"
        );
        require(
            (
                keccak256(abi.encodePacked(requestId, data))
                    .toEthSignedMessageHash()
            ).recover(signature) == airnode,
            "Invalid signature"
        );
        (callSuccess, callData) = address(0).call(
            abi.encodeWithSelector(fulfillFunctionId, requestId, data)
        );
        emit FulfilledRequest(airnode, requestId, data);
    }
}
```
to be implemented and deployed on all chains that AirnodeRrpV0.sol is deployed on

bbenligiray:
We may want to cover https://github.com/api3dao/airnode/issues/1717 first to not require users to specify the AirnodeRrpV0DryRun address in their config

bbenligiray:
I don't think this needs documentation. This will be tested in https://github.com/api3dao/airnode/issues/1589

Issue #1734, opened 2023-04-12
Use @api3/chains in airnode-protocol


bbenligiray:
airnode-protocol is currently using the `credentials.json`-based system. It needs to be migrated to use the new `@api3/chains` package instead.

PR #1737, opened 2023-04-12
Add AirnodeRrpV0DryRun

bbenligiray:
Closes https://github.com/api3dao/airnode/issues/1716

See the contract comments. I'll address deployment separately, potentially along with https://github.com/api3dao/airnode/issues/1734


packages/airnode-protocol/contracts/rrp/AirnodeRrpV0DryRun.sol
```diff
@@ -0,0 +1,110 @@
...
+            ).recover(signature) == airnode,
+            "Invalid signature"
+        );
+        // We cannot call `fulfillAddress` below because (1) we do not want
+        // this function to actually fulfill the request (2) the fulfill
+        // function will be behind an `onlyAirnodeRrp` modifier and will reject
+        // the calls from AirnodeRrpV0DryRun.
+        // Instead, we call an address that we know to not contain any
+        // bytecode, which means this call will not revert or spend extra gas.
+        (callSuccess, callData) = ADDRESS_WITH_NO_BYTECODE.call( // solhint-disable-line avoid-low-level-calls
```
bbenligiray, line 96: I didn't want to use `address(0)` since some chains may have a contract at that address for whatever reason

acenolaza: I was wondering just that 👍🏻 
We can also safely assume that `airnode` and `msg.sender` are EOAs without bytecode, wouldn't those be an alternative as well?

packages/airnode-protocol/test/rrp/AirnodeRrpV0DryRun.sol.js
```diff
@@ -0,0 +1,176 @@
...
+          rrpRequester.address,
+          rrpRequester.interface.getSighash('fulfill'),
+          fulfillData,
+          signature
+        );
+        // The estimated gas will be ~40,000. This number can slightly change
+        // with future forks, and differ greatly between roll-ups and regular
+        // chains.
+        expect(estimatedGas.toNumber()).to.greaterThan(30000);
+        expect(estimatedGas.toNumber()).to.lessThan(50000);
```
acenolaza, line 114: 👍🏻 
Just being curious, what would the gas more or less be when calling `estimateGas` on `AirnodeRrpV0`? I wonder if there is any value in comparing the 2 values in another `expect()` just to be safe.

acenolaza (approved):
LGTM 👍🏻

packages/airnode-protocol/contracts/rrp/AirnodeRrpV0DryRun.sol
bbenligiray: I think you can have `msg.sender` be a contract (if you can get the requester and the Airnode to play along), but I see how `airnode` is guaranteed to be an EOA. Do you think that we should use that instead?

packages/airnode-protocol/test/rrp/AirnodeRrpV0DryRun.sol.js
acenolaza: Yes, I agree. I'm just trying to see if we can find a more future-proof way of testing the estimageGas() response by using a relative comparison instead of hardcoded numbers that may become obsolete on different chains, etc. But I guess we can do that once (and if) this becomes an issue.

packages/airnode-protocol/contracts/rrp/AirnodeRrpV0DryRun.sol
acenolaza: > would allow one to build trustless sponsor wallets using https://eips.ethereum.org/EIPS/eip-3074 in the future

🤯 

> Do you think that we should use that instead?

I don't have a strong opinion about it. It is very unlikely that ADDRESS_WITH_NO_BYTECODE will collide with a smart contract address anyways so I think it's fine too.

packages/airnode-protocol/test/rrp/AirnodeRrpV0DryRun.sol.js
bbenligiray: I was mainly testing that it returns some kind of a number without erroring here. I don't think the value is very important as long as it's something completely wrong like `-5` or `10000000000000000000`.

packages/airnode-protocol/contracts/rrp/AirnodeRrpV0DryRun.sol
```diff
@@ -0,0 +1,107 @@
...
+/// make reverts. In contrast, AirnodeRrpV0 does not revert when the external
+/// call during the fulfillment reverts, and instead fails gracefully by
+/// emitting a `FailedRequest` event. This event signals to the future
+/// invocations of the stateless Airnode to not retry the failed fulfillment.
+/// Although this approach meets the intended purpose, it disables Airnode from
+/// calling `eth_estimateGas` on `fulfill()` to estimate the gas amount that
+/// will be used to execute a fulfillment successfully. Specifically, since
+/// `eth_estimateGas` looks for the lowest gas limit that results in the
+/// transaction not reverting, and AirnodeRrpV0's `fulfill()` does not revert
+/// when its external call reverts (because it runs out of gas),
```
Ashar2shahid, line 18: are you saying it will always return the gas estimate for when the airnode emits `Failed` ?

Ashar2shahid (approved):

packages/airnode-protocol/contracts/rrp/AirnodeRrpV0DryRun.sol
bbenligiray: Yes. This was originally picked up by e2e tests for https://github.com/api3dao/airnode/issues/1589 was failing.

PR #1755, opened 2023-05-01
Make AirnodeRrpV0 and RequesterAuthorizerWithErc721 addresses optional

### LINKED ISSUE #1717: Make all protocol contract addresses optional
See https://api3workspace.slack.com/archives/C02AYRX8D89/p1680603934257579

Currently, the user needs to enter all protocol contract addresses in the config file. This is awkward because @api3/airnode-protocol already knows what these addresses are, and there is no realistic use-case that requires the user to specify these themselves. Therefore, we should make these optional, and default to our deployments of the protocol contract addresses when they are left unspecified.

dcroote:
Closes #1717


.changeset/selfish-experts-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+'@api3/airnode-validator': minor
+---
+
+Default to deployed AirnodeRrpV0 if contracts omitted from config
```
Siegrift, line 5:
```suggestion
Default to deployed AirnodeRrpV0 if contracts are omitted from config
```

packages/airnode-validator/src/config/config.test.ts
```diff
@@ -14,10 +15,13 @@ import {
   gasPriceOracleSchema,
   localOrCloudProviderSchema,
   chainConfigSchema,
+  crossChainRequesterAuthorizerSchema,
 } from './config';
 import { version as packageVersion } from '../../package.json';
 import { SchemaType } from '../types';
 
+const AirnodeRrpV0: { [chainId: string]: string } = references.AirnodeRrpV0;
```
Siegrift, line 23:
```suggestion
const AirnodeRrpV0Addresses: { [chainId: string]: string } = references.AirnodeRrpV0;
```

packages/airnode-validator/src/config/config.test.ts
```diff
@@ -642,6 +646,75 @@ describe('chainConfigSchema', () => {
...
+      chainIdField: 'chainId',
+    },
+  ].forEach((obj) => {
+    const { testName, schema, configObject, chainIdField } = obj;
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { contracts, ...objectMissingContracts } = configObject;
+
+    it(`fails if AirnodeRrp contract address within ${testName}.contracts is not specified and there is no deployment for the chain`, () => {
+      const missingChainId = '-5';
+      const chainMissingChainId = {
```
Siegrift, line 670: The naming feels a bit wierd (saying the chain is missing - I know you mean from deployed references, but the context is not there). I'd just call it `unknownChain` or `nonDeployedChain` or `nonApi3Chain`.

```suggestion
      const unknownChain = {
```

packages/airnode-validator/src/config/config.ts
```diff
@@ -153,39 +157,66 @@ export const chainOptionsSchema = z
   })
   .strict();
 
+const defaultAirnodeRrp: SuperRefinement<{
+  id?: ChainId;
+  chainId?: ChainId;
+  contracts: SchemaType<typeof airnodeRrpContractSchema>;
+}> = (chainConfig, ctx) => {
+  if (chainConfig.contracts.AirnodeRrp === '') {
```
Siegrift, line 165: Why no:

```suggestion
  if (!chainConfig.contracts.AirnodeRrp) {
```

packages/airnode-validator/src/config/config.ts
```diff
@@ -153,39 +157,66 @@ export const chainOptionsSchema = z
...
+        code: z.ZodIssueCode.custom,
+        message:
+          `AirnodeRrp contract address must be specified for chain ID '${id}'` +
+          `as there was no deployment for this chain found in @airnode/protocol/deployments/references.json`,
+        path: ['contracts'],
+      });
+      return;
+    }
+    // Default to the deployed AirnodeRrp contract address if contracts is not specified
+    chainConfig.contracts = { AirnodeRrp: AirnodeRrpV0[id] };
```
Siegrift, line 179: This seems to me is not the way refinement should be used (to mutate to object). Am I wrong about this?

packages/airnode-validator/src/config/config.ts
```diff
@@ -153,39 +157,66 @@ export const chainOptionsSchema = z
...
-  chainId: z.string(),
-  contracts: airnodeRrpContractSchema,
-  chainProvider: providerSchema,
-});
+export const crossChainRequesterAuthorizerSchema = z
+  .object({
+    requesterEndpointAuthorizers: requesterEndpointAuthorizersSchema.nonempty(),
+    chainType: chainTypeSchema,
+    chainId: chainIdSchema,
+    // The type system requires optional() be transformed to the expected type despite a value being provided by superRefine
```
Siegrift, line 194: Yeah, I was wondering about this in https://github.com/api3dao/airnode/pull/1755/files#r1182261043 . You should do:
1. Use `.default` to provide the default value
2. Use super refine to throw if you detect a chain for which we don't have deployment

Siegrift (changes requested):
I did a quick sweep and I think we can implement this a bit more cleanly - esp. https://github.com/api3dao/airnode/pull/1755/files#r1182262839

packages/airnode-validator/src/config/config.ts
dcroote: I tried this approach initially: `default` won't work here because the default address we're providing depends on another field, the ID of the chain, and that is only available via `superRefine`.

Siegrift: I see. Then I recommend doing [transform](https://zod.dev/?id=transform) after doing the refinements.

dcroote: Do you mean not use `transform` after `.optional()` with `contracts`? Sadly that is needed too because using `optional()` alone causes problems in `airnode-node` and elsewhere because the type system thinks that the value can be undefined (doesn't know it always will be present because of `superRefine`). In other words, `transform` provides assurance that there will be a correctly typed value for the `contracts` field.

packages/airnode-validator/src/config/config.test.ts
```diff
@@ -642,6 +646,75 @@ describe('chainConfigSchema', () => {
...
+    readFileSync(join(__dirname, '../../test/fixtures/interpolated-config.valid.json')).toString()
+  );
+
+  // The field used to specify chain ID varies by schema
+  [
+    { testName: 'chains', schema: chainConfigSchema, configObject: { ...config.chains[0] }, chainIdField: 'id' },
+    {
+      testName: 'crossChainRequesterAuthorizer',
+      schema: crossChainRequesterAuthorizerSchema,
+      configObject: { ...config.chains[0].authorizers.crossChainRequesterAuthorizers[0] },
```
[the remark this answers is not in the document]
dcroote: Here, two objects (schemas) within the config are tested (note the `forEach`): one corresponding to `{ ...config.chains[0] }` and the other to `{ ...config.chains[0].authorizers.crossChainRequesterAuthorizers[0] }` as each object has a `contracts` field that can be empty.

packages/airnode-validator/src/config/config.test.ts
```diff
@@ -2,6 +2,7 @@ import { readFileSync } from 'fs';
 import { join } from 'path';
 import { ZodError } from 'zod';
 import zip from 'lodash/zip';
+import references from '@api3/airnode-protocol/dist/deployments/references.json';
```
dcroote, line 5: Adding `dist/` was necessary for the `airnode-client` to find the file given the `airnode-protocol` outdir specified in `tsconfig.json` is `dist`:

https://github.com/api3dao/airnode/blob/c43779b969fff7bf6ab7213fd7a212dbcfbf156f/packages/airnode-protocol/src/tsconfig.json#L8C7-L10

but I wonder if there is a cleaner approach to export?

packages/airnode-protocol/package.json
```diff
@@ -47,7 +47,6 @@
     "typescript": "^5.0.4"
   },
   "dependencies": {
-    "@api3/airnode-utilities": "^0.11.0",
```
Siegrift, line 50: Why remove utilities dependendency?

packages/airnode-validator/src/config/config.test.ts
Siegrift: Not sure what "Adding dist/" you mean.

packages/airnode-validator/src/config/config.ts
Siegrift: I meant something like this https://github.com/api3dao/airnode/commit/4a752489b283cb610388d6e437c304ed049f2a3b

I believe should not use mutation inside superRefine, but you can validate during transform: https://zod.dev/?id=validating-during-transform

packages/airnode-protocol/package.json
dcroote: This PR adds `airnode-protocol` as dependency of `airnode-validator`, which created cyclic dependency graph that resulted in Airnode failing to build:

```
Lerna (powered by Nx)   The task graph has a circular dependency

   @api3/airnode-admin:build --> @api3/airnode-protocol:build --> @api3/airnode-utilities:build --> @api3/airnode-validator:build --> @api3/airnode-protocol:build
```

The cleanest way to solve this was to remove the `airnode-utilities` dependency since that was only imported for logging within a single script. h/t @amarthadan

Siegrift: I see. The utilities package probably should not depend on validator - but that should not be part of this issue.

packages/airnode-validator/src/config/config.ts
dcroote: Thanks for the template and link, I switched from superRefine to transform as you suggest

packages/airnode-validator/src/config/config.test.ts
dcroote: Ignore- I fixed this import by exporting references from airnode-protocol

dcroote:
Who would have thought providing a few defaults would be such a pain. Anyway, this PR now allows omission of the addresses for `AirnodeRrpV0` and `RequesterAuthorizerWithErc721` when we've deployed those contracts on a given chain. To do this, we look up those addresses exported from `@api3/airnode-protocol` and throw an error if a user omits the address for a chain where we haven't deployed. This still preserves the ability to specify the address for a chain independent of whether there is a deployment and so is backwards compatible.

Siegrift (approved):
👍 LGTM. After fixing the TS inference in validator, this should be good to go.

packages/airnode-validator/src/config/config.ts
```diff
@@ -153,25 +159,51 @@ export const chainOptionsSchema = z
   })
   .strict();
 
+const ensureValidAirnodeRrp = (
+  value: any, // Typing the value as any because specifying it will break Zod inference
```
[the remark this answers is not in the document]
dcroote: Thanks- fixing the type inference was a bit of a pain but I agree it was a necessity to preserve typing elsewhere in Airnode

Issue #1711, opened 2023-04-04
Implement the gas price strategy: Sanitized providerRecommendedGasPrice


bbenligiray:
We tended to use the gas price strategy `providerRecommendedGasPrice` in Airseeker. However, this caused some data feed updates with very large gas prices on Fantom. Upon realizing that many chains report a (correctly calculated) block base fee even if they don't support EIP1559 (such as Fantom), we came up with a new way to sanitize the provider recommended gas price as in
```
providerRecommendedGasPrice = recommendedGasPriceMultiplier * getGasPrice();
// Normally we would be done here
baseFee = getBlock().baseFee;
if (providerRecommendedGasPrice > baseFee * A) {
  return baseFee * B + C;
} else {
  return providerRecommendedGasPrice;
}
```
where example values are A=5, B=2, C=3 gwei.

Here's the reasoning: During normal operation, `providerRecommendedGasPrice` will always be less than `5 * baseFee`, in which case we continue using `providerRecommendedGasPrice` as usual. `providerRecommendedGasPrice` is more than `5 * baseFee` in two cases:
1. There is a momentary spike in gas prices, in which case ignoring that and using `baseFee * 2` is perfectly fine (and `baseFee * 2` is a rather aggressive gas price anyway). We don't use `baseFee * 5` because `providerRecommendedGasPrice` being this high means `providerRecommendedGasPrice` is not to be trusted at all, and in a situation we don't know anything,  `baseFee * 5` is unreasonably large.
2. `baseFee` is almost zero because the blocks have been empty for a long duration (typical on testnets and deserted mainnets). This means that any gas price should be acceptable if enough of a tip is added (which is why we have the 3 gwei, but this may need to be specified per-chain).

To cover both of these cases, we use `baseFee * 2 + 3 gwei`.

Nodary tested this in production to patch the issue, and this works well on a variety of chains in that it both prevents Airseeker from using spiking gas prices and it also doesn't cause data feed updates to be lagged. It would be good to have this added as an official gas price strategy because, for example, QRNG Airnodes would also benefit a lot from this.

bbenligiray:
If you're not aware, this is already implemented by @bdrhn9 in an Airseeker fork, it just needs to be transferred over to Airnode proper

bdrhn9:
See thread: https://api3workspace.slack.com/archives/C02AYRX8D89/p1684140487018909
For visibility: @bbenligiray

PR #1757, opened 2023-05-03
Update providerRecommendedGasPrice

cserb:
The actual price calculation for `providerRecommendedGasPrice` has not been tested (or I couldn't find it in any tests).
Closes: https://github.com/api3dao/airnode/issues/1711

### DIFF
--- a/.changeset/quiet-walls-repair.md
+++ b/.changeset/quiet-walls-repair.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-utilities': patch
+---
+
+Sanitized providerRecommendedGasPrice
--- a/.changeset/quiet-walls-repair.md
+++ b/.changeset/quiet-walls-repair.md
@@ -1,5 +1,0 @@
----
-'@api3/airnode-utilities': patch
----
-
-Sanitized providerRecommendedGasPrice
--- a/packages/airnode-utilities/src/evm/gas-prices/gas-oracle.test.ts
+++ b/packages/airnode-utilities/src/evm/gas-prices/gas-oracle.test.ts
@@ -301,7 +301,7 @@ describe('Gas oracle', () => {
       expect(gasTarget).toEqual(gasOracle.getGasTargetWithGasLimit(providerRecommendedGasPrice, fulfillmentGasLimit));
     });
 
-    it('returns baseFeePerGas if getGasPrice too high', async () => {
+    it('returns baseFeePerGas if getGasPrice is too high', async () => {
       const getGasPriceSpy = jest.spyOn(ethers.providers.StaticJsonRpcProvider.prototype, 'getGasPrice');
       const getGasPriceMock = ethers.BigNumber.from(1000);
       getGasPriceSpy.mockImplementation(() => Promise.resolve(getGasPriceMock));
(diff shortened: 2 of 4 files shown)


dcroote (commented):
Nice Cris! I haven't had the opportunity to thoroughly review, but in the meantime two quick points on the PR generally (besides what you already experienced in the form of CI Prettier formatting and a changeset 😅 )
1. For CI steps that fail, you can click on Details and rerun only the failed jobs. This is necessary occasionally, for example in your last commit where the docs CI failed for a spontaneous reason unrelated to the PR.
2. GitHub requires specific language in order to link a PR to an issue and we leverage that link in managing the status of issues as they move across the [Airnode board](https://github.com/orgs/api3dao/projects/4/views/24). For example, when a PR gets created that closes an issue, that issue will automatically be assigned the status of "In review". So instead of `This implements: #1711`, something like `Closes #1711` ([more examples here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))

Also is there a link to what is referenced in the issue comment?
> this is already implemented by @bdrhn9 in an Airseeker fork, it just needs to be transferred over to Airnode proper

cserb:
> Also is there a link to what is referenced in the issue comment?
> 
> > this is already implemented by @bdrhn9 in an Airseeker fork, it just needs to be transferred over to Airnode proper

Unfortunately no. This was done in a private repo, that @bdrhn9 made available to me as a download. I can share it with you as well if it helps with the review.

aquarat:
@cserb The PR contents now, apart from the inclusion of tests (I assume Bedirhan's work), isn't too different to your first implementation, so well done on that. Also, it's good that you reached out to Bedirhan.

I was surprised that the compiler allowed you to directly compare two BigNumber objects in your first attempt and that it worked, but this version that explicitly uses the `gt` function is less scary 😆 

Thanks for guiding @dcroote 🙏 ❤️

cserb:
> I was surprised that the compiler allowed you to directly compare two BigNumber objects in your first attempt and that it worked, but this version, which explicitly uses the `gt` function is less scary 😆

Honestly the first implementation was specifically to test the CI and to understand what to expect. The direct comparison was not meant to stay :)

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.test.ts
```diff
@@ -293,6 +301,69 @@ describe('Gas oracle', () => {
...
+        provider,
+        providerRecommendedGasPriceStrategy,
+        startTime
+      );
+
+      const expectedGasTarget = baseFeePerGasMock.mul(2).add(3);
+      expect(providerRecommendedGasPrice.gasPrice).toEqual(expectedGasTarget);
+    });
+
+    it('returns getGasPrice if not too high', async () => {
```
dcroote, line 323: "not too high" is a big vague here. Perhaps something along the lines of `returns getGasPrice if the multiplied gas price is less than a multiple of the baseFeePerGas`

dcroote (approved):
Nice - approving with a few minor nits

dcroote (approved):
Nice - approving with a few minor nits

.changeset/quiet-walls-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+'@api3/airnode-utilities': patch
```
dcroote, line 2: One more minor nit: I think this should be minor instead of patch given it is a change in behaviour rather than a bug fix.

dcroote (approved):
LGTM so feel free to merge, though I'd recommend the “Squash and merge" strategy given the WIP and formatting commits.

Issue #1675, opened 2023-03-16
Update RequesterAuthorizerWithErc721 once it is audited


bbenligiray:
RequesterAuthorizerWithErc721 support is being added to Airnode https://github.com/api3dao/airnode/pull/1670 However, this contract isn't audited yet. It is temporarily housed in a `dev/` directory in airnode-protocol, and should be updated and moved once the audit is complete.

bbenligiray:
We can ask someone from the ChainAPI team test this while testing their RequesterAuthorizerWithErc721 integration, for example @PrenaamD

PR #1771, opened 2023-05-16
Update RequesterAuthorizerWithErc721

bbenligiray:
Closes https://github.com/api3dao/airnode/issues/1675

@api3/airnode-protocol-v1@2.4.0 exports the audited RequesterAuthorizerWithErc721 and its addresses. This PR uses these instead of the contracts from `dev/`.

One question to @dcroote , is it a problem for RequesterAuthorizerWithErc721 to only have been deployed on ethereum and ethereum-goerli-testnet regarding https://github.com/api3dao/airnode/issues/1717 ?
Also, should https://github.com/api3dao/airnode/pull/1755 also handle  RequesterAuthorizerWithErc721?


.eslintrc.js
```diff
@@ -65,4 +65,11 @@ module.exports = {
     'require-await': 'error',
     'no-return-await': 'error',
   },
+  settings: {
+    'import/resolver': {
+      node: {
+        paths: ['src'],
+      },
+    },
+  },
```
bbenligiray, line 74: I had to apply https://stackoverflow.com/a/42498220 even though I'm not confident what it really does

packages/airnode-protocol/contracts/dev/RequesterAuthorizerWithErc721.sol
dcroote, line 1: With both this and the `MockErc721.sol` contract deleted, we should be able to remove the hardhat overrides too:

https://github.com/api3dao/airnode/blob/fa3593b7a7e413cee588e59c9df3afda182f9242/packages/airnode-protocol/hardhat.config.js#L58-L77

.eslintrc.js
dcroote: Not entirely sure either, but I know I also fought with ESLint import errors previously in related work so glad there is something that appears to have worked.

dcroote (approved):
Nice- 1 response and 1 minor comment.

As for the other questions:

> is it a problem for RequesterAuthorizerWithErc721 to only have been deployed on ethereum and ethereum-goerli-testnet

Not a problem per se - we'll default to these addresses if the contract address is omitted on these chains. Otherwise, users will have to (deploy and) specify it themselves.

> Also, should https://github.com/api3dao/airnode/pull/1755 also handle RequesterAuthorizerWithErc721?

Yes- this PR is perfectly timed as I can fix the TODOs in that PR once this is merged

packages/airnode-protocol/contracts/dev/RequesterAuthorizerWithErc721.sol
bbenligiray: Thanks, I wasn't aware of this

PR #1772, opened 2023-05-17
Sanitized provider recommended gas price

### LINKED ISSUE #1711: Implement the gas price strategy: Sanitized providerRecommendedGasPrice

cserb:
Closes #1711


cserb:
@bdrhn9 before I continue and also adapt the test, can you confirm this is going in the right direction

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts
```diff
@@ -257,6 +281,8 @@ export const attemptGasOracleStrategy = (
       return fetchLatestBlockPercentileGasPrice(provider, gasOracleConfig, startTime);
     case 'providerRecommendedGasPrice':
       return fetchProviderRecommendedGasPrice(provider, gasOracleConfig, startTime);
+    case 'sanitizedProviderRecommendedGasPrice':
+      return sanitizedProviderRecommendedGasPrice(provider, gasOracleConfig, startTime);
```
bdrhn9, line 285: For consistency, it would be better `sanitizeProviderRecommendedGasPrice` for function naming.

bdrhn9:
Nit again: alternative naming for baseFeePerGas could be baseFee. This leads to shorten all variables include baseFeePerGas like:

baseFeePerGasMultiplier => baseFeeMultiplier
baseFeePerGasMultiplierThreshold => baseFeeMultiplierThreshold

However this suggestion more than nit :-D so feel free to ignore.

cserb:
> Nit again: alternative naming for baseFeePerGas could be baseFee. This leads to shorten all variables include baseFeePerGas like:
> 
> baseFeePerGasMultiplier => baseFeeMultiplier baseFeePerGasMultiplierThreshold => baseFeeMultiplierThreshold baseFeeTip => additionalGasFee 😋
> 
> However this suggestion more than nit :-D so feel free to ignore.

Well maybe not even so much of a nit. I've only now seen that we are using everywhere else `baseFee` and what I called `baseFeeTip` is `priorityTip` for consistency I will change it.

cserb:
Final review @bdrhn9 😄 ?

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts
bbenligiray: `sanitizedProviderRecommendedGasPrice` is better

bdrhn9: I proposed this naming for function naming because all other functions start with the verb `fetch`. However having regard to your recommendation, it'd be okay to name function `fetchProviderRecommendedGasPriceSanitized`. But as you can see, naming for all strategies end with `GasPrice` so  `fetchSanitizedProviderRecommendedGasPrice` is also good option.

bbenligiray: Sorry, I thought the comment was for line 284 and not 285. In that case, yes, it should be `fetchSanitizedProviderRecommendedGasPrice()` for consistency.

`fetchProviderRecommendedGasPriceSanitized()` only makes sense if the strategy was named `providerRecommendedGasPriceSanitized`, which I was suggesting above, but thinking about it again, it doesn't sound right.

aquarat:
Pending manual testing.

packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts
```diff
@@ -104,7 +129,10 @@ export const fetchProviderRecommendedGasPrice = async (
   };
 };
 
-export const fetchBaseFeePerGas = async (provider: Provider, startTime: number): Promise<ethers.BigNumber | null> => {
+export const fetchBaseFeePerGas = async (
+  provider: Provider,
+  startTime: number
+): Promise<ethers.providers.Block['baseFeePerGas']> => {
   const goLatestBlock = await go(() => provider.getBlock('latest'), {
```
bdrhn9, line 136: This function throws error when block header couldn't be fetched. In this case, gasOracle tries to the following strategy if it is defined. In case it's defined, it'll be probably `providerRecommendedGasPrice` strategy and again we make another RPC request to get `providerRecommendedGasPrice`. Or the other case where there is no next strategy defined, gasOracle will use constant gas price which is not safe for all chain. So instead of throwing error when block header couldn't be fetched, returning already fetched and calculated providerRecommendedGasPrice is more efficient and safe way to go.

bdrhn9: Or better and more elegant solution is first making RPC call for block header (throws error if fail) and then making RPC call for `providerRecommendedGasPrice`. So in this case, if block header can't be fetched, gasOracle tries the next strategy which will be probably `providerRecommendedGasPrice`.

bdrhn9:
Because new implementation and testing is necessary, it's better to continue on new PR #1828.

PR #1814, opened 2023-06-22
fix: populate default contract addresses when loading config

### LINKED ISSUE #1809: Default contract addresses are not present after validation
As first reported by @aTeoke [here](https://github.com/api3dao/airnode/issues/1717#issuecomment-1576593582), there is a bug whereby a `config.json` which does not provide a contract address will now pass the validator (if we have a deployment on the specified chain) after https://github.com/api3dao/airnode/pull/1755, but the validator doesn't save the config with the default address added after validating. Therefore, when `loadTrustedConfig` ([src](https://github.com/api3dao/airnode/blob/de609436f911c1b2f82684057cade3c2376f9491/packages/airnode-node/src/config/index.ts#L39-L45)) is called, it loads the config that is still missing the contract address. This causes the [error](https://github.com/api3dao/airnode/issues/1717#issuecomment-1578068817) `ERROR TypeError: Cannot read properties of undefined (reading 'AirnodeRrp') that indicates the AirnodeRrp address is missing.

dcroote:
This PR closes #1809 by populating default contract address when a config file, which may omit contract addresses for which there is an existing deployment, is loaded via `loadTrustedConfig`. Rather than duplicate the code to populate default addresses, this code calls validating functions added to airnode-validator as part of #1755.

### DIFF
--- a/packages/airnode-node/src/config/index.test.ts
+++ b/packages/airnode-node/src/config/index.test.ts
@@ -1,0 +1,43 @@
+import { join } from 'path';
+import { readFileSync } from 'fs';
+import { mockReadFileSync } from '../../test/mock-utils';
+import dotenv from 'dotenv';
+import { references } from '@api3/airnode-protocol';
+import { loadTrustedConfig } from '../index';
+
+const AirnodeRrpV0Addresses: { [chainId: string]: string } = references.AirnodeRrpV0;
+const RequesterAuthorizerWithErc721Addresses: { [chainId: string]: string } = references.RequesterAuthorizerWithErc721;
+
+describe('config validation', () => {
+  const exampleConfigPath = join(__dirname, '../../config/config.example.json');
+  const exampleSecrets = dotenv.parse(readFileSync(join(__dirname, '../../config/secrets.example.env')));
+
+  it('loads the config and adds default contract addresses', () => {
+    const chainId = '4';
+    const invalidConfig = JSON.parse(readFileSync(exampleConfigPath, 'utf-8'));
+    invalidConfig.chains[0].id = chainId; // Need to use an actual chain not hardhat
+    delete invalidConfig.chains[0].contracts;
+    delete invalidConfig.chains[0].authorizers.crossChainRequesterAuthorizers[0].contracts;
+    delete invalidConfig.chains[0].authorizers.crossChainRequesterAuthorizersWithErc721[0].contracts;
+    delete invalidConfig.chains[0].authorizers.requesterAuthorizersWithErc721[0].RequesterAuthorizerWithErc721;
+
+    mockReadFileSync('config.example.json', JSON.stringify(invalidConfig));
+
+    const config = loadTrustedConfig(exampleConfigPath, exampleSecrets);
+
+    expect(config.chains[0].contracts.AirnodeRrp).toEqual(AirnodeRrpV0Addresses[chainId]);
+    expect(config.chains[0].authorizers.crossChainRequesterAuthorizers[0].contracts.AirnodeRrp).toEqual(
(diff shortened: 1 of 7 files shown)


packages/airnode-node/src/config/index.ts
```diff
@@ -25,6 +25,44 @@ const readConfig = (configPath: string): unknown => {
...
+        })
+      : [];
+
+    return {
+      // Add same-chain RequesterAuthorizerWithErc721 contract address, which operates
+      // on the entire chain object
+      ...configTypes.ensureRequesterAuthorizerWithErc721(
+        {
+          ...chain,
+          contracts: contracts,
```
amarthadan, line 51:
```suggestion
          contracts,
```

packages/airnode-node/src/config/index.ts
```diff
@@ -25,6 +25,44 @@ const readConfig = (configPath: string): unknown => {
...
+      ? chain.authorizers.crossChainRequesterAuthorizers.map((craObj) => {
+          return configTypes.ensureCrossChainRequesterAuthorizerValidAirnodeRrp(craObj, ctx);
+        })
+      : [];
+
+    const crossChainRequesterAuthorizersWithErc721 = chain.authorizers.crossChainRequesterAuthorizersWithErc721
+      ? chain.authorizers.crossChainRequesterAuthorizersWithErc721.map((craObj) => {
+          return configTypes.ensureCrossChainRequesterAuthorizerWithErc721(craObj, ctx);
+        })
+      : [];
```
amarthadan, line 43: Are the conditions necessary? The `chain.authorizers.crossChainRequesterAuthorizers` and `chain.authorizers.crossChainRequesterAuthorizersWithErc721` must be present but they can be empty arrays. So, mapping through them even if they are empty will lead to the same thing, an empty array.

packages/airnode-node/src/config/index.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  const exampleSecrets = dotenv.parse(readFileSync(join(__dirname, '../../config/secrets.example.env')));
+
+  it('loads the config and adds default contract addresses', () => {
+    const chainId = '4';
+    const invalidConfig = JSON.parse(readFileSync(exampleConfigPath, 'utf-8'));
+    invalidConfig.chains[0].id = chainId; // Need to use an actual chain not hardhat
+    delete invalidConfig.chains[0].contracts;
+    delete invalidConfig.chains[0].authorizers.crossChainRequesterAuthorizers[0].contracts;
+    delete invalidConfig.chains[0].authorizers.crossChainRequesterAuthorizersWithErc721[0].contracts;
+    delete invalidConfig.chains[0].authorizers.requesterAuthorizersWithErc721[0].RequesterAuthorizerWithErc721;
```
amarthadan, line 22: Why did you populate those fields in the example config if you're then emptying them here?

amarthadan (approved):
Haven't tested it but the code LGTM.

packages/airnode-node/src/config/index.test.ts
dcroote: The authorizers that have been added to the config are objects with several fields- here I'm only deleting the optional contract field in each before then confirming via the test that the deployed address is added.

packages/airnode-node/src/config/index.ts
dcroote: Why is this not a lint 😅

packages/airnode-node/src/config/index.ts
dcroote: Good call 👍

dcroote:
I've tested this with the `api3/airnode-client-dev:97e936799a4a7132c384731c7c78eb1035d125e5` image using the `coingecko-cross-chain-authorizer` integration and get a successful request fulfillment without the previous error `Cannot read properties of undefined (reading 'AirnodeRrp')`

PR #1817, opened 2023-06-26
Re-implement "Estimate gas for RRP fulfillments"

### LINKED ISSUE #1589: Have RRP fulfillments estimate gas

bdrhn9:
Closes #1589  

This PR basically:
- Adds back `Estimate gas for RRP fulfillments` which was reverted and reason had been nicely told in [the Slack thread](https://api3workspace.slack.com/archives/C02AYRX8D89/p1680531711817959)
- Refactors validator to check, validate, populate default address for AirnodeRrpV0DryRun contract
- Separates estimation of RRP overhead and external call (fulfillment call) overhead (AirnodeRrpV0DryRun is being leveraged to estimate overhead for RRP protocol)
- Fixes incorrect chainId where deployments were absent (although this issue allowed the tests to pass, it resulted in an undefined-to-undefined equality check)
- Fixes typo in comment

I think it'll be better to refer the provided Slack thread first. Btw I'm not sure about PR title created by GitHub when reverted the ancestor PR, let me know if it needs to be changed.

Sorry for force pushs, I would like to stick with @dcroote 's commit message protocol :-)


packages/airnode-validator/src/config/config.ts
```diff
@@ -62,7 +64,8 @@ export const chainTypeSchema = z.literal('evm');
 
 export const airnodeRrpContractSchema = z
   .object({
-    AirnodeRrp: evmAddressSchema,
+    AirnodeRrp: evmAddressSchema.optional(),
```
bdrhn9, line 67: Here is a clarification to make it clearer: Although AirnodeRrp shouldn't be optional, I have set both contracts to be optional in this case. The reason behind this decision is to accommodate situations where the user wants to use their own address for one of the contracts, while using default deployments for the other contract. In such cases, the user can define chains.contracts for the desired contract but leave it undefined for the contract they wish to use default deployments for. 

Actually this schema is input schema (see [here](https://github.com/colinhacks/zod#type-inference)) and supposed to parse input config.json. Enforcing this schema with transforms below, inferred output type for this schema will be {AirnodeRrp: evmAddressSchema, AirnodeRrpDryRun: evmAddressSchema.optional()}

packages/airnode-validator/src/config/config.ts
```diff
@@ -153,36 +156,71 @@ export const gasPriceOracleSchema = z
...
   })
   .strict();
 
-export const ensureValidAirnodeRrp = (
-  contracts: SchemaType<typeof airnodeRrpContractSchema>,
-  chainId: string,
-  ctx: RefinementCtx
-) => {
-  if (!contracts?.AirnodeRrp) {
+export const ensureValidAirnodeRrp = (airnodeRrp: string | undefined, chainId: string, ctx: RefinementCtx) => {
```
dcroote, line 165: Won't `airnodeRrp` always be either an address string or an empty string and not undefined given how you've implemented `defaults`?

packages/airnode-validator/src/config/config.ts
```diff
@@ -153,36 +156,71 @@ export const gasPriceOracleSchema = z
...
+  if (!airnodeRrp) {
     if (!AirnodeRrpV0Addresses[chainId]) {
       ctx.addIssue({
         code: z.ZodIssueCode.custom,
         message:
-          `AirnodeRrp contract address must be specified for chain ID '${chainId}'` +
+          `AirnodeRrp contract address must be specified for chain ID '${chainId}' ` +
```
dcroote, line 171: Nice catch here and for the same elsewhere.

packages/airnode-node/src/types.ts
```diff
@@ -194,6 +195,7 @@ export interface CoordinatorStateWithApiResponses extends CoordinatorState {
 export interface EVMContracts {
   // TODO: Rename to airnodeRrp for consistency
   readonly AirnodeRrp: string;
+  readonly AirnodeRrpDryRun?: string | undefined;
```
dcroote, line 198: The `undefined` is redundant to `AirnodeRrpDryRun` being optional and can be removed.

packages/airnode-node/src/evm/fulfillments/api-calls.ts
```diff
@@ -75,6 +77,128 @@ async function testFulfill(
...
+  request: Request<ApiCallWithResponse>,
+  options: TransactionOptions
+): Promise<LogsErrorData<ethers.BigNumber>> {
+  const noticeLog = logger.pend(
+    'DEBUG',
+    `Attempting to estimate AirnodeRrp overhead to fulfill API call for Request:${request.id}...`
+  );
+  if (!request.success) return [[], new Error('Only successful API can be submitted'), null];
+
+  const airnodeRrpDryRunAddress = options.contracts?.AirnodeRrpDryRun;
```
dcroote, line 124: Isn't the `?` character not needed here, particularly given `airnodeRrpDryRunAddress!` just below?

packages/airnode-node/src/evm/fulfillments/api-calls.test.ts
```diff
@@ -526,8 +260,1048 @@ describe('submitApiCall', () => {
...
+
+  describe('estimateGasAndSubmitFulfill', () => {
+    describe('Pending API calls', () => {
+      test.each([gasTargetWithoutGasLimit, gasTargetFallbackWithoutGasLimit])(
+        `successfully estimates gas and submits a fulfill transaction for pending requests - %#`,
+        async (gasTarget) => {
+          const txOpts = { ...gasTarget, nonce: 5 };
+          const provider = new ethers.providers.JsonRpcProvider();
+          estimateAirnodeRrpOverheadMock.mockResolvedValueOnce(ethers.BigNumber.from(73804));
+          estimateFulfillmentCallOverheadMock.mockResolvedValueOnce(ethers.BigNumber.from(290001));
```
dcroote, line 622: Maybe variables for `73804` and `290001`?

dcroote (approved):
Nice work and I have mostly minor comments. For the PR title, perhaps something like "Re-implement estimate gas for RRP fulfillments"?

I was hoping there was going to be a simple way to incorporate this into an E2E test, but I don't see one at the moment.  The `airnode-examples` E2E test might be a good candidate, but it would involve hardhat contract deployment similar to AirnodeRrp e.g. after [this line](https://github.com/api3dao/airnode/blob/0c0c3529a090976040d2ed7e23ab8939e09a6f3c/packages/airnode-examples/test/e2e/coingecko-local.feature.ts#L30).

packages/airnode-node/src/evm/fulfillments/api-calls.ts
```diff
@@ -151,6 +270,65 @@ async function testAndSubmitFulfill(
...
 }
 
+export async function estimateGasAndSubmitFulfill(
+  airnodeRrp: AirnodeRrpV0,
+  request: Request<ApiCallWithResponse>,
+  options: TransactionOptions
+): Promise<LogsErrorData<Request<ApiCallWithResponse>>> {
+  // Should not throw
+  const [estimateGasLogs, estimateGasErr, estimateGasData] = await estimateGasToFulfill(airnodeRrp, request, options);
+  // const [estimateGasLogs, estimateGasErr, estimateGasData] = [[], new Error('cartcut'), [ethers.BigNumber.from(55),ethers.BigNumber.from(65)]];
```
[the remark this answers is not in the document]
bdrhn9: Ups 😞 it shouldn't be here.

packages/airnode-validator/src/config/config.ts
bdrhn9: Yes, you're right. I fixed it.

packages/airnode-node/src/evm/fulfillments/api-calls.test.ts
bdrhn9: Because they are originally ethers.BigNumber, I'm summing them using `.add` method as in [here](https://github.com/api3dao/airnode/blob/64afdf08addbb32b12ed9e5b94047c301a09cb0a/packages/airnode-node/src/evm/fulfillments/api-calls.ts#L319). So using primitive numbers for variables causes error unless we mock them.

PR #1828, opened 2023-07-09
Implement gas price strategy `sanitizedProviderRecommendedGasPrice`

### LINKED ISSUE #1711: Implement the gas price strategy: Sanitized providerRecommendedGasPrice

bdrhn9:
Closes #1711

### DIFF
--- a/.changeset/orange-squids-flow.md
+++ b/.changeset/orange-squids-flow.md
@@ -1,0 +1,6 @@
+---
+'@api3/airnode-utilities': minor
+'@api3/airnode-validator': minor
+---
+
+Implement gas price strategy `sanitizedProviderRecommendedGasPrice`
--- a/packages/airnode-utilities/src/evm/gas-prices/gas-oracle.test.ts
+++ b/packages/airnode-utilities/src/evm/gas-prices/gas-oracle.test.ts
@@ -245,10 +245,6 @@ describe('Gas oracle', () => {
       const getGasPriceMock = ethers.BigNumber.from(11);
       getGasPriceSpy.mockImplementation(() => Promise.resolve(getGasPriceMock));
 
-      const fetchBaseFeePerGasMock = jest.spyOn(gasOracle, 'fetchBaseFeePerGas');
-      const baseFeePerGasMock = ethers.BigNumber.from(10);
-      fetchBaseFeePerGasMock.mockResolvedValue(baseFeePerGasMock);
-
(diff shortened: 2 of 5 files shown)


dcroote (approved):
Nice implementation and tests. My only suggestion is adding `sanitizedProviderRecommendedGasPrice` to the E2E test suite here: `packages/airnode-utilities/test/e2e/gas-oracle.feature.ts`

dcroote:
> I think it is mergeable at that point isn't it?

Agreed 👍

PR #1743, opened 2023-04-18
Bump hardhat from 2.10.2 to 2.14.0

dcroote:
Closes #1736 (I tested `yarn test:coverage` within `airnode-protocol` runs).

Issue #1585, opened 2022-12-02
Fix building of Docker images for project state without changesets


amarthadan:
We're using [Changeset's snapshot packages](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md) to build our dev Docker images. But the Changesets will make changes in `package.json`s **only** if there are some changesets file to process. Even for the snapshot packages. So, on a rare occasion when there's no changeset file, the package can't be released and the Docker image can't be built.

dcroote:
For the avoidance of doubt, an [empty changeset](https://github.com/api3dao/airnode/commit/b516a06f529a204ea4d668bff3875fc05962ffba) still results in failure of the images to build- a changeset needs to [affect a package](https://github.com/api3dao/airnode/commit/117daca77c66866b6c067c61700e38df6371abff) in order to succeed.

PR #1840, opened 2023-07-24
Fix packaging Docker image

amarthadan:
At some point `hardhat` switch from storing the compilers in `.cache/hardhat-nodejs/compilers/` to [`.cache/hardhat-nodejs/compilers-v2`](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/util/global-dir.ts#L93). But changing the directory was not enough. For some reason, providing the downloaded `amd64` version was not enough to avoid the download. I've removed the download completely and added the [`solc-linux-amd64-v0.8.9+commit.e5eed63a.does.not.work`](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/solidity/compiler/downloader.ts#L259) file which is how `hardhat` internally marks "broken" native compiler and switches to the `wasm` one. That one then works as expected, no download required :slightly_smiling_face:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -9,12 +9,13 @@ ENV appDir="/app" \
 RUN apk add --update --no-cache git rsync docker $([ $(arch) == "aarch64" ] && echo "python3 make g++") && \
     yarn global add npm && \
     # Download both solidity compilers as per: https://github.com/nomiclabs/hardhat/issues/1280#issuecomment-949822371
-    mkdir -p /root/.cache/hardhat-nodejs/compilers/wasm && \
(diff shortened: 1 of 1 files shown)


dcroote (approved):
This makes sense as the directory change occurred in v2.12.0 and we jumped from v2.10.2 to v2.14.0 in https://github.com/api3dao/airnode/pull/1743.

`.does.not.work` 😅

amarthadan:
The docker build failing is caused by https://github.com/api3dao/airnode/issues/1585 (we've just released v0.12.0, so there are no changes). It's not affecting anything, merging.

PR #1865, opened 2023-08-25
update github actions CI for protocol-verification

Ashar2shahid:
Protocol verification didn't seem to be working, the following changes should trigger it on `packages/airnode-protocol` changes

The condition:
```
github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
```

probably never gets met because :
- the event is always going to be 'pull-request'
- [github.event.pull_request.head.repo.full_name](https://docs.github.com/en/webhooks/webhook-events-and-payloads#example-delivery) should be matching [github.repository](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context).

There's no real way of testing github actions locally other than forking and testing it there, so if the error still persists then I will opt for that route and create another PR

### DIFF
--- a/.github/workflows/protocol-verify.yml
+++ b/.github/workflows/protocol-verify.yml
@@ -16,8 +16,6 @@ jobs:
   pre-build:
     name: Prepare build environment
     runs-on: ubuntu-latest
-    # Don't run twice for a push within an internal PR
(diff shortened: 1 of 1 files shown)


dcroote (approved):
I agree with your logic here. Currently, the workflow is `on` only pull requests, but the `if:` conditionals limit it to running only for `push` or when originating from a fork.

PR #1888, opened 2023-09-27
feat: make operationParameter optional within endpoint parameters

### LINKED ISSUE #1800: Allow endpoint parameters to not be mapped to operation parameters
We have an endpoint parameter called `name`. Our API response is as follows
```
{
  "ETH/USD": 1500,
  "BTC/USD": 25000
}
```
We want the endpoint to return `1500` when `name` is `ETH/USD`, and return `25000` when `name` is `BTC/USD`. This is easily done by the postprocessing step `output = input[endpointParameters.name]` (once https://github.com/api3dao/airnode/issues/1673 is implemented), yet the problem is when we declare the endpoint parameter `name`, we have to dummy-map it to an operation parameter, which is ugly. (Yes, we can already do this by using `_path`, but we don't want to do that because that would require us to change out template ID, which would require us to change our Beacon ID.)

In short, while an endpoint is being declared in OIS, `operationParameter` should be optional. I'm not sure if I should have created this issue on https://github.com/api3dao/ois but it affects Airnode implementation as well.

dcroote:
Closes #1800.


Siegrift (approved):
👍 LGTM

Issue #1559, opened 2022-11-25
Add error checking to airnode examples


aTeoke:
Airnode examples often don't warn when a preceding step hasn't been followed. Ideally, to improve UX and reduce future queries, the scripts should check for previous steps not having been followed. Example: the deployment of the RRP contract.

PR #1895, opened 2023-10-09
fix(examples): check for AirnodeRrp deployment in localhost flow

dcroote:
Very minor. Contributes to #1559.

The error without this check is either `Error: ENOENT: no such file or directory...` or `Error: invalid contract address or ENS name (argument="addressOrName", value=undefined, code=INVALID_ARGUMENT, version=contracts/5.7.0)` depending on whether the `deployments/localhost.json` file is present. Both of these don't make it clear that the `deploy-rrp` wasn't run.

### DIFF
--- a/.changeset/curvy-timers-kick.md
+++ b/.changeset/curvy-timers-kick.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-examples': patch
+---
+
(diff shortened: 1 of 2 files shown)


Siegrift (approved):
👍 LGTM

Issue #1896, opened 2023-10-10
Use processing implementation from commons


Siegrift:
Relates to https://github.com/api3dao/commons/issues/7

PR #1899, opened 2023-10-18
Import processing from commons

Siegrift:
Closes https://github.com/api3dao/airnode/issues/1896


packages/airnode-node/src/api/index.ts
```diff
@@ -295,9 +297,14 @@ export async function callApi(payload: ApiCallPayload): Promise<LogsData<ApiCall
...
   if (verificationResult) return verificationResult;
 
-  const processedPayload = await preProcessApiSpecifications(payload);
-  const ois = payload.config.ois.find((o: OIS) => o.title === payload.aggregatedApiCall.oisTitle)!;
-  const endpoint = ois.endpoints.find((e: Endpoint) => e.name === payload.aggregatedApiCall.endpointName)!;
+  const {
+    aggregatedApiCall: { parameters },
+  } = payload;
+  const ois = payload.config.ois.find((o) => o.title === payload.aggregatedApiCall.oisTitle)!;
+  const endpoint = ois.endpoints.find((e) => e.name === payload.aggregatedApiCall.endpointName)!;
```
dcroote, line 304: Why were the types (and their corresponding imports) removed from this line and the one above it?

packages/airnode-node/src/api/index.ts
```diff
@@ -240,7 +242,7 @@ export async function processSuccessfulApiCall(
   if (!goExtractAndEncodeResponse.success) {
     const log = logger.pend('ERROR', goExtractAndEncodeResponse.error.message);
     // The HTTP gateway is a special case for ChainAPI where we return data from a successful API call that failed processing
-    if (payload.type === 'http-gateway') {
+    if (type === 'http-gateway') {
```
dcroote, line 245: Nice find

dcroote (approved):
Looks good; one minor question

packages/airnode-node/src/api/index.ts
Siegrift: They were redundant. OIS is typed so there was no need to have them.

PR #1921, opened 2023-12-11
feat: Bump axios and api3 deps

dcroote:
Remove reliance on deployed RequesterAuthorizerWithErc721 addresses


bbenligiray:
I would leave the RequesterAuthorizerWithErc721 stuff in, it's just that the old deployments are no longer valid and should not auto-populate the config. When we deploy the new versions, we want them to auto-populate the config.

dcroote:
> I would leave the RequesterAuthorizerWithErc721 stuff in

Agreed and I should have described the PR better- I've taken out only the code / tests that rely on the existence of a deployment (for which there are none currently)

bbenligiray:
If it's going to make things easier later, I can do a Sepolia deployment. I don't expect the contract to change later.

dcroote:
> If it's going to make things easier later, I can do a Sepolia deployment. I don't expect the contract to change later.

If possible I think that would be easiest as then none of the tests or configs would change beyond the new default Sepolia address. I know it isn't high priority to deploy the contract and then release though.

bbenligiray:
I deployed RequesterAuthorizerWithErc721 on Sepolia and published @api3/airnode-protocol-v1@3.1.0

dcroote:
Thank you! I reverted back to the original configs and tests, with minor updates to use the new Sepolia deployment.

bbenligiray (approved):
Thanks, I had forgotten about this :pray:

PR #1928, opened 2023-12-31
Increase OEV gateway memory

Siegrift:
We've been discussing plans to increase the OEV gateway memory to reuse the OEV gateway implementation in https://github.com/api3dao/oev-auctioneer/issues/6. I've done some [experiments](https://api3workspace.slack.com/archives/C06A83ZN30D/p1703591954683099) and the cold start varies depending on the gateway memory. This is because the lambda gets CPU proportionally to the memory (and higher memory means higher CPU which decreases the work at cold start).

I am opening a PR which changes the GW memory to 2048 (not configurable).

### DIFF
--- a/.changeset/healthy-jeans-cheer.md
+++ b/.changeset/healthy-jeans-cheer.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-deployer': minor
+---
+
+Increase OEV gateway memory limit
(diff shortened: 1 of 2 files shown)


bbenligiray (approved):
I can't think of anything else to be done about this

Siegrift:
I was waiting for our discussion in Slack to end, which I think did and I think this is good to go.

PR #1937, opened 2024-01-20
Release v0.14.0

dcroote:


Siegrift (approved):
👍 LGTM.

Btw. I will be conducting a few more tests for Airnode OEV gateway, but that is independent from the release. (And I am optimistic it's going to work).

Siegrift:
I think it wouldn't hurt. But I also want to check that the interface is indeed compatible to avoid unnecessary redeploys for API providers.

PR #1940, opened 2024-01-27
docs: replace old Rinkeby reference

dcroote:
The main purpose of this PR is to add a changeset so that the Docker build step of CI stops failing (see https://github.com/api3dao/airnode/issues/1585#issuecomment-1420226393)

### DIFF
--- a/.changeset/many-flies-count.md
+++ b/.changeset/many-flies-count.md
@@ -1,0 +1,5 @@
+---
+'@api3/airnode-protocol': patch
+---
+
+Replace old Rinkeby reference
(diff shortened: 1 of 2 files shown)


Siegrift (approved):
👍 LGTM, but how did this error happen in the first place 🤔 ?

dcroote:
> how did this error happen in the first place 🤔 ?

Do you mean why does the absence of a changeset cause CI failure? Or why was there a Rinkeby reference in the README? If the former, this has existed for some time, but only crops up after a release followed by a PR without a changeset. If the latter, I'm guessing when other references to Rinkeby were being removed, this README instance was missed.

dcroote:
> We have a check that ensures that every PR has a changeset

Except for Renovate PRs, which is what I happened to merge first after the 0.14 release 🙃

Issue #1997, opened 2024-05-12
Investigate gas estimation with large amounts of gas


dcroote:
A [user on Discord is reporting](https://discord.com/channels/758003776174030948/1096629188397637733/1238753928653570078) gas estimation issues when fulfillment uses a large amount of gas:
 > for this kind of behavior, you must keep the gas of the fulfillment less than 1.6M. the problem is that; when the estimated gas increases, the estimation gets a little less than what the transaction uses.

> I don't think it's you guys problem because i've checked the airnode code and you guys are doing the estimation like everybody else does, but you can just start out by creating an fulfillment that has like 2M gas

bbenligiray:
As a note, the user reporting this seems to be using base-sepolia-testnet

https://sepolia.basescan.org/tx/0xff1400c4eb0fbcede0b01a456060f3702b0ae4504268d29fdcf44a8359671029#eventlog

tarik0:
> As a note, the user reporting this seems to be using base-sepolia-testnet
> 
> https://sepolia.basescan.org/tx/0xff1400c4eb0fbcede0b01a456060f3702b0ae4504268d29fdcf44a8359671029#eventlog

The problem does not only happen in Base Sepolia, It also happens in Ethereum Sepolia as well.

tarik0:
```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.20;

import "@api3/airnode-protocol/contracts/rrp/interfaces/IAirnodeRrpV0.sol";
import "@openzeppelin/contracts/access/Ownable.sol";

contract MockedRequester is
    Ownable
{
    address internal _airnodeRrp;
    address internal _airnode;
    address internal _sponsorWallet;
    bytes32 internal _endpointIdUint256;

    event Requested(bytes32 requestId);

    constructor() {}

    function setSettings(
        address airnodeRrp,
        address airnode,
        address sponsorWallet,
        bytes32 endpointIdUint256
    ) external onlyOwner {
        _airnodeRrp = airnodeRrp;
        _airnode = airnode;
        _sponsorWallet = sponsorWallet;
        _endpointIdUint256 = endpointIdUint256;

        IAirnodeRrpV0(_airnodeRrp).setSponsorshipStatus(address(this), true);
    }

    struct Request {
        uint256 fromId;
        uint256 toId;
        uint256 timestamp;
        bool hasReferral;
    }

    struct Response {
        uint256 rawSeed;
        uint256 timestamp;
        bool hasReferral;
[...]

PR #2001, opened 2024-06-20
Adds an extra to the tx estimate to account for EIP-150 reserved gas

acenolaza:
Closes https://github.com/api3dao/airnode/issues/1997

### What does this change

The way Airnode currently estimates gas for fulfill txs (`AirnodeRrpV0DryRun.fulfill()` + `target.fulfill()`) produces a gas estimate that might be inaccurate on txs where the `fulfill()` function requires a lot of gas to be executed. This is mainly because even thought we estimate a low level `.call(...)` on `AirnodeRrpV0DryRun.fulfill()`, the 1/64 gas reserve on the caller introduced in EIP-150 is not being considered on the estimate. This is fine for a `fulfill()` tx that doesn't need much gas to execute (because we are overestimating anyways) but for fulfillments that require large amounts of gas this reserved gas must be accounted for or we could end up with estimates that are less than actually needed.

### DIFF
--- a/packages/airnode-node/src/evm/fulfillments/api-calls.ts
+++ b/packages/airnode-node/src/evm/fulfillments/api-calls.ts
@@ -316,15 +316,20 @@ export async function estimateGasAndSubmitFulfill(
   }
 
   const [airnodeRrpGasCost, fulfillmentCallGasCost] = estimateGasData;
-  const totalCost = airnodeRrpGasCost.add(fulfillmentCallGasCost);
+  const subTotalCost = airnodeRrpGasCost.add(fulfillmentCallGasCost);
+  // https://github.com/ethereum/EIPs/issues/114
+  // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-150.md
+  const eip150GasReserve = subTotalCost.div(64);
+  const totalGasCost = subTotalCost.add(eip150GasReserve);
+
   // If gas estimation is success, submit fulfillment without making static test call
   const gasLimitNoticeLog = logger.pend(
     'INFO',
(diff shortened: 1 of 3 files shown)


packages/airnode-node/src/evm/fulfillments/api-calls.ts
```diff
@@ -316,15 +316,20 @@ export async function estimateGasAndSubmitFulfill(
   }
 
   const [airnodeRrpGasCost, fulfillmentCallGasCost] = estimateGasData;
-  const totalCost = airnodeRrpGasCost.add(fulfillmentCallGasCost);
+  const subTotalCost = airnodeRrpGasCost.add(fulfillmentCallGasCost);
+  // https://github.com/ethereum/EIPs/issues/114
+  // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-150.md
+  const eip150GasReserve = subTotalCost.div(64);
```
dcroote, line 322: Shouldn't we divide by 63 since our estimate is 63/64 of the true amount required?

acenolaza: Right, good catch 👏🏻

acenolaza: https://github.com/api3dao/airnode/pull/2001/commits/36d4325c7b45f5bf9a149b7bc68a626ee8ecf573

dcroote (approved):
The calculation looks good but I haven't tested- have you tested with large amounts of gas? If you'd like I could re-run a test I did previously.

dcroote (approved):
> Gas Limit & Usage by Txn:
5,030,630 | 4,886,991 (97.14%)

Nice!! 🎉 Definitely looks solved since I previously couldn't get close to a successful tx with that much gas.

Ashar2shahid:
great job @acenolaza

Issue #2028, opened 2025-02-05
Release & test v0.15 snapshot packages


dcroote:

dcroote:
Ran into the following terraform error (`Error: Command failed: terraform apply -var="aws_region=...` when testing an AWS deployment
```
Error: expected runtime to be one of [nodejs nodejs4.3 nodejs6.10 nodejs8.10 nodejs10.x nodejs12.x nodejs14.x nodejs16.x java8 java8.al2 java11 python2.7 python3.6 python3.7 python3.8 python3.9 dotnetcore1.0 dotnetcore2.0 dotnetcore2.1 dotnetcore3.1 dotnet6 nodejs4.3-edge go1.x ruby2.5 ruby2.7 provided provided.al2 nodejs18.x python3.10 java17], got nodejs20.x

with module.httpReq.aws_lambda_function.lambda,
```

Which, after some research, makes sense since our version is
https://github.com/api3dao/airnode/blob/4a8b87752019f7ca9c081f4a5b3ae2fe9d2fa0d6/packages/airnode-deployer/terraform/aws/providers.tf#L5

and nodejs20.x support was added by terraform [in v5.26.0](https://github.com/hashicorp/terraform-provider-aws/blob/v5.26.0/CHANGELOG.md).

Therefore, https://github.com/api3dao/airnode/pull/1780 will need to be tested & merged

dcroote:
Upgraded terraform-provider-aws in #1780 and terraform-provider-google in #2008. I tested each in the process so should be good to go now

dcroote:
Tested using `yarn test:release` against snapshot `snapshot-v0.15.0-rc.1` and it worked

PR #2030, opened 2025-02-10
Release v0.15.0

dcroote:
Tested in #2028 - tl;dr - tested, there was a terraform aws issue, I bumped that and the analogous gcp terraform versions, tested those, then retested the snapshot packages


Siegrift (approved):
👍 LGTM

PR #2045, opened 2026-01-07
Use shared Renovate preset

mcoetzee:
Relates to https://github.com/api3dao/tasks/issues/1777

I'm leaving `lockFileMaintenance` disabled, because enabling it with yarn v1 would increase the risk of supply chain attacks.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,31 +1,14 @@
 {
-  "extends": ["config:base", "docker:disable"],
-  "automerge": true,
-  "automergeType": "pr",
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["local>api3dao/renovate-config", "docker:disable"],
   "packageRules": [
     {
-      "matchPackageNames": ["chalk", "hardhat", "ora", "node", "@openzeppelin/contracts", "zod"],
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -1,31 +1,14 @@
 {
-  "extends": ["config:base", "docker:disable"],
-  "automerge": true,
-  "automergeType": "pr",
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["local>api3dao/renovate-config", "docker:disable"],
   "packageRules": [
     {
-      "matchPackageNames": ["chalk", "hardhat", "ora", "node", "@openzeppelin/contracts", "zod"],
+      "matchPackageNames": ["ora", "node", "@openzeppelin/contracts"],
```
dcroote: Could you retain `zod` too please? We manually update it with OIS package bumps because we've had compatibility issues when the versions are mismatched, see https://github.com/api3dao/ois?tab=readme-ov-file#make-sure-only-one-zod-version-is-installed

dcroote (approved):
One minor request, but otherwise looks good, and I agree with the lock file choice. 

As for pnpm, yes a transition from yarn is overdue.

